PR #19, opened 2014-06-21
Remove -c flag. Erroneous output caused on android

venkateshshukla:
While cross-building for android, the -c flag was causing errors.
This -c option ends up in Libs.private of libusb-1.0.pc file.
On its usage, it is interpreted as "Compile and assemble, but do not link"
option of gcc. Usage of -c in this way might be unintended.
Hence, removing this option.

Signed-off-by: Venkatesh Shukla <venkatesh.shukla.eee11@iitbhu.ac.in>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -121,9 +121,6 @@ linux)
 
 case $is_backend_android in
 yes)
-	dnl some pthread functions is in libc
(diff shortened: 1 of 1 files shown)


dickens:
Closed in c7401ca

PR #51, opened 2015-01-25
Update the linking path to @executable_path.

olunx:


LudovicRousseau:
I just fixed the Xcode project using the correct end-of-line.

Can you rebase your patch to the latest git version and submit it again?

LudovicRousseau:
@olunx I have no news since 1 year.
Can you document the problem your patch is trying to solve?

hjelmn:
Considering the translation of his commit message is "Change to a static library" I don't think we want this.

LudovicRousseau:
@hjelmn that is also what I found reading the patch/diff.
Closing.

PR #52, opened 2015-02-09
darwin: fix resource leak (cached devices are never freed)

parafin:
For each new device (processed by hotplug code) darwin_ref_cached_device
is called twice - in process_new_device and in darwin_get_cached_device
(which is called from the first function). But when the device is detached
only one darwin_deref_cached_device is done - in darwin_destroy_device.
So there is inbalance between ref and deref which causes darwin_cached_devices
list to monotonically grow and consume more and more resources over time.
This commit fixes the problem by adding darwin_deref_cached_device call
to darwin_devices_detached.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -279,6 +279,7 @@ static void darwin_devices_attached (void *ptr, io_iterator_t add_devices) {
 static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) {
   struct libusb_device *dev = NULL;
   struct libusb_context *ctx;
+  struct darwin_cached_device *old_device;
 
   io_service_t device;
   UInt64 session;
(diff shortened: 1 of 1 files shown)


hjelmn:
I will take a look at this tonight.

hjelmn:
This really isn't a resource leak as the disconnected cached devices will be destroyed at exit. That said, for a long running program with lots of attach/detach the cached device list will grow which is a real problem.

At some point I based the caching off the device location (which is not completely unique) and this wasn't a problem.

parafin:
Well, maybe by definition it's not a leak, since there is a cleanup code which is run on library unload (program exit), but consequences are the same as if it was a real leak - there is no point in keeping around information (and particularly open handle) about disconnected device. I found this issue because there is a report that real problems arise (like long USB bus scanning time) when program using libusb runs for a long time on OS X and a lot of USB connect/disconnect events happen in that time. So it's good that you accepted the patch;)

PR #59, opened 2015-03-16
Remove unnecessary \n from log messages

photron:
The final end-of-line marker is added in usbi_log_v() using the
USBI_LOG_LINE_END define.

Also don't mix hardcoded \n and USBI_LOG_LINE_END in usbi_log_v().


dickens:
Applied, thanks.

PR #60, opened 2015-03-16
Use standard struct timespec on vs2015 and above

shuffle2:
Currently libusb will fail to compile with vs2015, so it should be updated before vs2015 is released, hopefully...


shuffle2:
ping, vs2015 is released now.

BrentRector:
I've submitted a pull request with all the necessary changes so Visual Studio 2015 compiles libusb without warnings or errors.

PR #62, opened 2015-04-14
Remove unused member add_iso_packet_size from struct usbi_os_backend

tklauser:
The member add_iso_packet_size of struct usbi_os_backend is set to 0 in
all backends and thus has no actual use. This has been the case since
commit ad6e2b71 ("Linux: fire multiple URBs at once for split
transfers").

Also, the comment above the member actually states the fact that it is
unused and could potentially be removed, so remove it.

Signed-off-by: Tobias Klauser <tklauser@distanz.ch>

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -1249,8 +1249,7 @@ struct libusb_transfer * LIBUSB_CALL libusb_alloc_transfer(
 	int iso_packets)
 {
 	struct libusb_transfer *transfer;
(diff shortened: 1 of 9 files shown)


tklauser:
Rebased on top of master as of 4ec225e835167c9841a2bba00a252213dcb85a1e to resolve merge conflict in libusb/os/wince_usb.c

dickens:
Applied, thanks.

PR #64, opened 2015-04-28
Windows: Fixes for corner case leaks when reusing device references

projectgus:
I noticed my libusb-based program was leaking memory under Windows and tracked it down to leaks that occur when libusb_get_device_list() reuses existing device references.

A test program that demonstrates the problem is here:
https://gist.github.com/projectgus/c9e29f90397202dbba38

* When run on Windows with current libusb the debug output includes "some libusb_devices were leaked" near the end, even though all instances are released. (related to 0cbb891)
* When run under drmemory leaks are reported on close. (related to f6b348f)
* If run with the RUN_INFINITE variable changed to 1 then the memory usage will climb indefinitely.

The three commits in this PR fix these problems for me.

I've been cross-compiling with mingw and testing on Windows 7 and XP 32-bit. But I don't think those details are relevant.

BTW, thanks very much to everyone who works on libusb. It's marvellous being able to write cross-platform code for something as complex as USB, and have it run everywhere. I am even more admiring of libusb after today, having now become more familiar with the actual Windows USB stack. ;)

### DIFF
--- a/libusb/os/windows_usb.c
+++ b/libusb/os/windows_usb.c
@@ -1483,7 +1483,7 @@ static int set_hid_interface(struct libusb_context* ctx, struct libusb_device* d
 	for (i=0; i<priv->hid->nb_interfaces; i++) {
 		if (safe_strcmp(priv->usb_interface[i].path, dev_interface_path) == 0) {
 			usbi_dbg("interface[%d] already set to %s", i, dev_interface_path);
-			return LIBUSB_SUCCESS;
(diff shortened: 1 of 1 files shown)


dickens:
Applied the first two commits. The third is not needed, so it was omitted. Thanks for submitting this!

rezahousseini:
Where is this pull request merged?

mcuee:
@rezahousseini Please create a new issue if you still see the issue in 1.0.24 or latest git. Thanks.

PR #65, opened 2015-04-29
Miscellaneous fixes

tklauser:
- Replace open coded functions by the proper library function/header
- Remove unused function parameters or use them properly (instead of accessing globals)


dickens:
Applied, thanks!

PR #71, opened 2015-05-16
Added function to retreive context of a device. libusb_get_context()

kuldeepdhaka:
This will allow application code to get device context.
earlier, it is not possible to get device context without hacking device data structure.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1,6 +1,7 @@
 /* -*- Mode: C; indent-tabs-mode:t ; c-basic-offset:8 -*- */
 /*
  * Core functions for libusb
+ * Copyright © 2015 Kuldeep Singh Dhaka <kuldeepdhaka9@gmail.com>
  * Copyright © 2012-2013 Nathan Hjelm <hjelmn@cs.unm.edu>
  * Copyright © 2007-2008 Daniel Drake <dsd@gentoo.org>
  * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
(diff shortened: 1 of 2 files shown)


dickens:
Hi,

I don't see the utility of having such a function. Most libusb functions
require either a context, device, or device handle. I can only find one
function (libusb_get_port_path) where you explicitly need a device's
context in addition to the device.

It assumed that the application or library using libusb either
exactly knows or doesn't care about the context. It would be a
bit strange to have more than one context in use and not internally keep
track of which device belongs to which context.

For me, this is a NAK.

Regards,
Chris

kuldeepdhaka:
Hello @dickens 
I was designing a library where i need to accept a libusb_device from application code.
now, since their is no method to extract libusb_context from libusb_device, i have to accept a libusb_device and a libusb_context as parameter.
my function already has about 2-3 parameters (variations) and parameter {libusb_device, libusb_context} is also required, a total of 4-5 parameters, why not just remove libusb_context since it is already referenced by libusb_device.

their is one more usecase, in order to poll the device, i need the context.
this is like, libusb_get_device(libusb_device_handle) which return libusb_device
and i added this libusb_get_context(libusb_device) which libusb_context
not everybody realise the need to the 2nd function, but it *is* useful in some case. 

yes, you can say, why not ask for context and device but a simple question: is it bad to make a flexible API? (it is just returning the referenced context)

kuldeepdhaka:
Please accept or reject this PR.

kevinmehall:
I also have a use case where I'd like to accept a `libusb_device_handle`, start some asynchronous transfers on it, and call `libusb_handle_events_timeout_completed` to wait for their completion, but that function requires `libusb_context`.

The libusb synchronous API is implemented that way. As the synchronous functions are simple wrappers around the asynchronous API, why shouldn't library users be able to do the same?

dickens:
Thanks for submitting and discussing this PR. After some deliberation, the consensus among the maintainers is to reject this. We feel that the context is a crucial piece of information that any caller of libusb should have explicit knowledge about.

kuldeepdhaka:
@kevinmehall yup!

PR #83, opened 2015-07-25
Add Visual Studio 2015 support.

BrentRector:


Oukache:
hotplug capabilites are supported on windows 10 platform, for this version ?

mcuee:
Currently hotplug is not supported under libusb Windows.

xlz:
Tests OK with libfreenect2, a downstream libusb application.

libusb/libusbi.h
```diff
@@ -164,7 +164,7 @@ static inline void *usbi_reallocf(void *ptr, size_t size)
 #ifndef TIMESPEC_TO_TIMEVAL
 #define TIMESPEC_TO_TIMEVAL(tv, ts)                                     \
         do {                                                            \
-                (tv)->tv_sec = (ts)->tv_sec;                            \
+                (tv)->tv_sec = (long) (ts)->tv_sec;                            \
```
iHateInventNames: this seems wrong.
long is 32bit we should not made destructive casts.
consider changing type of lvalues

hjelmn: On 64-bit Linux and OS X long is 64-bit but I agree this is a weird cast since tv_sec is of type time_t.

hjelmn:
BTW, can you redo this without the merge? For PRs please rebase your work on top of the current libusb master.

libusb/libusbi.h
xlz: long is 32-bit on Windows. time_t is 64-bit. timeval.tv_sec is long from winsock.h, it can't be changed.

There are definitely Y2038 problems, but other places are littered with `(long)` cast. Fixing Y2038 bugs would be of a future PR.

xlz:
I rebased it in #120.

libusb/libusbi.h
dickens: I committed a modified version of this patch that does the cast either to long or time_t depending on the platform. I feel better about doing it this way than blindly using long across all platforms.

PR #99, opened 2015-09-06
add interrupt.{c,h} from chromium

rezso:
https://code.google.com/p/chromium/issues/detail?id=266149


LudovicRousseau:
Can you also provide a Doxygen documentation?

dickens:
This patch is not valid for the latest libusb code. Also, this function
does not require its own file--it can be placed in io.c with the rest of
the event handling functions.

Regards,
Chris

PR #100, opened 2015-09-08
New libusb_fdopen(dev, fd, handle) API to provide ability to associate the file descriptor with a device handle.

dmitrykos:
The proposed API would act by idea similar to fdopen(), e.g. it would be possible to associate a created device handle with the existing file descriptor which was obtained earlier for the USB device.

It would be helpful to handle device opening on Android OS for instance, where file descriptor is obtained when user granted the permission to access a USB device. That file descriptor could then be passed to libusb API via libusb_fdopen() call.

The Linux back-end was modified to handle libusb_fdopen() call and use supplied file descriptor instead of trying to call _get_usbfs_fd(). Potentially libusb_fdopen() could be used to provide external file handles to a Windows back-end too. For that case intptr_t type is used for the 'fd' parameter of libusb_fdopen() function in order to guarantee the sufficient parameter width for passing the raw pointer.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1280,7 +1280,6 @@ int API_EXPORTED libusb_open(libusb_device *dev,
  * This function allows to associate a device handle with the existing file 
  * descriptor.
  *
- *
  * \param dev the device to open
  * \param fd the file descriptor to associate with the device 
  * \param handle output location for the returned device handle pointer. Only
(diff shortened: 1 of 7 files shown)


LudovicRousseau:
You should create a branch on your repository to ONLY contain the libusb_fdopen() changes.

Your PR contains unrelated changes. Please create a new one and close this one.

PR #101, opened 2015-09-11
New libusb_fdopen(dev, fd, handle) API to provide ability to associate the file descriptor with a device handle.

dmitrykos:
The proposed API would act by idea similar to fdopen(), e.g. it would be possible to associate a created device handle with the existing file descriptor which was obtained earlier for the USB device.

It would be helpful to handle device opening on Android OS for instance, where file descriptor is obtained when user granted the permission to access a USB device. That file descriptor could then be passed to libusb API via libusb_fdopen() call.

The Linux back-end was modified to handle libusb_fdopen() call and use supplied file descriptor instead of trying to call _get_usbfs_fd(). Potentially libusb_fdopen() could be used to provide external file handles to a Windows back-end too. For that case intptr_t type is used for the 'fd' parameter of libusb_fdopen() function in order to guarantee the sufficient parameter width for passing the raw pointer.

edit:
Reverted all unrelated changes and merged the proposal into the 'libusb_fdopen' branch.


dickens:
I do not think the core should be aware of any of this. The fd or HANDLE or
whatever is a backend-specific detail, thus I don't think the libusb_device
struct should be modified to accommodate this, and the core should simply
have a function that calls the backend function, or returns appropriately
if the backend does not provide this.

Regards,
Chris

hjelmn:
I agree.

Also, in the spirit of fdopen I would prefer that libusb_fdopen take ownership of the file descriptor, IOUSBDevice, etc. The backend would be responsible for closing and releasing it. Also, since not all systems use a file descriptor it might be worth coming up with a new name.

-Nathan

dmitrykos:
> dickens
> the core should simply have a function that calls the backend function

The problem with Android is that obtaining USB device file descriptor can not be done in one call. E.g. application must first request the permission to use a device and then user confirms or rejects it. It is all done asynchronously and can take indefinite amount of time.

Developing some Android specific function which would extract device file descriptor in one call looks quite impossible to me in normal Android usage circumstances unless device is rooted and USB devices can be opened in a traditional way how libusb is handling it currently.

libusb_fdopen() (or with another name if fdopen) could be #ifdefed for Android only. The idea behind libusb_fdopen() was to offer API which would not break binary compatibility and on the other hand would provide possibility to use libusb on Android right from the box.

dickens:
Closing this as the PR has many more commits unrelated to the subject and because it is obsolete based on the current master. If you'd like, please open another PR with only commits related to the subject, and provide full documentation supporting any new APIs.

aalmada:
Are the file descriptor and SELinux Android-related issues already handled on the latest libusb build? If so, how do I pass the file descriptor to libusb?  There are multiple forks around but would like to use the official one with the latest source. Thanks.

PR #109, opened 2015-10-08
Isochronous transfer support for a Windows back-end when libusbK back-end is used for a device driver.

dmitrykos:
Isochronous transfer support for a Windows back-end when libusbK back-end is used for a device driver. The initial code is obtained from a http://github.com/JoshBlake/libusbx.git repository and adopted to the current libusb code along with some necessary fixes to make it work properly.

The implementation was tested (audio output) and is working properly under Windows 7 on two different USB DAC devices with isochronous support. Device driver was installed by Zadig tool (http://zadig.akeo.ie) version 2.1.2 which installed libusbK (v3.0.7.0) back-end.


dickens:
Closing this as the PR has many more commits unrelated to the subject and because it is obsolete based on the current master.

PR #112, opened 2015-10-15
darwin: fix occasional dead-lock on libusb_exit

parafin:
CFRunLoopStop() isn't thread-safe, see for example this bugreport:
https://github.com/joyent/libuv/issues/799
Use CFRunLoopSource instead. Test-case:
```C
#include <libusb.h>
#include <stdio.h>
int main() {
  int i = 0;
  while(1) {
    libusb_init(0);
    printf("%d\n", i++);
    libusb_exit(0);
  }
  return 0;
}
```
will hang after some time (may take a while) without this patch.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -50,6 +50,7 @@ static clock_serv_t clock_realtime;
 static clock_serv_t clock_monotonic;
 
 static CFRunLoopRef libusb_darwin_acfl = NULL; /* event cf loop */
+static CFRunLoopSourceRef libusb_darwin_acfls = NULL; /* shutdown signal for event cf loop */
 static volatile int32_t initCount = 0;
 
 static usbi_mutex_t darwin_cached_devices_lock = PTHREAD_MUTEX_INITIALIZER;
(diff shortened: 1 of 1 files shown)


hjelmn:
Interesting. Apple has this to say about run loops:

***Thread Safety and Run Loop Objects***
Thread safety varies depending on which API you are using to manipulate your run loop. The functions in Core Foundation are generally thread-safe and can be called from any thread. If you are performing operations that alter the configuration of the run loop, however, it is still good practice to do so from the thread that owns the run loop whenever possible.

https://developer.apple.com/library/prerelease/mac/documentation/Cocoa/Conceptual/Multithreading/RunLoopManagement/RunLoopManagement.html#//apple_ref/doc/uid/10000057i-CH16-SW1

I didn't think CFRunLoopStop as a call that alters the configuration of a run loop so I assumed it was thread-safe. It might be worth asking Apple to clarify (and potentially open a bug on radar).

Even if the routine is supposed to be thread-safe and Apple fixes it libusb needs to support unpatched versions of OS X for some time. So, I will review this patch for inclusion in 1.0.21.

dickens:
@hjelmn how about this one?

hjelmn:
I may want to take this one. As I said, there is no indication this call shouldn't be thread safe. I will decide tomorrow.

Mitch9192:
I am still experiencing this issue.  Built from Master on OS X 10.15.1.  Example code provided at the top of this issue will exhibit the behavior.  Appears OS specific.  (Linux doesn't exhibit the issue)

* Before updating my libusb (pre this fix) it would occur in < 85k iterations. 
* New version seems more stable.  In fact, I have seen almost 2 million iterations without issue; however, it does still occur.


Version 1.0.20 (I think)
Screen Shot 2020-02-25 at 2 03 15 PM


Version 1.0.23
Screen Shot 2020-02-26 at 2 57 46 PM

PR #120, opened 2015-10-27
Add Visual Studio 2015 support, rebased

xlz:
Rebased for #83.

Just want the support to be merged quickly.

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -164,7 +164,7 @@ static inline void *usbi_reallocf(void *ptr, size_t size)
 #ifndef TIMESPEC_TO_TIMEVAL
 #define TIMESPEC_TO_TIMEVAL(tv, ts)                                     \
         do {                                                            \
-                (tv)->tv_sec = (ts)->tv_sec;                            \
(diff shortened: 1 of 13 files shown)


libusb/libusbi.h
```diff
@@ -164,7 +164,7 @@ static inline void *usbi_reallocf(void *ptr, size_t size)
 #ifndef TIMESPEC_TO_TIMEVAL
 #define TIMESPEC_TO_TIMEVAL(tv, ts)                                     \
         do {                                                            \
-                (tv)->tv_sec = (ts)->tv_sec;                            \
+                (tv)->tv_sec = (long) (ts)->tv_sec;                            \
```
iHateInventNames: We have to process overflow case. Throw exception if cast result wrong would be good enough.
if((long) ts->tv_sec != ts->tv_sec) throw <bad_cast>;

xlz: This was discussed in #83.

hjelmn: Also keep in mind this is C. There is no throw.

PR #122, opened 2015-11-13
configure: remove AM_MAINTAINER_MODE

rossburton:
AM_MAINTAINER_MODE, contrary to it's name, disables the rebuild rules unless you enable them explicitly.  There are very few situations where this is desirable and can lead to unpredictable or broken builds.  In fact even the creator of AM_MAINTAINER_MODE now recommends against using it.

http://www.gnu.org/software/automake/manual/html_node/maintainer_002dmode.html has more analysis.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -27,7 +27,6 @@ lt_age="1"
 LTLDFLAGS="-version-info ${lt_current}:${lt_revision}:${lt_age}"
 
 AM_INIT_AUTOMAKE
-AM_MAINTAINER_MODE
 
 AC_CONFIG_SRCDIR([libusb/core.c])
 AC_CONFIG_MACRO_DIR([m4])


hjelmn:
I am leaning towards taking this one. Will see if there is a consensus.

hjelmn:
@LudovicRousseau, @mcuee, @pbatard Any comments on this one. Looks good to me.

LudovicRousseau:
No objection.
If even the creator of AM_MAINTAINER_MODE says it is a bad idea...

Some years ago I had problems with timestamps (of files extracted for a ``.tar.gz`` archive IIRC) and ``make`` wanted to regenerate auto-generated files. This was really a problem for _normal_ users wihout the autotools installed. Maybe it was because I used CVS at that time.

hjelmn:
At the same time as removing AM_MAINTAINER_MODE I think we should increase the minimum autoconf version. Requiring 2.69 would be ideal and would only affect maintainers.

Issue #63, opened 2015-04-14
Cannot build using Visual Studio 2015 CTP


deisenhut:
I am attempting to build libusb-1.0.19 on Windows using Visual Studio 2015 CTP.  It builds fine when using Visual Studio 2013 with just a couple of deprecation warnings, but using Visual Studio 2015 CTP I get the following errors.  Note, I used the 2013 solution file and let VS upgrade the projects.  The only difference I see in the solution and project files are changes in the version number.

1>------ Build started: Project: libusb-1.0 (static), Configuration: Debug Win32 ------
2>------ Build started: Project: getopt, Configuration: Debug Win32 ------
3>------ Build started: Project: libusb-1.0 (dll), Configuration: Debug Win32 ------
1>  core.c
3>  core.c
1>D:\Perforce\my-project\3rdParty\Libusb\OriginalSource\libusb-1.0.19\libusb\os/threads_windows.h(47): error C2011: 'timespec': 'struct' type redefinition
1>  C:\Program Files (x86)\Windows Kits\10\Include\ucrt\time.h(39): note: see declaration of 'timespec'
1>  descriptor.c
2>  getopt1.c
2>  getopt.c
2>  Generating Code...
3>D:\Perforce\my-project\3rdParty\Libusb\OriginalSource\libusb-1.0.19\libusb\os/threads_windows.h(47): error C2011: 'timespec': 'struct' type redefinition
3>  C:\Program Files (x86)\Windows Kits\10\Include\ucrt\time.h(39): note: see declaration of 'timespec'
3>  descriptor.c
1>D:\Perforce\my-project\3rdParty\Libusb\OriginalSource\libusb-1.0.19\libusb\os/threads_windows.h(47): error C2011: 'timespec': 'struct' type redefinition
1>  C:\Program Files (x86)\Windows Kits\10\Include\ucrt\time.h(39): note: see declaration of 'timespec'
1>  hotplug.c
3>D:\Perforce\my-project\3rdParty\Libusb\OriginalSource\libusb-1.0.19\libusb\os/threads_windows.h(47): error C2011: 'timespec': 'struct' type redefinition
3>  C:\Program Files (x86)\Windows Kits\10\Include\ucrt\time.h(39): note: see declaration of 'timespec'
3>  hotplug.c
2>  getopt_2015.vcxproj -> D:\Perforce\my-project\3rdParty\Libusb\OriginalSource\libusb-1.0.19\msvc\..\Win32\Debug\lib\getopt.lib
[...]

PR #134, opened 2015-12-29
Support for Visual Studio 2015 & Windows 10 SDK

johannrichard:
This patch adds a Solution for VisualStudio 2015 including a fix needed
in the preprocessor definitions (`HAVE_STRUCT_TIMESPEC`). This should fix Issue https://github.com/libusb/libusb/issues/63. 

Furthermore, the references in the Solution are those for the latest Windows 10 SDK
(V14.0 instead of v12.0 of the Windows SDK).

With these changes (and nothing else), a successful build of libusb under Windows was possible.


johannrichard:
I just realise that #120, #83 and #60 handle this, although from different angles. Since I'm probably making a fool of myself by asking for yet another different change, I prefer to close the pull request.

PR #138, opened 2016-01-07
io: don't submit transfer if canceled asynchronous

FounderSG:
if a transfer been canceled by another thread just before submit, this transfer is still submitted, `libusb_cancel_transfer` need to be called by application code again. with this patch, application code need to call `libusb_cancel_transfer` only once without worry about race condition.

Signed-off-by: Founder Fang <founder.fang@gmail.com>

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -1470,6 +1470,7 @@ int API_EXPORTED libusb_submit_transfer(struct libusb_transfer *transfer)
 	struct usbi_transfer *itransfer =
 		LIBUSB_TRANSFER_TO_USBI_TRANSFER(transfer);
 	int remove = 0;
+	int canceled = 0;
 	int r;
 
 	usbi_dbg("transfer %p", transfer);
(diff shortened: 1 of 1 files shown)


dickens:
NAK. A transfer that has not been submitted yet cannot by definition be
cancelled and in my opinion should not cause a subsequent submission of
that same transfer to be aborted. I see this is an application issue.

FounderSG:
in callback function 
{
    //process data received.
    ....
    // re-submit transfer
}
what will happen if another thread call libusb_cancel_transfer just before re-submit transfer?

dickens:
That's the application's responsibility. If transfers may be cancelled asynchronously by other threads, the application should have a synchronization mechanism around the transfer to instruct whether or not to resubmit.

Callback function:
`{
    ...
    mutex_lock()
    if (!cancelled)
        libusb_submit_transfer(...)
    mutex_unlock()
    ...
}`

FounderSG:
the purpose of this patch is to simplify application code. 
of course application code can handle this situation。

dickens:
As the application code rightfully should.

From the docs:

libusb_transfer_cancel() returns:

    0 on success 
    LIBUSB_ERROR_NOT_FOUND if the transfer is *not in progress*, already complete, or already cancelled.

Your example falls into the "not in progress" category. This patch changes the behavior of the function and, as I stated, pushes application responsibility into the library.

FounderSG:
let me show you a real race condition example, assume use winusb backend.
below code was from examples/dpfp_threaded.c, if usb data coming just before call libusb_cancel_transfer, because polling thread already stopped, libusb internal flags remains unchanged, libusb_cancel_transfer return success because WinUsb_AbortPipe won't complain, then when the program runs to while loop, it will loop forever.
    r = libusb_cancel_transfer(img_transfer);
    if (r < 0) {
        request_exit(1);
        goto out_deinit;
    }
    while (img_transfer || irq_transfer)
        if (libusb_handle_events(NULL) < 0)
            break;

i believe dpfp_threaded.c was written carefully and reviewed by a lot of people, but it sill have this kind of problem.
if the library can help people avoid common mistake, why not?

PR #143, opened 2016-01-19
Android as a separate OS, plus a test_app for easy testing

hutorny:
- Refactoring build for separating Android platform as first-class OS
- Add android/autogen.sh for proper auto-configuration of cross build
- Remove android/config.h in favour of autogenerated ./config.h
- Changed build scripts to target latest available platform by default
- Added APP_PIE flag to build position-independent executables for Android
- Add Android test_app for easy deploying and running tests and examples


hjelmn:
I don't see why autogen.sh has anything to do with CMake. We don't currently plan to support CMake.

Anyway, for this to even be considered it needs to be rebased off the current master. Please no 'merge' commits.

mcuee:
There may be something useful in this pull request but it needs work to remove irrelevant parts.

dickens:
Closing this as there are commits not relevant to libusb and it no longer merges cleanly.

PR #145, opened 2016-01-28
Windows: Fix Cygwin build.

hiyuh:
SSIA.


dickens:
Thanks for the PR. I only picked commit f6901d6 as I had already addressed the other issues.

PR #155, opened 2016-02-25
bootstrap: use autoreconf

hjelmn:
This PR cleans up the bootstrap script to use autoreconf. It has the same effect as the code it replaces.

Signed-off-by: Nathan Hjelm <hjelmn@me.com>


hjelmn:
Pushed manually.

PR #174, opened 2016-03-17
Replaced __android_log_write by __android_log_print

aalmada:
The use of __android_log_write causes problems in execution time.


aalmada:
Not able to reproduce issue. Closing until further notice...

PR #181, opened 2016-04-07
libusb_set_log_handler() API proposal to re-direct all logs to a custom log handling function

dmitrykos:
Hi! The proposed API allows to re-direct libusb log output to a custom log handling function. It is useful if project is using own logging mechanism which also includes file logging and thus such re-direction allows to intercept libusb logs easily.


dickens:
This is an interesting proposal, but once again there are many, many commits unrelated to the PR. Please open another containing only relevant commits.

PR #183, opened 2016-05-02
[winusb] faulty devices can cause priv->config_descriptor[index] being NULL

H2OBrain:
This fixed an access violation exception I just experienced


dickens:
After reviewing the patch, I'm not sure how you could end up in this scenario. When cache_config_descriptors() is called, any failure will set the "num_configurations" field to 0 [1]. Your patch adds a NULL check inside a loop in which the condition is checking against this same "num_configurations" field. So I'm not seeing how you could possibly end up in this loop if cache_config_descriptors() actually failed.

[1] https://github.com/H2OBrain/libusb/blob/4a12d6a1316a659550509fbcfe94ad3ec5efa934/libusb/os/windows_winusb.c#L1084-L1088

H2OBrain:
Well it happened to me in a VM with Windows 7 32-bit running, my device was faulty and crashed during configuration, so this fix is for a very special case which should never occur in real applications.
Stepping through the code, numConfigurations was 1 and the configuration was NULL.
Maybe cache_config_descriptors does not fail all the time.

I can also apply this fix each time I compile a new version of libusb, so it doesn't has to go into upstream. I only wanted to document (mention) this somewhere..

dvj:
I have the same issue and was going to propose a similar patch. Granted, the device is faulty, but shouldn't cause access violations in this case.

PR #184, opened 2016-05-10
fix timerfd initialization

blogic:
sys/timerfd.h defines TFD_NONBLOCK as 0x800 but in kernel TFD_NONBLOCK
is an alias for O_NONBLOCK which is defined in arch-specific fcntl.h.
While it's still 0x800 for most of archs but for mips it's 0x80. So
timerfd_create(..., TFD_NONBLOCK) returns -EINVAL because of that. Fix
this by using O_NONBLOCK instead.

Signed-off-by: Alexander Gordeev <lasaine@lvk.cs.msu.su>

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -34,6 +34,7 @@
 #include <sys/time.h>
 #endif
 #ifdef USBI_TIMERFD_AVAILABLE
+#include <fcntl.h>
 #include <sys/timerfd.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


blogic:
This patch has been lingering inside openwrt fro 5+ years. Sorry for not sending it upstream earlier.

hjelmn:
How is this not a kernel bug? The man page for timerfd_create doesn't not list O_NONBLOCK as a valid flag.

blogic:
ok, this was a bug in the libc which has been fixed at some point, sorry for the noise

PR #190, opened 2016-05-30
Redirect all log messages to a custom log handling function.

dmitrykos:
A proposal to redirect log messages to a user-side log handling function. It is useful for the debugging when libusb log output can be redirected to an application central log handler which then redirects all app logs to a file and etc.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -431,6 +431,7 @@ if (cfg != desired)
   * - libusb_set_auto_detach_kernel_driver()
   * - libusb_set_configuration()
   * - libusb_set_debug()
+  * - libusb_set_log_handler()
   * - libusb_set_interface_alt_setting()
   * - libusb_set_iso_packet_lengths()
   * - libusb_setlocale()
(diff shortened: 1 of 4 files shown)


jwrdegoede:
Seems like a sensible API addition to me and the implementation looks good to. hjelmn, if you agree feel free to merge this one.

dmitrykos:
Unfortunately my proposed enhancement has deficiency - it is impossible to redirect logs from libusb_init() call. If user wants to see a reason of failure of libusb_init() then such logs will not be redirected to a custom logging function until libusb_set_log_handler() is made.

My new proposal is to modify libusb_set_log_handler() and remove dependency on libusb_context. libusb_log_handler_cb() would be static file-global function then and thus user will be able to set log function before libusb_init() that will allow to catch possible error logs with libusb_log_handler_cb().

Please comment if I should update the related code to the above proposed global logging solution instead of per context logging.

Youw:
As another user, that requires such functionality, I'd say, that per context logging is more useful.

Regarding libusb_init() - we can always use libusb_strerror() to get failure reason for logging purpose in this particular case.

I'd vote to merge it to master as is as soon as possible.

hjelmn:
@jwrdegoede I agree. Looks good. @mcuee Do we want to bring this in for 1.0.23?

hjelmn:
This has gotten a little stale (sorry for the delay). Can the author update this so it merges properly?

dmitrykos:
@hjelmn, in my implementation I have modified this approach to avoid using the ```libusb_context``` in order to be able to catch log messages before ```libusb_init``` function is called to be able to catch important log messages when initialization of libusb fails (was important to catch it in case of Android OS).

So my usage scenario is a follows:
```c
int r;

libusb_set_log_cb(LogHandler);

if ((r = libusb_init(&context)) < 0)
{
    USBErrLog("libusb_init() failed: error[%d]", r);
    return false;
}

...
```

Should I update implementation to this approach, because it is much more useful? The implementation works for around 2 years already without any issue.

dmitrykos:
@hjelmn, created new pull request here: http://github.com/libusb/libusb/pull/486

PR #228, opened 2016-11-08
add testlibusb solutions and fix testlibusb.c can't compile by snprin…

Tody-Guo:
…tf()


mcuee:
This is not a valid fix.

Issue #233, opened 2016-11-22
libusb is not working with Android 7.1


kennethtang4:
To start with, it is not a problem of libusb itself but more to the Android system but I am looking if any one have an idea on how to fix this problem.

The library is working perfectly up to Android 6.0 (haven't tested with 6.1 and 7.0 yet).

The problem is happening in the function "sysfs_get_device_list" in the file "linux_usbfs.c". I am getting an errno code of 13 which is permission denied. It seems that Android has changed the permission of accessing the directory /sys/bus/usb/devices. USB Permission has been asked and granted through the USB Manager beforehand (needed for previous Android versions as well) and I tried to request access for READ_EXTERNAL_STORAGE too but this has no effect on it.

So I wonder if any one can come with a workaround for this?

mcuee:
Please refer to this one as well. https://github.com/libusb/libusb/issues/188

mcuee:
You may want to try Pull Request #242.

kennethtang4:
#242 can solve the problem. The naming of the function is confusing though. The function libusb_wrap_fd is escentially opening the usb by file descriptor. Also, it will not add the usb device the to device list in libusb, so mind that some libraries depend on libusb may need to be modified to make it working.

mcuee:
I will close this for now. But the comment is good. I will add this link to the #242 .

PR #242, opened 2016-12-13
Add Android 5-7 support

vianney:
This patch series adds support for recent Android devices (tested up to 7.1.1).  On such devices, SELinux policies prevent the use of sysfs and reading /dev/bus/usb.  The only way to use libusb as a regular user is to get a file descriptor through the Android Platform API and pass it on.  All operations must use that file descriptor.

This series adds an API function to libusb to wrap an open file descriptor as a libusb_device_handle.  To ease reviewing, it is decomposed as follows:

* Patches 1-3 contain light refactoring and small utility extensions needed later on.
* Patch 4 adds the public API function.
* Patch 5 implements the new API for the Linux platform.
* Patch 6-7 ensure the libusb library is able to initialize on Android.
* Patch 8 is a minor warning fix as a bonus.

Please note that patch 7 (disable hotplug events on Android) means that libusb cannot be used anymore standalone on a rooted Android device.  Is that an issue?  If so, I guess the way to go is to make it a compile-time option.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1216,6 +1216,78 @@ int usbi_clear_event(struct libusb_context *ctx)
 }
 
 /** \ingroup libusb_dev
+ * Wrap a platform-specific system device handle and obtain a libusb device
+ * handle for the underlying device. The handle allows you to use libusb to
+ * perform I/O on the device in question.
+ *
+ * On Linux, the system device handle must be a valid file descriptor opened
+ * on the device node.
+ *
+ * The system device handle must remain open until libusb_close() is called.
+ * The system device handle will not be closed by libusb_close().
+ *
+ * Internally, this function creates a temporary device and makes it
+ * available to you through libusb_get_device(). This device is destroyed
+ * during libusb_close(). The device shall not be opened through libusb_open().
+ *
+ * This is a non-blocking function; no requests are sent over the bus.
+ *
+ * \param ctx the context to operate on, or NULL for the default context
+ * \param sys_dev the platform-specific system device handle
+ * \param dev_handle output location for the returned device handle pointer. Only
+ * populated when the return code is 0.
+ * \returns 0 on success
+ * \returns LIBUSB_ERROR_NO_MEM on memory allocation failure
+ * \returns LIBUSB_ERROR_ACCESS if the user has insufficient permissions
+ * \returns LIBUSB_ERROR_NOT_SUPPORTED if the operation is not supported on this
+ * platform
+ * \returns another LIBUSB_ERROR code on other failure
+ */
+int API_EXPORTED libusb_wrap_sys_device(libusb_context *ctx, intptr_t sys_dev,
+	libusb_device_handle **dev_handle)
+{
+	struct libusb_device_handle *_dev_handle;
+	size_t priv_size = usbi_backend.device_handle_priv_size;
+	int r;
+	usbi_dbg("wrap_sys_device %p", sys_dev);
+
+	USBI_GET_CONTEXT(ctx);
+
+	if (!usbi_backend.wrap_sys_device)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	_dev_handle = malloc(sizeof(*_dev_handle) + priv_size);
+	if (!_dev_handle)
+		return LIBUSB_ERROR_NO_MEM;
+
+	r = usbi_mutex_init(&_dev_handle->lock);
+	if (r) {
+		free(_dev_handle);
+		return LIBUSB_ERROR_OTHER;
+	}
+
+	_dev_handle->dev = NULL;
+	_dev_handle->auto_detach_kernel_driver = 0;
+	_dev_handle->claimed_interfaces = 0;
+	memset(&_dev_handle->os_priv, 0, priv_size);
+
+	r = usbi_backend.wrap_sys_device(ctx, _dev_handle, sys_dev);
+	if (r < 0) {
+		usbi_dbg("wrap_sys_device %p returns %d", sys_dev, r);
+		usbi_mutex_destroy(&_dev_handle->lock);
+		free(_dev_handle);
+		return r;
+	}
+
+	usbi_mutex_lock(&ctx->open_devs_lock);
+	list_add(&_dev_handle->list, &ctx->open_devs);
+	usbi_mutex_unlock(&ctx->open_devs_lock);
+	*dev_handle = _dev_handle;
+
+	return 0;
+}
+
+/** \ingroup libusb_dev
  * Open a device and obtain a device handle. A handle allows you to perform
  * I/O on the device in question.
  *
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1387,6 +1387,7 @@ int LIBUSB_CALL libusb_get_max_packet_size(libusb_device *dev,
 int LIBUSB_CALL libusb_get_max_iso_packet_size(libusb_device *dev,
 	unsigned char endpoint);
 
+int LIBUSB_CALL libusb_wrap_sys_device(libusb_context *ctx, intptr_t sys_dev, libusb_device_handle **dev_handle);
 int LIBUSB_CALL libusb_open(libusb_device *dev, libusb_device_handle **dev_handle);
 void LIBUSB_CALL libusb_close(libusb_device_handle *dev_handle);
 libusb_device * LIBUSB_CALL libusb_get_device(libusb_device_handle *dev_handle);
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -702,6 +702,34 @@ struct usbi_os_backend {
 	 */
 	void (*hotplug_poll)(void);
 
+	/* Wrap a platform-specific device handle for I/O and other USB
+	 * operations. The device handle is preallocated for you.
+	 *
+	 * Your backend should allocate any internal resources required for I/O
+	 * and other operations so that those operations can happen (hopefully)
+	 * without hiccup. This is also a good place to inform libusb that it
+	 * should monitor certain file descriptors related to this device -
+	 * see the usbi_add_pollfd() function.
+	 *
+	 * Your backend should also initialize the device structure
+	 * (dev_handle->dev), which is NULL at the beginning of the call.
+	 *
+	 * This function should not generate any bus I/O and should not block.
+	 *
+	 * This function is called when the user attempts to wrap an existing
+	 * platform-specific device handle for a device.
+	 *
+	 * Return:
+	 * - 0 on success
+	 * - LIBUSB_ERROR_ACCESS if the user has insufficient permissions
+	 * - another LIBUSB_ERROR code on other failure
+	 *
+	 * Do not worry about freeing the handle on failed open, the upper layers
+	 * do this for you.
+	 */
+	int (*wrap_sys_device)(struct libusb_context *ctx,
+		struct libusb_device_handle *dev_handle, intptr_t sys_dev);
+
 	/* Open a device for I/O and other USB operations. The device handle
 	 * is preallocated for you, you can retrieve the device in question
 	 * through handle->dev.
--- a/libusb/os/linux_udev.c
+++ b/libusb/os/linux_udev.c
@@ -232,7 +232,7 @@ static int udev_device_info(struct libusb_context *ctx, int detached,
 	}
 
 	return linux_get_device_address(ctx, detached, busnum, devaddr,
-					dev_node, *sys_name);
+					dev_node, *sys_name, -1);
 }
 
 static void udev_hotplug_event(struct udev_device* udev_dev)
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -534,8 +534,10 @@ static int linux_start_event_monitor(void)
 {
 #if defined(USE_UDEV)
 	return linux_udev_start_event_monitor();
-#else
+#elif !defined(__ANDROID__)
 	return linux_netlink_start_event_monitor();
+#else
+	return LIBUSB_SUCCESS;
 #endif
 }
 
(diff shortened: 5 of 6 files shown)


mcuee:
Please refer to the comments at https://github.com/libusb/libusb/issues/233

>#242 can solve the problem. The naming of the function is confusing though. The function libusb_wrap_fd is escentially opening the usb by file descriptor. Also, it will not add the usb device the to device list in libusb, so mind that some libraries depend on libusb may need to be modified to make it working.

vianney:
@mcuee @kennethtang4 Thanks for the feedback.  I am willing to improve this patchset to get it merged upstream.  So let's get the discussion started.

Concerning the naming, I have chosen "wrap" over "open" to make it clear that we do not open a new file descriptor (and hence, that the provided file descriptor must not be closed), but I would not mind naming it "libusb_open_fd" or anything else.  Suggestions are welcome.

I also have the concern that file descriptors are backend-specific (event though most OS have such concept).  Having them exposed in the user API is not very nice, but I am afraid there is no way around it.  Maybe we should replace the int argument by something more platform-agnostic, but what?

About the device list: yes, this feels a bit hacky.  I would appreciate some input from libusb developers on the best way to handle this.  The requirement for Android is that we should not rely on anything besides the provided file descriptor.  This means that in the worst case, if the link in /proc/self/fd is unreadable or non-standard, we do not know the bus number of the device (afaik, there is no ioctl for that).  Maybe we could do the following:
* If we got the complete device address: reuse existing libusb_device if it was enumerated.  Otherwise create a new one, add it to the list, and remove it when the handle is closed.
* If the bus number is unknown: create a new libusb_device using the file descriptor as session id, add it to the list, and remove it when the handle is closed.

To libusb maintainers: is there any interest in merging this kind of patches?  If so, should we keep compatibility with using libusb in the standard way on a rooted Android device (i.e., with hotplug enumeration and libusb_open)?

mcuee:
@vianney It will be helpful that you can subscribe to libusb mailing list and post there to seek comments. Thanks.

Ulez:
Do you have solved this problem?I have used your code(commit on 14 Dec 2016 ) to build SO files and use in my application,however ,The same problem occurs.i use in this way : ret =libusb_wrap_fd (NULL,g_dev_fd, &g_dev_handle),Did i use it wrong?

reflog:
@vianney Thanks for the patch! If I'm calling libusb_wrap_fd instead of libusb_open, does it mean I shouldn't call libusb_close?

barak353:
Hey, i want to compile the libusb folder with the new pathces into .so file.
How can i do this? I tried using cygwin and the ndk-build, but it failled, then i tried to do this with android studio but i got to many error messages in this files.
Can you please include the Makefile that you have used, or explain how you compile the libusb folder into .so file?

chipweinberger:
Could someone explain how this should be used? What to call and in what order?

dirkhh:
This looks like it might be what I'm looking for in order to be able to connect to USB dive computers from Android... but just like @chipweinberger I am wondering how to use these patches in order to be able to open an FTDI device from Android... Help would be appreciated.

lenaschimmel:
I'm also currently trying to use these patches, and `libusb_wrap_fd` seems to work, basically. However, I'm unsure what else is supposed to work. 

I'm trying to use existing code ([libmaxtouch](https://github.com/atmel-maxtouch/mxt-app/tree/master/src/libmaxtouch)) that relies on the contents of the `libusb_device_descriptor` struct. It seems to me as if the common way to get that descriptor would be to use `libusb_get_device_list` and then `libusb_get_device_descriptor` and I do not know if this just fails for me and my devices, or if this is not supported at all on Android.

I tried the [`listdevs.c` example](https://github.com/libusb/libusb/blob/master/examples/listdevs.c) - embedded into an Android App via the NDK - and I get 0 devices. I tried this on Android 7.0 on x86 (a 64 bit system, but with 32 bit binaries) and Android 5.0 on armeabi-v7a. In both cases a USB device was attached and my app had the needed permission and already acquired the file descriptor from the Java API. On both devices, libusb reports: `found usbfs at /dev/bus/usb`, but I understand that `find_usbfs_path` will always print this message when it actually **cant'** find any usbfs. So I checked for myself:

* On the Android 7.0 system and the device of interest is actually present at `/dev/bus/usb/001/004` as I can confirm using `adb shell`. The device is rooted.
 * On the 5.0 system, I cannot even `ls /dev/bus/usb` via `adb shell` because of unsufficient permissions. The device is not rooted.

If it was - at least on some devices - really impossible to get the `libusb_device_descriptor` from C, I think it should be possible to get most of that from the [Java API](https://developer.android.com/guide/topics/connectivity/usb/host.html) and somehow pass it back to libusb, however that seems overly complicated.

NickAtAccuPS:
@reflog If I'm reading the comment in [this commit](https://github.com/libusb/libusb/pull/242/commits/98d80100f48bbf6706980659c2456979e2752aff) you should call libusb_close to close the libusb object, but you would still need to manually close the OS file descriptor manually afterward.

lenaschimmel:
@NickAtAccuPS On a first glimpse, that seems very useful. The android documentation says nothing about its endianness, but that should be an easy trial-end-error process. Also, it's unclear to me _which_ descriptor(s) is/are returned. If it's just the 18 byte device descriptor, that's not entirely useful, because there seems to be no API to get the other descriptors (Configuration Descriptors, Interface Descriptors, Endpoint Descriptors, String Descriptors), but I guess this will return all the descriptors concatenated, just as if we read them from the usbfs? (Sorry, didn't have time to just try it on Android).

If so, I think we could just add a few more parameters to `libusb_wrap_fd` so that we get:

	int LIBUSB_CALL libusb_wrap_fd (libusb_context *ctx,
					int fd,
					libusb_device_handle **dev_handle,
					uint8_t bus_number,
					uint8_t port_number,
					char* raw_descriptor,
					int desc_length);

Am I missing anything?

vianney:
Guys, did you read the documentation? Let me quote the doc for `libusb_wrap_fd()`:

> Wrap an open file descriptor and obtain a device handle for the underlying
> device. A handle allows you to perform I/O on the device in question.
>
> The file descriptor must remain open until `libusb_close()` is called.
> The file descriptor will not be closed by `libusb_close()`.
>
> Internally, this function creates a temporary device and makes it
> available to you through `libusb_get_device()`. This device is destroyed
> during `libusb_close()`. The device shall not be opened through `libusb_open()`.

and what I wrote in the description of this pull request:

> This patch series adds support for recent Android devices (tested up to 7.1.1). On such devices, SELinux policies prevent the use of sysfs and reading /dev/bus/usb. The only way to use libusb as a regular user is to get a file descriptor through the Android Platform API and pass it on. All operations must use that file descriptor.

It follows that:

1. You have to use the Java Android API to open a file descriptor to the USB device before you can use libusb.
2. As a *regular* user, you cannot use libusb's enumeration functions to get a `libusb_device` because you cannot read sysfs and /dev/bus/usb.
3. After obtaining a `libusb_device_handle`, you can call `libusb_get_device()` to get a `libusb_device` which can be used to get all descriptors you want. No need for additional trickery with the Android API.
4. You must of course properly clean up the `libusb_device_handle` by calling `libusb_close()` when you are done.

To maintainers: when I get some time, I will rebase this patch onto master and post to the mailing list. Unfortunately, other projects keep me quite busy right now.

lenaschimmel:
@vianney Thank you very much! I was (mostly) aware of your points 1, 2 and 4 but I completely missed 3 (probably because the existing code I worked with assumed to have access to the descriptors before opening the device). A quick test on my device confirms that I understood it correctly this time.

akshaytaru2007:
Hi All,
First it is not issue, I am asking for help. I am trying to connect a Hardware(IoT based) via USB to Android. USB Manager not detecting the devices and not listing the device. The Android is non-rooted, 6.1 It does support OTG.
As all of you successfully used libusb in Android, I am still struggling as there is no enough documentation or support.
I have built the .SO file using NDK and added those under jniLibs/AIBs/libusb1.0.SO

Now in my native-lib.cpp I am trying to use libusb's functions to init or get list of device. But I am failing.
I belive there must be some configuration needs to be done.
Those who have succeeded kindly help.

I have also posted Question on stackoverflow related to this: https://stackoverflow.com/q/44281689/1468354

If you need any more info. I will provide

akshaytaru2007:
Hi All,
First it is not issue, I am asking for help. I am trying to connect a Hardware(IoT based) via USB to Android. USB Manager not detecting the devices and not listing the device. The Android is non-rooted, 6.1 It does support OTG.
As all of you successfully used libusb in Android, I am still struggling as there is no enough documentation or support.
I have built the .SO file using NDK and added those under jniLibs/AIBs/libusb1.0.SO

Now in my native-lib.cpp I am trying to use libusb's functions to init or get list of device. But I am failing.
I belive there must be some configuration needs to be done.
Those who have succeeded kindly help.

I have also posted Question on stackoverflow related to this: https://stackoverflow.com/q/44281689/1468354

If you need any more info. I will provide

lenaschimmel:
@akshaytaru2007 As stated in [this readme](https://github.com/libusb/libusb/tree/master/android), the usual `libusb` will not work on you unrootet device and you must use _"`android.hardware.usb.UsbManager` to request permission to use the UsbDevice and then opening the device"_. 

Also, you need a patched version of `libusb` to actually handle the device descriptor. This PR that you are commenting on contains the necessary changes, but in your [StackOverflow question](https://stackoverflow.com/questions/44281689/use-libusb1-0-so-functions-inside-native-library) you did not mention that you are basing your work on this PR. So are you on the `master` branch or on this PR? 

Regarding the list of devices, I think it is not possible to use `libusb_get_device_list` on Android 6.1 without root, no matter if you use the code of this PR or not. Instead, you have to use [Android's Java API for USB Host](https://developer.android.com/guide/topics/connectivity/usb/host.html). In your Java code, you can list the devices, request permission, open the device and then give the file descriptor to your native library, which can then call `libusb_wrap_fd`.

@all Maybe I can setup a sample Android project somewhere next week and post a link. This should make it easier for now users to get started.

@vianney I respect that you don't have much time right now. I just want to say that it seems that maybe rebasing might be a matter of 5 minutes work - I just had to fix two trivial merge conflicts. Anyway, I'm looking forward to the day when this goes upstream!

akshaytaru2007:
@lenaschimmel I tried using `android.hardware.usb.UsbManager` to give the list of device but it is giving me empty list though the device is plugged in via USB. So I can not request permission to _UsbDevice_
I am using libusb-1.0.20 version and the code is not part of this PR.
I do have more info about Hardware. The Hardware has the OS Windows CE 7.0.  It has ActiveSync USB device class selected. 

@lenaschimmel As you guys know much about USB communication, I am asking your assistance that is there any way I can talk to this device
PS. I can not disclose the Device details as it's client's hardware

lenaschimmel:
@akshaytaru2007 I'm very new to USB communication and `libusb` myself, so I can't really help that much, but I'll try...

I think if you get an empty list from `android.hardware.usb.UsbManager` there's some problem either with you Android / Java code, or your hardware / Android OS version does not support USB host functionality like it should. Anyway, that has nothing to do with `libusb` and should IMHO be discussed elsewhere, for example by posting another question on StackOverflow.

Once you get the `fileDescriptor` from the `UsbDeviceConnection`, it's time to try `libusb` again. Also, I think you will not be successful if you use the standard version of `libusb-1.0.20`, you will need to compile it from this branch / PR. (Exception: if some time passes, and this PR gets merged into `master` by then, you can of course just grab the newest version of `libusb`.)

FDmitnick:
I modification bu your code ,now can use the libusb in android 7.0,THX ;
but ,by the test in android 4.0 and 5.0 not can ok .

ao2:
> Guys, did you read the documentation? Let me quote the doc for libusb_wrap_fd():
> 
>     Wrap an open file descriptor and obtain a device handle for the underlying
>     device. A handle allows you to perform I/O on the device in question.
> 
>     The file descriptor must remain open until libusb_close() is called.
>     The file descriptor will not be closed by libusb_close().
> 
>     Internally, this function creates a temporary device and makes it
>     available to you through libusb_get_device(). This device is destroyed
>     during libusb_close(). The device shall not be opened through libusb_open().

Hi,

since the function basically replaces `libusb_open`on Android I guess it can use a different name to highlight that.

Maybe following the style of `libusb_open_device_with_vid_pid` it may be called `libusb_open_from_fd`, this should make it clearer that `libusb_close` can and should be called on the returned handle.

The **open** in libusb functions usually refers to the handle, so I would expect users to understand that it does not refer to the file descriptor (which must be opened ad closed by the user code).

BTW I think you should really write to the libusb mailing list to get more attention from core developers.

Thanks for your code.

Ciao,
   Antonio

hjelmn:
My only objection to this is the name of the new function and the argument type. The only question is if we need something similar on a future system when it is neither a file descriptor nor an event. It's very unlikely we will run into that situation. If everyone else is happy I say we merge this for the next release.

NielsHygum:
This is awesome! Thanks for adding Android support, works super on my S9 with Oreo 8.0.

dirkhh:
Indeed, that is outstanding news. We have quite a few users who would benefit from this.

hjelmn:
Any update on this? I would like to try and get this into v1.0.23 since were are already add APIs. What I need:

1) A clean set of patches (squash this down to the minimum number). I would expect 2-3 commits will be sufficient.
2) Email the patch set to the libusb-devel mailing list to get some feedback. I would give ~ 1 week for feedback from other developers. Those that are active should respond fairly quickly if they have concerns. I will take those I know who watch github to look once this is updated.

ao2:
IMHO the current commits separation is quite sensible, it makes review easier than having unrelated changes squashed together.

hjelmn:
@ao2 Alright. It looks like things have been cleaned up so that it can be properly reviewed.

hjelmn:
Looked through this just now and I keep coming back to one thing: should the API be more generally named and take something like intptr_t instead of int. This would allow someone to get a device from their system on any platform and pass it to libusb. It will by definition be backend specific (IOUSBDevice on macOS for example).

Maybe:

```c
int API_EXPORTED libusb_wrap_sys_device (libusb_context *ctx, intptr_t sys_dev, libusb_device_handle **dev_handle);
```

The Linux backend would continue to use int and I could add something to the darwin backend that gives similar functionality without more APIs.

Any thoughts?

hjelmn:
Could even make the old call be a helper:

```c
static inline int libusb_wrap_fd (libusb_context *ctx, int fd, libusb_device_handle **dev_handle) {
    return libusb_wrap_sys_device (ctx, (intptr_t) fd, dev_handle);
}
```

hjelmn:
@vianney It will once I approve it. I will keep an eye out for the email.

hjelmn:
@vianney Approved it on the mailing list. Will see what (if anything) other people have to say.

PR #243, opened 2016-12-13
darwin: Replace bzero() with memset()

vianney:
The bzero() function is a deprecated BSD function that can be replaced
by the standard C89 memset() function.  When compiling for POSIX 2008,
the bzero() function is undefined on the OS X platform.


LudovicRousseau:
It is not`` bcopy()`` but ``bzero()``.
Maybe you should update your patch.

vianney:
Thanks Ludovic.  I wonder how that got through my mind 😄.

LudovicRousseau:
Fixed in f3a9060d19cac341581f4a4e1877054dbb606a00

Thanks

PR #245, opened 2016-12-28
Windows: cache_config_descriptors doesn't return error code

dvj:
currently always returning `LIBUSB_SUCCESS` even if error was found

this can cause all sorts of trouble, including the root cause behind https://github.com/libusb/libusb/pull/183

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -981,7 +981,7 @@ static int cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handle
 			LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
 		memcpy(priv->config_descriptor[i], cd_data, cd_data->wTotalLength);
 	}
-	return LIBUSB_SUCCESS;
+	return r;
 }
 
 /*


mcuee:
I am not that sure why this was implemented in such a way, it could be due to save some faulty device as well under Windows. I could be wrong though.

From Pete: `And the way we cache the descriptors on Windows is by issuing a formal
GET_DESCRIPTOR request (through the Windows ioctl
IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION where we fill in the
low byte of the wValue with the index, exactly as documented in section
9.4.3 of the USB 3.0 specs.`

This could fail for quite a number of device as far as I know.

dvj:
It seems like the proper thing to do is return the error code, and then other functions can take appropriate action. As it is currently, there is no way for user code to detect this sort of error, which ultimately can cause segfaults, as seen in #183. 

Even if the current behavior does allow some class of faulty devices to work, it should be a decision higher up in the stack to ignore the error code rather than not return it at all at this level.

dickens:
Closed via 92093a59

Issue #46, opened 2014-12-18
Windows: Add isochronous support for libusbK and WinUSB


piedar:
@JoshBlake proposed [this patch](https://github.com/libusbx/libusbx/pull/164) to libusbx, but it was never merged.  A cherry-pick still applies cleanly to libusb HEAD.

```
git clone https://github.com/libusb/libusb.git
cd libusb
git remote add jblake https://github.com/JoshBlake/libusbx.git
git fetch jblake
git cherry-pick c5b0af4 1c74211
```

I tested the patch and it appears to work with libusbK v3.0.7.0.  Any chance this can be merged into libusb?  It would make libfreenect a lot easier to build on Windows by replacing our emulation layer that is woefully out of date.

QianJin2013:
from Windows 8.1, WinUSB also support isochronous mode. Hope this be implemented in future.

mcuee:
It will be good to add WinUSB isoc support (directly to libusb WIndows backend or through libusbK). Travis mentioned he intended to add WinUSB isoc support to libusbK recently but I am not so sure when he will start that work.

mcuee:
Another option is to look at usbdk backend which has not yet been integrated.

https://github.com/dmitryfleytman/libusb/tree/usbdk-backend-v3

mkassner:
https://github.com/ktossell/libuvc would finally work on windows if isochronous transfers where to be implemented with winusb8.1: https://github.com/ktossell/libuvc/issues/12

Any chance for this to happen in the near future?

xlz:
@mcuee 
What is the plan now?
Is the usbdk backend favored over the libusbk patch? WinUSB? What about Windows 7 users?

mcuee:
Not sure when this support will be added to Windows backend which is now
lagging behind Linux and Mac OS X (isoc support and hotplug support). It
depends on the time and priority of the main Windows developer now.

BTW, usbdk works under Windows 7.

diabolo38:
with new Microsoft policy on driver signing i'm scared no new libusbk release will ever happen.
So usbdk and winusb  seam the only possible candidate to enhance support on windows.

I will not be able to participate actively on development but i can participate and provide some punctual help with testing on win 7  (uvc cam,  gadget 0 target) and bus level debug if needed.

xlz:
[...]

PR #246, opened 2016-12-29
add isochronous transfer per Josh Blake

cuisinart:
support of isochronous transfer is required by libfreenect2 to support xbox kinect and other interactive devices. This is tested under windows 10, built with msvc 2015, working.


mcuee:
This may need to be better documented.
Firstly refer to this original link: https://github.com/JoshBlake/libusbx/tree/winiso
Secondly you may want to mention that this is supported on the libusbK driver only.

cuisinart:
OK, sounds fine with me.  Is there a way to edit the comments? Is that something you can do when pulling?

xlz:
UsbDk seems preferred now, while the native WinUSB is also getting improved. FYI: https://github.com/libusb/libusb/issues/46#issuecomment-149013612

xlz:
I rebased the patches on latest libusb master here: https://github.com/xlz/libusb/tree/winiso.

Issue #144, opened 2016-01-24
 error C2011: 'timespec': 'struct' type redefinition (Windows 10 issue)


j796160836:
Hi, I got some error during install `bleno` and I found that might be some issue about installing libusb module.
https://github.com/sandeepmistry/bleno

```
lenovo-PC:bleno$ npm install usb

> usb@1.1.1 install C:\Users\lenovo\Documents\node\bleno\node_modules\usb
> node-pre-gyp install --fallback-to-build

Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
  core.c
  descriptor.c
  hotplug.c
  io.c
C:\Users\lenovo\Documents\node\bleno\node_modules\usb\libusb\libusb\os/threads_windows.h(47): error C2011: 'timespec':
'struct' type redefinition (compiling source file ..\libusb\libusb\descriptor.c) [C:\Users\lenovo\Documents\node\bleno\
node_modules\usb\build\libusb.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\time.h(39): note: see declaration of 'timespec' (com
  piling source file ..\libusb\libusb\descriptor.c)
C:\Users\lenovo\Documents\node\bleno\node_modules\usb\libusb\libusb\os/threads_windows.h(47): error C2011: 'timespec':
'struct' type redefinition (compiling source file ..\libusb\libusb\io.c) [C:\Users\lenovo\Documents\node\bleno\node_mod
ules\usb\build\libusb.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\time.h(39): note: see declaration of 'timespec' (com
  piling source file ..\libusb\libusb\io.c)
..\libusb\libusb\io.c(1227): warning C4244: '=': conversion from 'time_t' to 'long', possible loss of data [C:\Users\le
novo\Documents\node\bleno\node_modules\usb\build\libusb.vcxproj]
C:\Users\lenovo\Documents\node\bleno\node_modules\usb\libusb\libusb\os/threads_windows.h(47): error C2011: 'timespec':
'struct' type redefinition (compiling source file ..\libusb\libusb\hotplug.c)strerror.c [C:\Users\lenovo\Documents\node
\bleno\node_modules\usb\build\libusb.vcxproj]

  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\time.h(39): note: see declaration of 'timespec' (com
[...]

PR #248, opened 2016-12-30
Fix issue #144

SkyLined:
Issue was claimed to be fixed, but still happens with Windows SDK 10.0.10240.0 and latest usb 1.2.0 (latest on npm).

The first definition is in time.h in the Windows SDK, which also does a #DEFINE _INC_TIME. Checking for this define allows us to skip redefining the struct and avoid this error.

### DIFF
--- a/libusb/os/threads_windows.h
+++ b/libusb/os/threads_windows.h
@@ -34,7 +34,7 @@ typedef struct usbi_cond {
 } usbi_cond_t;
 
 // We *were* getting timespec from pthread.h:
-#if (!defined(HAVE_STRUCT_TIMESPEC) && !defined(_TIMESPEC_DEFINED))
+#if (!defined(HAVE_STRUCT_TIMESPEC) && !defined(_TIMESPEC_DEFINED) && !defined(_INC_TIME))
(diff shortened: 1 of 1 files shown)


mcuee:
I am not so sure how npm (node.js) is related to libusb. Probably you are building the node libusb module but I am not so sure how you build libusb by using Windows SDK.

Usually you build libusb using Visual Studio (community version of VS2013/2015 is free) I am actually wondering if this is because you use an old version of the Windows SDK. The latest version is version 10.0.14393.0.

FYI, I am not using the VS2013/VS2015 but I can build using libusb using the Enterprise WDK rs1 version 10.0.14393.0 which is supposed to be a superset of the SDK.

mcuee:
As I replied, I do not think this is a valid fix.

mcuee:
@SkyLined Please update to latest version of Windows SDK to see if it fixes your issue. I do not think this is an issue with libusb.

mcuee:
In any case, this fix is not passing the checks. So I will close it for now. Please reopen this once you have tried with the latest version of Windows SDK and fix the check.

Issue #251, opened 2017-01-05
libusb doesn't find virtual USB devices


nickich:
I'm using Eltima Network Gate on Windows 7 to access remote devices, but libusb doesn't find them properly.
When I run listdevs, the virtual devices came up with Linux VID/PID.
I was looking at 'windows_winusb.c' and it seems that it thinks the devices are hubs...

> 8086:8c2d (bus 2, device 1)
> 8086:8c26 (bus 1, device 1)
> 046d:c52b (bus 3, device 2) path: 2
> 1d6b:0001 (bus 4, device 1) path: 1
> 0c45:64d0 (bus 2, device 3) path: 1.5
> 8087:07da (bus 1, device 3) path: 1.5
> 8087:8000 (bus 1, device 2) path: 1
> 8087:8008 (bus 2, device 2) path: 1
> 8086:8c31 (bus 3, device 1)
>

mcuee:
I think this is the correct behavior since it acts a virtual USB hub for the network device. Why do you think it is a bug? What is the issue you are facing?

nickich:
It's actually a virtual USB device. The libusb thinks it is a hub and since it can't get its VID/PID, libusb assigns Linux ones. 
Same virtual device is recognised and enumetated properly by other sw.

mcuee:
No I do not think this is true. It is not a virtual USB device. Rather those remote USB device connect to it is a virtual USB device. Itself is a virtual HUB. If it is a USB device, virtual or real, it will have a USB VID/PID.

Ref: http://www.eltima.com/products/usb-over-ethernet/

nickich:
The remote device connects as a virtual device on a local machine. This virtual device has PID/VID, but the libusb doesn't recognize it as such.

mcuee:
Please refer to Pull Request 257. https://github.com/libusb/libusb/pull/257

PR #257, opened 2017-01-11
Fix enumerating Eltima's USB STUB virtual device

jiez:
This patch should fix issue #251. Eltima's USB STUB virtual device behaves like a USB HUB. But getting its SPDRP_ADDRESS fails, which causes it being skipped. Thus all virtual USB devices connected to it cannot be enumerated because their parent device has been skipped. This patch allows such a device to be enumerated even it does not have a port number. It also adds Eltima's USB STUB driver in USB HUB drivers list so it will be enumerated as a USB HUB.


mcuee:
Maybe this can be merged. Not so sure if there are any bad consequences here.

diabolo38:
what would the impact on any call to get port number , get path and alike call ?  Am not experiences and familiar enought in that part to fully answer.
But as first look this patch could cause some minor regression in app that baldy "blindyl expect" a port number (as it always use to be one) , more severe issues with "unique device path" that more app may rely on to identify .

PR #258, opened 2017-01-12
Misc fixes

dickens:
This is a collection of miscellaneous fixes, mostly for Windows. If all the builds pass and there are no objections, I plan to merge this weekend.

### DIFF
--- a/Xcode/config.h
+++ b/Xcode/config.h
@@ -6,9 +6,6 @@
 /* Message logging */
 #define ENABLE_LOGGING 1
 
-/* Define to 1 if you have the `gettimeofday' function. */
-#define HAVE_GETTIMEOFDAY 1
-
 /* Define to 1 if you have the <poll.h> header file. */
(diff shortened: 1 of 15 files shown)


hiyuh:
IMHO, use `AX_PTHREAD` from `autoconf-archive`?

diabolo38:
while looking at some of the patch , i was a bit surprises to see timing uisage of CLOCK_REALTIME i would have expected CLOCK_MONOTONIC.

dickens:
Why are you surprised? The gettimeofday() function reads the realtime clock, so this is a direct replacement. The actual clock source for the logging doesn't matter much as we use a reference time to start at 0 anyway.

diabolo38:
yes perfectly fine for log we don"t quiet care of some jump or roll back  , after more in depth sight i find that  CLOCK_MONOTONIC is well used for all time out . sorry for the stupid remarks  :( .

PR #274, opened 2017-02-22
Silence some warnings.

jmgao:
This PR silences some warnings (-Wsign-compare, -Wswitch) that are on by default in the Android platform build system. (We could turn them off, but IMO they're generally useful warnings, despite being false positives here.)


dickens:
Applied, thanks!

jmgao:
It seems like you missed a commit: https://github.com/libusb/libusb/pull/274/commits/c26c909ae5b706b44792f2ea0ae606b84e063b53

mcuee:
Just want to mention that the commits are all included.
https://github.com/libusb/libusb/blob/master/libusb/os/linux_usbfs.c#L814

PR #277, opened 2017-03-03
Fix Null-Pointer exception as described in the mailing list

mbussjaeg:
Fix Null-Pointer exception for HCD hubs. Concerning
libusb_get_device_descriptor.


mbussjaeg:
I will open a new pull-request with some more github background....

PR #283, opened 2017-03-16
darwin: use clock_gettime() on 10.12+

hjelmn:
Apple implemented the POSIX.2 clock_gettime () call in macOS
10.12. This commit updates the darwin backend to not use the mach
clock if clock_gettime() is available.

Signed-off-by: Nathan Hjelm <hjelmn@mac.com>


hjelmn:
Just letting travis see this. Will rebase-merge when done.

PR #284, opened 2017-03-23
Integrating isochronous support in the WinUSB backend

philemonf:
This change implements the isochronous transfers in the WinUSB backend for Windows 8.1 and above.

The limitations of my implementation are the one inherited from the WinUSB API, i.e. isochronous transfers must be aligned to 1-ms frame boundaries for IN endpoints. I tried to design the code so that, if it is used on an older Windows platform than 8.1, i.e. the ones with no isochronous support in WinUSB, the behavior is as it was before: the user gets a LIBUSB_ERROR_NOT_SUPPORTED error code if he calls the libusb isochronous API.

As a testing side-effect, this change also includes support for USB 3.0 SuperSpeed Endpoint Companion Descriptor in libusb_get_max_iso_packet_size.

I am happy to submit it to peer review and to improve it as you might request :-).

-- Phil

EDIT: this change has been discussed here https://sourceforge.net/p/libusb/mailman/message/35704907/.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -511,6 +511,7 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, int
 	if (if_desc->bNumEndpoints == 0) {
 		usbi_dbg("no endpoints found for interface %d", iface);
 		libusb_free_config_descriptor(conf_desc);
+		priv->usb_interface[iface].current_altsetting = altsetting;
 		return LIBUSB_SUCCESS;
 	}
 
@@ -530,6 +531,9 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, int
 	// Extra init may be required to configure endpoints
 	if (priv->apib->configure_endpoints)
 		r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
+
+	if (r == LIBUSB_SUCCESS)
+		priv->usb_interface[iface].current_altsetting = altsetting;
 
 	return r;
 }
@@ -1710,11 +1714,25 @@ static void winusb_destroy_device(struct libusb_device *dev)
 static void winusb_clear_transfer_priv(struct usbi_transfer *itransfer)
 {
 	struct winusb_transfer_priv *transfer_priv = usbi_transfer_get_os_priv(itransfer);
+	struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
+	struct winusb_device_priv *priv = _device_priv(transfer->dev_handle->dev);
+	int sub_api = priv->sub_api;
 
 	usbi_close(transfer_priv->pollable_fd.fd);
 	transfer_priv->pollable_fd = INVALID_WINFD;
 	transfer_priv->handle = NULL;
 	safe_free(transfer_priv->hid_buffer);
+
+	if (transfer->type == LIBUSB_TRANSFER_TYPE_ISOCHRONOUS && sub_api == SUB_API_WINUSB) {
+		if (transfer_priv->isoch_buffer_handle != NULL) {
+			if (WinUSBX[sub_api].UnregisterIsochBuffer(transfer_priv->isoch_buffer_handle)) {
+				transfer_priv->isoch_buffer_handle = NULL;
+			} else {
+				usbi_dbg("Couldn't unregister isoch buffer!");
+			}
+		}
+	}
+
 	safe_free(transfer_priv->iso_context);
 
 	// When auto claim is in use, attempt to release the auto-claimed interface
@@ -1987,6 +2005,14 @@ const struct windows_usb_api_backend usb_api_backend[USB_API_MAX] = {
 			pLibK_GetProcAddress((PVOID *)&WinUSBX[i].fn, i, KUSB_FNID_##fn);	\
 	} while (0)
 
+#define NativeWinUSBOnly_Set(fn)									\
+	do {											\
+		if (native_winusb)								\
+			WinUSBX[i].fn = (WinUsb_##fn##_t)GetProcAddress(h, "WinUsb_" #fn);	\
+		else                                                                            \
+			WinUSBX[i].fn = NULL;                                                   \
+	} while (0)
+
 static int winusbx_init(struct libusb_context *ctx)
 {
 	HMODULE h;
@@ -2040,6 +2066,11 @@ static int winusbx_init(struct libusb_context *ctx)
 		WinUSBX_Set(WritePipe);
 		WinUSBX_Set(IsoReadPipe);
 		WinUSBX_Set(IsoWritePipe);
+		NativeWinUSBOnly_Set(RegisterIsochBuffer);
+		NativeWinUSBOnly_Set(UnregisterIsochBuffer);
+		NativeWinUSBOnly_Set(WriteIsochPipeAsap);
+		NativeWinUSBOnly_Set(ReadIsochPipeAsap);
+		NativeWinUSBOnly_Set(QueryPipeEx);
 
 		if (WinUSBX[i].Initialize != NULL) {
 			WinUSBX[i].initialized = true;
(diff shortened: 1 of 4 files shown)


hjelmn:
Can you squish this down into two commits: one with the core changes, and one to add WinUSB isoc support. This will make it easier for the Windows maintainer to review.

philemonf:
Noted. I will do it ASAP.

philemonf:
I did what you suggested. I splitted the core change into two commits as they are not related. Let me know if you have any other requests.

philemonf:
Any news :-)? Any other changes you would want me to do to merge the change?

philemonf:
Any news on this one ? I would love to get your feedback and reviews :). Thank you all!

PS: I received this pull request on my fork https://github.com/philemonf/libusb/pull/1 if anyone has comments about it.

aospan:
I have tested this patches with my project (https://tv.jokersys.com) and it works fine. Please accept this patches to mainline libusb.
Thanks a lot !

laran002:
I ran into an issue where libusb_submit_transfer() returned LIBUSB_ERROR_IO when submitting a transfer to an isochronous OUT endpoint. If I use the same transfer using an IN endpoint it returns fine. I also ensured to use the result from libusb_get_max_iso_packet_size( device, ep_iso_out) to set the packet length with libusb_set_iso_packet_lengths().

After some debugging I found that the WriteIsochPipeAsap at windows_winusb.c(3106) throw a ERROR_INVALID_USER_BUFFER.

philemonf:
Ok, I think it might be due to an error in the way the size of OUT transfer is computed compared to the Linux version of the API. Can you please try to set the actual_length of all your isochronous packet manually (instead of using libusb_set_iso_packet_lengths that sets their length field) and tell us if it solves the issue? If it does solve the issue, I'll submit a fix. Otherwise, did you make sure your user buffer is still alive when your transfer are asynchronously submitted?

philemonf:
PS: please note the edit of the last message: you need to set the actual_length field in addition of the length field. I am looking forward to seeing if it fixes your issue.

philemonf:
You can find a attempted fix on the branch `fix-out-isoch-length` of my fork. Here is the change: https://github.com/philemonf/libusb/commit/c97d809ad5b59a9737791d4e50aa714dc5dddb58. In this case you would only require to set the length of your packets (libusb_set_iso_packet_lengths). I hope this helps and please tell us if it fixes the issue.

laran002:
It looks like transfer->iso_packet_desc[idx].actual_length in windows_winusb.c(3075) returns zero for some reason. When I force out_transfer_length = 256 (my set packet length) I GetLastError() == ERROR_IO_PENDING is true on windows_winusb.c(3114). Then the transfer appears to happen as expected, not yet confirmed in a loop back test yet though.

philemonf:
Great, the problem is the one I expected. Please tell me if the fix solves your issue.

laran002:
Yes, now it works.

philemonf:
Did you test it with the attempted fix or by setting `actual_length` yourself? In the former case, I'll merge the fix.

laran002:
Sorry for being unclear, I updated mesage above

laran002:
For some reason I start getting libusb_isochronous_transfer() failed: LIBUSB_ERROR_NOT_SUPPORTED issued by winusbx_do_additional_checks(sub_api). As far as I remember I have not done anything except checked out your merged fix and recompiled it. Is it more likelly it is related to the driver? I am using the WinUSB driver from Zadig-2.3.

philemonf:
Are you sure WinUSB is loaded and your system does not have an instance of the libusbK driver on your system as it preempts the use of WinUSB (this is not part of my change and was there prior to it). Please put a breakpoint in the `winusbx_init` function to see. I always tested with the WinUSB provided with the OS.

laran002:
Thanks, the libusbk driver was causing this. I remember accidentally installing the libusbk using zadig some days ago. I uninstalled the device and chose the "Delete the driver software ..." option. Apparently the .dll's are still on my system. I just NULL:ed the `LoadLibraryA("libusbK")` handle in `winusbx_init` and recompiled. That did the trick!

philemonf:
You can remove the dll in the `System32` and `SysWOW64` folders in the `c:\Windows\` folder to avoid altering the code if you want.

libusb/os/windows_winusb.c
```diff
@@ -2871,6 +2906,228 @@ static int winusbx_submit_bulk_transfer(int sub_api, struct usbi_transfer *itran
...
+	}
+
+	// Important note: the WinUSB_Read/WriteIsochPipeAsap API requires a ContinueStream parameter that tells whether the isochronous
+	// stream must be continued or if the WinUSB driver can schedule the transfer at its conveniance. Profiling subsequent transfers
+	// with ContinueStream = FALSE showed that 5 frames, i.e. about 5 milliseconds, were left empty between each transfer. This
+	// is critical as this greatly diminish the achievable isochronous bandwidth. We solved the problem using the following strategy:
+	// - Transfers are first scheduled with ContinueStream = TRUE and with winusbx_iso_transfer_continue_stream_callback as user callback.
+	// - If the transfer succeeds, winusbx_iso_transfer_continue_stream_callback restore the user callback and calls its.
+	// - If the transfer fails, winusbx_iso_transfer_continue_stream_callback reschedule the transfer and force ContinueStream = FALSE.
+	if (!transfer_priv->iso_break_stream) {
```
codido: It seems like the first request should have ContinueStream be set to FALSE according to https://msdn.microsoft.com/en-us/library/windows/hardware/dn376866(v=vs.85).aspx

At least for a small use case I ran here, setting the first transaction to true resulted in an ERROR_INVALID_PARAMETER error.

philemonf: Let me motivate this one as it has been thought through. The whole thing was thought so that each transfer is first tried with `ContinueStream` set to true. If the transfer fails as a result, the callback detects it and retries it with `ContinueStream` set to false.

Say one schedules N libusb transfers (as this is generally the case with such an async API, e.g. libuvc). The first transfer will fail and be rescheduled but all the subsequent ones will continue the stream and, in this manner, this strategy results in full bandwidth usage. I agree that the first transfer will always fail but so will any transfers that are scheduled too late for the stream to be continued. In some sense, the first transfer is just a special case of not being able to continue a stream and this case is handled in the callback, i.e. `
winusbx_iso_transfer_continue_stream_callback`.
Does this make sense?

philemonf:
According to this guideline https://msdn.microsoft.com/en-us/library/windows/hardware/dn376866(v=vs.85).aspx, it seem to me that `WinUSB_QueryPipeEx` should already be doing the calculation on `Interval` but I agree it is unclear.

libusb/os/windows_winusb.c
codido: Yes, but for some reason I kept getting the same error over and over again. Re-tested it now and couldn't reproduce it so might have been something else.

In any case, since the two errors are quite different, and this would mask real invalid parameters, any reason not to set iso_break_stream to true on initialization?

philemonf: Thanks for taking the time to review my code by the way! For this one, I still have concerns. Let me consider two different use cases. Before that let `T` be the time WinUSB waits before scheduling a transfer with `ContinueStream` set to `FALSE`. In my tests, `T` was about 5 milliseconds.

The first use case is the one I mentioned above where one submits `N` transfers and resubmit a transfer in the transfer callback. This is a typical use case for isochronous transfers as it typically consists in periodic transfers and therefore the programmer want to ensure that enough transfer buffers are ready to receive the incoming data. If we set `iso_break_stream` to true on initialization, the first time each of this `N` transfers are scheduled, they are followed by a "blank time" of `T`. In total, this result in `N` times `T` timing overhead at start up. In the current implementation, there is only a timing overhead of `T` for the first transfer.

The second use case I want to consider is the one where `N` is determined at runtime using an adaptive procedure that monitors the current throughput and allocates new transfers if it measures that the bandwidth usage is poor. If we set `iso_break_stream` to true on initialization, each time a new transfer is allocated, this would result in a drop in throughput as it would break the stream.

Does this make sense? What do you think?

codido: np, and thanks for your work btw!

I'm not entirely sure when winusb returns this error, but what I had in mind was to make sure ContinueStream==FALSE for the first ever read/write operation on that endpoint. Other operations, even if issued long after, can still have ContinueStream==TRUE and deal with failures accordingly.

That being said, if it doesn't have an ill effects, it's probably not that important. Just resulted in a red herring when debugging other invalid parameter errors :)

philemonf:
@codido Sorry for the delay: September was a very busy month. I hope this last commit fixes what you wanted. I couldn't properly test the interval computation due to hardware issues. Therefore, please confirm that it works for your Android use case. Thanks again for the review!

mirh:
Rebase needed.

aospan:
BTW, high bandwidth isochronous transfers works fine too. It's 187.5 Mbit/sec as expected (3 data packets per microframe 1024 bytes each - this is maximum). I have tested on Windows 10. More info in my post:
https://jokersys.com/2018/01/15/high-bandwidth-usb-isochronous-transfers/

Set process priority "above normal" required. Otherwise we can miss USB isoch
transfers (commit here https://github.com/aospan/libjokertv/commit/0a853cd4bf7d2ee946313bd6e2b9d62ea8d2d16d )

philemonf:
I merged my change with the recent change of libusb. Thank you very much for the feedback @aospan. @aospan does the new version work for you?

aospan:
@philemonf yes, I have tested your latest tree and it works fine.

I tried to increase amount of isoch packets in one urb (hit the limit near 400 packets. but revert ot my current value 128). Also I have tried to increase amount of libusb transfers (currently I have about 16 transfers). No luck ;( 
Only increasing priority does help.

mirh:
Mhh.. updates?

philemonf:
I was waiting on a merge decision. Am I missing something?

If you are referring to the hack @aospan had to make for USB 2 high bandwidth, I could not repro due to lack of hardware.

philemonf:
@aospan Did you have any update on your issue?

hjelmn:
Please resolve the merge conflicts then we can take a look.

philemonf:
Sorry for the delay. I just had a look and don't see conflict. Were you referring to something in particular?

hjelmn:
@philemonf I see:

```
This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.
```

Also, I don't want to see any merge commits. I have the cherry-pick any commits one by one to make sure our nano versioning is followed. Please keep the total number of commits down to a minimum.

hjelmn:
I will close this PR if I don't see any update in the next week. Need a clean PR with no merge commits.

philemonf:
Sorry, I missed your last message. It should be alright now.

hjelmn:
Ok, thanks for your patience. I seem to be the only active maintainer right now. I don't have any windows systems to test with but I will assume this is ok given the feedback in this thread.

philemonf:
Great!

@aospan Do you confirm that your scenarios are still working with the post-merge version in master :)? There should be no difference with the previous tree as I verified that there was no diff using git.

philemonf:
@hjelmn Note that there is a change to libusb_get_max_iso_packet_size that does not only affect Windows. It adds support for wBytesPerInterval in the USB SuperSpeed Companion Descriptor for USB3 devices.

hjelmn:
@philemonf Yup. Noticed that. Looks ok to me.

aospan:
@philemonf LGTM :)

PR #285, opened 2017-03-23
Fix HCD hubs Null-Pointer exception in windows_get_config_descriptor_…

mbussjaeg:
…by_value.
As described on the libusb-devel list this Null-Pointer check will fix problems with HCD hubs.
This fix just adds the same check that were used in version 1.0.20 and are necessary for some HCD hubs regarding not cached config descriptors.


dickens:
Thanks for the PR. This check is no longer necessary with commit 92093a59.

PR #286, opened 2017-03-23
Fix linux builds (<2.6.23 not supporting O_CLOEXEC)

mbussjaeg:
Fix for older linux builds (<2.6.23) because of used flag O_CLOEXEC introduced with commit aa73b2ecb81f1fe4f503d3ac693679632587ad50.

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -27,6 +27,10 @@
 #include <ctype.h>
 #include <dirent.h>
 #include <errno.h>
+/* Older builds: O_CLOEXEC requires min. Linux 2.6.23 */
+#ifndef O_CLOEXEC
+#define O_CLOEXEC 0
+#endif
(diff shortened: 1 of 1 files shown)


hjelmn:
I am a little torn about accepting this. On one hand it is simple but on the other 2.6.x is really really old. I don't think we should claim support for anything older than 3.0.

mbussjaeg:
Yes, it's really simple and ensures the compile works again.
I didn't know there's a minimum for any 2.6 version. I mean, even Windows XP is currently supported and that's from 2001.
Anyway, with the version a month ago the compile worked and now it doesn't. If it's not accepted I think you should clarify the minimum version is 2.6.23 from now on.
Just discuss this for the libusb master, I need this fix anyway.

dickens:
I addressed this in v2 of my patch, but somehow I managed to apply and push v1 to master. Thanks for reporting this.

PR #287, opened 2017-03-30
Don't ignore platform config.h

fadden:
The android and Xcode config.h headers were being ignored.


LudovicRousseau:
Fixed in https://github.com/libusb/libusb/commit/b06b4d101b5ddd95e86a6ba34b526eda6fc839ba

Thanks

PR #293, opened 2017-04-21
Fixed USB_CONFIGURATION_DESCRIPTOR struct alignment issue

L-proger:
Fixed USB_CONFIGURATION_DESCRIPTOR struct alignment issue (10 bytes size instead of 9)


dickens:
Merged, thanks!

PR #304, opened 2017-05-25
[darwin_usb] Fix ISOC processing on Mac

aospan:
not all isoc frames are ok in one transfer
so we need to traverse all isoc frames one-by-one
otherwise we can loose some good isoc frames

Here is my observation on real hw:

I have ISOC transfer with 16 frames inside. After submitting this
transfer I have received callback with result=0xe000400e.
This error code is kIOUSBNotSent1Err and not fatal and mean that the
requested frame number was in the past, or lies too far into the
future. But some frames inside this transfer is ok and we can loose
it if we rely only on transfer result.

Signed-off-by: Abylay Ospan <aospan@jokersys.com>

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1968,18 +1968,21 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   usbi_dbg ("handling %s completion with kernel status %d",
              isControl ? "control" : isBulk ? "bulk" : isIsoc ? "isoc" : "interrupt", tpriv->result);
 
-  if (kIOReturnSuccess == tpriv->result || kIOReturnUnderrun == tpriv->result) {
-    if (isIsoc && tpriv->isoc_framelist) {
-      /* copy isochronous results back */
-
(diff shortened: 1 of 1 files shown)


aospan:
similar patch for windows isoc processing:
https://github.com/philemonf/libusb/pull/1

Interestingly, Linux doesn't need this type of fixes. Linux usb stack return SUCCESS (LIBUSB_TRANSFER_COMPLETED) for transaction even there is some isoc packets not filled with data. As I understand this behavior is expected "by design" - otherwise we don't need status for every isoc packet inside transaction :)

aospan:
Ok, I have fixed this issue on FPGA side (device side). Here is commit:
https://github.com/aospan/joker-tv-fpga/commit/6fb99196d1d62e64ec275875a9de465cde6ead6b

Now I send zero length isoc packets to host if I don't have data inside FPGA FIFO.
Before I have not send any packets back to host. This caused situation I have tried to fix in this pull request.
So, this pull request doesn't need any more. Please do not apply it.

P.S.
I have made same for Windows  - https://github.com/philemonf/libusb/pull/1

PR #308, opened 2017-05-30
Usbdk

sameehj:
UsbDk dynamic force selection is now optional and dynamic. This approach leaves Libusb's forward and backward compatibility as it is while making it easier for forcing UsbDk as a backend.

Moreover, an enum was added for configuring the use of a specific backend for easier
future extention of other backends force selection as it was suggested by
"scumcoder at yandex.ru" here: https://sourceforge.net/p/libusb/mailman/message/35859213/.

Inorder to force the use of UsbDk as a backend the function "libusb_force_use_backend"
should be called along with the enum "LIBUSB_USBDK" before calling libusb_init.


sameehj:
I have opened a new pull request.

PR #312, opened 2017-06-09
Windows: Support LIBUSB_TRANSFER_ADD_ZERO_PACKET on winusb

### LINKED ISSUE #91: Windows: Add LIBUSB_TRANSFER_ADD_ZERO_PACKET flag to indicate need for ZLP
Ref: http://libusb.org/ticket/6

Linux and Mac OS X backend are ready. But not Windows and OpenBSD/NetBSD backend.

patstew:
Fixes #91

### DIFF
--- a/libusb/os/windows_common.h
+++ b/libusb/os/windows_common.h
@@ -265,12 +265,19 @@ struct usbdk_device_handle_priv {
 	// Not currently used
 	char dummy;
 };
+ 
+enum WINUSB_ZLP {
+	WINUSB_ZLP_UNSET = 0,
+	WINUSB_ZLP_OFF = 1,
+	WINUSB_ZLP_ON = 2
+};
 
 struct winusb_device_handle_priv {
 	int active_interface;
 	struct {
 		HANDLE dev_handle; // WinUSB needs an extra handle for the file
 		HANDLE api_handle; // used by the API to communicate with the device
+		uint8_t zlp[USB_MAXENDPOINTS]; // Current per-endpoint SHORT_PACKET_TERMINATE status (enum WINUSB_ZLP)
 	} interface_handle[USB_MAXINTERFACES];
 	int autoclaim_count[USB_MAXINTERFACES]; // For auto-release
 };
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2024,8 +2024,6 @@ static int winusb_submit_transfer(struct usbi_transfer *itransfer)
 		break;
 	case LIBUSB_TRANSFER_TYPE_BULK:
 	case LIBUSB_TRANSFER_TYPE_INTERRUPT:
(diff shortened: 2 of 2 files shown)


dickens:
This implementation has one particular shortcoming--the policy is turned on/off globally for an endpoint.

Consider the following scenario:
1) User submits a transfer and requests the ADD_ZERO_PACKET flag. This makes its way to winusbx_submit_bulk_transfer() and the policy gets turned on.
2) Before the first transfer completes, the user submits a second transfer without the ADD_ZERO_PACKET flag. This makes its way to winusbx_submit_bulk_transfer() and the policy gets turned off.
3) After the second transfer is done submitting, the first transfer makes its way out of the WinUSB queue and does not get a ZLP added because the policy is now off.

It would be better to instead manually submit a second zero-length transfer if this flag is present AND the transfer length is a multiple of the endpoint max packet size. The second part is an important point, because a ZLP isn't supposed to be sent unconditionally.

hjelmn:
@patstew Can you update based on @dickens comments?

patstew:
I'd sort of filed this under 'it's good enough for me' for the time being.
I was a little bit concerned about what to do with the overlapped IOs and errors in going about it that way. You've got two transfers in flight now, and what if one fails and the other doesn't? And is it necessary to wait on both, or can you assume that the ZLP completing means that the actual transfer is complete?

patstew:
Even so, you'd now need to have two overlapped structures for each winfd, and decide which to look at to get bytes transferred/complete based on whether it's an emulated ZLP or not. I think that it will end up adding a bunch of complexity in the winusb code, unless I'm missing an easy solution.

I think in most cases you either want ZLP behaviour on an endpoint or not, so I could change this PR so that each endpoint basically adopts the policy of the first transfer, and you get an error if any future transfer tries to change it. It'd be better than the status quo of failing on all ZLPs, it'd be a much smaller change with less to go wrong and I think it would cover 99% of usecases. Would that be acceptable?

hjelmn:
@patstew Sounds good to me. That would address one of the @dickens comments. I can't comment on any of the other comments as I don't know anything about the Windows USB stack.

patstew:
@dickens Do you have any comments about this way of doing it?

mcuee:
I agree that most of the people will want ZLPs always or never, if they are only deal with one device in the application.

patstew:
It's per endpoint, per device, so I don't think it matters how many devices you use in the application.

mcuee:
I see. Thanks for the clarifications.

hjelmn:
Should we take this for 1.0.25? ZLP is a pretty big missing feature from the windows backend but the backend has been bit rotting.

mcuee:
@dickens Just want to know if you still have concerns for the merging of this pull request. Please advice. Thanks.

mcuee:
I am fine with this pull request myself. I think we can take a bit of risk and merge the pull request. It will not really affect existing Windows user since it is a new feature.

PR #318, opened 2017-07-05
Windows: Fix possible null pointer dereference

johnkeeping:
This may be null if we have failed to read a configuration descriptor in
cache_config_descriptors().


dickens:
Thanks for the PR. As of commit 92093a59243f01819482d53659b40898504bf5d3, failing to read a configuration descriptor will force dev->num_configurations to 0, so we don't need this check.

PR #321, opened 2017-07-11
core: Introduce libusb_set_option() API function

dickens:
This new function allows more flexibility in extending the library to
support more user-configurable options. It is intended to provide a
single API that can support a wide variety of needs and eliminates the
need for new API functions to set future options.

The function is introduced with a single option (LIBUSB_OPTION_LOG_LEVEL)
that replaces the libusb_set_debug() function. Documentation relating to
libusb_set_debug() and the uses of this function in the examples and
tests have been updated accordingly.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -116,15 +116,17 @@ struct list_head active_contexts_list;
  * libusb uses stderr for all logging. By default, logging is set to NONE,
  * which means that no output will be produced. However, unless the library
  * has been compiled with logging disabled, then any application calls to
- * libusb_set_debug(), or the setting of the environmental variable
- * LIBUSB_DEBUG outside of the application, can result in logging being
- * produced. Your application should therefore not close stderr, but instead
- * direct it to the null device if its output is undesirable.
+ * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level), or the setting of the
+ * environmental variable LIBUSB_DEBUG outside of the application, can result
+ * in logging being produced. Your application should therefore not close
+ * stderr, but instead direct it to the null device if its output is
+ * undesirable.
  *
- * The libusb_set_debug() function can be used to enable logging of certain
- * messages. Under standard configuration, libusb doesn't really log much
- * so you are advised to use this function to enable all error/warning/
- * informational messages. It will help debug problems with your software.
+ * The libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) function can be
+ * used to enable logging of certain messages. Under standard configuration,
+ * libusb doesn't really log much so you are advised to use this function
+ * to enable all error/warning/ informational messages. It will help debug
+ * problems with your software.
  *
  * The logged messages are unstructured. There is no one-to-one correspondence
  * between messages being logged and success or failure return codes from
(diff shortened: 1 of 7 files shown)


libusb/core.c
```diff
@@ -116,15 +116,17 @@ struct list_head active_contexts_list;
...
- * libusb_set_debug(), or the setting of the environmental variable
- * LIBUSB_DEBUG outside of the application, can result in logging being
- * produced. Your application should therefore not close stderr, but instead
- * direct it to the null device if its output is undesirable.
- *
- * The libusb_set_debug() function can be used to enable logging of certain
- * messages. Under standard configuration, libusb doesn't really log much
- * so you are advised to use this function to enable all error/warning/
- * informational messages. It will help debug problems with your software.
+ * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL), or the setting of the
```
LudovicRousseau: The log level is missing the call ``libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL)``.
I guess it should be ``libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, log_level)`` instead.

The same problem is also present many times in other comments.

dickens: Yes, this was intentional, for brevity. The previous comments only had libusb_set_debug() (note that all arguments are missing). Here I wanted to highlight that the ctx and correct option must be provided. I can change them all though, if it helps make it more clear.

LudovicRousseau: I prefer to avoid brevity in API documentation. It was obvious for you but looked like a bug to me.

dickens:
New PR with changes suggested by @LudovicRousseau

libusb/core.c
```diff
@@ -2039,6 +2024,48 @@ void API_EXPORTED libusb_set_debug(libusb_context *ctx, int level)
...
+	va_list ap;
+
+	USBI_GET_CONTEXT(ctx);
+
+	va_start(ap, option);
+	switch (option) {
+	case LIBUSB_OPTION_LOG_LEVEL:
+#if defined(ENABLE_LOGGING) && !defined(ENABLE_DEBUG_LOGGING)
+		if (!ctx->debug_fixed)
+			ctx->debug = va_arg(ap, int);
```
LudovicRousseau: You could check that the ``level`` argument has a valid value (between ``LIBUSB_LOG_LEVEL_NONE`` and ``LIBUSB_LOG_LEVEL_DEBUG``) and return ``LIBUSB_ERROR_INVALID_PARAM`` if it is not the case.

libusb/core.c
```diff
@@ -2039,6 +2024,48 @@ void API_EXPORTED libusb_set_debug(libusb_context *ctx, int level)
...
+#endif
+		break;
+	default:
+		r = LIBUSB_ERROR_INVALID_PARAM;
```
LudovicRousseau: I think the returned value should be ``LIBUSB_ERROR_NOT_SUPPORTED`` as the documentation says: "the option is not supported".

dickens: The documentation says "returns LIBUSB_ERROR_INVALID_PARAM if the option or arguments are invalid". Hitting the default case in the switch means libusb doesn't know anything about the option.

The other return value LIBUSB_ERROR_NOT_SUPPORTED will be for an option like the upcoming LIBUSB_OPTION_USE_USBDK, which will only be supported on Windows.

libusb/core.c
dickens: Thanks, I'll add that and update the PR.

libusb/core.c
LudovicRousseau: Ah ok.
I propose to change the documentation:
``\returns LIBUSB_ERROR_NOT_SUPPORTED if the option is valid but not supported on this platform``

dickens:
PR updated based on latest feedback. Thanks @LudovicRousseau!

dickens:
Any more feedback, comments, or objections? Plan on merging soon.

PR #327, opened 2017-07-27
Fix typo in exit() function signature

jonner:
This typo caused a bunch of compile warnings about incompatible pointer types.


dickens:
Applied, thanks!

PR #362, opened 2017-11-07
darwin: fix reporting of USB3 device speed when kIOUSBDeviceInterfaceID320 is used

parafin:
Even when version 320 of IOUSBDeviceInterface is used new enough OS can report
kUSBDeviceSpeedSuper as device speed. Without this fix libusb will return
LIBUSB_SPEED_UNKNOWN in that case (e.g. when deployment target was set to 10.7).


seanm:
You should conditionalize this against the SDK kUSBDeviceSpeedSuper was added in (10.7), like so:

#if MAC_OS_X_VERSION_MAX_ALLOWED >= 1070

seanm:
I propose this as better: https://github.com/libusb/libusb/commit/e4851c38946560ac0fe3eeea848b37c264b4a9c8

PR #365, opened 2017-11-23
Fix isochronous transfers in Linux 4.13.8

jceel:
Sometime between 4.13.0 and 4.13.8, a new check appeared in `USBDEVFS_SUBMITURB` ioctl code:

```
		for (totlen = u = 0; u < number_of_packets; u++) {
			/*
			 * arbitrary limit need for USB 3.0
			 * bMaxBurst (0~15 allowed, 1~16 packets)
			 * bmAttributes (bit 1:0, mult 0~2, 1~3 packets)
			 * sizemax: 1024 * 16 * 3 = 49152
			 */
			if (isopkt[u].length > 49152) {
				ret = -EINVAL;
				goto error;
			}
			totlen += isopkt[u].length;
		}
		u *= sizeof(struct usb_iso_packet_descriptor);
		if (totlen <= uurb->buffer_length)
			uurb->buffer_length = totlen;
		else
			WARN_ONCE(1, "uurb->buffer_length is too short %d vs %d",
				  totlen, uurb->buffer_length);
		break;
```

Since libusb is not setting `uurb->buffer_length` for isochronous transfers, they all end up with 0 bytes transferred and the `uurb->buffer_length is too short %d vs %d` warning is printed in the dmesg. This PR makes libusb set `buffer_length` to be the sum of all iso descriptor lengths.

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -2084,6 +2084,7 @@ static int submit_iso_transfer(struct usbi_transfer *itransfer)
 		for (j = 0, k = packet_offset - urb_packet_offset;
 				k < packet_offset; k++, j++) {
 			packet_len = transfer->iso_packet_desc[k].length;
+			urb->buffer_length += packet_len;
 			urb->iso_frame_desc[j].length = packet_len;
 		}
 


pointcontrols:
I found that issue yesterday. In fact, in 4.14 that  change has been reverted:

>
 commit 845d584f41eac3475c21e4a7d5e88d0f6e410cf7
Author: Hans de Goede <hdegoede@redhat.com>
Date:   Mon Oct 16 16:21:19 2017 +0200

    USB: devio: Revert "USB: devio: Don't corrupt user memory"

    Taking the uurb->buffer_length userspace passes in as a maximum for the
    actual urbs transfer_buffer_length causes 2 serious issues:

    1) It breaks isochronous support for all userspace apps using libusb,
       as existing libusb versions pass in 0 for uurb->buffer_length,
       relying on the kernel using the lenghts of the usbdevfs_iso_packet_desc
       descriptors passed in added together as buffer length.

jceel:
Glad to see that! Anyway, I guess it won't hurt to correctly set `buffer_length` in libusb anyway? :)

pointcontrols:
Yes. I was just about to submit a PR myself, when I saw yours. :)

mkassner:
Our project like many that use libusb and isochronous transfers need this fix to work. Please consider merging soon!

PR #374, opened 2017-12-28
Misc warning fixes and cleanup

seanm:


seanm:
hmmm, not sure why this shows "20 commits"...  It's the 7 commits from yesterday that I'd like to have merged.  Did I create this PR incorrectly?

examples/dpfp.c
```diff
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 
 #include "libusb.h"
+#include "libusbi.h"
```
dickens: NAK, libusbi.h is for library use only.

examples/dpfp_threaded.c
```diff
@@ -32,6 +32,7 @@
 #include <fcntl.h>
 
 #include "libusb.h"
+#include "libusbi.h"
```
dickens: NAK, libusbi.h is for library use only.

examples/hotplugtest.c
```diff
@@ -22,12 +22,18 @@
 #include <stdio.h>
 
 #include "libusb.h"
+#include "libusbi.h"
```
dickens: NAK, libusbi.h is for library use only.

seanm:
Yes, it was for unused macro.

Generally, in my code, I suppress that warning like this:

```
(void)paramName;
```

Would that be ok?

Also, what is the correct way to revisit a change with the GitHub workflow?  To edit the commit?

dickens:
Yes, that approach is fine and is essentially what UNUSED does.

To update this, amend the appropriate commit and then force push to update the PR.

examples/hotplugtest.c
```diff
@@ -22,12 +22,18 @@
...
 
 int done = 0;
 libusb_device_handle *handle = NULL;
 
 static int LIBUSB_CALL hotplug_callback(libusb_context *ctx, libusb_device *dev, libusb_hotplug_event event, void *user_data)
 {
+	UNUSED(ctx);
```
dickens: This is also causing a build failure for earlier versions of Visual Studio. Variable declarations must come before any code in the current block scope.

seanm:
For consistency, shall I just get rid of the UNUSED macro entirely, and do everything the same way?

dickens:
Do you mean to get rid of the UNUSED macro from libusbi.h and all the uses of it within the library? If so, no I don’t think that’s what should be done. If anything, I would ensure that all unused variables within the library use the macro and all outside the library use the direct (void) approach.

seanm:
Yes, that's what I meant. It seems to me to make more sense to have a single approach, to spare one from having to think about which file one is editing.

dickens:
I’m not opposed to making a common header that the examples/tests can use to get useful macro definitions. This could even take care of things like _snprintf for VS2013 and earlier.

seanm:
Well, maybe subsequently... for now I've just changed the examples to not use the macro.

Everything else look ok?

dickens:
Still need to move the suppression statements below all variable declarations. This will fix the Visual Studio builds.

seanm:
I was hoping the lack of do/while (as in the macro) would solve that, but I guess not.  Odd that I can't repro that building as C89 with clang...

dickens:
PR merged via the following commits:

8e1a282 examples: ezusb: Renamed header guard variable to not use reserved identifier
830a9cb Fix unused parameter warnings
f66c63e examples: xusb: replaced insecure sprintf with snprintf
7338c2d core: Fixed deprecated attribute to work on clang
c1d8c8d Removed accidental double semi colons at end of line
0b3d4c6 Fixed various trivial cppcheck 1.80 warnings
96b89f9 WinCE: Fixed incorrect check of return value, found by cppcheck

Thanks for the work!

PR #378, opened 2018-01-09
Some WIP with 2 small fixes to merge

seanm:
@dickens @hjelmn I see that you are planning a release shortly, so you are probably being very conservative in what you merge at this point, but please at least cherry pick these two:

e4851c3
e6fbd27

anything else you can merge now or after the release... thx.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -3,7 +3,7 @@
  * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
  * Copyright © 2007-2008 Daniel Drake <dsd@gentoo.org>
  * Copyright © 2012 Pete Batard <pete@akeo.ie>
- * Copyright © 2012 Nathan Hjelm <hjelmn@cs.unm.edu>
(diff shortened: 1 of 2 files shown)


dickens:
I've merged the following four commits:

568274d extend enum liusb_speed for 10000MBit/s case (as 3f6a48d)
e4851c3 Test use of kUSBDeviceSpeedSuper against SDK version (as c8069cd)
e6fbd27 Suppress deprecation warning when building against 10.13 SDK (as 4f9d585)
026e630 Improve error checking to prevent (very unlikely) null dereference (as fef88bf)

The rest we'll take for the next release. Thanks!

hjelmn:
@seanm What in this PR still needs to be brought into master? Its kind of a mess of commits right now.

hjelmn:
@seanm Let me try just cherry-picking those commits into a clean PR. Will need to do so anyway to get the nano updated.

hjelmn:
Looks like all the Dec 29 commits are already merged with exception of "extend enum liusb_speed for 10000MBit/s case".

PR #380, opened 2018-01-22
Added libusb_hotplug_get_user_data

gudenau:
I have a project that needs to be able to fetch `user_data` from a registered libusb callback, this patch adds `libusb_hotplug_get_user_data`.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1993,6 +1993,17 @@ int LIBUSB_CALL libusb_hotplug_register_callback(libusb_context *ctx,
 void LIBUSB_CALL libusb_hotplug_deregister_callback(libusb_context *ctx,
 						libusb_hotplug_callback_handle callback_handle);
 
+/** \ingroup libusb_hotplug
+ * Gets the user_data associated with a hotplug callback.
+ *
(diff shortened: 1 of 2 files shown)


hjelmn:
I see no problem with adding this API. Anyone else have any thoughts?

libusb/libusb.h
```diff
@@ -1993,6 +1993,17 @@ int LIBUSB_CALL libusb_hotplug_register_callback(libusb_context *ctx,
 void LIBUSB_CALL libusb_hotplug_deregister_callback(libusb_context *ctx,
 						libusb_hotplug_callback_handle callback_handle);
 
+/** \ingroup libusb_hotplug
+ * Gets the user_data associated with a hotplug callback.
+ *
+ * Since version v1.0.23 \ref LIBUSB_API_VERSION >= 0x01000103
+ *
```
hjelmn: Current API version (of v1.0.23) will be 0x01000107.

a1ien:
I very like this API. Without this function if we alloc user data befor call libusb_hotplug_register_callback we need track this allocation for call free if we want call libusb_hotplug_deregister_callback  later

a1ien:
@hjelmn Can we hope that this PR will be merged

PR #386, opened 2018-01-31
Windows: Fix race condition between submit and handle events on windows

patstew:
Check the event object for completion in poll instead of using HasOverlappedIoCompleted
When we submit a transfer on windows the fd is added to the poll list before the read/write begins, so HasOverlappedIoCompleted can be called before overlapped.Internal is set to ERROR_IO_PENDING if events are being being handled concurrently, so the fd will be marked as completed before it has actually started.

### DIFF
--- a/libusb/os/poll_windows.c
+++ b/libusb/os/poll_windows.c
@@ -149,8 +149,7 @@ static int check_pollfds(struct pollfd *fds, unsigned int nfds,
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


UniversalSerialBummer:
Can confirm issue.
Patch works for me. Thanks Patstew.

dickens:
Applied with a slight modification. Thanks!

Issue #394, opened 2018-02-21
wIndex always set to 0 with libusb_control_transfer()


kuro68k:
I'm trying to debug an issue with dfu-util and I think it's a problem with libusb.

dfu-util calls libusb_control_transfer() with wIndex set to 1, but my device (a custom USB device) receives the packet with wIndex set to 0. The rest of the packet is fine, it's just that field that is wrong.

I verified with Network Monitor packet capture and a logic analyzer that the sent byte is in fact zero.

```
libusb_control_transfer(
	0x005b7800,		// handle
	0xA1,			// bRequestType
	DFU_GETSTATUS,		// bRequest (0x03)
	0,			// wValue
	1,			// wIndex
	buffer,			// data*
	6,			// wLength
	5000);			// timeout
```

The captured packet is:

```
   - ControlTransfer: 
    + Urb: Function = URB_FUNCTION_CONTROL_TRANSFER, Length = 6, Flags 0xb
    - SetupPacket: 
     + bmRequestType: (Class request) 0xa1
       bRequest: 0x3
       wValue: 0 (0x0)
       wIndex: 0 (0x0)
       wLength: 6 (0x6)

A1 03 00 00 00 00 06 00 
```

I was only able to test on Windows, I don't have a suitable Linux box at the moment. wIndex is usually used as the interface number for control transfers, and it's unusual to use values other than zero but in this case it's necessary for a runtime DFU interface.

pbatard:
You are most likely running into the WinUSB limitation that is documented in the [Know Restrictions](https://github.com/libusb/libusb/wiki/Windows#Known_Restrictions) for Windows on the Wiki:

> With WinUSB, when `LIBUSB_RECIPIENT_INTERFACE` is used for the transfer, the WinUSB DLL forces the low byte of wIndex to the interface number, regardless of what you set it to.

You will have to use a different driver than WinUSB. However, be mindful that libusb can only go as far as the underlying driver allows it to, and there are restrictions associated with the use of each driver, so it's always a good idea to check the documentation before opening an issue.

As I believe your question should have been answered, I will close this request. Feel free to comment further if you think this is not the case.

kuro68k:
Ah yes, that's most likely it. Thanks, I will look at work-arounds.

kuro68k:
[...]

PR #398, opened 2018-02-26
Fix control transfer interface selection

kuro68k:
When sending a control transfer to an interface (via the recipient in bmRequestType) WinUSB requires wIndex to match the interface passed in to ControlTransfer(). Previous behaviour was to always use the first valid interface (always 0?). Now when the target is an interface the wIndex value is used as the target interface number.

The USB spec requires wIndex to be the interface number when the target is an interface in bmRequestType. It is possible some buggy software does not do that, in which case it would not work on WinUSB but might work okay under Linux. As such I don't think making this change will break anything that isn't already broken.

This addresses issue https://github.com/libusb/libusb/issues/394

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2308,6 +2308,7 @@ static int winusbx_claim_interface(int sub_api, struct libusb_device_handle *dev
 				return LIBUSB_ERROR_ACCESS;
 			}
 		}
+		handle_priv->interface_handle[iface].dev_handle = handle_priv->interface_handle[0].dev_handle;
 	}
 	usbi_dbg("claimed interface %d", iface);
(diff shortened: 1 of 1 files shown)


madcowswe:
We have a project that uses libusb to do firmware update with DfuSe. We have people getting [weird issues](https://github.com/madcowswe/ODrive/issues/179) that seems to vary between machines, and we have not been able to track down a pattern.

Following from [this ticket](https://sourceforge.net/p/dfu-util/tickets/55/) I would like to lend my support to merging this fix, so it can make it into a pyusb release.
We don't use dfu-util, but the [implementation we use](https://github.com/madcowswe/ODrive/blob/master/tools/odrive/dfuse/DfuDevice.py#L47) also uses control transfers with libusb.

Thanks.

PR #401, opened 2018-03-04
S1 fastboot adbusb.inf

jc46:
android s1 fastboot adbusbdriver,inf


mcuee:
Why this has anything to do with libusb?

PR #414, opened 2018-03-30
Move to brew bundle

stephengroat:
improved dependency management with brew bundle

`bundle` allows for the package to be upgrade or install, doesn't return 1 if package is already installed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,9 +41,8 @@ addons:
             - ubuntu-toolchain-r-test
 
 before_install:
-    # brew upgrade sets the return code to 1 if the latest version is already installed. the true command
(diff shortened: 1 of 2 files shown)


stephengroat:
the `.travis.yml` could also be updated with `brew bundle cleanup --force` [to remove unneeded preinstalled `brew` formulas](https://github.com/Homebrew/homebrew-bundle#cleanup), which include `libusb`

hjelmn:
Please squish all of this into a single commit.

stephengroat:
sorry about that, now a single

hjelmn:
Ok, looks good. Replacing by #451 which increments the nano version.

PR #423, opened 2018-04-27
Adding support for ARM & ARM64 Windows Platform

DjArt:
Configuring projects for compilation for these platforms. Only vs2017, because ARM64 supported in these version only.


mcuee:
Just wondering what is the platform you use to test. Is it the Snapdragon 835 based system?

DjArt:
No. I tested that on Tegra 3/4, Qualcomm 808, 810 platforms.

DjArt:
And BCM2837 too.

Issue #290, opened 2017-04-10
Last version (1.0.21/1.022) does not list all connected devices on macOS Sierra


amok:
On master branch (seems version 1.0.21) device (STM32VLDiscovery board in my case) doesn't show up in list of connected devices, but on tag v1.0.9 it does.

Related to texane/stlink/issues/574 ([and comment](https://github.com/texane/stlink/issues/574#issuecomment-291313832))

```
➜  ~ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.3
```

libusb@master devices list

```
➜  libusb git:(master) ./examples/testlibusb
Dev (bus 20, device 5): Broadcom Corp. - Bluetooth USB Host Controller
```

libusb@v1.0.9 devices list

```
➜  libusb git:(f98eaca) ✗ ./examples/testlibusb
Dev (bus 20, device 0): 05AC - 8007
Dev (bus 20, device 5): Broadcom Corp. - Bluetooth USB Host Controller
Dev (bus 20, device 10): STMicroelectronics - STM32 STLink

➜  libusb git:(f98eaca) ✗ ./examples/listdevs
05ac:8007 (bus 20, device 0)
05ac:8290 (bus 20, device 5)
0483:3744 (bus 20, device 10)
```

Please let me know if you need some more debugging info.

xor-gate:
I think the device is bound to a kernel driver as mentioned here:

https://github.com/libusb/libusb/wiki#supported-environments

> Note: the device should not be bound to a kernel driver. You can use a codeless kext to achieve that if there is a kernel driver (eg: HID or USB Mass Storage) which already binds to the device.

marc-groundctl:
This also worked in 1.0.20.  It appears to be caused by 871eb29 and b7526c1, and can be worked around by applying the following patch:

```diff
diff --git a/libusb/os/darwin_usb.c b/libusb/os/darwin_usb.c
index 9946222..42fcc6e 100644
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -516,7 +516,7 @@ static void darwin_check_version (void) {
   errno = 0;
   version = strtol (version_string, NULL, 10);
   if (0 == errno && version >= 15) {
-    darwin_device_class = "IOUSBHostDevice";
   }
 }

@@ -882,7 +882,7 @@ static int get_device_parent_sessionID(io_service_t service, UInt64 *parent_sess

[...]

PR #428, opened 2018-05-21
Fix for Apple devices not showing in normal mode

matteyeux:
Hey there, here is a fix for Apple devices not showing in normal mode (device is detected in Recovery/DFU mode).

People were talking about this patch in #290 but has not been submitted yet

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -528,7 +528,6 @@ static void darwin_check_version (void) {
   errno = 0;
   version = strtol (version_string, NULL, 10);
   if (0 == errno && version >= 15) {
-    darwin_device_class = "IOUSBHostDevice";
   }
 }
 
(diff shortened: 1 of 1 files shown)


hjelmn:
The new IOUSBFamily code has been a complete disaster. It was buggy as hell from the getgo and required a bunch of odd changes (IOUSBDevice -> IOUSBHostDevice). Now it looks like IOUSBHostDevice doesn't catch everything (WTF Apple). This commit appears to essentially get us back to detecting the devices in the same way as before the IOUSBFamily rewrite. I don't know if this will break an older macOS release (between the rewrite and now).

I am tempted to just take this PR and fix the older releases if they break.

hjelmn:
For reference. I didn't say how it was broken but detection was broken in El Capitan: b7526c19a57e9c584bdc689fa2447804580dbf94

PR #430, opened 2018-05-24
transfer_lock should be released before usbi_handle_transfer_completi…

pyhalov:
…on() call

### DIFF
--- a/libusb/os/sunos_usb.c
+++ b/libusb/os/sunos_usb.c
@@ -1360,9 +1360,15 @@ solaris_submit_ctrl_on_default(struct libusb_transfer *transfer)
 	}
 	usbi_dbg("Done: ctrl data bytes %d", ret);
 
-	/* sync transfer handling */
+	/**
+	 * Sync transfer handling.
+ 	 * We should release transfer lock here and later get it back
+	 * as usbi_handle_transfer_completion() takes its own transfer lock.
+	 */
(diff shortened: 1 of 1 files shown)


libusb/os/sunos_usb.c
```diff
@@ -1361,6 +1361,7 @@ solaris_submit_ctrl_on_default(struct libusb_transfer *transfer)
 	usbi_dbg("Done: ctrl data bytes %d", ret);
 
 	/* sync transfer handling */
+	usbi_mutex_unlock(&LIBUSB_TRANSFER_TO_USBI_TRANSFER(transfer)->lock);
 	ret = usbi_handle_transfer_completion(LIBUSB_TRANSFER_TO_USBI_TRANSFER(transfer),
```
hjelmn: Where is the corresponding lock? Why is unlock not called on error conditions. Does the lock need to be re-obtained after calling ```usbi_handle_transfer_completion()```?

Just want to make sure I understand what is wrong.

pyhalov: This lock is taken in libusb_submit_transfer(), which calls usbi_backend.submit_transfer() , which in its turn calls solaris_submit_ctrl_on_default(). 

usbi_handle_transfer_completion() takes its own lock on itransfer->lock, so without this we hang.

pyhalov: We return immediately after usbi_handle_transfer_completion. Perhaps, we should re-obtain lock after calling usbi_handle_transfer_completion(), have to check.

PR #433, opened 2018-05-24
Use more portable di_prop_lookup_ints() interface

pyhalov:


hjelmn:
Out of curiosity, how is ```di_prop_lookup_ints()``` more portable? I see nothing in the documentation. I don't use Solaris so I have no way to validate that this is a correct patch.

pyhalov:
I don't know if di_prop_exists() is available in Oracle Solaris 10, but it's definately not available in illumos.

PR #435, opened 2018-05-24
Don't use list_* functions, when they are not available

pyhalov:
Solaris port uses list-management functions, like list_create(),
list_remove_head() and so on. They are not available
for userland applications on illumos.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -161,6 +161,25 @@ sunos)
 	LIBS="-pthread -ldevinfo"
 	AC_CHECK_HEADERS([poll.h])
 	AC_DEFINE([POLL_NFDS_TYPE],[nfds_t],[type of second poll() argument])
+	AC_MSG_CHECKING([whether public list API is available])
+	AC_LINK_IFELSE([AC_LANG_PROGRAM(
+	[[#include <sys/list.h>
+		#include <stdlib.h>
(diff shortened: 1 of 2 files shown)


pyhalov:
So far I have no Solaris system, which implements these interfaces, so it would be interested to test correctness of detection.
@alanc

jwrdegoede:
libusb has its own copy of the linux-kernel list code which is used elsewhere, see the list_ declarations in libusb/libusbi.h.

IMHO it would be better to simply port the libusb/os/sunos_usb.c code to use that list code.

hjelmn:
I agree with @jwrdegoede. @pyhalov can you make the requested changes?

kentsuji:
Please look at the report #586 I submited.  I attached my sample fix for this issue in the report.

Mno-hime:
With https://github.com/libusb/libusb/pull/627 merged perhaps this can be closed?

kentsuji:
I confirmed that tests and examples could be successfully built on OpenIndiana with #627 change.  No reference to list_* functions in libusb-1.0.so.

Mno-hime:
Thanks @kentsuji, I didn't know there are tests. The `stress` program works fine on OpenIndiana with #627 changes.

PR #439, opened 2018-06-05
configure.ac: fix detection of clock_gettime library

baruchsiach:
glibc before 2.17 requires link with librt for clock_gettime(). The
AC_SEARCH_LIBS check in configure.ac should detect this dependency.
Unfortunately commit cb77a25e51 (configure.ac: Remove obsolete AC_ERROR
and make formatting consistent) inadvertently renamed to clock_gettime2,
thus breaking librt detection.

Restore the correct clock_gettime() name.

Signed-off-by: Baruch Siach <baruch@tkos.co.il>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -107,7 +107,7 @@ case $backend in
 linux)
 	AC_DEFINE(OS_LINUX, 1, [Linux backend])
 	AC_SUBST(OS_LINUX)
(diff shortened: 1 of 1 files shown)


hjelmn:
Nice. This one should be good to go. Will push now.

PR #446, opened 2018-06-18
Add German translation

alexander255:
Most strings were copied from the german `strerror` output on equivallent `errno` values and then adapted as necessary.


hjelmn:
Verified that this looks ok using Google Translate.

PR #452, opened 2018-07-09
Fix BOS enumeration in testlibusb.c

fabiensanglard:
The code for BOS did not account for multiple entries. Added a loop.


hjelmn:
You might have to add a ```#include <stdint.h>``` line to make windows happy. I have no idea why it doesn't like uint8_t (other than Microsoft was last to implementing any of C99).

hjelmn:
Hmm, on closer inspection it looks like the MS compiler doesn't like ```for (type foo = ;...)``` statements. It's a pain but it looks like you will have to move the variable declaration up to the top of the function.

PR #473, opened 2018-09-18
Allow set configuration control requests when using libusb0.sys

qmfrederik:
`libusb0.sys` supports setting the active configuration on Windows. (It is the only open source Windows driver I know of that does.)

This PR allows sending of the set configuration control request when using `libusb0.sys`.

This PR is intended to be used together with https://github.com/mcuee/libusbk/pull/15


qmfrederik:
Related: #94

qmfrederik:
Awesome, thanks!

PR #476, opened 2018-09-20
Attempt to load libusbk.dll from the same directory as libusb-1.0.dll

qmfrederik:
There are scenarios (such as providing native libraries for .NET Core applications using NuGet packages) where `libusb-1.0.dll` may be loaded from a directory which is not in the default search path.

In these scenarios, libusb-1.0 will fail to load libusbK if libusbK is located in the same directory as libusb-1.0, because that directory is not in the search path.
libusbK may be located from another location (such as C:\Windows\system32, for example), causing unexpected behavior.

This PR amends the code which probes for libusbK to add the directory in which libusb-1.0 is located to the DLL search path.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1996,6 +1996,49 @@ static int winusbx_init(struct libusb_context *ctx)
 	LibK_GetProcAddress_t pLibK_GetProcAddress = NULL;
 	LibK_GetVersion_t pLibK_GetVersion;
 
+	// Try to load libusbK.dll. We'll attempt to look for libusbK.dll in the same directory as libusb-1.0.dll,
+	// so first find the path of libusb-1.0.dll and then try to load the library from there.
+	// SetDllDirectoryA is available on Windows XP SP1 and above, so don't attempt to do this on Windows XP RTM or
+	// earlier
+#if NTDDI_VERSION >= NTDDI_WINXPSP1
+	char path[MAX_PATH];
+	char drive[_MAX_DRIVE];
+	char dir[_MAX_DIR];
+	HMODULE hm = NULL;
+
+	if (GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
+		GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+		(LPCSTR)&winusbx_init, &hm) == 0)
+	{
+		int ret = GetLastError();
+		usbi_warn(ctx, "GetModuleHandle failed, error = %d\n", ret);
+	}
+	else
+	{
+		if (GetModuleFileNameA(hm, path, MAX_PATH) == 0)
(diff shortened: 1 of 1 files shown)


qmfrederik:
>  Normaly libusbk.dll shall have been installed separately and the DLL be in system path. 

That's part of the problem, anyone can drop libusbk.dll in C:\Windows\system32, and I prefer using a copy of libusbk.dll that I ship with my application (and control) over a copy in system32 that can come from anywhere.

> Also libsub can be static then how does it end up? That is my typical use case to exactly avoid the DLL loadbpath issue

libusb can be static but it will dynamically load libusbk.dll in winusbx_init using LoadLibraryA.
If you statically link libusb, this patch will cause the code to look for libusbk.dll in the directory of your executable.

> My take is that your not adressing the intial problems why is the DLL not in system path! Or in a place that satisfy load rule where you loss at first the exe path then cwd then system path.

libusbk.dll is a user-mode library and I want to load it from the directory in which my app is installed to avoid [DLL Hell](https://en.wikipedia.org/wiki/DLL_Hell#DLL_stomping). Not depending on libusbk.dll in the system path is an advantage for me.

> If you need to explicitly load that DLL then you shall chip it with your app exe (where you shalk have also place lubusb dll) and it will be loads from here.

Not always, .NET Core and NuGet packages is one example where libusb and libusbk would not be in the same directory as the app exe.

Similarly, consider you have an app like this:

```
C:\myapp\myapp.exe
C:\myapp\libusb- 1.0.dll
C:\myapp\libusbk.dll
```

and also have a copy of (a different version of) libusbK in win32:

```
C:\Windows\system32\libusbk.dll
```

Now if you do this:

```
cd C:\myapp
myapp.exe
```

you'll end up with libusbk.dll from `C:\myapp\libusbk.dll`

But if you do:
```
cd \
C:\myapp\myapp.exe
```

you'll end up with libusbk.dll from `C:\Windows\system32\libusbk.dll`.

That's something I want to avoid. I always want to use `C:\myapp\libusbk.dll`.

> At worst first do a load without any abs path then only itbif fail try your mod so you do not ceate any bacward incompatibity.

This will give preference to libusbk.dll in C:\Windows\system32, something I explicitly want to avoid. I prefer my copy of libusbk.dll.

diabolo38:
the rule is dll shall be in  the  app folder  or in system path.
if not deal with it ... for instance your app  can pre-load  libusk  dependencies prior to load libusb 
from whatever abs path you knows (argv[0] base registry etc ...) 
so when libusb  need to load dll whatever  path it use it will  succees cos it is already loaded in memory. That is one solution we also used (we preload all  dll fep from specific app sub- folder) prior to load  the depandant  dll

Again to me but i may be wrong  you try to push as solution to an  app problem into the lib . that will create more issue in all existing app   when it is so easy to solve it in your app...

mcuee:
Looks like not a libusb problem. Closing this for now.

PR #486, opened 2018-10-23
Core: new API (libusb_set_log_cb) to redirect log messages.

dmitrykos:
The proposed API allows to redirect all log messages to a custom log handler function that can be useful in situations when libusb is used by the application with own centralized logging (which can log as to console, so to the file and etc.). These logs can be collected from the users of the application for the debugging purpose of some failures/problems and thus having possibility to redirect logs is very important for such case.

The implementation does not require ```libusb_init``` be called forehand and if callback is provided before the ```libusb_init``` then logs from the possible failure to initialize will also be forwarded to the callback function that can be useful in debugging the reason of the failure.

The usage example can be as such:
```c
void my_log_handler(enum libusb_log_level level, const char *str)
{
...
}

...

int r;

libusb_set_log_cb(my_log_handler);

if ((r = libusb_init(&context)) < 0)
{
    USBErrLog("libusb_init() failed with error: %d", r);
    return false;
}

...
```

It is a refactored log redirection proposal from this pull request:
http://github.com/libusb/libusb/pull/190

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -51,6 +51,9 @@ static const struct libusb_version libusb_version_internal =
 static int default_context_refcnt = 0;
 static usbi_mutex_static_t default_context_lock = USBI_MUTEX_INITIALIZER;
 static struct timespec timestamp_origin = { 0, 0 };
+#ifndef USE_SYSTEM_LOGGING_FACILITY
+static libusb_log_cb log_handler = NULL;
+#endif
 
 usbi_mutex_static_t active_contexts_lock = USBI_MUTEX_INITIALIZER;
 struct list_head active_contexts_list;
@@ -424,6 +427,7 @@ if (cfg != desired)
   * - libusb_set_auto_detach_kernel_driver()
   * - libusb_set_configuration()
   * - libusb_set_debug()
+  * - libusb_set_log_cb()
   * - libusb_set_interface_alt_setting()
   * - libusb_set_iso_packet_lengths()
   * - libusb_set_option()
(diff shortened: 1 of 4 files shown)


hjelmn:
@mcuee @LudovicRousseau @dickens  Any comments on this? I see no reason not to add this API call. Any objections?

Youw:
I think it make sense to have both options - global log redirect, and per-context, as in #190.

LudovicRousseau:
No objection from me.

Keep in mind that programs using this new API will not be able to use an older version of libusb. The shared library visioning on Linux should reflect that. Or something like that.

hjelmn:
@LudovicRousseau Yes, indeed we need to update the library triplet. Will include that when this is merged.

@dmitrykos Can you update this to also support setting the log redirect per-context as @Youw requests?

dmitrykos:
@hjelmn, it would be great if @Youw could clarify what is rationale behind the per context logging? I would like to remind that per context logging has major weakness - it is not possible to log ```libusb_init``` failure (or behavior) with per context logging because ```libusb_init``` creates this context object.

@Youw would you please clarify the usage case with per context logging?

Youw:
I have a several physical devices (some of them are optional) that are treated as a single logical device.
For each logical device I create a separate context, and I'd like not only store libusb logs, but also distinguish libusb logs related to different devices.

@dmitrykos, I totally understand your concern, and I don't suggest replacing global log redirect with per-context one.
I see no reason, why there couldn't be _both_, like:
```cpp
if (ctx && ctx->log_handler != NULL)
    ctx->log_handler(level, str);
else if (log_handler != NULL)
    log_handler(level, str);
else
    fputs(str, stderr);
```
Or, even non exclusive:
```cpp
if (ctx && ctx->log_handler != NULL)
    ctx->log_handler(level, str);

if (log_handler != NULL)
    log_handler(level, str);
else
    fputs(str, stderr);
```

The fact that log level can be setup on context level, makes another good reason, why log redirect on context makes perfect sense too.

dmitrykos:
@Youw thank you very much for the clarification. Ok then I propose to modify ```libusb_set_log_cb(libusb_log_cb callback)``` to ```libusb_set_log_cb(libusb_context *ctx, libusb_log_cb callback)``` where if ```ctx``` is ```NULL``` then callback will be set as global log handler and if not ```NULL``` then as per context log handler. So there will be just one API function to set log callback.

@hjelmn would it be suitable for libusb API or I should split it to 2 independent functions, would you please comment?

hjelmn:
@dmitrykos A single function is sufficient. Just want to make sure we cover the use cases before adding any APIs. Once you push the update I will take a look again.

Youw:
> `libusb_set_log_cb(libusb_context *ctx, libusb_log_cb callback)`
Looks very good to me.

dmitrykos:
@hjelmn my proposal for 

``` c 
libusb_set_log_cb(libusb_context *ctx, libusb_log_cb callback)
``` 

is not sufficient because if ```ctx``` is ```NULL``` then libusb assumes that user is working with default context which is ```usbi_default_context``` and thus I propose to add ```global``` parameter which will explicitly set log handler as global handler.

``` c 
libusb_set_log_cb(libusb_context *ctx, libusb_log_cb callback, int global)
``` 

So, we get use cases:

1.  **Global log handler**: ```global``` is ```1``` then ```callback``` is always set as global log handler, ```ctx``` is ignored.
1.  **Default context log hanler**: ```ctx``` is ```NULL``` then ```callback``` is set to the default context (if it exists), ```global``` is ignored.
3.  **Context log handler**: ```ctx``` is ```not NULL``` then ```callback``` is set to the default context (if it exists), ```global``` is ignored.

It looks to me that all cases are covered. If this proposal is ok I will proceed with implementation.

LudovicRousseau:
I propose to rename the `global` parameter into `mode` and use it as a bitmap field with values like `LIBUSB_CB_GLOBAL` for a global callback.

This opens the possibility to expand the function to other modes if needed.

dmitrykos:
@LudovicRousseau thank you for your proposal it looks good to me. So it would be then

``` c 
enum libusb_log_cb_mode {
	LIBUSB_LOG_CB_GLOBAL= 1 << 0,
	LIBUSB_LOG_CB_CONTEXT = 1 << 1
};

libusb_set_log_cb(libusb_context *ctx, libusb_log_cb callback, int mode)
``` 

or you propose to have generic enumeration for other possible uses in other functions, so it would not be just for use with ```libusb_set_log_cb```

``` c 
enum libusb_cb_mode {
	LIBUSB_CB_GLOBAL= 1 << 0,
	LIBUSB_CB_CONTEXT = 1 << 1
};
``` 

?

Would you please clarify.

Youw:
Another suggestion from my side:
```cpp
typedef void (LIBUSB_CALL *libusb_log_cb)(libusb_context *ctx, enum libusb_log_level level, const char *str);
```
Again, this is to be able to distinguish different contexts on the application side.

dmitrykos:
@Youw, yes ```libusb_context``` will be in the callback, similar how it was in the initial per context implementation (or just like you proposed).

Will be waiting for the clarification from @LudovicRousseau regarding enum in order to proceed with changes.

LudovicRousseau:
Since the constants are for log only the name `LIBUSB_LOG_CB_...` is better for me.

hjelmn:
Looks good. Use the enum and the naming @LudovicRousseau suggested and this will be good to go.

dmitrykos:
The changes are implemented and committed as per our productive discussion.

hjelmn:
Great. Thanks! Please squish this all down to a single commit and I think we are good to go.

dmitrykos:
@hjelmn looks like all is done finally!

PR #495, opened 2018-11-13
add simply CMakeLists.txt

seker:


hjelmn:
Again, probably not.

seker:
> Again, probably not.

but，why？？？

seker:
> Again, probably not.

Maybe you can let me to maintain the cmake configuration ?

hjelmn:
Not really interested in that.
  1) There is no  guarantee you will stick around after the merge,
  2) others will be expected to pick up support because "its in the repo",
  3) I see no value-add. We have MSVC and xCode projects embedded and this is a small enough project that those are not difficult to maintain, and
  4) those who imbed libusb inside their project can easily write their own CMake or whatever file.

I am no fan of CMake so unless I get overruled or something changes to make CMake actually useful then this will likely stand.

Youw:
> 4. those who imbed libusb inside their project can easily write their own CMake or whatever file

Yes, and that's what we (users of libusb) do - not the most user-friendly integration strategy.

> I see no value-add. We have MSVC and xCode projects

Both xCode and MSVC project files are not intended to be read by humans. For those who needs their own CMake integration - takes time to understand what are required compilation keys/defines/source files.

Also, you need to maintain _multiple_ projects. A simple case (no examples or tests):
- autotools;
- libusb.xcodeproj;
- libusb_dll.dsp;
- libusb_dll_2005.vcproj;
- libusb_dll_2012.vcxproj;
- libusb_dll_2013.vcxproj;
- libusb_dll_2015.vcxproj;
- libusb_dll_2017.vcxproj;
- libusb_dll_wince.vcproj;
- libusb_static.dsp;
- libusb_static_2005.vcproj;
- <and 5 more static msvc>;

16 project files and 3 different build systems. Maybe that's not difficult to maintain (even though personally I cannot agree with such statement), but that looks like a mess.
Also, three different `config.h` - does not makes things easier.

All these projects can be replaced by a single `CMakeLists.txt file`. Of course it will contain platform specific sections, but mostly because of platform-specific sources, and compiler flags (in some cases).

Personally I think there is almost everything is wrong regarding current PR (that would never pass my code-review).

But I don't see any great reason not to consider migration to CMake as a long-term strategy. It will take effort to perform the migration (I could do it in a few days, for all platforms), and I didn't take into account other platforms (like Android), and already existing external integration, but all of the above can be generated by CMake (including pkg-config) in a fashion manner.

hjelmn:
I see no reason to migrate away from autotools. CMake provides no benefit, in my opinion, for any platform other than Windows.

hjelmn:
I should also point out you can not embed libusb inside another project without making it GPL. You can, however, build libusb and link Amity project again it.

Youw:
> You can, however, build libusb and link any project again it.

Don't see what is the relation to current discussion, but thanks.

At least consider generating [cmake-package](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html) for CMake users, to make our life easier.

hjelmn:
@Youw Is pkg-config not sufficient? Kind of strange that we would need to generate anything since with autotools adding libusb checks is trivial.

```
PKG_CHECK_MODULES([libusb], [libusb-1.0])
```

But I guess this is just another one of the idiosyncrasies of CMake that will prevent it from ever being anything but a solution in search of a problem on Unix-based systems :p.

Youw:
pkg-config is not available by-default on Windows or macOS.

Even when it is, that is not a perfect solution. If you need to deliver you application to end-user machine. You have to find where the shared library physically located and bundle it to your installer. Much easier to have it as a part of your build system.

Also, I had multiple projects which where Windows/macOS only, and I did some conclusions quite a while ago:
- macOS is barely Unix-like from developer perspective (hello Objective-C++);
- without CMake, you life feels like pain;
- Unix way is good for Unix-like systems only :)

Youw:
>  Is pkg-config not sufficient?
After giving it a second thought, maybe it would work as good as cmake-packages.

hjelmn:
Being a macOS developer myself I have to disagree. It's very Unix-like and all the standard tools work well on it even if they are not installed by default. Installing is painless with macports or homebrew. Once it is up its as simple as brew install libusb then setting up my xCode project to slurp the libusb.dylib into the Resources folder of my app. CMake is totally useless, but that is IMHO.

hjelmn:
And Objective-C++ is just another language that is also supported by autotools. Nothing special there.

Youw:
For reference, to help those who are trying to google this one: now there is a community-supported https://github.com/libusb/libusb-cmake

PR #505, opened 2018-12-12
windows: get usb device until usb configuration is set

nxpfrankli:
There are low possiblity to open device before windows complete usb
configuration. Then claim interface will be always failure.
this patch workaround this problem, make sure return device until
windows finish usb configuration.

Signed-off-by: Tom Zheng <haidong.zheng@nxp.com>


nxpfrankli:
Any update about this pull request?

PR #506, opened 2018-12-15
Added USBI_PRINTFLIKE to allow finding printf format string mistakes

seanm:
Fixed one such misktake

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -2598,7 +2598,7 @@ int API_EXPORTED libusb_get_next_timeout(libusb_context *ctx,
 		timerclear(tv);
 	} else {
 		timersub(&next_timeout, &cur_tv, tv);
-		usbi_dbg("next timeout in %d.%06ds", tv->tv_sec, tv->tv_usec);
+		usbi_dbg("next timeout in %ld.%06lds", (long)tv->tv_sec, (long)tv->tv_usec);
 	}
 
 	return 1;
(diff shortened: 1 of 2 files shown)


LudovicRousseau:
OK for me

hjelmn:
Looks good. Will pull it in.

hjelmn:
Still planning to merge this and several others. Been busier than expected the last couple weeks.

seanm:
Don't take this the wrong way, but: you and another core developer have both already reviewed the code. It appears everyone agrees it's acceptable.

Isn't it just a matter of pressing "Merge pull request" in github?  Is there an additional complication because of that 'nano' version thing?  Perhaps some kind of workflow improvement or additional automation could be beneficial here...?

hjelmn:
Yup. Just the nano thing. Just a cherry-pick by one of us. I will try to do it tonight.

hjelmn:
Might be github hooks we can use but it takes me about 10 mins to close out 10 PRs. Would take longer to find the right github hook documentation :).

seanm:
Thanks for the explanation.

I'd argue it's better to figure out the right github hook, though I grant that'll likely take more time in the short term.

I can think of a few downsides to batching:
- it increases the time between someone writing a patch, and other people testing it in mainline. If someone finds a problem, the patch writer's memory will be farther back.
- it makes it harder to test incrementally. ex: I have a nightly script that builds a bleeding-edge version of my app, that uses that day's code of all my dependencies. If everything appears one day to the next, and something breaks, it's harder to regress.

hjelmn:
Yeah. I need to figure out how to use the github tools to handle this. Will try to find the time in the next couple of months. For now I have the first batch of PRs closed. Will work on more tomorrow with an aim for release in two weeks.

seanm:
Thanks!

PR #507, opened 2018-12-15
Removed unneeded checks for passing null to free()

seanm:


libusb/io.c
```diff
@@ -2755,15 +2752,12 @@ const struct libusb_pollfd ** LIBUSB_CALL libusb_get_pollfds(
  * Since version 1.0.20, \ref LIBUSB_API_VERSION >= 0x01000104
  *
  * It is legal to call this function with a NULL pollfd list. In this case,
- * the function will simply return safely.
+ * the function will simply no nothing.
```
LudovicRousseau: I guess a "return" word is missing here.

libusb/os/darwin_usb.c
```diff
@@ -1675,8 +1675,8 @@ static int submit_iso_transfer(struct usbi_transfer *itransfer) {
 
   struct darwin_interface *cInterface;
 
-  /* construct an array of IOUSBIsocFrames, reuse the old one if possible */
-  if (tpriv->isoc_framelist && tpriv->num_iso_packets != transfer->num_iso_packets) {
+  /* construct an array of IOUSBIsocFrames, reuse the old one if the sizes are the same */
+  if (tpriv->num_iso_packets != transfer->num_iso_packets) {
```
LudovicRousseau: This change should NOT be part of the same PR.

libusb/io.c
seanm: Whoops, I meant "do". :)

libusb/os/darwin_usb.c
seanm: Why? It's the same thing: removing a check for null before calling free().

LudovicRousseau: Exact. Sorry.

libusb/io.c
seanm: Fixed.

seanm:
So with the comment fixed, will you be merging this?

hjelmn:
Looks good to me.

LudovicRousseau:
ok for me

seanm:
So is there something I need to do here?  I don't see any button that would allow me to merge it...

hjelmn:
@seanm We do this all by hand to make sure the nano gets updated. I plan to go through all the PRs today or tomorrow.

PR #510, opened 2018-12-16
Tried to untangle mixing of int, enum libusb_error, kern_return_t, an…

seanm:
…d IOReturn.

- Replace all kern_return_t with IOReturn, they are basically the same, but IOReturn is more common in IOUSBLib
- improve consistency by using the name 'kresult' for all IOReturn-type results, and not using that name for libusb_error return values.
- changed a few functions to use real 'bool' (this file already uses C99)
- replaced some raw '0's with LIBUSB_SUCCESS or kIOReturnSuccess.
- replaced some raw 'int's with either IOReturn or enum libusb_error.
- replaced ambiguous "if (kresult)" with more explicit "if (kresult != kIOReturnSuccess)"
- replaced incorrect use of darwin_to_libusb with darwin_transfer_status

Most of these were found by increasing the compiler warning level.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -98,11 +98,11 @@ static int darwin_release_interface(struct libusb_device_handle *dev_handle, int
 static int darwin_reset_device(struct libusb_device_handle *dev_handle);
 static void darwin_async_io_callback (void *refcon, IOReturn result, void *arg0);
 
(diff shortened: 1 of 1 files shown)


LudovicRousseau:
I would prefer to have different commits for different issues.
and also the compiler warning the commit is fixing.

hjelmn:
@LudovicRousseau I think this is fine. Mostly a cleanup in the Darwin support code.

PR #531, opened 2019-02-07
windows_winusb: Add support for reporting super speed plus devices

hjmallon:
Tested with USB-C 3.1 device on Windows 10 latest with Dell Laptop.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -881,6 +881,8 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 				&conn_info_v2, sizeof(conn_info_v2), &conn_info_v2, sizeof(conn_info_v2), &size, NULL)) {
 				usbi_warn(ctx, "could not get node connection information (V2) for device '%s': %s",
 					  priv->dev_id,  windows_error_str(0));
+			} else if (conn_info_v2.Flags.DeviceIsOperatingAtSuperSpeedPlusOrHigher) {
+				conn_info.Speed = 4;
 			} else if (conn_info_v2.Flags.DeviceIsOperatingAtSuperSpeedOrHigher) {
 				conn_info.Speed = 3;
 			}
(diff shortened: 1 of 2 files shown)


libusb/os/windows_winusb.c
```diff
@@ -898,6 +900,7 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 		case 1: dev->speed = LIBUSB_SPEED_FULL; break;
 		case 2: dev->speed = LIBUSB_SPEED_HIGH; break;
 		case 3: dev->speed = LIBUSB_SPEED_SUPER; break;
+		case 4: dev->speed = LIBUSB_SPEED_SUPER_PLUS; break;
```
philemonf: There are some checks in the code that checks for speed == LIBUSB_SPEED_SUPER. You might want to make sure that they also check for LIBUSB_SPEED_SUPER where this applies.

philemonf: I know that libusb_get_max_iso_packet_size is doing it for instance. There might be other places.

philemonf: PS: l am not a maintainer, so you might want to wait to get a feedback from them.

hjelmn:
I have no way to check if this does what it says it does. If anyone else can verify that would be helpful.

libusb/os/windows_winusb.c
hjmallon: I have added that as a separate possible bug. It is not caused by this change. https://github.com/libusb/libusb/issues/553

hjmallon:
@hjelmn is there anyone about here who can confirm with windows?

Maybe (sorry for tagging - I just pulled names from recent windows commits):
@gogoer , @dmitrykos , @nxpfrankli , @DjArt , @dickens ?

nxpfrankli:
code itself look good. but I have not device to test it.

dmitrykos:
@hjelmn, unfortunately do not have super-speed USB device to test.

libusb/os/windows_winusb.c
philemonf: It's probably just a matter of changing the == to >= if the spec for isochronous packet size has not changed in SuperSpeed Plus (I don't know if it has).

PR #537, opened 2019-02-18
Offer an internal hardware ID for more complex device matching

hjmallon:
* Working on Linux, macOS and WinUSB

On Linux we return `sysfs_dir` which can be used with `udev_device_new_from_subsystem_sysname(,"usb",id)` to get a udev_device. On macOS we return an IOServicePath which can be used with `IORegistryEntryFromPath(,id)` to make an `io_service_t`. On Windows WinUSB we return the `dev_id` which can be used with `CM_Locate_DevNodeA(,id,)` to get device information.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -637,6 +637,40 @@ static int darwin_get_config_descriptor(struct libusb_device *dev, uint8_t confi
   return (int) len;
 }
 
+static int darwin_get_platform_device_id (libusb_device *dev, char *data, int length) {
+  struct darwin_cached_device *priv = DARWIN_CACHED_DEVICE(dev);
+  io_iterator_t deviceIterator;
+  io_service_t service;
+  io_string_t path;
+  kern_return_t kresult;
+  size_t s;
+
+  kresult = usb_setup_device_iterator (&deviceIterator, priv->location);
+  if (kresult != kIOReturnSuccess)
+    return darwin_to_libusb (kresult);
+
+  service = IOIteratorNext (deviceIterator);
+  if (service == IO_OBJECT_NULL) {
+    IOObjectRelease(deviceIterator);
+    return LIBUSB_ERROR_NOT_FOUND;
+  }
+
+  kresult = IORegistryEntryGetPath(service, kIOServicePlane, path);
+
+  IOObjectRelease(service);
+  IOObjectRelease(deviceIterator);
+
+  if (kresult != kIOReturnSuccess)
+    return darwin_to_libusb (kresult);
+
+  s = strlen(path) + 1;
+  if (s > length)
+    return LIBUSB_ERROR_OVERFLOW;
+
+  memcpy(data, path, s);
+  return LIBUSB_SUCCESS;
+}
+
 /* check whether the os has configured the device */
 static enum libusb_error darwin_check_configuration (struct libusb_context *ctx, struct darwin_cached_device *dev) {
   usb_device_t **darwin_device = dev->device;
@@ -697,7 +731,7 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
   } else
     /* not configured */
     dev->active_config = 0;
-  
+
   usbi_dbg ("active config: %u, first config: %u", dev->active_config, dev->first_config);
 
   return LIBUSB_SUCCESS;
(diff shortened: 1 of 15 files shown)


hjmallon:
I believe this needs adding to the .def file for DLL on windows but I do not know how to do that

nxpfrankli:
look like 

libusb_get_platform_id
libusb_get_platform_id@12=libusb_get_platform_id

The compiler also decorates C functions that use the __stdcall calling convention with an underscore (_) prefix and a suffix composed of the at sign (@) followed by the number of bytes (in decimal) in the argument list.

My project need work with difference type device, HID and customized USB device.  It work well at Windows and linux. but face issue at MAC OS.  MAC OS's HID driver can't unload easily.  Maybe HIDAPI is quick solution.  I need match port between HIDAPI and libusb. May this API can help me much. 

Perfectly, I hope libusb can support detach kernel driver in MAC OS or work as like windows. Added another hook for HID.

libusb/os/darwin_usb.c
```diff
@@ -637,6 +637,40 @@ static int darwin_get_config_descriptor(struct libusb_device *dev, uint8_t confi
...
+
+  s = strlen(path) + 1;
+  if (s > length)
+    return LIBUSB_ERROR_OVERFLOW;
+
+  memcpy(data, path, s);
+  return LIBUSB_SUCCESS;
+}
+
 /* check whether the os has configured the device */
```
hjelmn: Seems a little overkill to go through the IO registry again. Would be better to just cache the path when the device is detected.

hjelmn:
I am not sure I understand the benefit of this PR. How does exposing internal location data help with matching devices?

hjmallon:
We are using libusb for device searching and communicating. One question we needed to solve was 'which disk is in this usb storage device?'. To answer that question I needed to get an identifier from the libusb device that could then be used for matching the device.

On macOS I had to get the IORegistry node of the usb device and traverse the children looking for kIOBSDNameKey
On Linux I had to get the udev device of the usb device and iterate the block devices, matching if the block device had the usb device as a parent
On Windows I had to use CM_Locate_DevNodeA to get the device and CM_Get_Child to get the disk, followed by a HUGE amount of messing about to turn that into a PhysicalDrive string

All of these methods required access to the initial matching identifier string of the usb device and having my method rescan the usb seemed unnecessary when libusb had many of those identifiers internally already.

mcuee:
This seems to be a popular request.

mcuee:
@hjmallon 

Just wonder if you are still interested in getting this PR updated?

mcuee:
> My project need work with difference type device, HID and customized USB device. It work well at Windows and linux. but face issue at MAC OS. MAC OS's HID driver can't unload easily. Maybe HIDAPI is quick solution. I need match port between HIDAPI and libusb. May this API can help me much.
> 
> Perfectly, I hope libusb can support detach kernel driver in MAC OS or work as like windows. Added another hook for HID.

This is an old post. But just to make it clear, there will be NO HID backend for libusb under macOS. HIDAPI is the way to go.

mcuee:
> We are using libusb for device searching and communicating. One question we needed to solve was 'which disk is in this usb storage device?'. To answer that question I needed to get an identifier from the libusb device that could then be used for matching the device.
> 
> On macOS I had to get the IORegistry node of the usb device and traverse the children looking for kIOBSDNameKey On Linux I had to get the udev device of the usb device and iterate the block devices, matching if the block device had the usb device as a parent On Windows I had to use CM_Locate_DevNodeA to get the device and CM_Get_Child to get the disk, followed by a HUGE amount of messing about to turn that into a PhysicalDrive string
> 
> All of these methods required access to the initial matching identifier string of the usb device and having my method rescan the usb seemed unnecessary when libusb had many of those identifiers internally already.

Indeed there are requests to match  'which disk is in this usb storage device?' and 'which serial port is linked to the USB to Serial devices', etc.

libusb/os/darwin_usb.c
Youw: BTW: something we've learned in HIDAPI - RegistryPath is not always unique. We've seen different devices with the same path, but different Registry IDs:
https://github.com/libusb/hidapi/commit/4d63a5085ac6e6a5be6f0f29a42a4e739da30881
Not sure is this is applicable to LIBUSB (I believe those where BLE devices), but still, something to consider.

mcuee:
Close this in favor of #1617.

PR #539, opened 2019-02-28
Initialize the non-control endpoint fds in netbsd_open()

aymeric-nb:
Not all users of libusb use libusb_claim_interface(), and in that
case, the NetBSD backend fails to open() the necessary /dev/ugenX.Y
and will try to send ioctl()s to fd 0, which fortunately fails.

This unbreaks e.g. openocd on NetBSD.


hjelmn:
Looks fine to me.

PR #559, opened 2019-04-15
Linux non hotplug for containers

jxzgithub:
Dear reviewers and maintainers,

Please refer to the commit message in each patch for issues fixed.

The new added non-hotplug discovery won't be popular on most linux boxes, but will be useful when we running apps in non-privileged containers where uevents are blocked for some reasons like security.

Once this series is accepted, I can also try to turn the conditional compiling into a runtime option so the apps can work on the same libusb release in and outside containers seamlessly.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2356,6 +2356,19 @@ err_unlock:
 	return r;
 }
 
+void usbi_unref_device(libusb_device *dev) {
+	if (!dev)
+		return;
+
+	usbi_mutex_lock(&dev->lock);
+	if (dev->refcnt)
+		--dev->refcnt;
+	usbi_mutex_unlock(&dev->lock);
+
+	if (dev->parent_dev)
+		usbi_unref_device(dev->parent_dev);
+}
+
 /** \ingroup libusb_lib
  * Deinitialize libusb. Should be called after closing all open devices and
  * before your application terminates.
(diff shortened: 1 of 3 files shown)


hjelmn:
Hmm. This changes how devices are enumerated in the netlink (non-udev) case. Is that intentional?

jxzgithub:
> Hmm. This changes how devices are enumerated in the netlink (non-udev) case. Is that intentional?

It is not intentional to replace the existing netlink. I just put some thoughts here.

For the case this series tries to address, I believe the netlink (non-udev) cannot help since it is still based on the blocked events from kernel. But it provides a blocking manner to the applications that this series doesn't cover.

We have to treat these cases respectively. The existing code doesn't seem have a way to tell this 3rd option. I think a new conditional compiling switch like NO_NETLINK can be a short-term answer. It might be depended by udev switch too (I think eventually the udev uses the same source of netlink event from kernel, right?)

Let me know if the above can be acceptable.

hjelmn:
Finally had a chance to take another look. Seems ok to me. Please rebase and resolve any conflicts.

jxzgithub:
Thanks for the feedback. I could need some time to resume this halted work and give it an update.

themarpe:
Hi @jxzgithub - does this PR address the following scenario? https://community.intel.com/t5/Intel-Distribution-of-OpenVINO/NCS2-in-docker-container-on-RPI-results-in-RuntimeError-Can-not/m-p/1191905/highlight/true#M19803

How would one pass the USB device to the docker container with this changes applied? Are the cgroup rules still required or just the volume mount of /dev/bus/usb?

jxzgithub:
@themarpe,
This patch series only deal with the connection/disconnection notification, not covering how a device gets into to container. Refer to the commit msg in the first change. Interestingly, the link solved the same issue this work to address since reconnection works. I guess someone  in Openvino figured it out later.

This work won't be needed.

Thanks

mcuee:
@jxzgithub 
Does the pull request help in this case?
https://github.com/libusb/libusb/issues/595

jxzgithub:
> @jxzgithub
> Does the pull request help in this case?
> #595

I am not sure because it seems a different issue.

mcuee:
@jxzgithub Thanks. I asked because I did not quite understand what you say by "Interestingly, the link solved the same issue this work to address since reconnection works. I guess someone in Openvino figured it out later." What exactly does OpenVino did to sort out the "connection/disconnection notification" notification issue this pull request is addressing?

What about the solution for other containers? Maybe this pull request is still useful.

jxzgithub:
@mcuee,
- " Works fine, even if you remove and reinsert the stick." in the link "https://community.intel.com/t5/Intel-Distribution-of-OpenVINO/NCS2-in-docker-container-on-RPI-results-in-RuntimeError-Can-not/m-p/1191905/highlight/true#M19803" gives me the impression that workaround solved the issue my series fixed without a change in the libusb source, so I ditched my series.

mcuee:
> * " Works fine, even if you remove and reinsert the stick." in the link "https://community.intel.com/t5/Intel-Distribution-of-OpenVINO/NCS2-in-docker-container-on-RPI-results-in-RuntimeError-Can-not/m-p/1191905/highlight/true#M19803" gives me the impression that workaround solved the issue my series fixed without a change in the libusb source, so I ditched my series.

Thanks for the clarification. 

So it is not very clear how they do it but it seems to be possible to fix this without changing libusb.

mcuee:
@jxzgithub In this case, do you want to continue the work, or close the pull-request?

themarpe:
@mcuee @jxzgithub 

I think that the main "issue" is that for containers like Docker in above post, libusb has to be recompiled with udev disabled and linked to app for this workaround to work.

I am under impression that this PR would eliminate this requirement and still be able work within containers by passing the cflags and `/dev/bus/usb`, which would communication with devices using libusb from containers simpler.

mcuee:
@jxzgithub 

jxzgithub:
Sorry. I still don't have much time to come back to this work. Please feel free to take any useful bits in this series. I am going to close this as a clean-up.

Apologize for the long delay of no response.

PR #562, opened 2019-04-17
windows: fix return type of usbi_{inc,dec}_fds_ref.

jmgao:
These functions were declared as returning int, but failed to actually
return anything, and no one was using their return values anyway.

Some compilers will "helpfully" emit a function that just falls through
to whatever happens to be below it (and MSVC seems to reject it entirely?): https://gcc.godbolt.org/z/mUcf2f


hjelmn:
LGTM

PR #565, opened 2019-04-30
Various fixes for the Haiku port

pulkomandy:
- Use native API for getting full configuration descriptors instead of
  using a raw request
- Implement some callbacks that are mandatory and result in crashes if
  not available


pulkomandy:
Hi,
Why was this closed? We would like to get the Haiku support upstream up to date.

Issue #207, opened 2016-09-10
Error C2001: Newline in constant


Grix:
I get this error when trying to build the unmodified libusb_2013.sln VS2013 solution included in your releases. 

Error C2001: Newline in constant

It refers to line 111 of strerror.c, which is part of the russian error message translations. I can't get the library to build without deleting russian from the locale arrays. Not sure if there is a syntax error you missed if or if there is some unicode problems caused by the russian characters on my system.

LudovicRousseau:
It looks like the libusb_2013.sln solution is working fine on the Continuous Integration system: https://ci.appveyor.com/project/LudovicRousseau/libusb

Maybe it is a problem on your side?

PS: I am not a Windows expert or even user.

Grix:
>Maybe it is a problem on your side?

Quite possibly. I wouldn't know how to fix it though

Grix:
It is caused by my system locale it seems. If I change locale to english (US) it works. But my usual locale, Japanese, results in the error. You should  be able to reproduce it if you change locale to Japanese and try building the project.

LudovicRousseau:
The file https://github.com/libusb/libusb/blob/master/libusb/strerror.c#L111 looks like a valid UTF-8 encoded file.
It is strange the compiler complains about the Cyrillic text but not about the French accents in the section just before.

It looks like VS2013 does not handle UTF-8 correctly. Maybe you should report the problem to Microsoft.

undisputed-seraphim:
It fails to compile on VS2015 as well.

However, I have found that it can be very easily resolved simply by putting one space at the back of the two offending strings: [Lines 111](https://github.com/libusb/libusb/blob/master/libusb/strerror.c#L111) and 117.

Kindly consider my suggestion or inform me if you prefer a pull request. This should be a rather painless solution.

mcuee:
What is your system locale? If you use English locale, it should be okay. 
For your solution, have you tested using a few different locale? Say the Japanese locale the original poster mentioned. Thanks.

undisputed-seraphim:
I'm on the Japanese locale as well.

NoviceLive:
Hope this comment helps.

Environment: Windows 8.1 + Visual Studio 2015, with Chinese being the system locale and UTF-8 being the source code encoding.

[...]

PR #566, opened 2019-05-02
Minor changes

RipleyTom:
-On windows compiling with visual studio 2017 results in a compilation error:
strerror.c(111): error C2001: newline in constant
(that's the first russian error message)

Adding the /utf-8 option to compiler parameters seems to fix it(thanks pbatard), fixes https://github.com/libusb/libusb/issues/207.

-I also changed all the $(SolutionDir) to $(ProjectDir) which make including the existing project files into a different solution much easier.

-Added unistd.h for the mingw build(only if USBI_TIMERFD_AVAILABLE is defined) as the CI seemed to have trouble finding close function.

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -31,6 +31,7 @@
 #include <sys/time.h>
 #endif
 #ifdef USBI_TIMERFD_AVAILABLE
+#include <unistd.h>
 #include <sys/timerfd.h>
 #endif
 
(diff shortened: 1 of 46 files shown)


pbatard:
I think you want to combine this with the proposed solution to #207 which is to add `/source-charset:utf-8` (or even better `/utf-8`) to the VC compiler options.

I recently went through [something like this in my Rufus project](https://github.com/pbatard/rufus/commit/7e21f8a60c8d508fef62a13c32e5cb1849dd0757), without having to add BOMs, and it seems to have solved the `error C2001: newline in constant` error once and for all.

RipleyTom:
Thanks a lot, no errors, no warnings, beautiful :).

hjelmn:
LGTM

PR #567, opened 2019-05-03
windows_winusb: Mask wIndex correctly when checking interfaces

leoetlino:
According to the USB 3.0 spec (9.3.4), the interface number is in the
lower 8 bits of wIndex.

Note that wIndex must not be modified because some (non-compliant?)
devices actually use the higher 8 bits for custom data with
recipient-interface requests.


leoetlino:
Looks like this is a regression from #398.

PR #569, opened 2019-05-04
windows_winusb: Set policy ISO_ALWAYS_START_ASAP for libusbK

leoetlino:
In some cases, this is required to avoid eventually getting a
USBD_STATUS_BAD_START_FRAME error back from the Windows USB stack.

This makes the libusbK code match the behaviour of the Linux backend.

It appears that the libusbK backend tried to get this behaviour by
setting StartFrame to 0. However, libusbK docs state that:

"Specifing 0 for KISO_CONTEXT::StartFrame (start transfer ASAP) is
restricted to the first transaction on a newly opened or reset pipe."


riking:
Build failure seems unrelated.

PR #592, opened 2019-07-12
windows: remove total fds (256) limitations

nxpfrankli:
When queue more usb requests, and more devices working at the same time
256's limitation is easy to reach.

This patch remove such limitation of windows version.

dymatic allocate map fd_table.
each time increase 256 entry if request more fds.


nxpfrankli:
build failure should not cause by my patches.  I see previous build also failure.

PR #593, opened 2019-07-12
OpenBSD: Several fixes

worr:
bdc4c15: Any program on OpenBSD calling into libusb would segfault, since `clock_gettime` was a `NULL` pointer due to a struct initialization mistake
9582481: The debug statement in `obsd_clock_gettime` would recurse infinitely
beb12ab: `%d` -> `%zu` in places where a `size_t` was passed in rather than an `int`


seanm:
The AppVeyor failures are on Windows, and so presumably unrelated to this.

Changes LGTM.

@hjelmn can you merge?

PR #603, opened 2019-08-02
Windows: Workaround for composite devices with U2F on Win10 1903

EchterAgo:
On Windows 10 version 1903 and later Windows blocks attempts to open HID devices with a U2F usage unless running as administrator. This patch makes the `composite_open` function ignore when opening the HID part of a composite device fails.

Signed-off-by: Axel Gembe <derago@gmail.com>

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -4115,8 +4115,17 @@ static int composite_open(int sub_api, struct libusb_device_handle *dev_handle)
 		}
 	}
 
-	if (available[SUB_API_MAX]) // HID driver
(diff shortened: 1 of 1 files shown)


EchterAgo:
I wanted to make this conditional based on OS version, but the `get_windows_version` seems to not work correctly. Running Windows 10 in the `testlibusb` program it detects it as Windows 8. When I add a manifest file that makes GetVersionEx return correct values, it detects I'm running Windows 11 or later.

EchterAgo:
Not sure why the AppVeyor build failed. It errors in code I didn't touch. Seems unrelated to this change as other PRs have the same error.

PR #604, opened 2019-08-03
Windows: Fix detection of Windows 10 when manifest declares support

EchterAgo:
On Windows 10 `GetVersionEx` will return version 10.0 when the running application has a manifest that declares Windows 10 compatibility.

https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#supportedOS

When no such manifest is found, the call will report version 6.2, which is Windows 8.

The 6.4 version is returned by early Windows 10 Insider Previews and Windows Server 2017 Technical Preview 1.

Signed-off-by: Axel Gembe <derago@gmail.com>

### DIFF
--- a/libusb/os/windows_nt_common.c
+++ b/libusb/os/windows_nt_common.c
@@ -470,7 +470,8 @@ static void get_windows_version(void)
 	case 0x61: windows_version = WINDOWS_7;	    w = (ws ? "7" : "2008_R2");	  break;
 	case 0x62: windows_version = WINDOWS_8;	    w = (ws ? "8" : "2012");	  break;
 	case 0x63: windows_version = WINDOWS_8_1;   w = (ws ? "8.1" : "2012_R2"); break;
-	case 0x64: windows_version = WINDOWS_10;    w = (ws ? "10" : "2016");	  break;
+	case 0x64: // Early Windows 10 Insider Previews and Windows Server 2017 Technical Preview 1 used version 6.4
+	case 0xA0: windows_version = WINDOWS_10;    w = (ws ? "10" : "2016");	  break;
 	default:
 		if (version < 0x50) {
 			return;


EchterAgo:
Alternatively we might want to use `GetFileInfoVersion` on `kernel32.dll`, which is [one of the methods Python uses](https://github.com/python/cpython/blob/master/Python/sysmodule.c#L1400), should work on all Windows versions the `get_windows_version` function supports and does not require a manifest.

Another alternative is to use `RtlGetVersion`, though we'd have to resolve that symbol form `ntoskrnl.exe`. It also is present starting from Windows 2000 and does not require a manifest.

EchterAgo:
Oh, another note, the 6.3 version for Windows 8.1 is correct when the manifest is present. Without manifest Windows 8.1 also reports version 6.2.

EchterAgo:
A minimal manifest file that declares support for all Windows OS released so far looks like this:

```
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
        <application>
            <!-- Windows Vista -->
            <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/>
            <!-- Windows 7 -->
            <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
            <!-- Windows 8 -->
            <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
            <!-- Windows 8.1 -->
            <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
            <!-- Windows 10 -->
            <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
        </application>
    </compatibility>
</assembly>
```

Name it `myprogram.exe.manifest` and place it alongside `myprogram.exe` to use it. Alternatively it can be embedded in the resource section of an exe file.

EchterAgo:
@dickens as the author of the code, did you actually get a value of 0x64 on Windows 10 from this code? I'm downloading the earliest Windows 10 I have (version 1507) to test it right now.

EchterAgo:
Looks like the first technical preview of Windows Server 2016 had a version of 6.4:

https://en.wikipedia.org/wiki/Windows_Server_2016#Technical_Preview

EchterAgo:
Ok, so early Windows 10 Insider Previews and Windows Server 2016 Technical Preview 1 used the kernel version 6.4, so I left the 0x64 value in there as well.

EchterAgo:
I tested the earliest public Windows 10 release, version 1507, and it also reports version 10.0.

hjelmn:
Looks reasonable to me. I don't use Windows so I can't verify the patch. Will go ahead and take it and revert it if it causes any issues.

EchterAgo:
There isn't any code that checks for windows 10 yet, but if this is merged I'd make #603 test the windows version to only apply the fix on windows 1903. There are other devices that have been restricted similarly though, so maybe a partial open should be possible.

Issue #602, opened 2019-07-25
SONAME change in 1.0.23 rc


jakirkham:
Appears there is [a SONAME change in the rc]( https://abi-laboratory.pro/index.php?view=objects_report&l=libusb&v1=1.0.22&v2=1.0.23-rc1 ). Curious if this is sticking around for the full release of 1.0.23 or if that is getting changed back.

LudovicRousseau:
This change was done in https://github.com/libusb/libusb/commit/fde20bb9b7cba0ea6e96db920d0a0169c361ca92

But the commit does not provide a justification.

I don’t remember if we broke the ABI with the next release.

jakirkham:
Any thoughts, @hjelmn? 🙂

flibitijibibo:
We just found this while integrating libusb into SDL, seems the only difference between 1.0.22 and master is exclusively new APIs in libusb.h, so there shouldn't be any breakage (unless I've missed something):

https://github.com/libusb/libusb/compare/v1.0.22...master

sezero:
The correct way of handling libtool versioning is incrementing lt_age
along with lt_current, because only new apis are added.  This is way the
libraries stay compatible, i.e. the main libusb library names stay the same.
Something like (against current git):
```
diff --git a/configure.ac b/configure.ac
index 76f67d9..da8a158 100644
--- a/configure.ac
+++ b/configure.ac
@@ -24,5 +24,5 @@ AC_INIT([libusb],[LIBUSB_MAJOR[.]LIBUSB_MINOR[.]LIBUSB_MICRO[]LIBUSB_RC],[libusb
 lt_current=2
 lt_revision=0
-lt_age=1
+lt_age=2
 LTLDFLAGS="-version-info ${lt_current}:${lt_revision}:${lt_age}"
 
```

With this, it compiles for Linux like:
`libusb-1.0.so -> libusb-1.0.so.0.2.0` and
`libusb-1.0.so.0 -> libusb-1.0.so.0.2.0`

.. and for Mac OS X like:
`libusb-1.0.dylib -> libusb-1.0.0.dylib`
The dylib compatibility version is incremented though (which is a true annoyance with libtool):
For libusb-1.0.22:
`/usr/local/lib/libusb-1.0.0.dylib (compatibility version 2.0.0, current version 2.0.0)`
For current git after above patch:
`/usr/local/lib/libusb-1.0.0.dylib (compatibility version 3.0.0, current version 3.0.0)`

[...]

PR #605, opened 2019-08-07
Unix: fix library version for next release

LudovicRousseau:
https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
says:

4. If any interfaces have been added, removed, or changed since the last
update, increment current, and set revision to 0.

5. If any interfaces have been added since the last public release, then
increment age.

So both current and age must be incremented. The patch in
fde20bb9b7cba0ea6e96db920d0a0169c361ca92 was not complete.

Signed-off-by: Ludovic Rousseau <ludovic.rousseau@free.fr>


flibitijibibo:
Confirming that this fixes #602 for me!

hjelmn:
@LudovicRousseau Thanks for fixing that. I will get another rc out shortly.

PR #606, opened 2019-08-11
Change the errcode argument in libusb_strerror to be of type int

VladimirTechMan:
It would be more convenient in practice if the type of error code
argument in `libusb_strerror` was declared as **int**, rather than
`enum libusb_error`: In practice, the value for the argument almost
always comes from a value returned by libusb function, thus its
type is **int** at that point. In which case, depending on the
checks enabled when compiling the C or C++ code of applications
using libusb's API, every call to `libusb_strerror` often requires
explicit type casting – while not a biggie, it gets inconvenient
as the number of logging calls using `libusb_strerror` grows.

It is worth of noting that the "peer" function `libusb_error_name`
already takes the **int**-type argument, thus eliminating the need in
extra value casting. Aligning the signature of `libusb_strerror`
to be the same would also improve the consistency between those
two parts of the libusb API.

Given that `libusb_strerror` does a range validation of its argument
value, there should not be any real drawbacks or negative effects
from "relaxing" the argument type from enumeration to integer.
The existing application code that already does explicit casting
to enum, will keep compiling and running like before.


hjelmn:
Is this ABI compatible? C leaves the underlying type to be defined by the compiler. The type must be at least large enough to hold the largest value in the enumeration. In practice I think the type is always int if the largest value is less than 2^32-1. If we take this change we are *probably* ok not bumping the .so version.

VladimirTechMan:
That's a good point, @hjelmn. Well, as you already noted, the C standard does not mandate an `enum` to be of specific size except being large enough to hold the biggest value in that enumeration. Thus, specifically for this case, I am not sure if we would be able to guarantee the backward compatibility of the ABI for the function on all combinations of supported platforms and compilers. That being said, I did a quick test on the platforms immediately available to me: Using **clang** on **macOS Mojave** (Apple LLVM version 10.0.1), and using **gcc 8.3** on **Ubuntu 18.10** (64-bit) and the recent version of **Raspbian** (32-bit). In those cases, the results of `sizeof` on `int` and on `enum libusb_error` were the same.

```c
#include "libusb/libusb.h"
#include <stdio.h>

int main(void) {
    printf("sizeof(enum libusb_error) = %lu\n", sizeof(enum libusb_error));
    printf("sizeof(int) = %lu\n", sizeof(int));
    return 0;
}
```

dickens:
Given that no sane compiler is going to allocate larger than an int for an enum that has a range of -99..0, this change is safe.  Even if the compiler was crazy enough to allocate a single byte for this value, the compiler will still promote the argument to an int (or larger) in order to pass it as an argument to the `libusb_strerror()` function.

PR #611, opened 2019-08-14
windows: fix build

hjelmn:
Make code compliant with C89 so it will compile.


hjelmn:
I think this will fix travis. Would prefer not to ship with a broken Windows build but will if we can't get the issue resolved today.

PR #612, opened 2019-08-15
windows: workaround WaitForMultipleObjects max 64 events limitation.

nxpfrankli:
WaitForMultiObjects have limiation. Only can wait max 64 events.

but usbi_poll may pass more than 64 fds. In previous implement,
only wait for first 64 events. if previous 64 events were not trigger
usbi_poll will wait for about 10s timemout eventhough other event triggered.

This patch workaround this limitation.

If max events less than 64, call WaitforMultiObjects directly.
If max events more than 64, group every 63 events into one work thread.

This thread call waitformulitobjects wait for this groug events and one
addtional thread exit events.
If any events trigger, this thread will trigger main notify events.

The main usbi_poll thread call waitforsingleobject wait for notify
events. If this events trigger, that means any of work threads get events.
Then call exit notify events let all working thread exit safely.

Return value changed, 0 means timeout. 1 - N means which event
triggered.

### DIFF
--- a/libusb/os/poll_windows.c
+++ b/libusb/os/poll_windows.c
@@ -224,10 +224,6 @@ static int check_pollfds(struct pollfd *fds, unsigned int nfds,
 			fds[n].revents = fds[n].events;
 			nready++;
 		} else if (wait_handles != NULL) {
-			if (*nb_wait_handles == MAXIMUM_WAIT_OBJECTS) {
-				usbi_warn(NULL, "too many HANDLEs to wait on");
-				continue;
-			}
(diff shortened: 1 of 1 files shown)


awzjcasey:
I have tested the the latest version with commits through Mar 16 2020 and while it works on Windows with up to 64 buffers it hangs if an attempt is made to allocate more than 64 buffers on Windows 10.

mcuee:
@awzjcasey Maybe you want to provide more details.

mcuee:
@awzjcasey Sorry you have have to create a new ticket with reference to this one. This is a pull request merged.

Issue #86, opened 2015-08-10
Windows: libusb Windows hotplug support


mcuee:
This is one of the long standing missing feature for Windows.

mcuee:
Last known tree back in the libusbx time.
https://github.com/dickens/libusbx-hp/tree/windows-hotplug-3

changyp6:
libusb hotplug for Windows is really important.
I have a program running on Linux/Window/MacOS, but Windows doesn't have hotplug support, I have to write my own device detection code for Windows.
But Windows only reports that there is a device arrived or there's a device removed, you have to find which one is added or removed.
So when I receive device remove event, I usb libusb_get_device_list() to get current device present in the system, and compare this list with the old one to find out which one is removed, and I use the same method to find which one is added, after I find the difference, my program will handle the removed/added device.

But if I unplug the device and plug it back to the same USB port very quickly, and Windows's driver searching  program is disabled(to make system recognize device faster), my program can get device remove and device add event correctly, but libusb_get_device_list() will get a usb device list the same as the previous saved one.

Here is the reason:
When my program receives device remove event, and start to list all the USB devices, because the unplug/plug action is so fast, libusb_get_device_list() will give a list after the device is plugged back. So this USB device list is the same as the list before unplug/plug action, which means my program cannot find out which device is removed and which is added, there will be 2 missing actions for my program.

Because my program is event driven, this causes 2 events missing.

I really hope that hotplug feature can be implemented for Windows.

mcuee:
https://github.com/dickens/libusbx-hp/tree/windows-hotplug-3

You may want to try the above old branch to see the hotplug code there
works for your application. I am not so sure how much efforts need to bring
that branch up to date though.

changyp6:
I tried that before, I even tried to merge the change into libusb 1.0.19, but I failed to merge it to newer code. I still need to use latest branch, because there are tons of bug/leak fixes for Windows platform.

Hope Windows hotplug support can be implemented soon.

dickens:
Will be part of the next release. Unfortunately it is a complicated set of
changes.

Regards,
Chris

changyp6:
This is a great news!

chrisba77:
[...]

PR #617, opened 2019-08-26
Windows Hotplug Support and Driver Name API

ids-imaging:
This adds hotplug support for the two windows backends, as well as an API to get the name of the driver that is used for a device (this is only relevant for windows).

This implementation uses the already existing enumeration functionality which means that this could be made more efficient in the future.

Feedback is very much appreciated!

#86

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -614,7 +614,7 @@ libusb_free_device_list(list, 1);
  * itself. */
 #define DISCOVERED_DEVICES_SIZE_STEP 8
 
-static struct discovered_devs *discovered_devs_alloc(void)
+struct discovered_devs *usbi_discovered_devs_alloc(void)
 {
 	struct discovered_devs *ret =
 		malloc(sizeof(*ret) + (sizeof(void *) * DISCOVERED_DEVICES_SIZE_STEP));
@@ -626,7 +626,7 @@ static struct discovered_devs *discovered_devs_alloc(void)
 	return ret;
 }
 
-static void discovered_devs_free(struct discovered_devs *discdevs)
+void usbi_discovered_devs_free(struct discovered_devs *discdevs)
 {
 	size_t i;
 
@@ -660,7 +660,7 @@ struct discovered_devs *discovered_devs_append(
 	new_discdevs = realloc(discdevs,
 		sizeof(*discdevs) + (sizeof(void *) * capacity));
 	if (!new_discdevs) {
-		discovered_devs_free(discdevs);
+		usbi_discovered_devs_free(discdevs);
 		return NULL;
 	}
 
@@ -807,7 +807,7 @@ struct libusb_device *usbi_get_device_by_session_id(struct libusb_context *ctx,
 ssize_t API_EXPORTED libusb_get_device_list(libusb_context *ctx,
 	libusb_device ***list)
 {
-	struct discovered_devs *discdevs = discovered_devs_alloc();
+	struct discovered_devs *discdevs = usbi_discovered_devs_alloc();
 	struct libusb_device **ret;
 	int r = 0;
 	ssize_t i, len;
@@ -861,7 +861,7 @@ ssize_t API_EXPORTED libusb_get_device_list(libusb_context *ctx,
 
 out:
 	if (discdevs)
-		discovered_devs_free(discdevs);
+		usbi_discovered_devs_free(discdevs);
 	return len;
 }
 
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1153,9 +1153,32 @@ out:
 DEFAULT_VISIBILITY
 libusb_device * LIBUSB_CALL libusb_ref_device(libusb_device *dev)
 {
+	libusb_context *ctx;
+
+	if (!dev)
+		return NULL;
+
+	/* NOTE There is still a data race on the device and thus on the contexts
+	 * mutex */
+	ctx = dev->ctx;
+	/* Per context there can only be one ref or unref happening at the same time
+	 * NOTE There is still a data race on the device and thus on the contexts
+	 * mutex */
+	usbi_mutex_lock(&ctx->ref_devs_lock);
+
+	/* If the ref count is 0 the device has already been freed, don't touch it
+	 * again
+	 * NOTE This may be a read after free */
+	if (dev->refcnt <= 0) {
+		usbi_mutex_unlock(&ctx->ref_devs_lock);
+		return NULL;
+	}
+
 	usbi_mutex_lock(&dev->lock);
 	dev->refcnt++;
 	usbi_mutex_unlock(&dev->lock);
+
+	usbi_mutex_unlock(&ctx->ref_devs_lock);
 	return dev;
 }
 
@@ -1166,10 +1189,27 @@ libusb_device * LIBUSB_CALL libusb_ref_device(libusb_device *dev)
  */
 void API_EXPORTED libusb_unref_device(libusb_device *dev)
 {
+	libusb_context *ctx;
 	int refcnt;
 
 	if (!dev)
 		return;
+
+	/* NOTE There is still a data race on the device and thus on the contexts
+	 * mutex */
+	ctx = dev->ctx;
+	/* Per context there can only be one ref or unref happening at the same time
+	 * NOTE There is still a data race on the device and thus on the contexts
+	 * mutex */
+	usbi_mutex_lock(&ctx->ref_devs_lock);
+
+	/* If the ref count is 0 the device has already been freed, don't touch it
+	 * again
+	 * NOTE This may be a read after free */
+	if (dev->refcnt <= 0) {
+		usbi_mutex_unlock(&ctx->ref_devs_lock);
+		return;
+	}
 
 	usbi_mutex_lock(&dev->lock);
 	refcnt = --dev->refcnt;
@@ -1191,6 +1231,7 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
 		usbi_mutex_destroy(&dev->lock);
 		free(dev);
 	}
+	usbi_mutex_unlock(&ctx->ref_devs_lock);
 }
 
 /*
@@ -2295,6 +2336,7 @@ int API_EXPORTED libusb_init(libusb_context **context)
 
 	usbi_mutex_init(&ctx->usb_devs_lock);
 	usbi_mutex_init(&ctx->open_devs_lock);
+	usbi_mutex_init(&ctx->ref_devs_lock);
 	usbi_mutex_init(&ctx->hotplug_cbs_lock);
 	list_init(&ctx->usb_devs);
 	list_init(&ctx->open_devs);
@@ -2347,6 +2389,7 @@ err_free_ctx:
 	usbi_mutex_unlock(&ctx->usb_devs_lock);
 
 	usbi_mutex_destroy(&ctx->open_devs_lock);
+	usbi_mutex_destroy(&ctx->ref_devs_lock);
 	usbi_mutex_destroy(&ctx->usb_devs_lock);
 	usbi_mutex_destroy(&ctx->hotplug_cbs_lock);
 
@@ -2432,6 +2475,7 @@ void API_EXPORTED libusb_exit(struct libusb_context *ctx)
 		usbi_backend.exit(ctx);
 
 	usbi_mutex_destroy(&ctx->open_devs_lock);
+	usbi_mutex_destroy(&ctx->ref_devs_lock);
 	usbi_mutex_destroy(&ctx->usb_devs_lock);
 	usbi_mutex_destroy(&ctx->hotplug_cbs_lock);
 	free(ctx);
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -918,6 +918,25 @@ uint8_t API_EXPORTED libusb_get_port_number(libusb_device *dev)
 }
 
 /** \ingroup libusb_dev
+ * Get the name of the driver that a device is used with.
+ *
+ * The function will write the maximum number of characters possible into the
+ * buffer while terminating the string correctly.
+ * \param dev a device
+ * \param driver buffer to write the driver name into
+ * \param size of the buffer to write the driver name into
+ * \returns the required buffer size to write the complete driver name
+ * 					(inclusive termination) or LIBUSB_ERROR_NOT_SUPPORTED.
+ */
+int API_EXPORTED libusb_get_driver(libusb_device *dev, char *driver, int size)
+{
+	if (usbi_backend.get_device_driver) {
+		return usbi_backend.get_device_driver(dev, driver, size);
+	}
+	return LIBUSB_ERROR_NOT_SUPPORTED;
+}
+
+/** \ingroup libusb_dev
  * Get the list of all port numbers from root for the specified device
  *
  * Since version 1.0.16, \ref LIBUSB_API_VERSION >= 0x01000102
--- a/libusb/os/netbsd_usb.c
+++ b/libusb/os/netbsd_usb.c
@@ -97,6 +97,7 @@ const struct usbi_os_backend usbi_backend = {
 	netbsd_open,
 	netbsd_close,
 
+	NULL,				/* netbsd_get_device_driver */,
 	netbsd_get_device_descriptor,
 	netbsd_get_active_config_descriptor,
 	netbsd_get_config_descriptor,
--- a/libusb/os/poll_windows.c
+++ b/libusb/os/poll_windows.c
@@ -60,7 +60,7 @@ static size_t fd_count;
 static size_t fd_size;
 #define INC_FDS_EACH 256
 
-static void usbi_dec_fd_table()
+static void usbi_dec_fd_table(void)
 {
 	fd_count--;
 	if (fd_count == 0) {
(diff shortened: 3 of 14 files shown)


changyp6:
The patch cannot be built in linux MinGW environment.
Error is Windows.h no such file or directory.

The reason is that in Windows system, "Windows" and "windows" are the same, but in Linux system, they are different.

ids-imaging:
Locally compiled with MSVC 2010, 2013, 2015, 2017, 2019 and msys2/mingw. Once all feedback is incorporated the changes will be squashed.

changyp6:
```
libusb/os/windows_winusb.c:28:10: fatal error: Windows.h: No such file or directory
 #include <Windows.h>
          ^~~~~~~~~~~
```
Some capital letters still there.
```
libusb/os/windows_winusb.c:3484:12: warning: unknown conversion type character 'z' in format [-Wformat=]
   usbi_dbg("invalid size (%zu)", *size);
            ^~~~~~~~~~~~~~~~~~~~
libusb/os/windows_winusb.c:3563:12: warning: unknown conversion type character 'z' in format [-Wformat=]
   usbi_dbg("invalid size (%zu)", *size);
            ^~~~~~~~~~~~~~~~~~~~
```
These are just some ugly warnings.
`%zu` can only work in POSIX system, in Windows, it should be `%Iz`

changyp6:
Sorry, my mistake.
It should be `%Iu` not `%Iz`. `%Iz` is a typo.

Please also remove `usbi_dbg("")` empty log from files. these lines are causing ugly warnings when compiling in MinGW.

And some other warnings saying that `static void windows_destroy_event_monitor() is NOT function prototype`,
The correct one should be `static void windows_destroy_event_monitor(void)`
and `static int windows_start_event_monitor(void)`
and `static void windows_destroy_event_window(void)`
and `static int windows_event_window_setup(void)`

It seems that `void` should be added to the parameter list, when there's no parameter of a function.

changyp6:
After all these modifications, I can successfully compile libusb with hotplug support.
And I have tested, hotplug works as expected.
The registered device with specific VID can be reported.

But there's another issue.
`libusb_control_transfer()` doesn't seem to work.
but `libusb_bulk_transfer()` works.

ids-imaging:
I don't think that the empty debug logs should be removed by this pull request as I'm not sure that they were added to implicitly print the function call flow.

Which backend are you using and has the controll transfer been working before? At the moment I'm unsure what could cause that.

changyp6:
I'm using libusb-win32.
The backend should be libusb0.sys

libusb/core.c
```diff
@@ -1151,11 +1167,31 @@ int API_EXPORTED libusb_get_max_iso_packet_size(libusb_device *dev,
  * \returns the same device
  */
 DEFAULT_VISIBILITY
-libusb_device * LIBUSB_CALL libusb_ref_device(libusb_device *dev)
+libusb_device *LIBUSB_CALL libusb_ref_device(libusb_device *dev)
```
LudovicRousseau: Why this change?

libusb/os/darwin_usb.c
```diff
@@ -2247,19 +2247,29 @@ static int darwin_free_streams (struct libusb_device_handle *dev_handle, unsigne
...
         .name = "Darwin",
         .caps = 0,
         .init = darwin_init,
         .exit = darwin_exit,
         .get_device_list = NULL, /* not needed */
+        .hotplug_poll = darwin_hotplug_poll,
         .get_device_descriptor = darwin_get_device_descriptor,
         .get_active_config_descriptor = darwin_get_active_config_descriptor,
         .get_config_descriptor = darwin_get_config_descriptor,
-        .hotplug_poll = darwin_hotplug_poll,
```
LudovicRousseau: Why moving this line?

libusb/os/haiku_usb_raw.cpp
```diff
@@ -209,6 +209,15 @@ haiku_clock_gettime(int clkid, struct timespec *tp)
...
 
+static int haiku_get_device_driver(struct libusb_device *device, char *driver,
+																 int size)
+{
+	if (size > 0 && device) {
+		driver[0] = '\0';
+	}
+	return 1;
+}
+
```
LudovicRousseau: Maybe it is possible to have a `default_get_device_driver() ` to not have the define the same function in 6 different backends. Only the windows backend uses a different implementation.

libusb/os/sunos_usb.c
```diff
@@ -1666,6 +1675,7 @@ const struct usbi_os_backend usbi_backend = {
         .hotplug_poll = NULL,
         .open = sunos_open,
         .close = sunos_close,
+        .get_device_driver =sunos_get_device_driver,
```
LudovicRousseau: missing space character after the "="

libusb/core.c
ids-imaging: Because I use clang format with the linux kernel style as that is what most code in libusb seems to use. If this whitespace bothers you I an revert it.

libusb/os/darwin_usb.c
ids-imaging: Because other platfoms have it there.

libusb/os/haiku_usb_raw.cpp
ids-imaging: I thought about it, but where would I put that function?

LudovicRousseau: Maybe in `core.c` since it is common code?

I am not sure the function should return an empty string on Unix systems. Maybe returning `LIBUSB_ERROR_NOT_SUPPORTED` is better.
In fact I don't know why this new API is needed. Is it mandatory for hotplug?
Maybe you should split the patch and move the addition of `libusb_get_driver()` in a different Pull Request.

ids-imaging: The addition to the API is important for us, not hotplugging.

**Why?**
Because some driver result in blue screens and that is something that has to be prevented. So this API is usefull for all windows users that have to make sure that a sane driver is used as backend. I thought I added that to the commit message but I must have forgotten it.

**API**
Ok that can be done.

ids-imaging:
I found a deadlock introduced with the datarace commit, which would explain why the control transfer failed. I'll fix that tomorrow.

changyp6:
> I found a deadlock introduced with the datarace commit, which would explain why the control transfer failed. I'll fix that tomorrow.

Thank you very much!

LudovicRousseau:
I propose you to separate the different changes in different PR.
For example your commits https://github.com/libusb/libusb/pull/617/commits/5c1fbde5dc8176f111f38aee0a020aa2708ff67f and https://github.com/libusb/libusb/pull/617/commits/6b028baf1eb78193cbc6b9daf1ff6b589b2bac71 should be in 2 different PR.

1 PR per issue solved.

changyp6:
I've just cross-compiled this patch in MinGW environment, there are some warnings:
```
../../libusb/os/windows_winusb.c:3486:12: warning: unknown conversion type character 'l' in format [-Wformat=]
   usbi_dbg("invalid size (%Ilu)", *size);
            ^~~~~~~~~~~~~~~~~~~~~
../../libusb/libusbi.h:238:71: note: in definition of macro '_usbi_log'
 #define _usbi_log(ctx, level, ...) usbi_log(ctx, level, __FUNCTION__, __VA_ARGS__)
                                                                       ^~~~~~~~~~~
../../libusb/os/windows_winusb.c:3486:3: note: in expansion of macro 'usbi_dbg'
   usbi_dbg("invalid size (%Ilu)", *size);
   ^~~~~~~~
../../libusb/os/windows_winusb.c:3486:12: warning: too many arguments for format [-Wformat-extra-args]
   usbi_dbg("invalid size (%Ilu)", *size);
            ^~~~~~~~~~~~~~~~~~~~~
../../libusb/libusbi.h:238:71: note: in definition of macro '_usbi_log'
 #define _usbi_log(ctx, level, ...) usbi_log(ctx, level, __FUNCTION__, __VA_ARGS__)
                                                                       ^~~~~~~~~~~
../../libusb/os/windows_winusb.c:3486:3: note: in expansion of macro 'usbi_dbg'
   usbi_dbg("invalid size (%Ilu)", *size);
   ^~~~~~~~
../../libusb/os/windows_winusb.c: In function '_hid_set_report':
../../libusb/os/windows_winusb.c:3565:12: warning: unknown conversion type character 'l' in format [-Wformat=]
   usbi_dbg("invalid size (%Ilu)", *size);
            ^~~~~~~~~~~~~~~~~~~~~
```

`%Ilu` should be changed to `%Iu`

And
```
../libusb/os/poll_windows.c:63:13: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 static void usbi_dec_fd_table()
             ^~~~~~~~~~~~~~~~~
```
`static void usbi_dec_fd_table()` should be `static void usbi_dec_fd_table(void)`

And
```
../../libusb/os/windows_usbdk.c:31:5: warning: "_MSC_VER" is not defined, evaluates to 0 [-Wundef]
 #if _MSC_VER >= 1800
     ^~~~~~~~
../../libusb/os/windows_usbdk.c:33:7: warning: "_MSC_VER" is not defined, evaluates to 0 [-Wundef]
 #elif _MSC_VER >= 1700
       ^~~~~~~~
```

I think these codes should be protected by 
```
#ifdef _MSC_VER
#endif
```

I think following code is sufficient enough
```
#ifdef _MSC_VER
 #if _MSC_VER >= 1700
 #include <usbiodef.h>
 #endif
#endif
```

Because `_MSC_VER >= 1700` has already covered `_MSC_VER >= 1800`

Anyhow, this pull request has made my 2018/2019 new year's wish come true!
Great job!!

ids-imaging:
> ...
> `%Ilu` should be changed to `%Iu`

That was an ugly issue - depending on the system the `l` was missing or not.

> ...
> `static void usbi_dec_fd_table()` should be `static void usbi_dec_fd_table(void)`

Not from this pull request, but I'll fix it.

> ...
> Because `_MSC_VER >= 1700` has already covered `_MSC_VER >= 1800`

Actually that block can be removed completely, since the fallback from `windows_nt_common.h:123` is good enough. Good catch!

> Anyhow, this pull request has made my 2018/2019 new year's wish come true!
> Great job!!

Thanks for testing and feedback!

changyp6:
> > ...
> > `%Ilu` should be changed to `%Iu`
> 
> That was an ugly issue - depending on the system the `l` was missing or not.
> 
In my build environment, Fedora 30 with MinGW, `%Iu` is sufficient enough for both i686 and x86_64. Because `%Iu` is just like `%zu`. `I` works as `z`

I don't know whether other compiler complains about it or not.

libusb/os/linux_usbfs.c
```diff
@@ -2857,6 +2857,7 @@ const struct usbi_os_backend usbi_backend = {
 	.wrap_sys_device = op_wrap_sys_device,
 	.open = op_open,
 	.close = op_close,
+	.get_device_driver = NULL,
 	.get_configuration = op_get_configuration,
```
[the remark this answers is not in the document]
ids-imaging: Should I also remove line 2850, for consistency purposes? Or do you want the smallest change sets possible?

libusb/os/openbsd_usb.c
```diff
@@ -95,6 +95,7 @@ const struct usbi_os_backend usbi_backend = {
 	.open = obsd_open,
 	.close = obsd_close,
 
+	.get_device_driver = NULL,
 	.get_device_descriptor = obsd_get_device_descriptor,
```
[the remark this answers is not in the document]
ids-imaging: What about haiku and darwin?

libusb/os/netbsd_usb.c
```diff
@@ -97,6 +97,7 @@ const struct usbi_os_backend usbi_backend = {
 	netbsd_open,
 	netbsd_close,
 
+	NULL,				/* netbsd_get_device_driver */,
 	netbsd_get_device_descriptor,
```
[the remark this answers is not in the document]
ids-imaging: Should I change that? With redundant NULLs, so it's easier to work with in the future or without?

libusb/os/sunos_usb.c
```diff
@@ -1666,6 +1666,7 @@ const struct usbi_os_backend usbi_backend = {
         .hotplug_poll = NULL,
         .open = sunos_open,
         .close = sunos_close,
+        .get_device_driver = NULL,
         .get_configuration = sunos_get_configuration,
```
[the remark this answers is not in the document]
ids-imaging: Should I also remove line 1666, 1678, 1679 and 1686 (wrong indent too?) for consistency purposes? Or do you want the smallest change sets possible?

libusb/os/netbsd_usb.c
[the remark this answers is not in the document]
LudovicRousseau: netbsd_usb.c does not use the `.field = ` notation so the extra NULL in the correct place is mandatory here.

libusb/os/linux_usbfs.c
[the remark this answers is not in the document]
LudovicRousseau: Keep the patch as small as possible.
If you want to do other changes then open another Pull Request.

ids-imaging:
Am I supposed to close the requests for change? Should I rebase/squash the changes so this can be merged?

LudovicRousseau:
You should create different PR for your different/unrelated changes.
See my previous comment https://github.com/libusb/libusb/pull/617#issuecomment-525346327

It is much more easy to review small PR.

hjelmn:
Yeah. Separate out each feature or bug fix into its own PR then squash the PRs down to as small a number of commits as possible. Try to make sure that the code still builds between commits so as not to break bisect.

libusb/core.c
```diff
@@ -626,7 +626,7 @@ static struct discovered_devs *discovered_devs_alloc(void)
 	return ret;
 }
 
-static void discovered_devs_free(struct discovered_devs *discdevs)
+void discovered_devs_free(struct discovered_devs *discdevs)
```
hjelmn: This needs to be renamed to ```usbi_discovered_devs_free``` to be consistent with the internal API naming.

libusb/core.c
```diff
@@ -614,7 +614,7 @@ libusb_free_device_list(list, 1);
  * itself. */
 #define DISCOVERED_DEVICES_SIZE_STEP 8
 
-static struct discovered_devs *discovered_devs_alloc(void)
+struct discovered_devs *discovered_devs_alloc(void)
 {
```
hjelmn: This needs to be renamed to ```usbi_discovered_devs_alloc``` to be consistent with the internal API naming.

hjelmn:
Thanks for working on this. It is getting very close.

ids-imaging:
I have incorporated the feedback and squashed all commits.

I see however no reason why there should be 6 PRs, 3 of those would be commits with less than 5 lines changed and one with about 40 – each PR containing one commit. The other two PRs still contain only one and two commits, though they are longer. They are easily reviewed in that form, and since no more feedback is expected, it makes even less sense to split them as context would be lost now. Moreover does this project use a merge strategy where no merge commits are made, meaning that the existence of more PRs and  their branches is not recorded in the git history, making them useless. 
Please explain why I should waste my time splitting this change set.

libusb/os/poll_windows.c
```diff
@@ -60,7 +60,7 @@ static size_t fd_count;
 static size_t fd_size;
 #define INC_FDS_EACH 256
 
-static void usbi_dec_fd_table()
+static void usbi_dec_fd_table(void)
 {
```
hjelmn: This are semantically equivalent. No problem with the change though.

libusb/core.c
hjelmn: thanks!

hjelmn:
I can't comment on any of the backed changes. Otherwise LGTM.

ids-imaging:
There was a race condition between windows events and context initialization / scan of already present devices. This should be fixed with the last commit - on request I'll squash that commit.

Please also considder #635 and #634 for 1.0.24

ids-imaging:
Our QA is finding a troubeling amount of race condition - partially caused by this PR. I'll be closing this PR as well as #634 and #635 as result of that. The progress can be viewed and testet in our fork.

hjelmn:
@ids-imaging When do you expect to open a new PR? I wanted to get this into v1.0.24. I am willing to hold up the release until mid-November if you think it will be ready by then.

DiscreteLogarithm:
@ids-imaging Thank you for your work. Can you give us an update on the status of windows hotplug support, please. Also how serious are the race conditions? Can we use the library with locking on our side and if yes for which functions we need to do that? Thanks a lot in advance.

dickens:
Thank you for the effort on this patch.  Windows hotplug is a long-awaited feature.  Unfortunately as  you discovered this implementation is not going to work.  There are also some things that are not completely addressed in this approach.

With that said, I'll be working on rebasing my previous hotplug work onto master.

mcuee:
Great to see Chris to be working on this again. This feature is really the major gap for the Windows backend. And Windows backend is really more complicated than the others.

liziqi-matrixrealm:
libusb_open or libusb_open_device_with_vid_pid  failed and return -5 LIBUSB_ERROR_NOT_FOUND after hotplug libusb_hotplug_register.

liziqi-matrixrealm:
Report a bug In windows platform.  There is a race when hotplug in then use winusb_get_device_list immediately. The USB device's emulation or some initialization has not finished so return -5 LIBUSB_ERROR_NOT_FOUND. We can add a short sleep before winusb_get_device_list to solve it.

PR #625, opened 2019-09-20
fix constant not in range of enumerated type

mrstock:
fix "Integer constant not in range of enumerated type 'enum libusb_transfer_status'"

LIBUSB_ERROR_NO_DEVICE doesn't exist on enum libusb_transfer_status


LudovicRousseau:
Fixed in https://github.com/libusb/libusb/commit/906e706f8abdff626370e59b880f340cf54e3632

Thanks

PR #632, opened 2019-10-04
Add not implemented guard for linux like systems

tinti:
When compiling for android and disable udev the following functions do
nothing: linux_scan_devices() and op_hotplug_poll(). The result is that
libusb is not able to list devices.

This patch raise a compile time error if the config.h results in no
implementation for those functions.

For instance if USE_UDEV is not defined and __ANDROID__ is defined the
following error will be raised "<function> not implemented for this config".

Signed-off-by: Vinicius Tinti <vinicius.tinti@almg.gov.br>

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -562,6 +562,8 @@ static int linux_scan_devices(struct libusb_context *ctx)
 	ret = linux_udev_scan_devices(ctx);
 #elif !defined(__ANDROID__)
 	ret = linux_default_scan_devices(ctx);
+#else
+#error "linux_scan_devices not implemented for this config"
 #endif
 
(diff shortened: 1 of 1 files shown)


tinti:
If you need a working libusb for android please use https://github.com/tinti/libusb/tree/android.

hjelmn:
@tinti Your version is way out of date. libusb works fine with android except for a known issue listing devices using libusb. Will be fixed in 1.0.24-rc1 in the next week.

tinti:
@hjelmn indeed. I just wanted to provide some workaround for anyone that is reading this thread.

Good to know that it will be fixed. Thanks.

tinti:
@hjelmn How about this one.

https://github.com/tinti/libusb/tree/feature/fix_linux_scan_devices_in_android

I confirm that it is working fine for me in `android-25` platform.

PR #634, opened 2019-10-07
winusb: Fix port querying

ids-imaging:
SPDRP_ADDRESS does not provide the port of a usb device, but the
address. Most of the time this seems to be no issue, but under certain
circumstances they differ. This patch fixes this issue by relying on the
registry value LocationInformation of the device by default and falling
back to SPDRP_ADDRESS in case there is no such entry in the registry.

One example would be the Dell Precision 7530 with the TB18DC docking
station. In that configuration it can be observed that devices on the
first port can not be detected with libusb. The reason for that is, that
the "port" (obtained with SPDRP_ADDRESS) is 3. Since there is no device
connected on port 3, which is used as ConnectionIndex,
DeviceIoControl(..., IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, ...)
will report that there is no device connected on that port and the
device is not listed, event though it should.


mcuee:
Maybe this is still of some value. So I reopen it again.

mcuee:
Please refer to https://github.com/libusb/libusb/pull/867

mcuee:
It seems to me #867 is better so I will close this one.

PR #635, opened 2019-10-07
Fix datarace on context initialization

ids-imaging:


mcuee:
Not so sure if this is still of some values.

dickens:
There's no description of what this is solving.  We can't take a patch that does not explain anything.

mcuee:
@ids-imaging Close this for now. Please reopen if you think this is still useful. Please explain the goal you are trying to achieve.

PR #637, opened 2019-10-07
Fix linux_scan_devices() in Android

tinti:
Use linux_default_scan_devices() in Android platform.

When building for Android the USE_UDEV is false and __ANDROID__ is true
resulting in no implementation for linux_scan_devices(). This commit fix
it by using linux_default_scan_devices() for Android.

Signed-off-by: Vinicius Tinti <vinicius.tinti@almg.gov.br>

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -560,7 +560,7 @@ static int linux_scan_devices(struct libusb_context *ctx)
 
 #if defined(USE_UDEV)
 	ret = linux_udev_scan_devices(ctx);
-#elif !defined(__ANDROID__)
+#else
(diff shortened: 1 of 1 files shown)


tinti:
Check https://github.com/libusb/libusb/pull/632 for more info.

laran002:
I am not sure if this is related or not. Since this change I actually get a list of devices when calling libusb_get_device_list() in Android which I did not before. So far so good.
However, when I use libusb_get_device_list()  the returned list does not change after the application is started. If I have a device connected at application start it continues to list it, even though it is unplugged. Also, if I plug in a device after application start it is never listed. I don't get the hot plug to respond either.
Can this be related to the access request procedure, or does that only have an impact on access of the opening of the handle?

Issue #638, opened 2019-10-08
Solaris: format mismatch in some usbi_dbg() 


kentsuji:
In sunos_usb.c, some usbi_dbg() calls have format mismatch.  For example, "%d" is specified for a size_t variable and "%llx" is specified for a uint64_t variable, in LP64 mode.  For those specifications that are incompatible between ILP32 and LP64, format macro defined in <inttypes.h>, such as PRIuPTR and PRIx64 should be used.

Following is an expample of the warnings in LP64 mode:
os/sunos_usb.c:204:11: warning: format '%d' expects argument of type 'int', but argument 6 has type 'size_t' {aka 'long unsigned int'} [-Wformat=]
  204 |  usbi_dbg("%p, %d", iocdata.nvl_user, iocdata.nvl_usersz);
      |           ^~~~~~~~                    ~~~~~~~~~~~~~~~~~~
      |                                              |
      |                                              size_t {aka long unsigned int}

And following is an example of the warnings in ILP32 mode:
os/sunos_usb.c:610:11: warning: format '%lx' expects argument of type 'long unsigned int', but argument 5 has type 'uint64_t' {aka 'long long unsigned int'} [-Wformat=]
  610 |  usbi_dbg("session id org:%lx", session_id);
      |           ^~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~
      |                                 |
      |                                 uint64_t {aka long long unsigned int}

They should be rewritten as the following:

  204      usbi_dbg("%p, %" PRIuPTR, iocdata.nvl_user, iocdata.nvl_usersz);

  610      usbi_dbg("session id org:%" PRIx64, session_id);

PR #640, opened 2019-10-11
Solaris: format mismatch in some usbi_dbg()

kentsuji:
In sunos_usb.c, some usbi_dbg() calls have format mismatch. For example, "%d"
is specified for a size_t variable and "%llx" is specified for a uint64_t
variable. In LP64 mode, they are format mismatch.  For those specifications
that are incompatible between ILP32 and LP64, format macros defined in
<inttypes.h>, such as PRIuPTR and PRIx64 should be used.

Confirmed this can be built on Solaris 11.4 and OpenIndiana using gcc7.

### DIFF
--- a/libusb/os/sunos_usb.c
+++ b/libusb/os/sunos_usb.c
@@ -38,6 +38,7 @@
 #include <errno.h>
 #include <sys/usb/usba.h>
 #include <sys/pci.h>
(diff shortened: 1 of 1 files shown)


kentsuji:
This fixes #638.

PR #645, opened 2019-11-01
Linux backend: fix ressource leak

LudovicRousseau:
Issue detected by Coverity:
22. leaked_handle: Handle variable fd going out of scope leaks the handle.


LudovicRousseau:
Nobody complained so I pushed my change in https://github.com/libusb/libusb/commit/1bb774ca7e5cb100e1fb7ac287e561b2155da70e

PR #649, opened 2019-11-12
linux_udev: silently ignore "bind" action

johnkeeping:
When a driver is bound to a device, udev emits a "bind" action that
causes libusb to log an error message:

	libusb: error [udev_hotplug_event] ignoring udev action bind

Since we know this action is not relevant for libusb, silently ignore it
to avoid people thinking there is something wrong.

There is already a debug log entry for the action, so there is no need
to add a duplicate here.

Signed-off-by: John Keeping <john@metanate.com>


LudovicRousseau:
Fixed in https://github.com/libusb/libusb/commit/f20f2be7825e34b5273af17fc0740c60bd352b32

Thanks a lot

PR #657, opened 2019-12-03
Improved system out of memory handling

SlashGordon:
Maps ENOMEM system error to LIBUSB_ERROR_NO_MEM. When the USBFS memory limit is reached, I get a LIBUSB_ERROR_IO, which is a bit confusing. I think LIBUSB_ERROR_NO_MEM makes the situation clear.


LudovicRousseau:
Fixed in https://github.com/libusb/libusb/commit/97ece77c4a6eaafd0a74fd3b73e290ec97d71eb5

Thanks a lot

PR #665, opened 2019-12-16
windows: Fix USBD_STATUS to LIBUSB_TRANSFER_STATUS translation

photron:
The MSDN documentaion says that the most significant 4 bits of the
USBD_STATUS value indicate success/pending/error state, but then gives
them as 2 bit values.

The broken translations code assumes these to be the lower 2 bits:

0b0000 for success
0b0001 for pending
0b0010 for error
0b0011 for error

But actually it's the higher 2 bits:

0b0000 for success
0b0100 for pending
0b1000 for error
0b1100 for error

The USBDK code already deals with USBD_STATUS and gets it correct.

Another problem is that the broken translations code then masks off
the most significant 4 bits of the USBD_STATUS value, but then compares
it to the full 32 bit error codes to figure out the actual error. This
switch will always jump to the default case, because all checked error
codes have their most significant 2 bits set, but the values they are
compared against have those bits masked off.

Move the working code from the USBDK backend to the shared header and
reuse it in the WinUSB backend too.

### DIFF
--- a/libusb/os/windows_nt_shared_types.h
+++ b/libusb/os/windows_nt_shared_types.h
@@ -34,6 +34,19 @@ typedef struct USB_CONFIGURATION_DESCRIPTOR {
 
 #include <poppack.h>
 
+// https://msdn.microsoft.com/en-us/library/windows/hardware/ff539136(v=vs.85).aspx
+typedef LONG USBD_STATUS;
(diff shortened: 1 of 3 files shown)


dickens:
Applied, thank you!

PR #669, opened 2019-12-27
Add Null POSIX backend

pinotree:
Add a simple null backend for POSIX platforms that reports no available devices, and provides no capabilities. Make use of this new backend on all the OSes without an existing backend, so libusb can be built even on OSes without USB support.

The primary case here is to build on GNU/Hurd, which does not support USB yet: the idea is to be able to build libusb anyway, so applications built on top of it (or using it for features) can built fine with it. In the end, applications do not care much about the actual internal backend of libusb, and they already need to cope with the lack of USB devices anyway (i.e. `libusb_get_device_list()` returning no devices, or not the wanted device).
We have been using a patch similar to this one already downstream in Debian GNU/Hurd, so there is less need to disable USB support in applications only on Hurd, and create differences w.r.t. Linux architectures in Debian.

I understand this patch might be controversial, and that this has high chance of being rejected. A possible alternative I had in mind was to allow this backend unconditionally with a `--enable-backend-null` (or so) configure option, so even on OSes with a real backend it is possible to build a libusb that does nothing (maybe to to ease deployment of stuff using libusb in resource-limited environments/containers).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -100,7 +100,9 @@ case $host in
 	threads=posix
 	;;
 *)
-	AC_MSG_ERROR([unsupported operating system $host])
+	AC_MSG_RESULT([Null])
+	backend="null"
+	threads="posix"
 esac
 
 case $backend in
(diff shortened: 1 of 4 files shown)


LudovicRousseau:
I like the idea.

mcuee:
I support this idea.

In OpenOCD they have a dumb debug adapter as well, kind of similar idea.

hjelmn:
Yup :+1:

LudovicRousseau:
Pushed in https://github.com/libusb/libusb/commit/53572d7e5eee79266139399924c7491174be1670
Thanks

PR #670, opened 2019-12-27
core: include sys/time.h on any GNU libc OS

pinotree:
GNU libc provides `sys/time.h`, so unconditionally include it when using that libc.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -54,7 +54,7 @@ typedef unsigned __int32  uint32_t;
 #include <sys/types.h>
 #endif
 
-#if defined(__linux__) || defined(__APPLE__) || defined(__CYGWIN__) || defined(__HAIKU__)
+#if defined(__linux__) || defined(__APPLE__) || defined(__CYGWIN__) || defined(__HAIKU__) || defined(__GLIBC__)
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


LudovicRousseau:
What _problem_ is solved with this patch?

pinotree:
> What problem is solved with this patch?

Mostly to make sure that the includes pulled by `libusb.h` are the same across different OSes; in practice, while testing #669 on Hurd, the build of the `sam3u_benchmark.c` example failed because of undefined `gettimeofday()`: on other OSes `sys/time.h` (required for `gettimeofday()` [1]) is included already.
While this can be a bug in the example (and potentially in other applications, although I find that unlikely), I wanted to minimize the differences in behaviour between different OSes.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/functions/gettimeofday.html

LudovicRousseau:
No objection.
Can you update the commit message so that your explanation above is included?

pinotree:
Done, although I kept a slightly more generic explanation (still to the point though, I hope).

LudovicRousseau:
Fixed in https://github.com/libusb/libusb/commit/1e9d26a753a741bfdb0769b4c895697d16edcc27
Thanks

PR #674, opened 2020-01-10
core: fix build warning on newer versions of gcc

gregkh:
When building libusb on a "newer" version of gcc (9.2), a lot of
warnings are thrown about zero-length messages as being part of a format
string.

An example of this is:

descriptor.c:546:11: warning: zero-length gnu_printf format string [-Wformat-zero-length]
  546 |  usbi_dbg("");
      |           ^~

Fix this up by replacing all calls of:
	usbi_dbg("");
with
	usbi_dbg(" ");
as obviously we still want to keep the implicit tracing message in the
log.

Signed-off-by: Greg Kroah-Hartman <gregkh@linuxfoundation.org>

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -813,7 +813,7 @@ ssize_t API_EXPORTED libusb_get_device_list(libusb_context *ctx,
 	int r = 0;
 	ssize_t i, len;
 	USBI_GET_CONTEXT(ctx);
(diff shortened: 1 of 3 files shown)


dickens:
Thanks @gregkh, this was something I was planning to look at.  Looks good, I will merge it soon.

PR #678, opened 2020-01-18
Darwin: Fix LIBUSB_TRANSFER_ADD_ZERO_PACKET flag

ghost:
The LIBUSB_TRANSFER_ADD_ZERO_PACKET flag does not work properly because the maximum packet size is determined incorrectly. The maximum packet size should be 1024 bytes for USB 3.0 and 512 bytes for USB 2.0. Without the suggested patch a maximum packet size of 16384 bytes was determined on my test machine and therefore no zero length packets were sent.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1746,7 +1746,7 @@ static int submit_bulk_transfer(struct usbi_transfer *itransfer) {
   IOReturn               ret;
   uint8_t                transferType;
   /* None of the values below are used in libusbx for bulk transfers */
-  uint8_t                direction, number, interval, pipeRef;
+  uint8_t                direction, number, interval, pipeRef, alternateSetting;
   uint16_t               maxPacketSize;
 
   struct darwin_interface *cInterface;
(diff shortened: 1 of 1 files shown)


hjelmn (dismissed):
Looks fine except that the new code should probably replace the old code. No reason to query the properties twice.

simongysi:
I guess you're talking about `GetPipeProperties`. IMHO this is still used to get the `number`, which is then passed to `GetEndPointProperties`.

libusb/os/darwin_usb.c
```diff
@@ -1766,6 +1766,23 @@ static int submit_bulk_transfer(struct usbi_transfer *itransfer) {
     return darwin_to_libusb (ret);
   }
 
+  ret = (*(cInterface->interface))->GetAlternateSetting (cInterface->interface, &alternateSetting);
+
+  if (ret) {
```
seanm, line 1771: Might be nice if these 3 error messages could be unique, so that if/when anyone sees one, they know which it's from.

tormodvolden:
My thoughts from looking at this:
- It is surprising that the maxPacketSize coming from GetPipeProperties() is wrong, in case of bulk endpoints and claimed interfaces I would have expected the same value as the endpoint property. But the proof is in the pudding, if the OP has seen this is wrong, it can be bugs or something we don't understand in the Darwin USB stack. It would be good to have more information, if only for reference, like a bug report with log files, maybe a test case, etc.
- The surrounding code has changed a bit since this was posted, also other parts regarding updating of endpoints etc. It would be good to have the issue retested on latest master (and the patch rebased).
- I think hjelmn's question about code duplication has been answered well by the OP.
- If the patch is otherwise fine and desired for 1.0.25, I am willing to rework it with more detailed/unique error messages in case the OP doesn't get to it.

tormodvolden:
I have rebased this for latest master and added InterfaceVersion >= 550 alternative code (pushed to my own repo at https://github.com/tormodvolden/libusb/commits/ghost-fix_zero_length_p_678), and have tried to reproduce the original issue, to no avail, although only on one device. In the second commit I added a warning if there is a mismatch. Please test this and attach a debug log if it can be reproduced.

mcuee:
Over the years there have no reports of issues related to LIBUSB_TRANSFER_ADD_ZERO_PACKET flag, either for macOS or Linux. Guess this feature is seldom used.

tormodvolden:
I think this has to wait until we see proof (as in debug logs) of the current code going wrong, with details on the device in question.

mcuee:
> I think this has to wait until we see proof (as in debug logs) of the current code going wrong, with details on the device in question.

I agree.

tormodvolden:
Relevant "discussion" at the [wMaxPacketSize](https://developer.apple.com/documentation/kernel/iousbendpointproperties/1546100-wmaxpacketsize) documentation, unfortunately without further explanation:
`The meaning of this value depends on whether GetPipePropertiesV3 or GetEndpointPropertiesV3 uses it.`

mcuee:
The user has deleted his github account. So probably we have to deal with this with low priority.

mcuee:
@Youw 

Would you help to ask Github Copilot to review this very old PR to see if it is worth keeping? Thanks.

mcuee:
@Youw 

Strange that Copilot does not start the review.

sonatique:
Closing because:
the fix for the described issue is now in master: Commit 7bc88c0f "macos: Fix Zero-Length Packet for multiple packets per frame" from #1582 does exactly what this PR proposed and more, fixing #1581

PR #680, opened 2020-01-21
Support for reporting USB 3.2 Gen2x2 speeds over libusb

hjmallon:
This PR is presented for comments only. I have not got a device so it is untested. I put it on the mailing list [here](https://sourceforge.net/p/libusb/mailman/libusb-devel/thread/191976D6-76DB-4864-B840-B6575625A5CA%40codex.online/#msg36900847).

USB 3.2 Gen1x2 and USB3.2 Gen2x2 have 2 tx and 2 rx channels running to the device, giving `10000 Mbit/s (USB SuperSpeedPlus Gen1x2)` and `20000 Mbit/s (USB SuperSpeedPlus Gen2x2)`. Support was tied into Linux in 4.18 and is shown in usbfs as `tx_lanes==2` and `rx_lanes==2`. Is this the way libusb should support reporting this (as separate functions to `libusb_get_device_speed` which would report the speed of a lane, `libusb_get_device_num_tx_lanes` and `libusb_get_device_num_rx_lanes`)?

macOS 10.15 SDK supports `kUSBDeviceSpeedSuperPlusBy2` but doesn't seem to have a way to identify `kUSBDeviceSpeedSuperBy2` (i.e. USB 3.2 Gen1x2 @ 2 x 5000Mbit/s).

See links:
* https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=460fd21618bf95a16b30500eb4b5452dab04e023
* https://github.com/Nebulino/MacOS-X-SKDs/blob/bbefac19f53aed1f1322d649bbb91591cb5b5006/MacOSX10.15.sdk/System/Library/Frameworks/IOKit.framework/Versions/A/Headers/usb/USB.h#L924

### DIFF
--- a/examples/xusb.c
+++ b/examples/xusb.c
@@ -808,6 +808,21 @@ static void print_device_cap(struct libusb_bos_dev_capability_descriptor *dev_ca
 	}
 }
 
+static const char* usb_speed_name(int speed, int num_rx_lanes, int num_tx_lanes)
+{
+	const char* const speed_name[6] = { "Unknown", "1.5 Mbit/s (USB LowSpeed)", "12 Mbit/s (USB FullSpeed)",
+		"480 Mbit/s (USB HighSpeed)", "5000 Mbit/s (USB SuperSpeed)", "10000 Mbit/s (USB SuperSpeedPlus)" };
+	const char* const speed_name_x2[2] = { "10000 Mbit/s (USB SuperSpeedPlus Gen1x2)", "20000 Mbit/s (USB SuperSpeedPlus Gen2x2)" };
+
+	if (num_rx_lanes == 2 && num_tx_lanes == 2) {
+		if ((speed<4) || (speed>5)) speed=0;
+		else return speed_name_x2[speed - 4];
+	}
+
+	if ((speed<0) || (speed>5)) speed=0;
+	return speed_name[speed];
+}
+
 static int test_device(uint16_t vid, uint16_t pid)
 {
 	libusb_device_handle *handle;
@@ -817,10 +832,10 @@ static int test_device(uint16_t vid, uint16_t pid)
 	struct libusb_config_descriptor *conf_desc;
 	const struct libusb_endpoint_descriptor *endpoint;
 	int i, j, k, r;
+	int num_rx_lanes, num_tx_lanes;
 	int iface, nb_ifaces, first_iface = -1;
 	struct libusb_device_descriptor dev_desc;
-	const char* const speed_name[6] = { "Unknown", "1.5 Mbit/s (USB LowSpeed)", "12 Mbit/s (USB FullSpeed)",
-		"480 Mbit/s (USB HighSpeed)", "5000 Mbit/s (USB SuperSpeed)", "10000 Mbit/s (USB SuperSpeedPlus)" };
+
 	char string[128];
 	uint8_t string_index[3];	// indexes of the string descriptors
 	uint8_t endpoint_in = 0, endpoint_out = 0;	// default IN and OUT endpoints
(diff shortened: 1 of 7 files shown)


hjelmn:
Seems reasonable to me. Just to verify, is this info not available through any current libusb call?

mvp:
Wouldn't it be better to create single function `libusb_get_device_lanes`?
It could return structure with `{rx_lanes, tx_lanes}`.
Alternatively, it could return 16 bit value, where high byte is number of rx lanes, and low byte is number of tx lanes - that itself could be wrapped as enum too for better readability.

hjmallon:
To be honest I don't really mind which API. I don't think there is a way to get this info at the moment. I still don't have any hardware to test this with (and 2x2 seems to be a bit esoteric still) so 🤷‍♂️ .

My original intent was to determine whether the common OSs had support for 2x2 and how they exposed this info, so much of this PR would be relevant whichever API we use.

wallentx:
Sorry if this is an out of place question, but do gen2x2 devices work at all on linux currently? I've got an asm3242 controller and a capable drive, but have not found a kernel that will show it running at expected speeds.

hjmallon:
> Sorry if this is an out of place question, but do gen2x2 devices work at all on linux currently? I've got an asm3242 controller and a capable drive, but have not found a kernel that will show it running at expected speeds.

I have no idea. I have no USB 3.2 gen 2x2 devices :)

wallentx:
> > Sorry if this is an out of place question, but do gen2x2 devices work at all on linux currently? I've got an asm3242 controller and a capable drive, but have not found a kernel that will show it running at expected speeds.
> 
> I have no idea. I have no USB 3.2 gen 2x2 devices :)
> 
> Do you get the `tx_lanes` and `rx_lanes` attrs that this change uses?

I built and installed the branch from your fork, but it doesn't seem to be doing anything for me. My guess is this has to be kernel related? I have another USB 3.1 PCIe expansion card  (ASM1051E controller), and I get SuperSpeed+¹⁰ working if I plug my gen 2x2 drive into that, but only 5000 if I plug it into the ASM3242. In dmesg, I see some sea errors referring to the scsi enclosure, and a message about it being recognized as a Gen1x2 device.

Meanwhile, my Samsung T7 USB 3.1 SSD gets SuperSpeed+¹⁰ on it no problem. 

PXL_20210224_085023541~3

PXL_20210224_083433950~3

wallentx:
PXL_20210224_204011840

Seems like kernel issue?

hjmallon:
The fact it is coming up at the wrong speed is not to do with libusb. As you say that is likely some kernel driver issue.

wallentx:
> The fact it is coming up at the wrong speed is not to do with libusb. As you say that is likely some kernel driver issue.

Thank you for your learned eye! Unfortunately, this is where I'm lost as to what to do next in order to surface this issue. I feel a bit stupid asking this, but what's the appropriate intake for reporting issues like this? Kernel.org?
I realize this is beyond the scope of this issue, and the libusb project.. but I guess the protocol has to work in order to be relevant in this space.

mcuee:
You can report to the Linux USB mailing list.
http://www.linux-usb.org/mailing.html

wallentx:
A follow-up: I ended up giving up on this, but this madlad got it working and left detailed instructions https://www.reddit.com/r/linux_devices/comments/lqe3lv/asm3242_drivers_to_get_usb_32_gen_2x2_superspeed/h9fwy4y

mcuee:
@hjmallon 

hjmallon:
I have updated the PR. My remaining comments:

* API discussion above not truely resolved
* I have still not got any USB gen2x2 devices and this is untested

tormodvolden:
Thanks for updating the PR. Yes, I agree the API discussion is not resolved. And if no one is around to test it and use it, I wouldn't give it priority for now. I would be happy to see an application that could need such a feature and what the developers thought about it. Like it was brought up in the ML thread, I initially thought the libusb_speed enum could have contained this, but I understand that through how the USB standard added this, it is more complicated.

hjmallon:
Yeah I added the `libusb_get_device_num_{t,r}x_lanes` functions as they mapped nicely to the Linux API, but I'm sure other opinions are available.

PR #682, opened 2020-01-24
threads_posix: Use thread-local storage to cache thread ID

dickens:
Trying to capture debug logs that reproduce a problem can be tricky.
Turning up the debug level will automatically make the program a bit
slower. This alone cane make timing-sensitive bugs "disappear" when
capturing logs. One of the hot paths for debug messages is fetching the
thread ID, which is immeasurably helpful in understanding thread
interactions within the library. Unfortunately most implementations
require a system call to fetch the executing thread's ID, which isn't
exactly going to help in the way of execution time.

Add a check for thread-local storage support when configuring the
library to build. If the toolchain provides this support, only one
system call will be required per thread. This check is only done for
non-Windows systems because thread-local storage is inefficiently
implemented on MinGW.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -323,16 +323,30 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])],
 	nopointersign_cflags="-Wno-pointer-sign", nopointersign_cflags="")
 CFLAGS="${saved_CFLAGS}"
 
-dnl check for -std=gnu99 compiler support
+dnl check for -std=gnu11 compiler support
 saved_CFLAGS="${CFLAGS}"
-CFLAGS="-std=gnu99"
-AC_MSG_CHECKING([whether CC supports -std=gnu99])
+CFLAGS="-std=gnu11"
+AC_MSG_CHECKING([whether CC supports -std=gnu11])
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])],
 	[AC_MSG_RESULT([yes])
-	 AM_CFLAGS="${AM_CFLAGS} -std=gnu99"],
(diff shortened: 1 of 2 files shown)


dickens:
@hjelmn @jwrdegoede @LudovicRousseau any thoughts?

hjelmn:
I would just use _Thread_local and require C11. macOS can be bumped to C11 as well. clang and gcc have been compliant for years. RHEL 7 users will have to install another compiler to build but that isn't a big deal.

dickens:
My concern with that is that it’s not just compiler support at that point.  Using _Thread_local requires the system to provide the threads.h header file, which wasn’t made available by glibc until 2.28.  Otherwise we have to jump through hoops in the header files to define _Thread_local if threads.h isn’t available, which to me isn’t much better than just detecting __thread support which has been available for many years.

hjelmn:
threads.h is not required for _Thread_local. It is only needed for the thread_local alias. Only need a new enough compiler (gcc 4.9+ for example).

LudovicRousseau:
No objection to the change.
I would prefer to use an explicit `HAVE_THREAD_LOCAL_STORAGE` instead of `HAVE_TLS`. TLS can be so many different things https://en.wikipedia.org/wiki/TLS

dickens:
I bumped the compiler check from `-std=gnu99` to `-std=gnu11`.  This isn't a hard requirement, the same as before.

I also renamed the config token `HAVE_CC_THREAD_LOCAL` to clarify that it is a compiler feature rather than an available function.

dickens:
Any further feedback or objections?  If not, I plan on merging this soon.

PR #683, opened 2020-01-24
Fix msvc 2019 project files (internal linkage)

winterrace:
The Visual Studio 2019 project files did internally still link to the 2017 ones.
As a consequence, compilation fails (might succeed if VS2017 build files are installed, though).

After these corrections it compiles flawlessly again.

### DIFF
--- a/msvc/fxload_2019.vcxproj
+++ b/msvc/fxload_2019.vcxproj
@@ -95,11 +95,11 @@
     <ClInclude Include="..\examples\ezusb.h" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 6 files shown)


dickens:
Of course, thank you. I wonder why it didn't cause issues on my installation.

winterrace:
Probably because you had the VS2017 build files installed. It prompted me to either install them or to upgrade the project (140 --> 142). However, the latter did not work somehow, the warning persisted. Finally I found the misleading links in the project files and redirecting them to the vs 2019 projects solved the issue.

PR #685, opened 2020-01-27
core: Kill usbi_backend.clock_gettime() function

dickens:
Out of all the backends supported by libusb, only two need to provide an
implementation of the clock_gettime() function. Windows completely lacks
such a function and versions of Mac OS prior to 10.12 do not provide it.
In all other cases the backend simply ends up calling the C library's
clock_gettime() function.

Let's optimize for the common case and check for the availability of
clock_gettime() during configure. If available, we will just call it
directly from any part of the library that needs it. If not available,
the backend is required to provide an implementation of
usbi_clock_gettime() that matches the current requirements.

### DIFF
--- a/Xcode/config.h
+++ b/Xcode/config.h
@@ -1,5 +1,12 @@
 /* config.h.  Manually generated for Xcode.  */
 
+/* On 10.12 and later, use newly available clock_*() functions */
+#include <AvailabilityMacros.h>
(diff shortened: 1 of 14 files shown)


dickens:
@hjelmn I'm not sure what the proper thing to do for defining HAVE_CLOCK_GETTIME on 10.12 or later for libusbi.h to see.

PR #689, opened 2020-02-05
linux_udev: provide an event thread name

AlexanderStein:
Instead of having just the application name as thread name, provide a more
descriptive one, which can e.g. read by htop.
If setting the name fails, the thread will still work as intended, just
raise a warning in this case.

Signed-off-by: Alexander Stein <alexander.stein@mailbox.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -143,6 +143,7 @@ linux)
 		THREAD_CFLAGS="-pthread"
 		LIBS="${LIBS} -pthread"
 	fi
+	AC_CHECK_FUNCS([pthread_setname_np])
 	;;
 netbsd)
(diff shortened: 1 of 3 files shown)


dickens:
Thanks for the patch.  Some points to address:

1. For Linux, it is uncommon for threads to contain a space in their name.  It should probably have an underscore or hyphen, the choice which I leave to you.
2. The netlink implementation can use this as well.
3. The `pthread_setname_np()` doesn’t exist in all versions of glibc (or whatever C library that might be there).  This will need to need a check added in the configure script.  This check should be limited to the Linux backend.
4. The event thread should itself call `pthread_setname_np()` with `pthread_self()` in order to avoid a spurious failure that can result from a race condition in procfs.

PR #699, opened 2020-02-22
fix MAX_PATH_LENGTH

a1exdandy:
The `MAX_PATH_LENGTH` in `libusb` `Windows` backend is used as size of `dev_id` buffer. This buffer used for retreiving `Device Instance Id` by `SetupDiGetDeviceInstanceIdA` function. Acording to [this](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/instance-ids), `Device Instance Id` must be less than `MAX_DEVICE_ID_LEN = 200`. So, value of `128` maybe not enough.

As real example, see [this comment](https://github.com/pgarba/King/issues/5#issuecomment-589983940).


dickens:
Applied, thanks.

PR #701, opened 2020-03-03
darwin: Fix race condition that results in segmentation fault

dickens:
Commit 763668cc92 ("darwin: fix occasional dead-lock on libusb_exit")
resolved the deadlock situation observed in #112, but unfortunately
there is a very rare race condition that can occur when the asynchronous
event thread exits the run loop before CFRunLoopWakeUp() is called. This
can occur when the shutdown source signal is processed as part of other
events in the run loop, in which case the thread was already "awake".

Prior to this change I was able to consistently trigger a segmentation
fault within 10,000 iterations of a libusb_init()/libusb_exit() loop.
With this change I reached over 4 million iterations without issue.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -47,23 +47,6 @@
   #include <objc/objc-auto.h>
 #endif
 
-#if MAC_OS_X_VERSION_MIN_REQUIRED >= 101200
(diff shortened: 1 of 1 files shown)


dickens:
@hjelmn any objections?

PR #710, opened 2020-04-03
netbsd: Recognize device timeouts.

niacat:
Recognize device timeout errno values returned by the kernel and translate them to the appropriate libusb error code.


LudovicRousseau:
Looks correct to me.

PR #712, opened 2020-04-09
Darwin: use IOUSBDevice as darwin_device_class for macOS Catalina

### LINKED ISSUE #693: macOS Catalina: wrong detection of parent devices
I'm working with `libusb` on my laptop. I need to get the whole USB location of my device. Device (_0781:5597_) is connected to MacBook over the USB hub. So, I expect that it will have more than one USB port in its location. But when I run `examples/listdevs` I got the following result:
```shell
➜  libusb git:(a36862b) ✗ ./examples/listdevs
05ac:027b (bus 128, device 5) path: 5
-> 0781:5597 (bus 20, device 11) path: 4
05e3:0608 (bus 20, device 10) path: 4
05e3:0608 (bus 20, device 9) path: 2
05e3:0751 (bus 20, device 6) path: 1
2109:0102 (bus 20, device 3) path: 4
1a40:0801 (bus 20, device 2) path: 2
0bda:8153 (bus 1, device 2) path: 3
2109:2817 (bus 20, device 1) path: 7
2109:0817 (bus 1, device 1) path: 3
```

If I checkout to this changes https://github.com/libusb/libusb/pull/282, I'll get the following output:
```shell
➜  libusb git:(871eb29) ✗ ./examples/listdevs
05ac:027b (bus 128, device 5) path: 5
-> 0781:5597 (bus 20, device 11) path: 7.2.2.4.4
05e3:0608 (bus 20, device 10) path: 7.2.2.4
05e3:0608 (bus 20, device 9) path: 7.2.2
05e3:0751 (bus 20, device 6) path: 7.2.1
2109:0102 (bus 20, device 3) path: 7.2.4
1a40:0801 (bus 20, device 2) path: 7.2
0bda:8153 (bus 1, device 2) path: 3.3
2109:2817 (bus 20, device 1) path: 7
2109:0817 (bus 1, device 1) path: 3
```
The second path looks much better, but I have a question. Does the Usb ports path should be equal the LocationID of the device? 
The second output returned 7.2.2.4.4 ports path, but device LocationID is `14222440` (2.2.2.4.4 is the ports path, right?)

Here is the output of `ioreg` for my device:
Screenshot 2020-02-14 at 19 59 15

I'm a newbie in MacOs and don't really understand what is going wrong. Could you help me, please?

Output of `uname -a` on my machine:
Darwin Igors-MacBook-Pro.local 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64

IgorAnohin:
`kIOUSBDeviceClassName` define from `IOUSBLib.h` file was changed from `IOUSBDevice` to `IOUSBHostDevice` in _macOS Catalina_. It might be checked by printing `darwin_device_class` value. As far as I understand, `libusb` worked with `IOUSBDevice` on macOS.

In previous macOS versions, it was always defined as `IOUSBDevice`:
```
 #define kIOUSBDeviceClassName		 "IOUSBDevice"
```
In macOS Catalina it looks as follows:
```
 #if MAC_OS_X_VERSION_MIN_REQUIRED <= MAC_OS_X_VERSION_10_14
 #define kIOUSBDeviceClassName           "IOUSBDevice"
 #define kIOUSBInterfaceClassName        "IOUSBInterface"
 #else
 #define kIOUSBDeviceClassName           kIOUSBHostDeviceClassName
 #define kIOUSBInterfaceClassName        kIOUSBHostInterfaceClassName
 #endif /* MAC_OS_X_VERSION_MIN_REQUIRED <= MAC_OS_X_VERSION_10_14 */
```

By default, macOS Catalina uses `IOUSBHostDevice`. The problem is that using the `IOUSBHostDevice` class misses some devices.  This has been described in 044a7ee commit by @hjelmn. To work only with `IOUSBDevice`, `darwin_device_class` has been explicitly set to `IOUSBDevice`.

Closes #693

For reference, here's the output of the 'listdevs' example before my changes:
```shell
05ac:027b (bus 128, device 2) path: 5
2109:0102 (bus 20, device 6) path: 4
05e3:0751 (bus 20, device 5) path: 1
1a40:0801 (bus 20, device 3) path: 2
0bda:8153 (bus 0, device 2) path: 3
2109:2817 (bus 20, device 1) path: 8
2109:0817 (bus 0, device 1) path: 4
```

And here's the output after my changes:
```shell
05ac:027b (bus 128, device 2) path: 5
2109:0102 (bus 20, device 6) path: 3.2.4
05e3:0751 (bus 20, device 5) path: 3.2.1
1a40:0801 (bus 20, device 3) path: 3.2
0bda:8153 (bus 0, device 2) path: 2.3
2109:2817 (bus 20, device 1) path: 3
2109:0817 (bus 0, device 1) path: 2
```

**P.s.**
I'm not sure about the commit's description. This is my first PR to `libusb`

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -68,7 +68,7 @@ static CFRunLoopSourceRef libusb_darwin_acfls = NULL; /* shutdown signal for eve
 
 static usbi_mutex_t darwin_cached_devices_lock = PTHREAD_MUTEX_INITIALIZER;
 static struct list_head darwin_cached_devices;
-static const char *darwin_device_class = kIOUSBDeviceClassName;
+static const char *darwin_device_class = "IOUSBDevice";
 
 #define DARWIN_CACHED_DEVICE(a) (((struct darwin_device_priv *)usbi_get_device_priv((a)))->dev)
 


mvp:
On MacOS prior to 10.14, `kIOUSBDeviceClassName` is defined as `"IOUSBDevice"`, and in 10.15 and later `kIOUSBDeviceClassName` is defined as `"IOUSBHostDevice"` (you can find it in header files at `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Versions/A/Headers/usb`).
In this PR, you are checking if MacOS is 10.15 and never just to make sure that `darwin_device_class` is `"IOUSBHostDevice"`, which would be equal to it on older MacOS anyway.

It seems that much easier approach would be simply always use `"IOUSBHostDevice"` for `darwin_device_class`.
I tried this, and it works for me:
```
iff --git a/libusb/os/darwin_usb.c b/libusb/os/darwin_usb.c
index 4639a7e..38cc5a3 100644
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -68,7 +68,7 @@ static CFRunLoopSourceRef libusb_darwin_acfls = NULL; /* shutdown signal for eve

 static usbi_mutex_t darwin_cached_devices_lock = PTHREAD_MUTEX_INITIALIZER;
 static struct list_head darwin_cached_devices;
-static const char *darwin_device_class = kIOUSBDeviceClassName;
+static const char *darwin_device_class = "IOUSBDevice";

 #define DARWIN_CACHED_DEVICE(a) (((struct darwin_device_priv *)usbi_get_device_priv((a)))->dev)
```

IgorAnohin:
@mvp, thank you for the feedback.
I suggested the same solution in my issue: https://github.com/libusb/libusb/issues/693#issuecomment-610997611

You're right, that's much easier! I decided to use a bit different solution because _"The similar changes were applied by @hjelmn , when he added IOUSBHostDevice in b7526c1"_. I wanted to add changes similar to the old ones.

If everyone will agree, that we may just replace `darwin_device_class = kIOUSBDeviceClassName` with `darwin_device_class = "IOUSBDevice"`, I'll change it. _To be honest, I'll prefer that variant, if it's possible!_

mvp:
Yes, I totally support this method. You should amend this PR with just one line patch:

    static const char *darwin_device_class = "IOUSBDevice";

That would make everyone happy! :-)

seanm:
In the macOS 10.14 SDK we see:

```
#define kIOUSBDeviceClassName		"IOUSBDevice"
#define kIOUSBInterfaceClassName	"IOUSBInterface"
```

in the macOS 10.15 SDK we see:

```
#if MAC_OS_X_VERSION_MIN_REQUIRED <= MAC_OS_X_VERSION_10_14
#define kIOUSBDeviceClassName           "IOUSBDevice"
#define kIOUSBInterfaceClassName        "IOUSBInterface"
#else
#define kIOUSBDeviceClassName           kIOUSBHostDeviceClassName
#define kIOUSBInterfaceClassName        kIOUSBHostInterfaceClassName
#endif /* MAC_OS_X_VERSION_MIN_REQUIRED <= MAC_OS_X_VERSION_10_14 */
```

Apple seems to have made a deliberate change here, and we should probably first understand why.

Also note that it's conditionalized on deployment target. @IgorAnohin what deployment target are you building with when you experience missing devices?

IgorAnohin:
> Apple seems to have made a deliberate change here, and we should probably first understand why.

I tried to find some description of these changes but didn't find anything. I'll be glad if someone may share a link with info for me.

> what deployment target are you building with when you experience missing devices?

I don't append any extra targets. My building steps:
1) `./autogen.sh`
2) `make`

By default `MAC_OS_X_VERSION_MIN_REQUIRED == 101500` on my PC (macOS Catalina). It's equal `MAC_OS_X_VERSION_10_15` define. My `uname -a`:
```shell
Darwin Igors-MacBook-Pro.local 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
```


**Possible alternative solution**
I think `MAC_OS_X_VERSION_MIN_REQUIRED` might be another solution to this problem. In fact, I do not know the side effects of explicitly using `darwin_device_class = IOUSBDevice`. And I don’t know if they exist.
But if we lower `MAC_OS_X_VERSION_MIN_REQUIRED` on macOS Catalina to use "#define kIOUSBDeviceClassName IOUSBDevice", the problem will be solved too. This can be a solution without side effects. @seanm what do you think about that?

seanm:
MAC_OS_X_VERSION_MIN_REQUIRED (aka the 'deployment target') isn't something that projects, especially libraries, should be specifying _within themselves_.  It is the job of the person building an app/library to decide: what is the oldest version of macOS that I want this to work with *at runtime*.  You specify that at compile-time by compiler flag or environment variable, see:

https://clang.llvm.org/docs/CommandGuide/clang.html

For example, my app requires macOS 10.11 or later, so when I build all my dependencies, like libusb, I specify a deployment target of 10.11.

So libusb shouldn't override the MAC_OS_X_VERSION_MIN_REQUIRED choice within its makefile/code.

In my opinion, the least bad change is this proposal:

```
-static const char *darwin_device_class = kIOUSBDeviceClassName;
+static const char *darwin_device_class = "IOUSBDevice";
```

But I don't know what side effects that might have.

I wonder if any recent WWDC video has info on 10.15 USB changes... or the Apple developer forums...

IgorAnohin:
@seanm , thank you for the feedback!

I force pushed the PR to use the one-line patch. PR description was changed too.

I checked the Apple developer forums and found nothing :(

dickens:
I do not have the Apple resources available to me to verify this change, so I will have to defer to @hjelmn.

hjelmn:
If this works with newer versions of macOS then this is fine. Let's go ahead with this mitigation then I will take some time to understand why the code isn't working as intended.  After this is in we should start the release process.

PR #713, opened 2020-04-13
core: provide function libusb_get_refcnt()

### LINKED ISSUE #355: Feature Request: Add ref count query function
In massively threaded apps that handle dozens of devices it sometimes gets tricky to debug the app when you don't know who's holding on to the device.

It would be nice to be able to query the refcount of a libusb_device or libusb_device_handle so I can get a hint to reorder my release functions.

xypron:
libusb relies on correct reference counting. For debugging a user
application it is useful to have a function to check the current reference
count of a device.

Closes #355

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -387,6 +387,7 @@ if (cfg != desired)
   * - libusb_get_port_number()
   * - libusb_get_port_numbers()
   * - libusb_get_port_path()
+  * - libusb_get_refcnt()
   * - libusb_get_ss_endpoint_companion_descriptor()
   * - libusb_get_ss_usb_device_capability_descriptor()
   * - libusb_get_string_descriptor()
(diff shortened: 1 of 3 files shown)


dickens:
NAK from me.  I see this as a specialized feature that buggy applications would need, not as a generally useful addition to the library.

I won't close this to give time for the other maintainers to weigh in.

xypron:
@dickens 
How do identify if a new application that you develop is handling references correctly? Is there any feature I missed that will give me a warning?

LudovicRousseau:
NAK also from me.
If it is used to debug your application then you can use a special version of libusb with this patch just for that. That is the benefit of Free Software: you can modify it to make it work better for you.

No need to include this change in the standard version of libusb.

PR #714, opened 2020-04-17
hotplug: Fix definition of libusb_hotplug_register_callback()

dickens:
The signature of the libusb_hotplug_register_callback() is slightly
incorrect in two regards:

  1) The 'events' parameter is meant to represent a bitwise OR of
     libusb_hotplug_event values. By OR'ing multiple values together,
     the result is not something that is actually a libusb_hotplug_event
     enumeration.

  2) The 'flags' parameter is meant to represent a bitwise OR of
     libusb_hotplug_flag values. The same considerations as above apply,
     though this has not practically been the case as there is currently
     only one flag value defined. However, a special value was already
     defined to represent the absence of any flags, which hinted at the
     problem with how it is currently defined.

Address these two issues by changing the types of the 'events' and
'flags' parameter to plain integers. As enumerations should already be
promoted to integers, this change should not cause any ABI concerns.

### DIFF
--- a/libusb/hotplug.c
+++ b/libusb/hotplug.c
@@ -143,6 +143,13 @@ int main (void) {
 \endcode
  */
 
+#define VALID_HOTPLUG_EVENTS			\
+	 (LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED |	\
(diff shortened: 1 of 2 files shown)


dickens:
I wanted to put this out there for review.  Other maintainers please review provide feedback.

PR #719, opened 2020-04-23
Adding libusb_get_device_by_dbcc_name

frederikvannoote:
Credits to @kurtpattyn


dickens:
NAK for several reasons:

1. This doesn't build.  The `sanitize_path()` function no longer exists.
2. The formatting of the source code is incorrect.
3. Exported libusb functions **never** live in the backend.
4. This is a function only useful for one OS.

The problem that is trying to be solved can very likely be done using already existing libusb functions.  If it cannot, let's discuss further.

Issue #721, opened 2020-04-28
Handle disconnected device on Windows 10


svonohr:
I want to detect when a user disconnects the device while the device is open. Libusb provides the error codes `LIBUSB_ERROR_NO_DEVICE`/`LIBUSB_TRANSFER_NO_DEVICE` for this. However, when testing with my Windows 10 machine and the WinUSB driver (v10.0.18362.1) I only get generic IO errors.

I’ve looked at the Windows backend and found that a currently not handled error is reported to `windows_transfer_callback`. The reported error is `ERROR_NO_SUCH_DEVICE` (Code: 433), which sounds like it does what I want. The error code is present in my `winerror.h` from the 10.0.18362.0 SDK, but not in any of the previous versions (e.g. not present in 10.0.17763.0). It seems like this is a new error code added in a recent Windows 10 update. I think this new error should be properly handled by libusb and mapped to `LIBUSB_TRANSFER_NO_DEVICE`.

PR #722, opened 2020-04-28
Windows: Translate ERROR_NO_SUCH_DEVICE to LIBUSB_TRANSFER_NO_DEVICE

svonohr:
Windows SDK 10.0.18362.0 adds a new error code ERROR_NO_SUCH_DEVICE which is returned when the device is disconnected.

Proposed change for #721

### DIFF
--- a/libusb/os/windows_common.c
+++ b/libusb/os/windows_common.c
@@ -430,6 +430,7 @@ static void windows_transfer_callback(const struct windows_backend *backend,
 		break;
 	case ERROR_FILE_NOT_FOUND:
 	case ERROR_DEVICE_NOT_CONNECTED:
+	case ERROR_NO_SUCH_DEVICE:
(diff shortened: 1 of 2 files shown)


svonohr:
Somehow Cygwin doesn't like my version check. Maybe just check for the existance of `ERROR_NO_SUCH_DEVICE`?

dickens:
I would just add the following to windows_common.h (a good place is around the USBD_STATUS stuff)

```
#ifndef ERROR_NO_SUCH_DEVICE
#define ERROR_NO_SUCH_DEVICE	433L
#endif
```

Then just unconditionally use it in the switch statement.

svonohr:
The checks are successful now. From my side I think this is ready.

dickens:
Applied, thanks.

Issue #723, opened 2020-04-29
[darwin] race condition during reset device (another process has device opened for exclusive access)


mringwal:
Since 1.0.23, libusb re-enumerates a device when libusb_reset_device is called. During re-enumeration, I get this error:

libusb: warning [darwin_open] USBDeviceOpen: another process has device opened for exclusive access

I'm certain that only my process is accessing the USB device - a Bluetooth controller.

(Background: [BTstack](https://github.com/bluekitchen/btstack/) first calls libusb_open, then  libusb_reset_device,  then starts to configure it. libusb_reset_device does not return an error).

The error only happens maybe with 50%. 

Logs:
- [device-open-ok.txt](https://github.com/libusb/libusb/files/4550617/device-open-ok.txt)
- [device-opened-exclusive-access-fail.txt](https://github.com/libusb/libusb/files/4550618/device-opened-exclusive-access-fail.txt)

The only difference is the 'exclusive access error'.

I have no experience with IOKit or libusb internals, but after adding a 10 ms delay in darwin_reset_device() after the ' while (dpriv->in_reenumerate) { } ', it the error does not show up.

seanm:
Intermittent errors that go away adding sleeps could be a threading bug.  Could you trying building and testing with ThreadSanitizer?

mringwal:
@seanm Do you have a link to 'ThreadSanitizer'. I did use the AddressSanitizer in BTstack before to find memory corruptions. Is this similar?

As the error is highly reproducible, it's easy to argue that if there's a thread switch where I've added the 10 ms sleep, it would work. What do you expect from the ThreadSanitizer?

I still don't understand what process/thread could access the device that has just been opened. There's a polling while() that checks if the enumeration is done, maybe that exits to early (flag is set, but enum not done).

seanm:
Yes, it's similar to AddressSanitizer. See https://clang.llvm.org/docs/ThreadSanitizer.html

mcuee:
@mringwal Can you try the latest git to see if the issue is still there?

mringwal:
Thanks for the notification and the update.

With current master/1.0.24, the error does not show up anymore. Excellent!

PR #724, opened 2020-04-29
[darwin] wait 10 ms before restoring state after re-enumeration

mringwal:
This is a minimal work-around/hack to avoid issue #723 - and I'm aware that this isn't a correct fix, but it improves usability until there's one.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1702,6 +1702,10 @@ static int darwin_reset_device(struct libusb_device_handle *dev_handle) {
     nanosleep (&delay, NULL);
   }
 
+  /* avoid race condition with 10 ms delay (hack) */
+  struct timespec delay = {.tv_sec = 0, .tv_nsec = 10000000};
+  nanosleep (&delay, NULL);
+
   /* compare descriptors */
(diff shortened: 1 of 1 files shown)


hjelmn:
10ms is a long time. I think better than trying to race with the system would be to add a couple of retries. If the device can not be reopened in 3-4 (maybe 5) attempts then reset should return an error indicating the device will need to be reopened by the caller.

hjelmn:
Actually, we already return the correct error code. ```LIBUSB_ERROR_NOT_FOUND``` is returned if the device could not be re-opened.

mringwal:
Hi, thanks for looking into this. If I remember correctly, the problem is that the reset call returns without an error and then something else fails later. There's already a `while (dpriv->in_reenumerate){..}` in there, but the flag re-enumeration isn't complete when the flag is set.

mcuee:
@mringwal Maybe you can try latest git to see if the issue is still there or not.

hjelmn:
Yes. Please verify.

mringwal:
Thanks, will do soon.

mringwal:
Thanks for the notification and the update.

With current master/1.0.24, the error does not show up anymore. Excellent!

PR #733, opened 2020-05-20
darwin: Fix stale descriptor information post reset

codido:
As part of the recent removal of the backend get_device_descriptor()
function, the darwin backend was modified to update the cached
device_descriptor when a new device is processed. However, this wasn't
done for cases where the device is reused rather than allocated.
This could potentially result in getting stale data when using the
device following a reset.

In addition, when the device is reused, its session ID should be kept
up-to-date in the cached devices list, so it could be found in subsequent
resets.

Signed-off-by: Ido Yariv <ido@wizery.com>


hjelmn:
Looks ok to me. The usage of static_assert cements the C11 requirement for Darwin. I may update the configury to enforce this requirement.

PR #734, opened 2020-05-21
Fix alignment for 32-bit words in parse_descriptor

saleemrashid:
`parse_descriptor` was aligning 32-bit words to 2 bytes, instead of 4 bytes. This didn't cause any issues before, because the only time the 32-bit word code path is used is from a 3 byte offset (which incidentally aligns to 4 bytes). However, a 1 byte offset would incorrectly align to 2 bytes.

Additionally, I added `static_assert`s to check the alignments are as expected because the C standard does not dictate an alignment. Alternatively, we could rewrite it to use `alignof` in the pointer arithmetic.

This pull request is completely untested.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -53,14 +53,14 @@ static void parse_descriptor(const void *source, const char *descriptor, void *d
 			*dp++ = *sp++;
 			break;
 		case 'w':	/* 16-bit word, convert from little endian to CPU */
-			dp += ((uintptr_t)dp & 1);	/* Align to word boundary */
+			dp += ((uintptr_t)dp & 1);			/* Align to word boundary */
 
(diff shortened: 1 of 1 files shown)


dickens:
I was working on a similar change to the parsing code that included this.

The static assertions for alignment are a bit overkill and IMO unnecessary.  If those assertions are ever false, I suspect much more than the parsing code will be affected.

hjelmn:
Agreed. The static asserts are overkill and unnecessary.

dickens:
Waiting for an update before this can be merged.

saleemrashid:
Updated. Sorry that you deemed the assertions unnecessary, I included them because I don't write code that depends on implementation-defined behaviour.

PR #737, opened 2020-05-25
Add release-all-interfaces functionality

Roman-Koshelev:


Roman-Koshelev:
Added function  libusb_release_all_interfaces().  I also think this should be called from libusb_close()

dickens:
I don't understand the purpose of this interface?  Closing a device implicitly releases all interfaces already.  What does this gain?

Roman-Koshelev:
"Closing a device implicitly releases all interfaces already." - without an explicit call, releasing and connecting the kernel driver takes several seconds

dickens:
I think I understand now.  You are referring to closing a device that has claimed one or more interfaces and the user has set auto-detach on that device?  If so, then this needs to be handled differently.  The close path should reattach any interfaces that were auto-detached and only those interfaces.

hjelmn:
Please do not update the nano version in a PR. This makes extra work when it is merged by a maintainer.

Roman-Koshelev:
"any interfaces that were auto-detached and only those interfaces". Since the auto_detach_kernel_driver flag is the same for all interfaces, if it is set, you will have to free all the interfaces.

dickens:
> "any interfaces that were auto-detached and only those interfaces". Since the auto_detach_kernel_driver flag is the same for all interfaces, if it is set, you will have to free all the interfaces.

That depends on _when_ `libusb_set_auto_detach_kernel_driver()` is called.  You could claim any number of interfaces prior to calling this function, and arguably those interfaces should _not_ have kernel drivers automatically reattached because libusb did not automatically detach said drivers.

hjelmn:
@dickens pointed out a flaw in this implementation. This change is a convenience over just closing the interfaces that were opened. I see no reason to pursue this change without further explanation on why this API change should be taken. Will close this for now. Please re-open if you are still interested in making this change.

PR #740, opened 2020-06-10
Fix risk error when discard_urbs

masc2008:
Fix risk error when discard_urbs, A URB cancelled return with "0" still has chance that finished successfully.

Need retune the reap_action based real URB status. The reason is a URB with status 0 will contain data from device, and need report this data to app level.
Otherwise, the data in URB will lost.

The cause is when  tpriv->reap_action != NORMAL, and  tpriv->reap_action ==
CANCELLED, then it will call  usbi_handle_transfer_cancellation(itransfer).
While when urb status is not -ENOENT, that stands a URB is not really
CANCELLED, and the we need further call usbi_handle_transfer_completion
otherthan usbi_handle_transfer_cancellation.


masc2008:
After futher review, this may not a bug with libusb.

PR #744, opened 2020-06-17
darwin: Fix invalid GetPipePropertiesV3 argument

codido:
GetPipePropertiesV3 seems to require that the bVersion field of the
properties argument be set before calling it:
"Version of the structure. Currently kUSBEndpointPropertiesVersion3.
Need to set this when using this structure"

Not doing so results in an invalid argument error.

Signed-off-by: Ido Yariv <ido@wizery.com>


hjelmn:
Ugh, will remember to run an IO test next time :). Thanks for the fix.

PR #751, opened 2020-07-08
Tests stress: fix printf zd

uril:
1. fix printf format for ssize_t
2. chmod 0644 a .c file


uril:
@dickens Thanks for reviewing. I updated the patch as you suggested.

Issue #743, opened 2020-06-13
libusb_set_configuration() returns LIBUSB_ERROR_INVALID_PARAM


thierer:
Calling libusb_set_configuration() with a configuration value of 0 or -1 returns a LIBUSB_ERROR_INVALID_PARAM error on win 10 with libusb-1.0.23 (haven't tried any other windows or libusb version; it does work on linux, though).

From what I understand, libusb can't set the configuration on windows (libusb_set_configuration() calls with a configuration value > 0 seem to be silently ignored).

Wouldn't it be best to just ignore errors with configuration values 0 and -1 as well?

I can see why you would return an error (because the function can't do what it was asked to), but the same is true for configuration values > 0 and I find "invalid parameter" not very appropriate as the documentation for libusb_set_configuration() [explicitly mentions 0 and -1 as valid configuration numbers](http://libusb.sourceforge.net/api-1.0/group__libusb__dev.html#ga785ddea63a2b9bcb879a614ca4867bed):

> A configuration value of -1 will put the device in unconfigured state. The USB specifications state that a configuration value of 0 does this, however buggy devices exist which actually have a configuration 0.

mcuee:
Probably the documentation needs to be updated. -1 should not be a valid configuration and 0 is the unconfigured state. The buggy device may be really legacy devices which we should not care now. Anyway, I think the current Windows code behavior is correct.

thierer:
Please note that not only using -1 returns LIBUSB_ERROR_INVALID_PARAM, but also a configuration value of 0 (I'm not sure if your comment "_... and 0 is the unconfigured state_" is meant to express that you consider both -1 and 0 illegal values).

I suspect this https://github.com/libusb/libusb/blob/ad0473413edec14805e59599ff537eba0f5b8d33/libusb/os/windows_winusb.c#L2454-L2461 to be the code that returns the LIBUSB_ERROR_INVALID_PARAM error.

If there are "issues" to select the configuration from WinUSB, as the comment suggests, I think LIBUSB_ERROR_NOT_SUPPORTED (what `hid_submit_control_transfer()` uses) would be more appropriate. 

https://github.com/libusb/libusb/blob/ad0473413edec14805e59599ff537eba0f5b8d33/libusb/os/windows_winusb.c#L3657-L3664

mcuee:
-1 is an illegal value. 
0 means un-configured state. But you will not be able to achieve what you want with WinUSB or libusbk driver.

[...]

PR #752, opened 2020-07-08
winusb: return NOT_SUPPORTED instead of INVALID_PARAM when trying to set a non-default configuration

thierer:
The PR I promised for #743 turned out to be very simple, as all the other platforms already seem to do the right thing.

That said, openbsd, netbsd and darwin don't seem to do any parameter checks for libusb_set_configuration(). Which could simple mean they *do* support setting non-default configurations and de-configuring a device. I can't check that for the former two, but I might for the latter.

I would have also added a comment to [the api docs](http://libusb.sourceforge.net/api-1.0/group__libusb__dev.html#ga785ddea63a2b9bcb879a614ca4867bed), but I couldn't find them in a repository?


mcuee:
For the API Docs, it is in core.c. 
https://github.com/libusb/libusb/blob/349baae0b5e5e8a0aeae9656b69afbcdd6acbe64/libusb/core.c#L1618

PR #756, opened 2020-07-27
libusb: fix a memory leak in sunos_new_string_list func

lzhq28:
In sunos_new_string_list func, if alloc list->string fails,
we will return NULL without free list, which has alread been
allocated successfully.

Fixes: 17348731b4 ('Solaris backend depends on Solaris private symbols')
Signed-off-by: Zhiqiang Liu <lzhq28@mail.ustc.edu.cn>


lzhq28:
@jwrdegoede

PR #760, opened 2020-07-29
Support unrooted android and rooted android

chrisju:
On unrooted Android, we can not operate usb directly and need use `libusb_wrap_sys_device`, we also will failed on `linux_scan_devices`. For 1.0.23, before commit `211d791`,  libusb can not work with rooted android. After commit `211d791`,  libusb can not work with unrooted android. Libusb can not know if it has authority to scan devices, so i add `libusb_init_ex` to specify it manually.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2188,12 +2188,16 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
 		break;
 
 	/* Handle all backend-specific options here */
-	default:
+	case LIBUSB_OPTION_USE_USBDK:
+	case LIBUSB_OPTION_WEAK_AUTHORITY:
 		if (usbi_backend.set_option)
 			r = usbi_backend.set_option(ctx, option, ap);
 		else
 			r = LIBUSB_ERROR_NOT_SUPPORTED;
 		break;
+
+	default:
+		r = LIBUSB_ERROR_INVALID_PARAM;
 	}
 	va_end(ap);
 
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1193,6 +1193,9 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
  * handle for the underlying device. The handle allows you to use libusb to
  * perform I/O on the device in question.
  *
+ * Must call libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY)
+ * before libusb_init if don't have authority to access the usb device directly.
+ *
  * On Linux, the system device handle must be a valid file descriptor opened
  * on the device node.
  *
@@ -2185,15 +2188,12 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
 		break;
 
 	/* Handle all backend-specific options here */
-	case LIBUSB_OPTION_USE_USBDK:
+	default:
 		if (usbi_backend.set_option)
 			r = usbi_backend.set_option(ctx, option, ap);
 		else
 			r = LIBUSB_ERROR_NOT_SUPPORTED;
 		break;
-
-	default:
-		r = LIBUSB_ERROR_INVALID_PARAM;
 	}
 	va_end(ap);
 
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -95,6 +95,9 @@ static int sysfs_available = -1;
 /* how many times have we initted (and not exited) ? */
 static int init_count = 0;
 
+/* is no authority to operate usb device directly */
+static int weak_authority = 0;
+
 /* Serialize hotplug start/stop */
 static usbi_mutex_static_t linux_hotplug_startstop_lock = USBI_MUTEX_INITIALIZER;
 /* Serialize scan-devices, event-thread, and poll */
@@ -381,6 +384,10 @@ static int op_init(struct libusb_context *ctx)
 		}
 	}
 
+    if (weak_authority) {
+        return LIBUSB_SUCCESS;
+    }
+
 	usbi_mutex_static_lock(&linux_hotplug_startstop_lock);
 	r = LIBUSB_SUCCESS;
 	if (init_count == 0) {
@@ -411,6 +418,21 @@ static void op_exit(struct libusb_context *ctx)
 		linux_stop_event_monitor();
 	}
 	usbi_mutex_static_unlock(&linux_hotplug_startstop_lock);
+}
+
+static int op_set_option(struct libusb_context *ctx, enum libusb_option option, va_list ap)
+{
+	UNUSED(ctx);
+	UNUSED(ap);
+
+	switch ((int)option) {
+	case LIBUSB_OPTION_WEAK_AUTHORITY:
+        usbi_dbg("set libusb has weak authority");
+        weak_authority = 1;
+        return LIBUSB_SUCCESS;
+	default:
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+	}
 }
 
 static int linux_scan_devices(struct libusb_context *ctx)
@@ -2688,6 +2710,7 @@ const struct usbi_os_backend usbi_backend = {
 	.caps = USBI_CAP_HAS_HID_ACCESS|USBI_CAP_SUPPORTS_DETACH_KERNEL_DRIVER,
 	.init = op_init,
 	.exit = op_exit,
+	.set_option = op_set_option,
 	.hotplug_poll = op_hotplug_poll,
 	.get_active_config_descriptor = op_get_active_config_descriptor,
 	.get_config_descriptor = op_get_config_descriptor,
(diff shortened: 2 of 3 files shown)


Peter-St:
Hello chrisju,

I was able to sucessfuly test your commit on unrooted Android 8. Thanks for publishing.

chrisju:
@mcuee  Can you add the bug label to this MR.? Current master not work in android apk, and `libusb_wrap_sys_device` is useless actually.

tmm1:
Instead of adding a new public initializer (`libusb_init_ex`), would it be possible to detect the failure and ignore it when building for android?

chrisju:
> Instead of adding a new public initializer (`libusb_init_ex`), would it be possible to detect the failure and ignore it when building for android?

If so, when developing other than android apk, you will not get the error correctly.

dickens:
Please, no.  There doesn't need to be a new function for this.  Something like `libusb_set_option()` would suffice.

mcuee:
Good point. We should use libusb_set_option() if possible, I remember that was introduced for usbdk.

mcuee:
@chrisju Would you please rework the patch based on the suggestion from Chris?

chrisju:
> Please, no. There doesn't need to be a new function for this. Something like `libusb_set_option()` would suffice.

`libusb_set_option` is after `libusb_init`, but `linux_scan_devices` is in `libusb_init`.  `libusb_set_option` can not work. The authority option must set before `libusb_init`.

hartcw:
I also have a fork of libusb that supports unrooted android.
The fork is here: https://github.com/kuvacode/libusb

My approach was to add a new API called libusb_import_android_fd. This allows an android app to get a valid device FD from the java android.hardware.usb.UsbManager API, and pass this into libusb. The next call to libusb_get_device_list() will then return a valid libusb_device that wraps the FD.

chrisju:
> I also have a fork of libusb that supports unrooted android.
> The fork is here: https://github.com/kuvacode/libusb
> 
> My approach was to add a new API called libusb_import_android_fd. This allows an android app to get a valid device FD from the java android.hardware.usb.UsbManager API, and pass this into libusb. The next call to libusb_get_device_list() will then return a valid libusb_device that wraps the FD.

You need consider rooted use on Android.

dickens:
> `libusb_set_option` is after `libusb_init`, but `linux_scan_devices` is in `libusb_init`. So `libusb_set_option` can not work. The authority option must set before `libusb_init`.

There is nothing requiring `libusb_set_option` to be called after `libusb_init`.  You simply pass `NULL` as the context pointer.  Since there is no default context at that point, the backend will receive `NULL` as the context pointer as well.  This is actually a good way to set global options that would apply to all contexts.

chrisju:
I have updated the pull-request, Now i add an option `LIBUSB_OPTION_WEAK_AUTHORITY` to work with Android apk. Must call `libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY)` before call `libusb_init` if use libusb in Android apk.

libusb/core.c
```diff
@@ -2185,15 +2188,12 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
 		break;
 
 	/* Handle all backend-specific options here */
-	case LIBUSB_OPTION_USE_USBDK:
+	default:
```
dickens, line 2191: You need to add a `case` for `LIBUSB_OPTION_WEAK_AUTHORITY`, not move the default up.  The default is meant to catch invalid options.

chrisju: `usbi_backend.set_option` will catch  invalid options.

dickens: No, that’s not the point or the design intent.  The core should validate the options.  The individual backends don’t need to validate, they just operate on what applies to them and return `LIBUSB_ERROR_NOT_SUPPORTED` otherwise.

dickens:
I also think this should be guarded in linux_usbfs.c with `#ifdef __ANDROID__`; it doesn’t make sense for standard Linux.

libusb/os/linux_usbfs.c
```diff
@@ -95,6 +95,9 @@ static int sysfs_available = -1;
 /* how many times have we initted (and not exited) ? */
 static int init_count = 0;
 
+/* is no authority to operate usb device directly */
+static int weak_authority = 0;
+
```
LudovicRousseau, line 100: Maybe the code should use `#ifdef __ANDROID__` around the declaration since `weak_authority` is only available for Android in the `op_set_option()` function?

This is to avoid dead code in non-Android version of the lib.

libusb/os/linux_usbfs.c
```diff
@@ -381,6 +384,10 @@ static int op_init(struct libusb_context *ctx)
 		}
 	}
 
+	if (weak_authority) {
+		return LIBUSB_SUCCESS;
+	}
+
```
LudovicRousseau, line 390: Same remark about `#ifdef __ANDROID__` here.

libusb/os/linux_usbfs.c
dickens: Agreed

libusb/os/linux_usbfs.c
dickens: Agreed

libusb/os/linux_usbfs.c
```diff
@@ -413,6 +420,23 @@ static void op_exit(struct libusb_context *ctx)
 	usbi_mutex_static_unlock(&linux_hotplug_startstop_lock);
 }
 
+static int op_set_option(struct libusb_context *ctx, enum libusb_option option, va_list ap)
+{
+	UNUSED(ctx);
+	UNUSED(ap);
+
+	switch ((int)option) {
```
dickens, line 428: Why is there a cast here?  Is it to avoid a compiler warning, or is it copied from windows_common.c?

chrisju: Copied from op_exit to avoid a compiler warning.

hjelmn: What is the warning?

chrisju: jni/../../libusb/os/linux_usbfs.c:427:49: warning: unused parameter 'ctx' [-Wunused-parameter]
static int op_set_option(struct libusb_context *ctx, enum libusb_option option, va_list ap)
                                                ^
jni/../../libusb/os/linux_usbfs.c:427:89: warning: unused parameter 'ap' [-Wunused-parameter]
static int op_set_option(struct libusb_context *ctx, enum libusb_option option, va_list ap)

LudovicRousseau: I think the question was about the cast in `(int)option`

hjelmn: Exactly. Casting to an int for the switch is odd.

libusb/os/linux_usbfs.c
```diff
@@ -95,6 +95,11 @@ static int sysfs_available = -1;
 /* how many times have we initted (and not exited) ? */
 static int init_count = 0;
 
+#ifdef __ANDROID__
+/* is no authority to operate usb device directly */
```
tmm1, line 99:
```suggestion
/* have no authority to operate usb device directly */
```

Mathew44:
I am using Android 9. I would like to use bulk transfer but I am getting error.
reap_for_handle returns errno 13 and after that returns errno 11.
Also usbfs_get_active_config returns 13 and initialize_handle returns 13.

Code:
static struct libusb_device_handle *handle = NULL;

static void print_devs(jint filedescr, jint intEnd)
{
    int rc = 0 ;
    libusb_context *context = NULL;
    libusb_device *device;
    libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY);
    libusb_init(&context);
    libusb_wrap_sys_device(context, (intptr_t) filedescr, &handle); 
    device = libusb_get_device(handle);
    unsigned char buffer2[64];
    int size;
    rc = libusb_bulk_transfer(handle, intEnd, (unsigned char*)&buffer2, 0x40, &size, 1000);
    if (rc < 0)
    {
        LOGD("bulk not ok");
    }
}

Mathew44:
Hello. I am using Android 9.
I would like to run bulk transfer or isochronous transfer using asynchronous approach from NDK.
How we can check in Java code using callback from C plus plus file when data has been received?
Could you give me an example of code, please?

libusb/os/linux_usbfs.c
chrisju: Sorry i misunderstood. Thank LudovicRousseau. The cast `(int)option` is just copied from windows_common.c.

libusb/os/linux_usbfs.c
chrisju: Fixed now. Thank you.

libusb/os/linux_usbfs.c
chrisju: I have removed the cast in commit `86997c8`.

PR #762, opened 2020-08-02
API docs: Elaborate on "NOT_SUPPORTED" status for set_configuration()

thierer:
Add remarks to the api docs for `libusb_set_configuration()` regarding the `LIBUSB_ERROR_NOT_SUPPORTED` return code now that @mcuee [kindly pointed me in the right direction](https://github.com/libusb/libusb/pull/752#issuecomment-657457136).

1. Explicitly list `LIBUSB_ERROR_NOT_SUPPORTED` as a possible return code so callers should consider handling it.
2. Point out that this probably isn't a serious error.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1603,6 +1603,11 @@ int API_EXPORTED libusb_get_configuration(libusb_device_handle *dev_handle,
  * causing most USB-related device state to be reset (altsetting reset to zero,
  * endpoint halts cleared, toggles reset).
  *
+ * Not all backends support setting the configuration from user space, which
+ * will be indicated by the return code LIBUSB_ERROR_NOT_SUPPORTED. As this
+ * suggests that the platform is handling the device configuration itself,
+ * this error should generally be safe to ignore.
+ *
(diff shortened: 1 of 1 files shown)


libusb/core.c
```diff
@@ -1603,6 +1603,11 @@ int API_EXPORTED libusb_get_configuration(libusb_device_handle *dev_handle,
  * causing most USB-related device state to be reset (altsetting reset to zero,
  * endpoint halts cleared, toggles reset).
  *
+ * Not all backends support setting the configuration from user space, which
+ * will be indicated by the return code LIBUSB_ERROR_NOT_SUPPORTED. As this
+ * suggests that the platform is handling the device configuration itself,
+ * this error should generally be save to ignore.
```
mcuee, line 1609: Spelling: "save" should be "safe".

thierer: Oops, thanks for the heads up - fixed!

mcuee: Thanks for the updates. Now it looks okay to me.

PR #763, opened 2020-08-05
OpenBSD: allow opening ugen devices multiple times

botovq:
Fix an OpenBSD backend bug where an existing open file descriptor is
overwritten if a libusb user attempts to open the same ugen(4) device
multiple times. This was observed with sane-backends and broke scanning.

Fix from stsp@openbsd.org

### DIFF
--- a/libusb/os/openbsd_usb.c
+++ b/libusb/os/openbsd_usb.c
@@ -225,15 +225,17 @@ obsd_open(struct libusb_device_handle *handle)
 	char devnode[16];
 
 	if (dpriv->devname) {
+		int fd;
 		/*
 		 * Only open ugen(4) attached devices read-write, all
 		 * read-only operations are done through the bus node.
 		 */
 		snprintf(devnode, sizeof(devnode), DEVPATH "%s.00",
 		    dpriv->devname);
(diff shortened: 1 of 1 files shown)


botovq:
More detailed description of the problem from stsp@ on the [OpenBSD ports mailing list][1]:

> This patch fixes my USB scanner. Tested with 'scanimage' from sane-backends.
>
> While Sane searches for scanners to use it loops over multiple backends.
> Each backend attempts to open a ugen(4) device node. An open file descriptor
> for e.g. /dev/ugen0.00 is stored in the libusb handle's OS-specific part,
> called dpriv->fd in our case (see patch below).
> 
> Once a backend succeeds in attaching a scanner, other backends may still
> probe the USB bus for additional scanners. This results in more attempts
> to open the same device node, all of which fail with EBUSY. When this
> happens, the current code will overwrite dpriv->fd with -1.
>
> Once Sane is done with identifying scanners, it will ask the backend which
> opened the device to scan an image. Access to the device node now fails
> with EBADF because libusb is passing -1 to the kernel instead of the open
> file descriptor for the device node.

[1]: https://marc.info/?l=openbsd-ports&m=159657201510355&w=2

PR #768, opened 2020-08-17
Memory leaks

klebertarcisio:
Fix memory leaks

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -2620,7 +2620,7 @@ int usbi_add_event_source(struct libusb_context *ctx, usbi_os_handle_t os_handle
 	if (ctx->fd_added_cb)
 		ctx->fd_added_cb(os_handle, poll_events, ctx->fd_cb_user_data);
 #endif
-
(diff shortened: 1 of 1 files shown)


dickens:
I do not think you understand the design and use of `ievent_source` here.  There is no memory leak.  Your change introduces a serious bug.

LudovicRousseau:
I think the problem of understanding is more with
```
list_add_tail(&ievent_source->list, &ctx->event_sources);
```

@klebertarcisio why do you think this is a memory leak? Is it reported by an automatic tool? or it is just a manual code review?

klebertarcisio:
Hi @LudovicRousseau and @dickens. I'm working on a technique to detect bugs in configurable systems. LibUSB is a great candidate because uses a lot of #ifdefs.
@LudovicRousseau, I used a tool. However, this tool only check if the function has **malloc/calloc** and **free**. As you can see, **usbi_add_event_source** has malloc and hasnt free.

I have more reports if you dont mind :)

Issue #791, opened 2020-09-30
WinUSB: Functionality broken if there is a "Device Descriptor Request Failed" device in the system


bohdan-tymkiv:
The 'libusb_get_device_list' stops enumerating devices and returns -1 if such device is present in the system.
![image](https://user-images.githubusercontent.com/10281082/94716735-f5742600-0357-11eb-9658-f519f629380c.png)

I've narrowed down the issue to commit https://github.com/libusb/libusb/commit/d21956dc

Simple change of this line 
https://github.com/libusb/libusb/blob/d21956dc3357bb40cde6d47eaf2497caf8d0a2de/libusb/os/windows_winusb.c#L844
to 
```
if (conn_info.ConnectionStatus == NoDeviceConnected || \
    conn_info.DeviceDescriptor.bNumConfigurations == 0) {
```
when applied on top of https://github.com/libusb/libusb/commit/d21956dc makes the issue go away.

However, such change leads to Segmentation fault when applied to current HEAD (https://github.com/libusb/libusb/commit/eee6998395184d87bd8e9c07ce2637caed1207f4).
I suspect that all special handling of LIBUSB_ERROR_NO_DEVICE is broken in current HEAD.

---- edit ---- 
```
> systeminfo | findstr /B /C:"OS Name" /C:"OS Version"
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.16299 N/A Build 16299

$ uname -a
MINGW64_NT-10.0-16299 host-name-removed 3.1.7-340.x86_64 2020-09-22 20:54 UTC x86_64 Msys
```

VitaSokhatska:
any updates here? We'd like to understand whether it is going to be addressed somehow in the nearest releases

Youw:
Confirm: seeing same issue on Windows 10.0.19041.572.
The patch suggested above does work for me too.

Youw:
> However, such change leads to Segmentation fault when applied to current HEAD (eee6998).

Don't see any issues with c3deb6dd6440fd467ce12965742f2d7a228b3bbe.

PR #806, opened 2020-11-02
Windows: ignore devices with broken descriptor during enumeration

Youw:
Fixes: https://github.com/libusb/libusb/issues/791

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -762,20 +762,31 @@ void usbi_disconnect_device(struct libusb_device *dev)
  * to the discovered device list. */
 int usbi_sanitize_device(struct libusb_device *dev)
 {
+    return usbi_sanitize_device_descriptor(DEVICE_CTX(dev), &dev->device_descriptor);
+}
+
+/* Perform sanity checks on a discovered device descriptor. If this
+ * function fails (negative return code), the device with this descriptor should not be added
+ * to the discovered device list. */
+int usbi_sanitize_device_descriptor(libusb_context *ctx, struct libusb_device_descriptor *device_descriptor)
+{
 	uint8_t num_configurations;
 
-	if (dev->device_descriptor.bLength != LIBUSB_DT_DEVICE_SIZE ||
-	    dev->device_descriptor.bDescriptorType != LIBUSB_DT_DEVICE) {
-		usbi_err(DEVICE_CTX(dev), "invalid device descriptor");
+	if (device_descriptor->bLength != LIBUSB_DT_DEVICE_SIZE ||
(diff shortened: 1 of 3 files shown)


Youw:
CI is failing for the same reason as on master. Not related to this change.

dickens (changes requested):
This change doesn't make sense to me.  There is no difference between `usbi_sanitize_device()` and `usbi_sanitize_device_descriptor()` and hence no reason to have a new function.

If you want to add a workaround in the Windows backend, then just check the device descriptor before the `memcpy()`.  You probably just need to validate the `bLength` and `bDescriptorType` fields and do your delay if they are not valid.  There is no reason for any changes to the libusb core for this.

Youw:
>  then just check the device descriptor

That makes code/logic duplication, if we're OK with that.

---
My motivation for the change:
The documentation for `usbi_sanitize_device` states that it checks whether the device is valid (a sane assumption: the device has to be fully constructed at that point), and if it is not - it has to be ignored during enumeration.
(The fact that so far each backend fails the entire enumeration if there is a single "broken" device - is a separate discussion. I do believe this has to be applied to all backends.)

My first suggestion for the fix was to return `LIBUSB_ERROR_NO_DEVICE` when `usbi_sanitize_device` fails in the original implementation here: https://github.com/libusb/libusb/blob/a016a082a8143409006d954112b5130bc7e9449c/libusb/os/windows_winusb.c#L1173
but it prints multiple log messages, whereas most of them describe only symptoms of the broken device, not the root reason (broken descriptor).

Simply reuse `usbi_sanitize_device` to check only a descriptor - doesn't seem a good option, because:
- the API/documentation for that function doesn't guarantee that it would only check for the descriptor, and that not fully constructed devices can be passed; (relying on the implementation, rather than on API/documentation is a source of future bugs);

> There is no difference between usbi_sanitize_device() and usbi_sanitize_device_descriptor() and hence no reason to have a new function
- the fact that `usbi_sanitize_device` checks only for a valid descriptor (yet) doesn't mean that it won't change in the future, and not reflected in its API/documentation;

Explicitly separating device check and descriptor check seem like a natural decision (keeps code consistency/gives the ability to reuse new function for other backends).

A cosmetic API addition to `usbi_sanitize_device_descriptor` of an optional context (as compared to `usbi_sanitize_device`) makes it possible to get a single log entry in case of broken device, instead of 20 of them.

---
If my motivation is not good-enough, I'll be happy to update this PR, to check for valid descriptor manually for Windows backend only.

Please confirm.

dickens:
Yes, please take care of this in the Windows backend and leave the core alone.  I'm not worried about code duplication when checking two fields of a structure.

PR #813, opened 2020-11-30
fix eol of msvc/*.vcxproj and msvc/*.vcxproj.filters files

kucharskim:
With `msvc/.gitattributes` in place, git automatically modifies `msvc/*.vcxproj` and `msvc/*.vcxproj.filters` files upon checkout. I want to get rid of those unrelated modifications.


pbatard:
You may also want to add the following to `.gitattributes` to avoid these kind of problems, since whatever consumes MSVC solution files will expect/handle CRLF:
```
*.sln           eol=crlf
*.vcxproj*      eol=crlf
```

dickens:
> You may also want to add the following to `.gitattributes` to avoid these kind of problems, since whatever consumes MSVC solution files will expect/handle CRLF:
> 
> ```
> *.sln           eol=crlf
> *.vcxproj*      eol=crlf
> ```

Yes this is already taken care of [here](https://github.com/libusb/libusb/blob/master/msvc/.gitattributes).  For some reason Git committed the files with CRLF line endings instead of plain LF line endings, hence this mess.

PR #821, opened 2020-12-10
autotools: enable installation of fxload tool

mriesch-wv:
The libusb examples contain a hidden gem, namely the fxload tool. But it seems one cannot install this tool, even with example build enabled. This would be beneficial, though, especially in packaging environments such as conda. Is there any particular reason *not* to deploy the tools?

This PR activates the installation of the fxload tool and serves as basis for discussion.
(Disclaimer: I am not used to autotools, but the change had the desired effect)

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -2,7 +2,9 @@ AM_CPPFLAGS = -I$(top_srcdir)/libusb
 LDADD = ../libusb/libusb-1.0.la
 LIBS =
 
-noinst_PROGRAMS = dpfp dpfp_threaded fxload hotplugtest listdevs sam3u_benchmark testlibusb xusb
+noinst_PROGRAMS = dpfp dpfp_threaded hotplugtest listdevs sam3u_benchmark testlibusb xusb
+
+bin_PROGRAMS = fxload
 
 dpfp_threaded_CPPFLAGS = $(AM_CPPFLAGS) -DDPFP_THREADED
 dpfp_threaded_CFLAGS = $(AM_CFLAGS) $(THREAD_CFLAGS)


mcuee:
I do not think this is a good idea. To libusb project, fxload is an example.

fxload is available in Linux distributions, probably an old version.
https://packages.ubuntu.com/groovy/fxload

Of course fxload maintainer can take the improvement from libusb.

mriesch-wv:
@mcuee Thanks for the prompt response. I didn't realize that this tool originated outside of this project. 

Now I wonder where it is developed these days, https://github.com/esden/fxload seems to be a promising candidate. Who is the current maintainer?

mcuee:
Sorry but I do not know who is the maintainer. But you can always talk to the distro maintainers (Fedora, Debian, Ubuntu, Arch, Gentoo, etc)

mriesch-wv:
OK, after some research it seems that the distros base on versions from around 2008-2010, when fxload was part of linux-hotplug. Conda packages base on https://github.com/mithro/fxload, which is a fork of https://github.com/esden/fxload without any further changes. The GitHub repos I found so far were updated in 2017 for the last time and do not support the FX3 model, which the libusb example version does. Hence, libusb features the most recent version I found so far. Glad for any pointers, though!

mcuee:
If you are keen, just fork the libusb fxload example to a new fxload project and then you can become the new maintainer of fxload. Then the distro can distribute it.

mcuee:
On the other hand, it other maintainers are okay with this idea, I will have no objection as well. 

It is just libusb is already short of developers and fxload is just a example of the libusb project, improved by Pete and federico-m back in 2012/2013 (including the FX3 support). I think it is still missing the EEPROM support feature.

mcuee:
I have just create an issue here. I am not so sure if this will trigger any action or not.
https://github.com/esden/fxload/issues/2

mriesch-wv:
@mcuee Thanks once more for your responses. I must say that I am not too keen on creating a new fork, as there are already several projects that do the same thing (more or less). Also, the libusb example does exactly what I need: loading a given firmware image into the RAM of a FX3 device. So I patched my local version and it works perfectly fine for now.

That said, I'll leave the PR open and will respond to any input on the subject. While I agree that the examples should not be installed by default, maybe providing the installation as optional feature could be a possibility.

LudovicRousseau:
Examples are not build by default. You need to configure with `--enable-examples-build`.
So I guess that examples will also _not_ be installed if they are not build.

I don't know if GNU/Linux distributions build libusb with `--enable-examples-build`. I would say no.

On he other hand, I don't know if it is a good idea to install fxload. I guess very few people will use it (I don't need it). The binary is still present in the examples/ directory if needed.

mriesch-wv:
@LudovicRousseau It absolutely makes sense that the examples are not built by default. But there is this nice `--enable-examples-build` switch, that activates the build if desired.

Of course, if the examples are not built, they cannot be installed. However, if they are built, there is still no possibility to install them. Therefore, if (granted, that is a big if) any distribution would want to provide `fxload` as package (separate package or together with libusb) they would have to patch the build system or craft another solution.

And I agree that most likely there are only few users of `fxload`, but still one can give them a chance as long as it doesn't break other things...

dickens:
In my opinion, libusb should not support installing any examples or tests.

mcuee:
@mriesch-wv 
https://github.com/esden/fxload/issues/2

It will be great that you create a pull request to the above fxload fork and esden is willing to review and integrate. Then distros can pick it up. That is probably the best solution.

mcuee:
Close this for now.

mriesch-wv:
OK, thank you all for your feedback. Given the history of the tool, I agree that reviving the standalone project in https://github.com/esden/fxload is the cleanest solution. I cannot guarantee anything but maybe I'll have a shot at reconciling the two versions.

Issue #717, opened 2020-04-19
Matching from android.hardware.usb.UsbDevice to a libusb_device


DiegoBM:
Hi everyone,

I'm trying to port a library that internally uses libusb from Linux to Android, the library managed to build apparently fine after previous build of all its dependencies, and while I have success obtaining permission for the USB device in my android application. Calling the library endpoints simply won't do anything. Then I came across this paragraph in libusb:

> The difficulties in this method is that there is no guarantee
> that it will continue to work in the future Android versions, it
> requires invoking Java APIs and running code to match each
> android.hardware.usb.UsbDevice to a libusb_device.

Could someone please describe how is that match done? I've searched around and for the love of Yoda that I cannot find anything useful or anyone describing such kind of example in some sort of detail, or some example of a working usage of the same case stated above.

Thank you in advance

XanaduAPI:
首先，你需要 用  Android Studio 构建一个 Android 项目
然后，你需要申请此App 运行时的权限

![image](https://user-images.githubusercontent.com/45027459/101230427-34b84d80-36e0-11eb-8863-17109173a4b9.png)
![image](https://user-images.githubusercontent.com/45027459/101230435-400b7900-36e0-11eb-839b-7eb486dad26f.png)

在主窗口的 OnCreate 中注册回调函数
![image](https://user-images.githubusercontent.com/45027459/101230455-5a455700-36e0-11eb-8585-08013f81a5d0.png)

在回调函数中 申请 USB 连接权限
![image](https://user-images.githubusercontent.com/45027459/101230483-7c3ed980-36e0-11eb-9f27-2411f8e8c174.png)

然后打开设备

	private void DeviceConnect(UsbDevice vDevice)
	{
		UsbManager			vManager = (UsbManager) getSystemService(Context.USB_SERVICE);
		//vManager.requestPermission(vDevice, PendingIntent.getBroadcast(MainActivity.this, 0, new Intent(ACTION_USB_PERMISSION), 0));

		UsbEndpoint			vPointRead = null;
		UsbEndpoint			vPointWrite = null;
		int				vInterfaceNumber = 0;

		if (false == vManager.hasPermission(vDevice))
		{
			vManager.requestPermission(vDevice, PendingIntent.getBroadcast(MainActivity.this, 0, new Intent(ACTION_USB_PERMISSION), 0));
		}
		INFO("USB 权限 : " + (vManager.hasPermission(vDevice) ? "有" : "无"));

		INFO("打开设备");
[...]

PR #830, opened 2020-12-21
Add unrooted Android example

Peter-St:
Adding another approach for connecting UsbDevices on Android.


Peter-St:
Using LibUsb on non rooted Android devices is explained in this PullRequest.
Should there eventually added a seperate C sample in the example folder for non rooted devices too?

mcuee:
If you can, please have another pull request to add the example for non root device. Thanks.

Peter-St:
So this is my Android Sample for Libusb.

The Sample shows how to create an Shared Object with the NDK which can be accessed over JNI or JNA from Java or Kotlin. --> with Granted Permissions on Java/Kotlin and obtained FileDescriptor.

The approach of including the full source of LibUsb to an Android Studio project is quite complicated. It is only possible over NDK for now I think. (eventually it's enough for only applying the precombiled LibUsb lib to an android device and accessing it over the unrooted sample ...)

avantgardnerio (approved):

avantgardnerio:
I only skimmed it, but it looks to be correct. I haven't been using libusb with android directly, but rather with the Rust FFI wrapper. I made a sample project that definitely worked and ran in CI [here](https://github.com/a1ien/rusb/pull/39), in case that is useful to compare.

dickens (changes requested):
The new file `examples/unRootedAndroid.c` is specific to Android.  As such I'd prefer it to live in the `android` directory.  However, since it is just a modified copy of `examples/testlibusb.c` then you can use preprocessor macros to build the same file for both.

Please also note the naming conventions used in the project.  We do not use Camel Case so please find a more suitable name for the file.

xloem:
I'm working on porting a library to android.  Here are my notes working off of this readme; I'll edit as I bump into things:
- It should be documented that `libusb_set_option(ctx_ptr, LIBUSB_OPTION_WEAK_AUTHORITY,0)` never actually uses ctx_ptr, so the user understands why it can be used before `libusb_init` and so that future developers defend the use-case.

Peter-St:
> The new file `examples/unRootedAndroid.c` is specific to Android. As such I'd prefer it to live in the `android` directory. However, since it is just a modified copy of `examples/testlibusb.c` then you can use preprocessor macros to build the same file for both.
> 
> Please also note the naming conventions used in the project. We do not use Camel Case so please find a more suitable name for the file.

Adding preprocessor macros for Android in the original examples files would effect then the functionality for rooted devices, because here we have a different approach of initializing LibUsb.
To guarantee booth ways to work, we need different files I think.

Peter

dickens (changes requested):

suriyo-cycron (approved):
😎😎😎😎

android/examples/unrooted_android.c
```diff
@@ -0,0 +1,300 @@
...
+ * LibUsb can only be included to Android Projects Source over NDK up to now. (CMake is not supported at the moment)
+ * (Android Studio)
+ * Clone the last LibUsb GitHub Repo to your Android Project and include the Android.mk file to your build.gradle.
+ */
+
+/*
+ Example JNA Approach:
+    public interface unrooted_sample extends Library {
+        public static final unrooted_sample INSTANCE = Native.load("unrooted_android", unrooted_sample.class);
+        public int main (int fileDescriptor);
```
ysard, line 46: Hi, this function should be renamed into unrooted_usb_description since e84f30f
This is a confusion mentioned here https://github.com/libusb/libusb/issues/717#issuecomment-753718501 

And It's not consistent with android/README of this PR https://github.com/libusb/libusb/pull/830/files#diff-16202c3941de7d6c550a2d075887e6c29f5ea083da72f53afd72ffc540687b95R71 

By the way, these examples deserve to be more complete rather than hints.
Maybe a link to https://github.com/Peter-St/unRootedSample (updated with the same names) ?

The same goes for #874 , without proper/full example it is difficult to test the code easily especially when you are not familiar with Android dev ;)

android/examples/unrooted_android.c
```diff
@@ -0,0 +1,300 @@
...
+
+
+// fileDescriptor = is the native File Descriptor obtained in Java and transfered to native over JNA for Example.
+int unrooted_usb_description(int fileDescriptor)
+{
+    libusb_context *ctx;
+    libusb_device_handle *devh = NULL;
+    int r = 0;
+    verbose = 1;
+    r = libusb_set_option(&ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);
```
ysard, line 279: The signature of `libusb_set_option` in libusb.h is:

`int LIBUSB_CALL libusb_set_option(libusb_context *ctx, enum libusb_option option, ...);`

Shouldn't we instead send ctx which is a pointer and not its address?
This code does not compile :o

android/examples/unrooted_android.c
```diff
@@ -0,0 +1,300 @@
...
+    if (r != LIBUSB_SUCCESS) {
+        LOGD("libusb_init failed: %d\n", r);
+        return -1;
+    }
+    r = libusb_init(&ctx);
+    if (r < 0) {
+        LOGD("libusb_init failed: %d\n", r);
+        return r;
+    }
+    r = libusb_wrap_sys_device(NULL, (intptr_t)fileDescriptor, &devh);
```
ysard, line 289: Using NULL in place of the context crashes my example based on unrooted_android.c of this PR.

This line is fine...
`r = libusb_wrap_sys_device(ctx, (intptr_t)fileDescriptor, &devh);`


Code used :
```
int unrooted_usb_description(int fileDescriptor) {
    libusb_context *ctx = NULL;
    libusb_device_handle *devh = NULL;
    int r = 0;
    LOGD("Hello here !!");
    r = libusb_set_option(ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);
    if (r != LIBUSB_SUCCESS) {
        LOGD("libusb_set_option failed: %d\n", r);
        return -1;
    }
    LOGD("libusb_set_option OK: %d\n", r);
    r = libusb_init(&ctx);
    if (r < 0) {
        LOGD("libusb_init failed: %d\n", r);
        return r;
    }
    LOGD("libusb_init OK: %d\n", r);
    r = libusb_wrap_sys_device(ctx, (intptr_t)fileDescriptor, &devh);
    if (r < 0) {
        LOGD("libusb_wrap_sys_device failed: %d\n", r);
        return r;
    } else if (devh == NULL) {
        LOGD("libusb_wrap_sys_device returned invalid handle\n");
        return r;
    }
    LOGD("libusb_wrap_sys_device OK: %d\n", r);
    ...
```

Here is the logcat:
```
27646 27646 I Libusb_Sample: USB test permission: "/dev/bus/usb/001/002": true
27646 27646 I Libusb_Sample: USB file descriptor: "/dev/bus/usb/001/002": 57
27646 27646 D LibUsb  : Hello here !!
27646 27646 D LibUsb  : libusb_set_option OK: 0
27646 27646 D LibUsb  : libusb_init OK: 0
27646 27646 W MY.APP: type=1400 audit(0.0:9118): avc: denied { read } for name="usb" dev="tmpfs" ino=122799 scontext=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:usb_device:s0 tclass=dir permissive=0
27646 27646 W MY.APP: type=1400 audit(0.0:9119): avc: denied { read } for name="/" dev="tmpfs" ino=13042 scontext=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:device:s0 tclass=dir permissive=0
27646 27646 F libc    : Fatal signal 11 (SIGSEGV), code 1, fault addr 0x1a8 in tid 27646 (MY.APP)
  441   441 W         : debuggerd: handling request: pid=27646 uid=10143 gid=10143 tid=27646
27905 27905 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
27905 27905 F DEBUG   : Build fingerprint: 'Xiaomi/santoni/santoni:7.1.2/N2G47H/V11.0.2.0.NAMMIXM:user/release-keys'
27905 27905 F DEBUG   : Revision: '0'
27905 27905 F DEBUG   : ABI: 'arm64'
27905 27905 F DEBUG   : pid: 27646, tid: 27646, name: MY.APP  >>> com.MY.APP <<<
27905 27905 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1a8
27905 27905 F DEBUG   :     x0   00000000000001a8  x1   0000000000000004  x2   0000007f80166800  x3   0000007f801677d4
27905 27905 F DEBUG   :     x4   0000007ff1fb2760  x5   0000000000000004  x6   0000007f6a0a09ca  x7   10c4400000000110
27905 27905 F DEBUG   :     x8   0000000000000000  x9   0000007f6a112780  x10  69440961db40bea4  x11  00000000fffffff4
27905 27905 F DEBUG   :     x12  00000000fffffff4  x13  0000007f80167340  x14  0000000000000041  x15  000000000000004c
27905 27905 F DEBUG   :     x16  0000007f80208038  x17  0000007f8b150b3c  x18  0000000000000000  x19  0000007f6a8dce38
27905 27905 F DEBUG   :     x20  0000007f6a8dcd30  x21  0000007ff1fb3568  x22  0000000000000000  x23  0000000000000008
27905 27905 F DEBUG   :     x24  0000007ff1fb2d38  x25  0000000000000001  x26  000000000000001e  x27  0000007ff1fb2ad0
27905 27905 F DEBUG   :     x28  0000007ff1fb2a10  x29  0000007ff1fb27c0  x30  0000007f80191aa4
27905 27905 F DEBUG   :     sp   0000007ff1fb27b0  pc   0000007f8b150b3c  pstate 0000000060000000
27905 27905 F DEBUG   : 
27905 27905 F DEBUG   : backtrace:
27905 27905 F DEBUG   :     #00 pc 0000000000068b3c  /system/lib64/libc.so (pthread_mutex_lock)
27905 27905 F DEBUG   :     #01 pc 0000000000071aa0  /data/app/com.MY.APP-2/lib/arm64/libnative_wrapper.so
27905 27905 F DEBUG   :     #02 pc 000000000007134c  /data/app/com.MY.APP-2/lib/arm64/libnative_wrapper.so
27905 27905 F DEBUG   :     #03 pc 00000000000798c8  /data/app/com.MY.APP-2/lib/arm64/libnative_wrapper.so
27905 27905 F DEBUG   :     #04 pc 000000000007648c  /data/app/com.MY.APP-2/lib/arm64/libnative_wrapper.so
27905 27905 F DEBUG   :     #05 pc 000000000006b5c8  /data/app/com.MY.APP-2/lib/arm64/libnative_wrapper.so (libusb_wrap_sys_device+184)
27905 27905 F DEBUG   :     #06 pc 00000000000676a8  /data/app/com.MY.APP-2/lib/arm64/libnative_wrapper.so (unrooted_usb_description+320)
27905 27905 F DEBUG   :     #07 pc 0000000000010154  /data/app/com.MY.APP-2/lib/arm64/libjnidispatch.so
27905 27905 F DEBUG   :     #08 pc 000000000000f938  /data/app/com.MY.APP-2/lib/arm64/libjnidispatch.so
27905 27905 F DEBUG   :     #09 pc 0000000000005cfc  /data/app/com.MY.APP-2/lib/arm64/libjnidispatch.so
27905 27905 F DEBUG   :     #10 pc 0000000000007b68  /data/app/com.MY.APP-2/lib/arm64/libjnidispatch.so (Java_com_sun_jna_Native_invokeInt+32)
27905 27905 F DEBUG   :     #11 pc 000000000056db04  /data/app/com.MY.APP-2/oat/arm64/base.odex (offset 0x512000)
```

Note that the 2 lines with `avc: denied` (SE Linux ?) are also present when It's working with the ctx given to the function.

Maybe this will help also for #874

android/examples/unrooted_android.c
Peter-St: right, we have to use the Pointer and not it's address.

android/examples/unrooted_android.c
Peter-St: right, I was too fast with my commits and forgot to rename this function.

About the links to complete repos outside of the libusb site I agree with you, that this would be easier for unfamiliar people, but up to now, no links for the other system platforms were given on this site. 
My unRootedSample project was a guidance for a single person. It was not designed for a general assistance scheme.

android/examples/unrooted_android.c
Peter-St: Ok, passing the context in this function makes sense.

About #874 I think this PR will work without libusb_wrap_sys_device.
There are modified versions of libusb, which uses other approaches to get the library working on Android:
https://github.com/saki4510t/UVCCamera

mcuee:
Just wondering whether we should merge just the improved Readme for 1.0.25RC1 and leave the examples for next release.

Peter-St:
> Just wondering whether we should merge just the improved Readme for 1.0.25RC1 and leave the examples for next release.

On the one hand:
This approach for Android is working. Including it to the pre-release would be a progress for the actual readme.

On the other hand:
[#874](https://github.com/libusb/libusb/pull/874) --> If this PR is going to be merged in future releases, the readme needs to be rewritten for including #874 too. So adding the samples I would not suggest for now.

I had no time to test #874 up to now, but hope to handle that soon.

So far,

Peter

mcuee:
@Peter-St Thanks for the insight. I am for the inclusion of the updated README to help the Android users out there. We can leave out the example.

And #874 may take some time to be merged.

harbulot:
This is a comment regarding `libusb_set_option(&ctx, ...)` vs. `libusb_set_option(ctx, ...)` as mentioned in earlier comments.

```c
libusb_context *ctx;
int r = 0;
r = libusb_set_option(ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);
r = libusb_init(&ctx);
```


In the code fragment above, it's `libusb_init(&ctx)` that initialises `ctx`, since its content is undefined before that.

If we need to set `LIBUSB_OPTION_WEAK_AUTHORITY` before `libusb_init`, shouldn't `ctx` always be `NULL`? Otherwise, `ctx` could be pointing to anywhere in memory (something that's likely not to have anything to do with `libusb_context`).

----

I've also encountered another problem, simply trying this in the Android Studio emulator, based on the simplest "Native C++" project: the app crashes if `ctx` is initialised to `NULL`.


```c
#include <jni.h>
#include <string>

#include <android/log.h>
#include "libusb.h"

extern "C" JNIEXPORT jstring JNICALL
Java_com_example_myapplication_MainActivity_stringFromJNI(
        JNIEnv *env,
        jobject /* this */) {
    int r;
    libusb_context *ctx;     // = NULL;

    __android_log_print(ANDROID_LOG_DEBUG, "MyApplication", "Before lib_usb_set_option...");

    r = libusb_set_option(ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);
    __android_log_print(ANDROID_LOG_DEBUG, "MyApplication", "lib_usb_set_option: %d", r);

    r = libusb_init(&ctx);
    __android_log_print(ANDROID_LOG_DEBUG, "MyApplication", "libusb_init: %d", r);
    if (r >= 0) {
        libusb_exit(ctx);
    }

    std::string hello = "Hello from C++";
    return env->NewStringUTF(hello.c_str());
}
```

When `libusb_context *ctx;` is left uninitialised, `libusb_init` is successful:

```
D/MyApplication: Before lib_usb_set_option...
D/MyApplication: lib_usb_set_option: 0
D/MyApplication: libusb_init: 0
```

When I set `libusb_context *ctx = NULL;`, the app crashes:

```
D/MyApplication: Before lib_usb_set_option...
D/MyApplication: lib_usb_set_option: 0
A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x4 in tid 10513 (o.myapplication), pid 10513 (o.myapplication)
```

It is very similar to the example given by @ysard in https://github.com/libusb/libusb/pull/830#discussion_r645929159 , but for me it fails at `libusb_init`.

Admittedly, I have only tried this in the emulator at this stage (so no actual USB peripheral) and I haven't done anything in the `AndroidManifest.xml` to address the permissions. However, the fact it somehow passes after using `libusb_set_option` using an uninitialised context value, but fails when it had been used with a `NULL` context suggests there might be a problem.

(This is using the latest `libusb`, commit e3dda19b31ee974fd749d55bc97d8bfc0e976bb4 .)

Any thoughts on this would be appreciated (I'm guessing this might lead to a separate issue if this isn't just about the documentation).

Peter-St:
Hello harbulot,

I cannot confirm the crash on an Android 8 device. I get no crash with `libusb_context *ctx;` nor   `libusb_context *ctx = NULL;`.

I think you can use `libusb_context *ctx;` as well, because lib_usb_set_option does not touch the context and it's then be initialized in `libusb_init(&ctx);`

harbulot:
Hello Peter,

Thank you, I agree it shouldn't matter if `libusb_set_option` doesn't touch the context, but if it only works with an uninitialised value (and not NULL), it's probably a sign there's something wrong elsewhere (and that `libusb_set_option` actually does rely on something uninitialised somehow, possibly by accident).

I've now tried on physical devices. Using the latest commit on the master branch (e3dda19b31ee974fd749d55bc97d8bfc0e976bb4), it crashes on all devices: Android 8.1 device, Android 9 device and emulator with Pixel 2 API 28.

However, when I revert libusb to the latest tagged version (v1.0.24), it doesn't crash.

I've tried to narrow it down, it seems to work with commit d551802db593436947a2464a7da1020edf0e7b56, but crash with commit a524555c987fec2c16417d2f58c8713efcbe11a9, which was precisely about allowing for initialisations using NULL.

Looking a bit more closely into it, the following change prevents the crash again:

```diff
diff --git a/libusb/core.c b/libusb/core.c
index 054bc6a..d20d5e0 100644
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2206,7 +2206,7 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,

        ctx = usbi_get_context(ctx);
        if (NULL == ctx) {
-               return LIBUSB_SUCCESS;
+//             return LIBUSB_SUCCESS;
        }

        switch (option) {
```

harbulot:
It would seem it has something to do with `usbi_default_context` being NULL when I called `libusb_set_option(ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL)` (where `ctx` was NULL too).

Perhaps I misunderstood the way `LIBUSB_OPTION_WEAK_AUTHORITY` had to be set up before the context was set up.

Does there need to be a default context initialised and then the custom context?

I've tried this (now back using commit e3dda19b31ee974fd749d55bc97d8bfc0e976bb4):

```c
int r;
libusb_context *ctx = NULL;

r = libusb_init(NULL);

r = libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);

r = libusb_init(&ctx);
if (r >= 0) {
    libusb_exit(ctx);
}

libusb_exit(NULL);
```

The app now crashes at `libusb_init(NULL)`

Peter-St:
ok,
`libusb_context *ctx;` generates a compiler error so it needs to be changed to `libusb_context *ctx = NULL;` --> I will change this.

Peter-St:
> It would seem it has something to do with `usbi_default_context` being NULL when I called `libusb_set_option(ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL)` (where `ctx` was NULL too).
> 
> Perhaps I misunderstood the way `LIBUSB_OPTION_WEAK_AUTHORITY` had to be set up before the context was set up.
> 
> Does there need to be a default context initialised and then the custom context?
> 
> I've tried this (now back using commit [e3dda19](https://github.com/libusb/libusb/commit/e3dda19b31ee974fd749d55bc97d8bfc0e976bb4)):
> 
> ```c
> int r;
> libusb_context *ctx = NULL;
> 
> r = libusb_init(NULL);
> 
> r = libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);
> 
> r = libusb_init(&ctx);
> if (r >= 0) {
>     libusb_exit(ctx);
> }
> 
> libusb_exit(NULL);
> ```
> 
> The app now crashes at `libusb_init(NULL)`

You have to run libusb_set_option before you call libusb_init.

harbulot:
Sorry, I was indeed going the wrong way trying to initialise a default context in my previous comment.

As you say (and it's also in the documentation indeed) `LIBUSB_OPTION_WEAK_AUTHORITY` should be set before calling `libusb_init`.

If my understanding is correct, calling `libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY)` should set `default_context_options[LIBUSB_OPTION_WEAK_AUTHORITY]` at [line 2200 of `core.c`](https://github.com/libusb/libusb/blob/a524555c98/libusb/core.c#L2200), but then return at line 2209 (in the absence of an initialised default context).

However, when calling `libusb_init` later , `default_context_options` [only seems to be used when used with a `NULL` context](https://github.com/libusb/libusb/blob/a524555c98/libusb/core.c#L2340):

```c
	if (!ctx) {
		for (enum libusb_option option = 0 ; option < LIBUSB_OPTION_MAX ; option++) {
			if (LIBUSB_OPTION_LOG_LEVEL == option || !default_context_options[option].is_set) {
				continue;
			}
			r = libusb_set_option(_ctx, option);
			if (LIBUSB_SUCCESS != r)
				goto err_io_exit;
		}
	} else
		*ctx = _ctx;
```

If we do pass a pointer (to a pointer) to a context, like so `libusb_init(&ctx)`, then `ctx` in `libusb_init` is not `NULL`, so `default_context_options[LIBUSB_OPTION_WEAK_AUTHORITY]` that was set before doesn't seem actually used at all, is it?

Peter-St:
> If we do pass a pointer (to a pointer) to a context, like so `libusb_init(&ctx)`, then `ctx` in `libusb_init` is not `NULL`, so `default_context_options[LIBUSB_OPTION_WEAK_AUTHORITY]` that was set before doesn't seem actually used at all, is it?

phew, for my understanding libusb_set_option to LIBUSB_OPTION_WEAK_AUTHORITY skipps the device enumeration. I must admit, I don't know if there were default contexts used or not.

harbulot:
I think the following patch would make `LIBUSB_OPTION_WEAK_AUTHORITY` have an effect, and it also fixes the app crash problems I was mentioning at first:

```diff
diff --git a/libusb/core.c b/libusb/core.c
index 054bc6a..90a1339 100644
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2205,14 +2205,11 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
        }

        ctx = usbi_get_context(ctx);
-       if (NULL == ctx) {
-               return LIBUSB_SUCCESS;
-       }

        switch (option) {
        case LIBUSB_OPTION_LOG_LEVEL:
 #if defined(ENABLE_LOGGING) && !defined(ENABLE_DEBUG_LOGGING)
-               if (!ctx->debug_fixed)
+               if (ctx && !ctx->debug_fixed)
                        ctx->debug = (enum libusb_log_level)arg;
 #endif
                break;
@@ -2230,6 +2227,10 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
        default:
                return LIBUSB_ERROR_INVALID_PARAM;
        }
+
+       if (NULL == ctx) {
+               return LIBUSB_SUCCESS;
+       }

        return LIBUSB_SUCCESS;;
 }
```

The purpose of that patch is to call `usbi_backend.set_option(...)` (which is what actually uses `LIBUSB_OPTION_WEAK_AUTHORITY` in `linux_usbfs.c`) when there is no default context initialised.

Please let me know if you'd like me to create a PR for this.

android/examples/unrooted_android.c
```diff
@@ -0,0 +1,298 @@
...
+        public static final unRootedSample INSTANCE = Native.load("unRootedAndroid", unRootedSample.class);
+        public int main (int fileDescriptor);
+    }
+    unRootedSample.INSTANCE.main( usbDeviceConnection.getFileDescriptor());
+ */
+
+#include <jni.h>
+#include <string.h>
+#include "libusb.h"
+#ifdef __ANDROID__
```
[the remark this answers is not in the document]
Peter-St: right

Peter-St:
> Please let me know if you'd like me to create a PR for this.

ok, try to create the PR. Eventually add `#ifdef __ANDROID__` to not influence the regular way of other platforms.

tormodvolden:
I extracted the README change here and fixed up formatting, in #996, so that that part can be merged while we wait for the example to be finalized.

android/examples/unrooted_android.c
```diff
@@ -0,0 +1,300 @@
...
+
+
+// fileDescriptor = is the native File Descriptor obtained in Java and transfered to native over JNA for Example.
+int unrooted_usb_description(int fileDescriptor)
+{
+    libusb_context *ctx = NULL;
+    libusb_device_handle *devh = NULL;
+    int r = 0;
+    verbose = 1;
+    r = libusb_set_option(ctx, LIBUSB_OPTION_WEAK_AUTHORITY, NULL);
```
tormodvolden, line 279: This must be changed to LIBUSB_OPTION_NO_DEVICE_DISCOVERY

tormodvolden (changes requested):

Peter-St:
The PR has been updated and the libusb_set_option method has been changed.

tormodvolden (approved):

mcuee (approved):
I do not see any issues merging this.

tormodvolden:
Only that it doesn't work... due to #942. I can push this as soon as 942 has been sorted out.

mcuee:
> Only that it doesn't work... due to #942. I can push this as soon as 942 has been sorted out.

#942 seems to be okay based on my limited understanding of libusb internals. Probably we can merge that first.

tormodvolden:
Please confirm that this builds and runs fine from git master now.

Peter-St:
Hello tormodvolden,

I can confirm that the last build compiles and runs fine.
I have tested the build with following Android Studio Project: https://github.com/Peter-St/unRootedSample

Thanks, Peter

tormodvolden:
@Peter-St, thanks a lot for verifying and, of course, for contributing this example.

PR #833, opened 2020-12-30
darwin: fix const array size to work with GCC compilers

michaelld:
Tested on macOS 11.2 20D5029f with Xcode 12.3 12C33 ARM64 and x86_64 both GCC and Clang, macOS 10.14.6 18G7016 and Xcode 11.3.1 11C504 both GCC and Clang, and macOS 10.5.8 9L31a Xcode 3.1.4 9M2809 with GCC. All tested Clang from early to current supports the original code, but all tested GCC from 5 to 10 show an error with setting the array size. Because the code related to this array does bounds checking, and the array is initialized and const, the actual array size makes no difference. Removing it allows for compiling using both Clang and GCC. See also https://trac.macports.org/ticket/61868#comment:17 .

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -2148,7 +2148,7 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
   struct darwin_transfer_priv *tpriv = usbi_get_transfer_priv(itransfer);
   const unsigned char max_transfer_type = LIBUSB_TRANSFER_TYPE_BULK_STREAM;
-  const char *transfer_types[max_transfer_type + 1] = {"control", "isoc", "bulk", "interrupt", "bulk-stream"};
+  const char *transfer_types[] = {"control", "isoc", "bulk", "interrupt", "bulk-stream", NULL};
   bool is_isoc = LIBUSB_TRANSFER_TYPE_ISOCHRONOUS == transfer->type;
 
   if (transfer->type > max_transfer_type) {


mcuee:
Indeed it will product an error under gcc. Tested under Homebrew with Mac Mini M1.

```
mcuee@mcuees-Mac-mini libusb % CC=gcc-10 ./configure --enable-examples-build  --enable-tests-build

mcuee@mcuees-Mac-mini libusb % make
/Library/Developer/CommandLineTools/usr/bin/make  all-recursive
Making all in libusb
  CC       core.lo
  CC       descriptor.lo
  CC       hotplug.lo
  CC       io.lo
  CC       strerror.lo
  CC       sync.lo
  CC       os/events_posix.lo
  CC       os/threads_posix.lo
  CC       os/darwin_usb.lo
os/darwin_usb.c: In function 'darwin_handle_transfer_completion':
os/darwin_usb.c:2151:3: error: variable-sized object may not be initialized
 2151 |   const char *transfer_types[max_transfer_type + 1] = {"control", "isoc", "bulk", "interrupt", "bulk-stream"};
      |   ^~~~~
make[2]: *** [os/darwin_usb.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

libusb/os/darwin_usb.c
```diff
@@ -2148,7 +2148,7 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
   struct darwin_transfer_priv *tpriv = usbi_get_transfer_priv(itransfer);
   const unsigned char max_transfer_type = LIBUSB_TRANSFER_TYPE_BULK_STREAM;
-  const char *transfer_types[max_transfer_type + 1] = {"control", "isoc", "bulk", "interrupt", "bulk-stream"};
+  const char *transfer_types[] = {"control", "isoc", "bulk", "interrupt", "bulk-stream", 0};
```
dickens, line 2151: Please use `NULL`, not 0.  This is a pointer and using `NULL` makes that explicit to the reader.

michaelld: Done

michaelld:
I also updated the commit message to be for all GCC, not just "older"

hjelmn:
Weird. C guarantees that any element where no initializer is provided will be initialized with 0. There is no such thing as partial initialization in C. Either no initialization or full initialization. So by initializing any member of this array the remissions member will be NULL.

hjelmn:
I am ok with this change but I think gcc is handling this wrong.

hjelmn:
Since this is a problem with using the const char as the size could also just use the defined constant directly.

michaelld:
> I am ok with this change but I think gcc is handling this wrong.

Indeed! Of course this issue impact -every- GCC that supports C11, from 5 to current ... so we need to find a workaround rather than hope for a GCC fix.

michaelld:
> Since this is a problem with using the const char as the size could also just use the defined constant directly.

Yes that's true. Either way should be be to be made to work. But, as I note in the description: Because the code related to this array does bounds checking, and the array is initialized and const, the actual array size makes no difference.

michaelld:
> Weird. C guarantees that any element where no initializer is provided will be initialized with 0. There is no such thing as partial initialization in C. Either no initialization or full initialization. So by initializing any member of this array the remaining member will be NULL.

True. If you're referring to the trailing `NULL` entry ... I add that for completion since it is good programming practice to terminate arrays. That said, because the related code does bounds checking, the added `NULL` is not technically required.

michaelld:
I'm good with whatever fix is deemed "the right thing to do" here, which allows GCC to build this code on Darwin. The fix in the PR was quicks and easy, which is why I chose it.

michaelld:
Ping on this PR ... any further thoughts?

dickens:
@hjelmn are you good to merge this?  I can do it if you want.

mcuee:
@hjelmn Please help to check if we can close this.

michaelld:
Thanks!

PR #842, opened 2021-01-10
linux_usbfs: Fix exit crash on Android

### LINKED ISSUE #841: [android] crash when exiting usb backend when weak_authority=1
In android I have to set weak authority:
`libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY);`

When unplugging the USB device, the app crashes in the exit callback:
```
2021-01-10 17:01:35.918 22547-22547/[myapp here] A/libc: path/to/libusb/os/linux_usbfs.c:424: void op_exit(struct libusb_context *): assertion "init_count != 0" failed
```

We should check weak_authority and skip doing the cleanup in op_exit.

starryalley:
In commit "89b810ec Android: Add option LIBUSB_OPTION_WEAK_AUTHORITY to support used in apk", we should also check weak_authority when exiting the backend.

This fixes #841

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -415,6 +415,11 @@ static int op_init(struct libusb_context *ctx)
 static void op_exit(struct libusb_context *ctx)
 {
 	UNUSED(ctx);
+#ifdef __ANDROID__
+	if (weak_authority) {
+        return;
+    }
+#endif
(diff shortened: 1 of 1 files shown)


mcuee:
This fix seems to be correct for Android but I do not have very good understanding of the internals of libusb.

Issue #838, opened 2021-01-05
libusb_set_interface_alt_setting returns kIOUSBPipeStalled on macOS


macfreek:
Hi, thanks for making libusb available!

I'm using [dfu-util](https://sourceforge.net/p/dfu-util/) to upgrade the firmware of a development board over USB. For some reasons, that fails on macOS, with the first reports from March 2020. It was suggested that a macOS security update triggered it, but I have not seen anything more clear than that. (Upstream tickets: [dfu-util](https://sourceforge.net/p/dfu-util/tickets/86/) and [pycom](https://github.com/pycom/pycom-micropython-sigfox/issues/422).)

Tracing down this behaviour, the root cause seems that a call to [`SetAlternateInterface()`](https://github.com/libusb/libusb/blob/eee6998395184d87bd8e9c07ce2637caed1207f4/libusb/os/darwin_usb.c#L1569) in `darwin_set_interface_altsetting()` (the `libusb_set_interface_alt_setting()` for macOS) returns a `kIOUSBPipeStalled` ("pipe is stalled") error. This is repeatable behaviour.

My question: is this known or expected behaviour? And how should either libusb or dfu-util handle this?

My apologies for not being more specific, I'm an end-user of this library, so am not sure about the intended behaviour.

Note that the documentation for [`SetAlternateInterface`](https://opensource.apple.com/source/IOUSBFamily/IOUSBFamily-630.4.5/Documentation/IOUSBLib/Functions/InterfaceInterface.html#SetAlternateInterface) that I could find, did not specifically mention `kIOUSBPipeStalled`. Only `kIOReturnSuccess`, `kIOReturnNoDevice`, and `kIOReturnNotOpen`.

Given the suggestion that this behaviour changed after a security update, I verified on an older macOS system. I can reproduce it on 10.14.6 (Mojave) and 10.11.6 (El Capitan), so I am not yet convinced that this was a trigger.

macfreek:
FYI, I can reproduce this with pyusb, so the issue is not specific to dfu-util.

```
import usb
dev = usb.core.find(idVendor = 0x04D8, idProduct=0xF014)
if dev:
    dev.set_configuration(1)
    dev.set_interface_altsetting(0, 0)
else:
    print("Can't find USB device")
```

raises `usb.core.USBError: [Errno 32] Pipe error`.

(I tried adding a `time.sleep(0.5)` just before the call to `set_interface_altsetting()` but got the same result).

mcuee:
[...]

PR #843, opened 2021-01-11
Fix for set_interface_alt_setting()

macfreek:
It should no longer return a pipe stalled error if the interface is successfully reset.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1567,8 +1567,15 @@ static int darwin_set_interface_altsetting(struct libusb_device_handle *dev_hand
     return LIBUSB_ERROR_NO_DEVICE;
 
   kresult = (*(cInterface->interface))->SetAlternateInterface (cInterface->interface, altsetting);
-  if (kresult != kIOReturnSuccess)
+  /* If a device only supports a default setting for the specified interface, then a STALL
+     (kIOUSBPipeStalled) may be returned. Ref: USB 2.0 specs 9.4.10.
+     Mimick the behaviour in e.g. the Linux kernel: in such case, reset all endpoints,
+     and hide errors.Current implementation resets the entire device, instead of single
+     interface, due to historic reasons. */
+  if (kresult != kIOReturnSuccess) {
+    usbi_warn (HANDLE_CTX (dev_handle), "SetAlternateInterface: %s", darwin_error_str(kresult));
     darwin_reset_device (dev_handle);
+  }
 
   /* update list of endpoints */
   ret = get_endpoints (dev_handle, iface);
@@ -1576,10 +1583,9 @@ static int darwin_set_interface_altsetting(struct libusb_device_handle *dev_hand
     /* this should not happen */
     darwin_release_interface (dev_handle, iface);
     usbi_err (HANDLE_CTX (dev_handle), "could not build endpoint table");
-    return ret;
   }
 
-  return darwin_to_libusb (kresult);
+  return ret;
 }
 
 static int darwin_clear_halt(struct libusb_device_handle *dev_handle, unsigned char endpoint) {


macfreek:
This is a fix for #838.

mcuee:
Maybe it is good to issue a warning before issuing the reset to warn that SetAlternateInterface is failing.
```
  kresult = (*(cInterface->interface))->SetAlternateInterface (cInterface->interface, altsetting);
  if (kresult != kIOReturnSuccess)
    darwin_reset_device (dev_handle);
```

macfreek:
> Maybe it is good to issue a warning before issuing the reset to warn that SetAlternateInterface is failing.

Fully agree. It now prints:
```
...
[ 1.090313] [002c0415] libusb: debug [libusb_set_interface_alt_setting] interface 0 altsetting 0
[ 1.091467] [002c0415] libusb: warning [darwin_set_interface_altsetting] SetAlternateInterface: pipe is stalled
[ 1.138173] [002c0415] libusb: debug [darwin_reset_device] darwin/reset_device: waiting for re-enumeration to complete...
...
```

hjelmn:
Can you squash these into a single commit? Once that is done I will merge this.

libusb/os/darwin_usb.c
```diff
@@ -1568,6 +1568,7 @@ static int darwin_set_interface_altsetting(struct libusb_device_handle *dev_hand
 
   kresult = (*(cInterface->interface))->SetAlternateInterface (cInterface->interface, altsetting);
   if (kresult != kIOReturnSuccess)
+    usbi_warn (HANDLE_CTX (dev_handle), "SetAlternateInterface: %s", darwin_error_str(kresult));
     darwin_reset_device (dev_handle);
```
tormodvolden, line 1572: Missing braces so only the warning is conditional, while the reset is always done.

macfreek: Bugger. I clearly have been exposed to Python recently :),.Thanks for the catch!

macfreek: Fixed in 6d7069f

tormodvolden:
I would have added to the commit message a reference to commit 065e586b which inadvertently changed this that you are now changing back.

tormodvolden:
If I am allowed to nitpick further, we are maybe not "fixing" the pipe error but ignoring it again, which from what I posted on the bug ticket is also done by the Linux kernel (at least in some cases), no idea what Windows does. And successfully has 3 s'es.

mcuee:
Good point. Typo needs to be fixed. And indeed we are not fixing the pipe error but rather ignore the error. Still the code does try to recover from the pipe error by reseting the device.

macfreek:
Yes, @tormodvolden, you _are_ a nitpicker. And I appreciate it. Thanks! :)

To avoid keeping to push, is the following fine. I'm getting verbose now :)

> MacOS: Silence pipe error in set_interface_alt_setting()
> 
> darwin_set_interface_altsetting() no longer returns the status of the underlying SetAlternateInterface(), but instead the result of a subsequent GetNumEndpoints() call. This reverts to the behaviour prior to commit 065e586.
> 
> Since darwin_set_interface_altsetting() resets the interface after SetAlternateInterface(), one of the consequences is that if SetAlternateInterface() returns a kIOUSBPipeStalled error, and the interface is successfully reset, darwin_set_interface_altsetting() now returns LIBUSB_SUCCESS instead of LIBUSB_ERROR_PIPE.
> 
> Closes: #838
> 
> Signed-off-by: Freek Dijkstra

I think that a comment in the code is warranted too, perhaps both that 
* `darwin_set_interface_altsetting()` returns the status of `GetNumEndpoints()`, not of `SetAlternateInterface()`.
* That the reset of the interface is macOS specific.

Unfortunately, I have no other reasons to specify other than "that's what happened before". Suggestions?

tormodvolden:
One reason is that it is closer to what happens on other platforms like Linux where the kernel hides the PIPE error (but resets all endpoints) if there is only one alternate setting in the interface.

macfreek:
I plan to add the following comment to the code. Please check if I got the terminology correct:

>  /* USB 2.0 specs 9.4.10: If a device only supports a default setting for the specified interface, 
>      then a STALL (kIOUSBPipeStalled) may be returned. Mimick behaviour on other platforms, 
>      e.g. Linux kernel: in this case reset all endpoints, and hide errors. Current implementation
>      resets entire device, instead of single interface, due to historic reasons. */

hjelmn:
@macfreek Looks good. Please add the comment to the commit and I will do the merge.

hjelmn:
Closed via bf943bc

macfreek:
@hjelmn Thanks for the merge, but above all your great care and attention. (And everyone else thanks too!)

(The last push was just after the merge, but no worries - it was literally a space missing after a full stop in the comments. After all the nitpicking, I couldn't let that one slip!)

Issue #844, opened 2021-01-11
Access violation in libusb v1.0.24 when using libusbK driver


dmitry-zakablukov:
Good day!

The following test code crashes with 0xC0000005 error code if libusb v1.0.24 is used for a device with libusbK driver:
```
#include "libusb.h"

#define VID 0x0951
#define PID 0x1600

libusb_device_handle* open_device( libusb_context* ctx, int vid, int pid )
{
    libusb_device_handle* handle = nullptr;

    libusb_device** devices = nullptr;
    ssize_t count = libusb_get_device_list( ctx, &devices );

    for( ssize_t index = 0; index < count; ++index )
    {
        libusb_device_descriptor desc{};
        libusb_get_device_descriptor( devices[ index ], &desc );

        if( desc.idVendor == VID && desc.idProduct == PID )
        {
            libusb_open( devices[ index ], &handle );
            break;
        }
    }

    libusb_free_device_list( devices, 1 );
    return handle;
}

int main()
{
    libusb_context* ctx = nullptr;
    libusb_init( &ctx );

    libusb_device_handle* handle = open_device( ctx, VID, PID );

    unsigned char buffer[ 256 ]{};

    // HERE: the following function call will crash the program
    libusb_control_transfer( handle,
        LIBUSB_ENDPOINT_IN | LIBUSB_REQUEST_TYPE_STANDARD | LIBUSB_RECIPIENT_DEVICE,
        LIBUSB_REQUEST_GET_DESCRIPTOR,
        LIBUSB_DT_STRING << 8,
        0,
        buffer,
        sizeof( buffer ),
        1000 );

    libusb_close( handle );
    libusb_exit( ctx );

    return 0;
}
```

Note that there is no crash if libusb v1.0.23 is used or a device has the WinUSB driver.

I know the problem is in `windows_iocp_thread() / usbi_signal_transfer_completion()` functions: access violation occurs in https://github.com/libusb/libusb/blob/master/libusb/io.c#L1734

When a `libusb_control_transfer()` is called, the libusbK driver is initialized, see https://github.com/libusb/libusb/blob/master/libusb/os/windows_winusb.c#L2531
This driver initialization triggers a return from `GetQueuedCompletionStatus()` in `windows_iocp_thread()`, but the data in `overlapped` variable is not the desired one (it is not the libusb library who prepared this data). This results in garbage pointed by `itransfer` variable here: https://github.com/libusb/libusb/blob/master/libusb/os/windows_common.c#L438

[...]

PR #848, opened 2021-01-14
Access violation fixed while processing foreign I/O completion packets

dmitry-zakablukov:
This PR fixes issue #844

In the order to distinguish libusb and foreign I/O completion packets, a marker was added to the `windows_transfer_priv` structure.
This approach still can cause an access violation in the case when a foreign I/O operation processor allocates a memory only for `OVERLAPPED` structure and the bytes after that memory are inaccessible for reading. But I think this is almost impossible situation.


dickens:
See my comment in #844 (should have made it here instead).  This approach will not work, for the reason you stated in your description.

dickens:
Issue has been addressed in b51c743, so I'll close this PR.  Thanks for the report.

dmitry-zakablukov:
@dickens , thank you!

Issue #850, opened 2021-01-14
Regression in libusb 1.0.24 under Linux --  buggy devices with configuration 0


bodograumann:
After an update from libusb 1.0.23-3 to 1.0.24-2 on Arch linux, my SmartCard reader does not work anymore.
I confirmed that it works again after a downgrade.
I am using gnupg with pcscd / ccid and get the following errors:

>Jan 14 14:16:13 MYHOSTNAME pcscd[5858]: 00008827 ccid_usb.c:598:OpenUSBByName() Can't get config descriptor on 1/3: LIBUSB_ERROR_NOT_FOUND
>Jan 14 14:16:13 MYHOSTNAME pcscd[5858]: 00000367 ifdhandler.c:150:CreateChannelByNameOrChannel() failed
>Jan 14 14:16:13 MYHOSTNAME pcscd[5858]: 00000020 readerfactory.c:1105:RFInitializeReader() Open Port 0x200000 Failed (usb:0a5c/5832:libudev:3:/dev/bus/usb/001/003)
>Jan 14 14:16:13 MYHOSTNAME pcscd[5858]: 00000003 readerfactory.c:376:RFAddReader() Broadcom Corp 5880 [Broadcom NFP] (0123456789ABCD) init failed.

Device: Bus 001 Device 003: ID 0a5c:5832 Broadcom Corp. 5880

Kernel: Linux 5.10.7-arch1-1 x86_64
Software:
- gnupg 2.2.26-1
- pcsclite 1.9.0-1
- ccid 1.4.33-1

tormodvolden:
For general debugging, please run the program with the debug environment variable set:
`env LIBUSB_DEBUG=4 pcscd -f -d
`
Of course if you could use git bisect to find which change broke it, it would be great.

dickens:
I suspect this is the same problem as #825.  Can you please confirm by testing the latest code from the repo?

mcuee:
https://gitlab.freedesktop.org/libfprint/libfprint/-/issues/88
lsusb -vvv result of the particular device.
```
Bus 001 Device 005: ID 0a5c:5834 Broadcom Corp. 
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 (Defined at Interface level)
  bDeviceSubClass         0 
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x0a5c Broadcom Corp.
  idProduct          0x5834 
[...]

PR #851, opened 2021-01-15
linux: Allow configuration value 0 in descriptor

tormodvolden:
Some buggy devices have a configuration with bConfigurationValue of 0.
This is a violation of the USB specification, but we will try to cope
with it.

Since this value normally means "unconfigured" for GET_CONFIGURATION and
SET_CONFIGURATION requests, your mileage may vary.

Commit e9eec3a6 had introduced stricter checking, and relaxing
it may fix issue #850.

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -808,10 +808,10 @@ static int op_get_active_config_descriptor(struct libusb_device *dev,
 		active_config = priv->active_config;
 	}
 
-	if (active_config == 0) {
-		usbi_err(DEVICE_CTX(dev), "device unconfigured");
-		return LIBUSB_ERROR_NOT_FOUND;
-	}
+	/* this might be a buggy device with a configuration value 0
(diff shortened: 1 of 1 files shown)


dickens:
This is a good start, but I'd rather verify that the device actually has a configuration descriptor with a bConfigurationValue of 0.  Then we can distinguish between an unconfigured device and a buggy device.

tormodvolden:
Yup. In the one case that was errorring out before, this is taken care of by the following op_get_config_descriptor_by_value(). In the second, op_get_configuration(), I was thinking the user can deal with it, we are not using that function internally. Does the API require us to do any checks for devices not being configured?

tormodvolden:
I have tested this by modifying a device firmware to have configuration value 0. libusb_get_active_config_descriptor() now returns success. I believe this solves the issue at hand, and as I reasoned in the previous comment, this is all that we have to do.

dickens:
Thanks for the PR.  I've addressed the issue in a more complete manner in f38f09d.

Issue #847, opened 2021-01-14
Potential libusb_set_interface_alt_setting improvement for macOS


mcuee:
As mentioned by #838, there is a macOS specific behavior if libusb_set_interface_alt_setting failed with a pipe stall. 

```
kresult = (*(cInterface->interface))->SetAlternateInterface (cInterface->interface, altsetting);
  /* If a device only supports a default setting for the specified interface, then a STALL
     (kIOUSBPipeStalled) may be returned. Ref: USB 2.0 specs 9.4.10.
     Mimick the behaviour in e.g. the Linux kernel: in such case, reset all endpoints,
     and hide errors.Current implementation resets the entire device, instead of single
     interface, due to historic reasons. */
  if (kresult != kIOReturnSuccess) {
    usbi_warn (HANDLE_CTX (dev_handle), "SetAlternateInterface: %s", darwin_error_str(kresult));
    darwin_reset_device (dev_handle);
  }
```

> From Nathan:
The reset was not present in libusb 0.1.12 so I specifically added it when implementing the libusb 1.0 backend. It was undoubtedly to work around some issue seen during development. If there was some sort of bug in that version of IOUSBFamily it is likely long gone (since at least 10.9.0 given the IOUSBFamily rewrite).

tormodvolden:
I have some suggestions at https://github.com/tormodvolden/libusb/tree/darwin_alt_intf
Not yet a merge requests because it wasn't tested much, and there is something going wrong in darwin_clear_halt(): it seems to me that it finds the correct pipeRef's but ClearPipeStall returns "pipe ref not recognized". Maybe I use the wrong parameter or an off-by-one. Ran out of time but maybe someone else spots it. Also not sure if get_endpoints() is needed any longer.

hjelmn:
get_endpoints should be required as the pipes may change.

mcuee:
One idea I mentioned in #838.
>We can probably check the current alternative setting and do nothing (most of the cases) if it is already at the correct alternative setting. Majority the device should only have the defaul alt setting 0. And for those which needs to be set a different alt settings, then the device should anyway supports the call propely.

mcuee:
BTW, I remember there are devices having problem with Clear_HALT feature request as well.

mcuee:
Of course there is another opinion that libusb should not do any check and the application should check. 

[...]

PR #852, opened 2021-01-17
darwin_set_interface_altsetting: Better handling if device doesn't support SET_INTERFACE

tormodvolden:


tormodvolden:
This should maybe be fixed up first:
If we now do the darwin_claim_interface() here it will implicitly call get_endpoints() so we don't have to do that explicitly at the end.

It seems the darwin_release_interface() is not needed. I don't know if a "double" claim without a proper release between is any problem. OTOH it is maybe better to not release if it is not necessary since it would leave it unclaimed for a brief period if we are scheduled.

tormodvolden:
The "fixup" commit addresses the two points in my previous comment. I would squash it with the previous commit if you agree.

tormodvolden:
BTW, I also tried to replace the full darwin_claim_interface() with only the USBInterfaceOpen() from it, but that didn't work - I assume the QueryInterface() plugInInterface stuff is needed.

tormodvolden:
This was the missing piece mentioned in #847: If the interface has several alternate interfaces, and the user actually wanted to change the value, we should return error if we get pipe error.

libusb/os/darwin_usb.c
```diff
@@ -1567,25 +1570,34 @@ static int darwin_set_interface_altsetting(struct libusb_device_handle *dev_hand
...
   if (ret) {
-    /* this should not happen */
     darwin_release_interface (dev_handle, iface);
-    usbi_err (HANDLE_CTX (dev_handle), "could not build endpoint table");
+    usbi_err (HANDLE_CTX (dev_handle), "could not reclaim interface");
   }
 
-  return ret;
+  /* Return error if we failed to change to another value */
+  kresult = (*(cInterface->interface))->GetAlternateSetting (cInterface->interface, &old_alt_setting);
```
dickens, line 1593: Does this cause a GET_INTERFACE request to be sent to the device?  If so this is not a good change as it will increase the chances of crashing buggy devices.

tormodvolden: No, it doesn't. I tested, and it seems to be cached by the OS.

tormodvolden:
An alternative to GetAlternateSetting (in case it would have caused a request) would have been:
      struct libusb_config_descriptor *config;
      int rc = libusb_get_active_config_descriptor (dev_handle->dev, &config);
      if (rc == LIBUSB_SUCCESS && config->interface[iface].num_altsetting > 1)
        return LIBUSB_ERROR_PIPE;
Not exactly the same logic but in practice the same: Only if the num_altsetting > 1 would another set value arrive through the upper layers.

hjelmn:
Please rebase and squash the commits into a minimal number of commits.

tormodvolden:
@hjelmn rebased, squashed and repushed.

Issue #855, opened 2021-01-19
Segfault at libusb_exit() -- libusb crashing on the wrong context


ctj-nooiwa:
Segfault occurs at libusb_exit() in my environment.
I can reproduce it and test program is follow.
Proper use is calling libusb_ext with non-NULL context.
But, I think causing segfault with wrongly given NULL context is too harsh.
```
$ cat libusb.c
#include <stdio.h>
#include <libusb-1.0/libusb.h>
#include <sys/types.h>
#include <unistd.h>
#include <pthread.h>
// gcc -lusb-1.0 -lpthread libusb.c -o libusb

void* libusb()
{
	libusb_context *ctx = NULL;
	while (1) {
		libusb_init(&ctx);
		libusb_exit(NULL);
	}
	return NULL;
}

int main(int argc, char **argv)
{
	int t;
	pthread_t threads[20];

	for (t = 0; t < 20; t++) {
		pthread_create(&threads[t], NULL, libusb, NULL);
	}
	sleep(100000);
	return 0;
}
```
Segfault occurs after a few seconds on Fedora 32(Intel i7-9700K CPU, libusbx-1.0.23-1 with
a5624b2: "core: Do not attempt to destroy a default context that doesn't exist").
```
$ gdb ./libusb core.libusb.3331047
[...]
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fe14640a8dc in list_del (entry=0x7fe128018088) at libusbi.h:163
163		entry->next->prev = entry->prev;
(gdb) bt
#0  0x00007fe14640a8dc in list_del (entry=0x7fe128018088) at libusbi.h:163
#1  libusb_exit (ctx=0x7fe128017e60) at core.c:2450
#2  0x000000000040117c in libusb () at libusb.c:14
#3  0x00007fe1463eb432 in start_thread () from /lib64/libpthread.so.0
#4  0x00007fe1463199d3 in clone () from /lib64/libc.so.6
```
I created the patch. usbi_get_context() need to hold default_context_lock.
usbi_dbg() also call usbi_get_context().
Please check my patch.
```
diff --git a/libusb/core.c b/libusb/core.c
index 051b4f7..737d4b0 100644
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2367,13 +2367,12 @@ void API_EXPORTED libusb_exit(libusb_context *ctx)
        struct timeval tv = { 0, 0 };
        int destroying_default_context = 0;
 
[...]

PR #856, opened 2021-01-20
Refactor initialization and how the default context is handled

dickens:
Highlights for this change:

- `usbi_default_context` is only set if `libusb_init()` is called with `NULL`
- All hotplug related functionality (e.g. initialization, processing) has been moved to hotplug.c
- Backends are simplified by removing initialization mutexes.  Mutual exclusion between `init()/exit()` is provided by `default_context_lock`
- Make hotplug types and functions part of libusbi.h with the common usbi_ prefixes (removes hotplug.h)

Addresses issue highlighted in #855


dickens:
If there are no objections I will merge this later today.  This also addresses #858.

mcuee:
The code changes are a bit complicated for me to review (after all I am not a programmer), but I agree with the general idea and the clean-up (refactor). So I have no objection with the changes.

hjelmn:
That is quite the PR to review. Would have been nice to have it broken down into multiple commits. Easier to review. Also a lot of unnecessary changes of variable names increases the delta making it more difficult to review.

Overall the changes look good but I have only had a chance to skim them. I need a chance to go through them in more detail.

Also, you have a good description in the PR but the commit message is very light. Can you update it with more detail?

matallui:
Is this fix going to make it to 1.0.25? Been waiting for the fix to be released.

mcuee:
Since Nathan is okay with the fix, I am for the fix to be merged and it is better to be included in the 1.0.25.

mcuee (approved):

mcuee:
@hjelmn This is the last major pull-request to be merged for 1.0.25 release. If Chris has no time to rework the patches, please help here. Thanks.

benzea:
Just for the record. This change broke gutenprint, as it incorrectly did a `libusb_init(&ctx)` and then continued to call `libusb_get_device_list` with a `NULL` context later on. The proposed fix is https://sourceforge.net/p/gimp-print/mailman/message/37614729/

Previously, this worked, because the created context was also the default, but now it does not work anymore.

See https://bugzilla.redhat.com/show_bug.cgi?id=2055504

For Fedora, I did the nasty thing and worked around it for now. That workaround is unfortunately much worse than the proper fix … but it seems to work https://src.fedoraproject.org/rpms/libusb1/blob/5e2dcc2c62b6603c3db85a783a09cb82af79499e/f/0001-core-Install-first-context-as-implicit-default.patch

PR #857, opened 2021-01-21
linux_usbfs: Accept sysfs attributes without trailing newline

benzea:
While Linux will always include a newline at the end of an attribute, it
does not really belong to the value itself. In fact, e.g. systemd (udev)
will strip all trailing newline characters, without relying on them to
be there.

Requiring newlines seems unecessary overall. And it caused some trouble
by breaking umockdev based CI tests where the trailing whitespace had
been stripped accidentally.

CC: @jwrdegoede

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -505,7 +505,7 @@ static int read_sysfs_attr(struct libusb_context *ctx,
 	if (fd < 0)
 		return fd;
 
-	r = read(fd, buf, sizeof(buf));
+	r = read(fd, buf, sizeof(buf)-1);
 	if (r < 0) {
 		r = errno;
 		close(fd);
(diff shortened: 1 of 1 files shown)


martinpitt:
Note that this was a behaviour change/regression in commit c3864c822b - libusb up to 1.0.23 didn't need the newlines either, just 1.0.24 does.

dickens:
This is not classifiable as a regression.  If something is going to simulate/mock sysfs, its behavior should actually mimic that of the real kernel.

benzea:
> This is not classifiable as a regression. If something is going to simulate/mock sysfs, its behavior should actually mimic that of the real kernel.

Of course. The main motivation for me is to avoid painful "fails to build from source" situations for distributions that are easily avoidable this way. And, to be honest, it also simply feels a bit cleaner to me, even if it should be acceptable to rely on the `\n` to be included.

martinpitt:
@dickens: Sure, and it does that now. I just wanted to point out in which version/commit it changed behaviour.

PR #867, opened 2021-02-03
Windows: Try alternative methods to query port number

darchons:
This commit is similar to #634 but I think it offers a better approach. The commit adds two more methods for querying the port number, in addition to the existing `SPDRP_ADDRESS` method, which may not actually return the port number when using USB drivers other than the Microsoft driver.

The first method uses `SPDRP_LOCATION_INFORMATION`, which may return a port in a format similar to "Port_#0002.Hub_#000D".

The second method uses `SPDRP_LOCATION_PATHS`, which returns the device location in a format similar to "PCIROOT(B2)#PCI(0300)#PCI(0000)#USBROOT(0)#USB(1)#USB(2)#USBMI(3)", where the port number is the number within the last "#USB()" token.

If both methods fail, `SPDRP_ADDRESS` is still used as a fallback.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1223,6 +1223,47 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 	return LIBUSB_SUCCESS;
 }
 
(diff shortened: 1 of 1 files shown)


patstew:
FWIW this fixes the issue with devices not being found on the Dell Thunderbolt docks for me.

mcuee:
The fix looks okay to me.

Issue #866, opened 2021-02-02
Getting the serial number without opening the device handle


mcuee:
Reference:
There are a few past discussions and recent discussions on getting the serial number without opening the device handle

Past discussion: 
https://marc.info/?t=142939771300004&r=1&w=4
https://sourceforge.net/p/libusb/mailman/message/33894596/

Recent discussion: Getting a serial number without opening a handle
Feb 2019: Accessing string descriptors when can't do libusb_open
https://marc.info/?t=155008592500001&r=1&w=4

Feb 2021: Detecting Multiple Devices Of Same Type
https://marc.info/?t=161158709600002&r=1&w=4

xloem:
I visited hjmallon's repository at https://github.com/hjmallon/libusb but did not find the fork mentioned at https://marc.info/?l=libusb-devel&m=155025463511502&w=4

mcuee:
I do not think the one materialized in the end. 

libusbp from David Grayson claims to have this feature.
https://github.com/pololu/libusbp

hjmallon:
@xloem I didn't think anyone was very interested. 😄 Let me do some git archeology and PR what I have this week!

hjmallon:
Here is my change from 2 years ago. I have rebased it on master but this has caused some things to need changing. If you are interested in helping test is (I have no Windows or linux machines atm) please comment on the PR.

https://github.com/libusb/libusb/pull/875

mliberty1:
I agree that this is an important feature.  I am using PR https://github.com/libusb/libusb/pull/875 in my code, and it would be great to have this upstreamed.

It seems like the API has been the sticking point in integrating this feature.  These USB string descriptors are processed by the OS, so the USB UTF-16 and ASCII formats are not guaranteed.  Perhaps a UTF-8 format would be better?  How about this API?

```
/** \ingroup libusb_desc
 * The OS string type for libusb_get_os_string().
 */
enum libusb_os_string_type_e {
   LIBUSB_OS_STRING_TYPE_MANUFACTURER,
   LIBUSB_OS_STRING_TYPE_PRODUCT,
   LIBUSB_OS_STRING_TYPE_SERIAL_NUMBER,
};

/** \ingroup libusb_desc
 * Retrieve a device string from the operating system.
 * This function works when the device is still closed since it relies
 * on the operating system to fetch the USB string descriptor.
 * Since the USB string descriptor could be processed by the OS,
 * this function returns a UTF-8 encoded string.  The string will
[...]

Issue #584, opened 2019-06-17
on the android 8 platfrom, libusb_init is ok，but libusb_get_device_list return 0 -- 1.0.22 is okay, regression in 1.0.23


laomengandroid:
Use the sample you provided，Step by step；
1.added log on the testlibusb.c adb libusb Source code
2.will binaryed libusb1.0 Library push   /system/lib64
3.will binaryed testlibusb  push  /system/bin
4.adb shell  /system/bin/ testlibusb
5.view log printf:
   >libusb_init is successfull
        r = libusb_init(ctx);
	@printf("libusb_init==%d",r);   printf r ==0;
   >libusb_get_device_list return 0
     cnt = libusb_get_device_list(NULL, &devs);
	  if (cnt < 0)
		return (int)cnt;
        @printf("libusb_get_device_list==%d\n",cnt); ---> cnt==0

wbing-xu:
I also found the same problem.
when i  reset to tag 'v1.0.22',  device can be found!

hjelmn:
What version are you running?

cnlohr:
It's broken in 1.0.23. I can't find any code path which can actually do enumeration on android in 1.0.23.

sicauyanglei:
What can I do for this problem?

sicauyanglei:
> ag 'v1.0.22

how can I do it

cnlohr:
use `git checkout` to specify the older branch.  I.e. `git checkout v1.0.22` in the folder where you've checked out the source.

mcuee:
https://github.com/libusb/libusb/commit/2f3bc98b0d0f4766496df53c855685a5f0e5e7cf

Maybe this commit has some problems.

mcuee:
#233 comment is relevant.

laran002:
Parts of 2f3bc98, related to linux_scan_devices(), is reverted back in #637

IXMSbhatt:
libusb_get_device_list still return 0 in android 7.1.1 too

cnlohr:
Just checking - are you running on a rooted device with selinux set to not enforce?

IXMSbhatt:
it's rooted device and selinux is not set. With libusb-1.0.22 it's working perfectly fine. I have revert back 1.0.23 to 1.0.22. Now my problem is solved with libusb-1.0.22.

green-green-avk:
Hmm...
[...]

Issue #596, opened 2019-07-16
Including libusb on Android using cmake


iliis:
This is one of these unfortunate cases where several things come together to cause a problem. I'm sharing it here to discuss it and get some comments and maybe spare someone else the pain of having to discover this the hard way.

## The Context

I'm porting a relatively large codebase to Android, which uses cmake to build everything and makes use of libusb to talk to some hardware. I'm cross-compiling this library by directly invoking cmake, i.e. basically `cmake .. -DANDROID_ABI=arm64-v8a -DCMAKE_TOOLCHAIN_FILE=$NDK/build/cmake/android.toolchain.cmake` (I then later include the resulting .so file in my Android project, together with some c++ glue code).

The official way to automatically find libusb is by using pkg-config. The autotools will put a `libusb-1.0.pc` file somewhere in `/usr/lib` that tells the user's build system where it can find the library and headers. We make use of that in our cmake scripts by `include(FindPkgConfig)` and `PKG_CHECK_MODULES(LIBUSB1 REQUIRED libusb-1.0>=1.0.17)`.

## The Problem

The Android build system provides no pkg-config at all, so when I build our library for Android cmake finds the host's pkg-config (or nothing at all) which then happily proceeds to include the host's libusb. Which is of course totatlly the wrong one.

## Possible Workarounds

Apart from just hardcoding a few absolute paths here and there, I've come up with a few ways to fix this on my side:

1. Write a `FindLibusb.cmake` script and include it in my project. This could fall back to pkg-config in the general case and add some magic in case we're compiling for Android. This means I have to modify our `CMakeLists.txt` to use `find_package()` instead.
2. Provide a stub that impersonates pkg-config and outputs the right values (which is what ffmpeg does to find the freetype library: https://github.com/rstiller/android-ndk-builder/blob/master/ffmpeg/fake-pkg-config)
3. Write a custom pkg-config `libusb-1.0.pc` and use environment variables to tell pkg-config where to look for it (PKG_CONFIG_{DIR,LIBDIR,SYSROOT}).

## Discussion

All these solutions are relatively simple to implement, but all require some work. I.e. it's not working out of the box. Maybe we can find a way to add support for this scenario? After all, libusb works very nicely on Android, even without root.

Maybe we can configure ndk-build to spit out a `libusb-1.0.pc` as well? I'm not familiar with ndk-build, but the newer, cmake-based Android build system would certainly be able to do this.

## P.S.

[...]

Issue #31, opened 2014-08-26
Android: Android build name correction


kuldeepdhaka:
android/jni/libusb.mk

LOCAL_MODULE := libusb1.0
this should be
LOCAL_MODULE := usb-1.0

"lib" should be dropped as it is not required
"-" should be placed between "usb" and "1.0".

also appropriately update  android/jni/examples.mk 
why important: afaik naming system on desktop is libusb-1.0.so\* not libusb1.0.so*

mcuee:
Refer to this pull request:
https://github.com/libusb/libusb/pull/143

mcuee:
Close this aging ticket since it is captured in the above pull request. Please re-open if you think it is wrongly closed.

iliis:
https://github.com/libusb/libusb/pull/143 was apparently never merged, so this issue still remains. I've ran into this when cross-compiling my own library that finds libusb via pkg-config (and thus looks for "libusb-1.0", i.e. with "-").

mcuee:
Reopen this since it is mentioned in another ticket.

mcuee:
Close this for now and please use #596 for further discussion.

mcuee:
@adisuciu Does #596 cover this? I think #596 covers this so I closed this ticket. If it is not true, I will reopen this ticket.

adisuciu:
I'm not an expert on the matter but I think this ticket is a prerequisite for #596 as you can see in the P.S at the end of #596. #596 proposes that the Android build script should also create a .pc file which would be even better for detecting the library.

xloem:
@adisuciu do you have any idea on how to prevent breaking existing buildsystems that rely on the wrong library name?  Maybe an environment or make variable, or outputting two libraries, or waiting until a major version release?

Did anyone ever make a PR for this change?  I think it might be a different issue from #874 but I do see that my pr is getting used and it would be good to help other issues out if I can.  It's been so many years with this issue here unaddressed.

adisuciu:
I think some kind of variable that changes the library name would be good.
When making without the variable, the script should say that the .so name would be deprecated and they should set the variable when making to be fully compatible in the future ?

mcuee:
Reopen as this is still needed.

PR #874, opened 2021-02-12
Better Android Support

### LINKED ISSUE #717: Matching from android.hardware.usb.UsbDevice to a libusb_device

### LINKED ISSUE #868: Automating permissions on Android
After https://github.com/libusb/libusb/issues/862#issuecomment-773726901 it looks like Android is somewhat supported by libusb, but the user still needs to write some JNI and Java to map between the devices listed in the android api and the actual device handles or files.  I'm opening this ticket just as a way to track and discuss the idea of including that code within libusb itself, as some forks have done in the past.

It should all be doable within C++ if the user passes a JNIEnv* pointer in, I think.

xloem:
**This code implements connection and device listing on unrooted android devices, without requiring any user-provided java.  **
This PR is usable now, but not all possible features are implemented, such as device relisting or forking away from the process spawned by android.  The hope is that the codebase is expandable by others.

Needs remerging and testing now #942 is merged.

The java native interface library is used, which is always available when building on android, to do the sdk calls within libusb itself.  The approach is modularised to aid expanding if new features are needed.

There is also an [RPC approach](https://github.com/libusb/libusb/pull/874#issuecomment-873322802) that I have not yet reviewed, for use in terminal emulator apps.  An RPC approach could handle more edge cases than this direct approach does.

I provided for device enumeration by generating usb descriptors from the android sdk, which requires no user permission for this information.

I also detect if permission is needed when an unpermissioned device is opened, and perform the `requestPermission` call if necessary.  This pops up a dialogue to the user asking if they will allow the connection.

Dispatching this request involves setting an action string on the intent for permission that the application can optionally subscribe to.  The permission intent I dispatch has action string `libusb.android.USB_PERMISSION`.

The code only takes effect if the user provides their JNIEnv* or JavaVM* pointer via a new libusb option.

Notes:
- [ ] Now that https://github.com/android/ndk/issues/1320 is fixed in android api 31, this could be detected to use nonrooted android code transparently when api >= 31, without libusb_set_option
- [x] Remove string descriptor generation since the result is discarded.  Paste a commit link to cherry pick and revert to #875 in case it's helpful.  #866
- [ ] The way the application context is acquired may not work if the application is a service,
- [x] If libusb functions are called from non-main threads, `DetachCurrentThread` must be called to close the environment pointer.  `GetEnv` returns a special value when this is needed.  Maybe use thread destruction as recommended by android docs: https://github.com/libusb/libusb/pull/874#issuecomment-841715680 .  Testing this means verifying thread detachment.
- [x] Remove the extra parameter added to the weak_authority option for compatibility with any existing code that does not pass an extra parameter.
- [ ] Now that there is android example documentation in the main tree, it would make sense to mention the jni approach too, in it.
- [ ] optional but desired: Appropriate changes to AndroidManifest.xml to use USB devices smoothly should be documented.  For example `<uses-feature>` `android.hardware.usb.host` `android:required="false"`.  Relate with the readme pull request regarding documentation?  See [android usb host guide](https://developer.android.com/guide/topics/connectivity/usb/host).
- [ ] optional: Let the javavm be detached from libusb so the user can fork() without producing issues accessing the android application context.
- [ ] optional: It seems it would be nice for there to be a way to generate a libusb device connection from an android intent.
- [ ] optional: Add support for hotplug / device_list to recognise new devices as they change. (was #584).  It may not be reasonable to respond to device insertion without accessory java, but the list of what's connected should be freely rescannable.
- [x] Rebase to reduce the commit clutter.  Keep a reference to the history tree somewhere.
- [ ] optional: refactor to match the common backend interface better
- [x] It might make more sense to accept a JavaVM pointer than a JNI pointer.  I think android is moving towards making this more normal, since it is safer.  See their [native application header file describing both](https://github.com/pfalcon/android-platform-headers/blob/master/android-4.2_r1/frameworks/native/include/android/native_activity.h#L50), which is often used from non-main threads.
- [ ] optional: Build a test app
- [X] A number of completed todo items were removed from this post to shorten it.  They are available in the post edit history.

fixes #868 fixes #717

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -2125,7 +2125,52 @@ enum libusb_option {
 	 */
 	LIBUSB_OPTION_WEAK_AUTHORITY = 3,
 
-	LIBUSB_OPTION_MAX = 4
+	/** Provide a JNIEnv* pointer for libusb to use on Android.  If this
+	 * pointer is nonzero, the Android SDK will be used for backend
+	 * functions that would otherwise require additional Java code to
+	 * ask for permission.  The pointer must be correct for the current
+	 * thread and process.
+	 *
+	 * This option should be set _before_ calling libusb_init(), and
+	 * specifies the java virtual machine only for new calls to
+	 * libusb_init() after the option is set.  The context pointer is
+	 * unused.
+	 *
+	 * When this option is set, the provided pointer is used for device
+	 * enumeration and connection for the entire lifetime of the libusb
+	 * context.  After connection, the normal linux backend is used via
+	 * the opened file descriptor.
+	 *
+	 * If a device does not have permission to connect, an Android Intent
+	 * is broadcast via UsbManager.requestPermission with action string
+	 * "libusb.android.USB_PERMISSION".  As documented in the Android
+	 * SDK, this Intent will have two "extras" indicating details of the
+	 * USB device in question, and whether or not the user granted
+	 * permission. The call to open the device returns LIBUSB_ERROR_ACCESS.
+	 *
+	 * Before a connection is made and the user provides permission for a
+	 * device, libusb generates fake descriptors for it from Android's API.
+	 * If these descriptors are used before connection, the correct ones
+	 * must be rerequested with an appropriate libusb API function, after
+	 * connection.
+	 *
+	 * A crash may happen if fork() is performed in a way Android doesn't
+	 * track, because the Application Context will no longer be valid.
+	 *
+	 * Only valid on Android.
+	 */
+	LIBUSB_OPTION_ANDROID_JNIENV = 4,
+
+	/** Like LIBUSB_OPTION_ANDROID_JNIENV, except a thread-agnostic JavaVM*
+	 * pointer is passed instead of a thread-specific JNIEnv* pointer.  If
+	 * you can pass a JavaVM* instead of a JNIEnv* pointer, the it does not
+	 * matter what thread it is passed from.
+	 *
+	 * Setting either option is equivalent.
+	 */
+	LIBUSB_OPTION_ANDROID_JAVAVM = 5,
+
+	LIBUSB_OPTION_MAX = 6
 };
 
 int LIBUSB_CALL libusb_set_option(libusb_context *ctx, enum libusb_option option, ...);
--- a/libusb/os/linux_android_jni.c
+++ b/libusb/os/linux_android_jni.c
@@ -1,0 +1,998 @@
+/* -*- Mode: C; c-basic-offset:8 ; indent-tabs-mode:t -*- */
+/*
+ * Android jni interface for libusb
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include "linux_android_jni.h"
+#include "libusb.h"
+
+#include <errno.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <string.h>
+
+#include <jni.h>
+
+static int android_jni_env(struct android_jni_context *jni, JNIEnv **jni_env);
+static int android_jni_fill_ctx_ids(struct android_jni_context *jni,
+                                    JNIEnv *jni_env);
+static int android_jni_gen_epoint(struct android_jni_context *jni,
+                                  JNIEnv *jni_env, jobject endpoint,
+                                  uint8_t **descriptors, size_t *descriptors_len);
+static int android_jni_gen_iface(struct android_jni_context *jni, JNIEnv *jni_env,
+                                 jobject interface, uint8_t *num_ifaces,
+                                 uint8_t **descriptors, size_t *descriptors_len);
+static int android_jni_gen_config(struct android_jni_context *jni,
+                                  JNIEnv *jni_env, jobject config,
+                                  uint8_t **descriptors, size_t *descriptors_len);
+
+struct android_jni_context
+{
+	JavaVM *javavm;
+	pthread_key_t detach_pthread_key;
+
+	/* jobjects need global references */
+	jclass Intent, PendingIntent;
+
+	jobject PackageManager__FEATURE_USB_HOST;
+
+	jobject application_context;
+	jobject package_manager;
+	jobject usb_manager;
+
+	jstring permission_action;
+
+	/* ids do not need global references */
+	jmethodID Collection_iterator;
+	jmethodID HashMap_values;
+	jmethodID Iterator_hasNext, Iterator_next;
+
+	int Build__VERSION__SDK_INT;
+	int Build__VERSION_CODES__P;
+	jmethodID Intent_init;
+	jmethodID PackageManager_hasSystemFeature;
+	jmethodID PendingIntent__getBroadcast;
+	jmethodID UsbConfiguration_getInterfaceCount, UsbConfiguration_getId,
+	          UsbConfiguration_getName, UsbConfiguration_isSelfPowered,
+	          UsbConfiguration_isRemoteWakeup, UsbConfiguration_getMaxPower,
+	          UsbConfiguration_getInterface;
+	jmethodID UsbDevice_getDeviceId, UsbDevice_getConfigurationCount,
+	          UsbDevice_getVersion, UsbDevice_getDeviceClass,
+	          UsbDevice_getDeviceSubclass, UsbDevice_getDeviceProtocol,
+	          UsbDevice_getVendorId, UsbDevice_getProductId,
+	          UsbDevice_getManufacturerName, UsbDevice_getProductName,
+	          UsbDevice_getSerialNumber, UsbDevice_getConfiguration;
+	jmethodID UsbDeviceConnection_close, UsbDeviceConnection_getFileDescriptor,
+	          UsbDeviceConnection_getRawDescriptors;
+	jmethodID UsbEndpoint_getAddress, UsbEndpoint_getAttributes,
+	          UsbEndpoint_getMaxPacketSize, UsbEndpoint_getInterval;
+	jmethodID UsbInterface_getEndpointCount, UsbInterface_getId,
+	          UsbInterface_getAlternateSetting, UsbInterface_getInterfaceClass,
+	          UsbInterface_getInterfaceSubclass, UsbInterface_getInterfaceProtocol,
+	          UsbInterface_getName, UsbInterface_getEndpoint;
+	jmethodID UsbManager_getDeviceList, UsbManager_hasPermission,
+	          UsbManager_openDevice, UsbManager_requestPermission;
+};
+
+int android_jnienv_javavm(JNIEnv *jni_env, JavaVM **javavm)
+{
+	if (jni_env == NULL) {
+		*javavm = NULL;
+		return LIBUSB_SUCCESS;
+	}
+	return (*jni_env)->GetJavaVM(jni_env, javavm) == JNI_OK ?
+		LIBUSB_SUCCESS : LIBUSB_ERROR_OTHER;
+}
+
+int android_jni(JavaVM *javavm, struct android_jni_context **jni)
+{
+	int r;
+	JNIEnv *jni_env;
+
+	jclass ActivityThread, Context;
+	jmethodID ActivityThread__currentActivityThread, ActivityThread_getApplication;
+	jmethodID Context_getPackageManager, Context_getSystemService;
+	jobject activity_thread, Context__USB_SERVICE;
+	jobject local_context, local_pkg_mgr, local_usb_mgr;
+	jstring local_perm_act;
+
+	*jni = malloc(sizeof(struct android_jni_context));
+	if (*jni == NULL)
+		return LIBUSB_ERROR_NO_MEM;
+
+	(*jni)->javavm = javavm;
+
+	r = pthread_key_create(
+		&(*jni)->detach_pthread_key,
+		(void(*)(void*))(*javavm)->DetachCurrentThread);
+	if (r != 0) {
+		free(*jni);
+		return r == ENOMEM ? LIBUSB_ERROR_NO_MEM : LIBUSB_ERROR_OTHER;
+	}
+
+	r = android_jni_env(*jni, &jni_env);
+	if (r != LIBUSB_SUCCESS) {
+		free(*jni);
+		return r;
+	}
+
+	r = android_jni_fill_ctx_ids(*jni, jni_env);
+	if (r != LIBUSB_SUCCESS) {
+		free(*jni);
+		return r;
+	}
+
+	/* ActivityThread activity_thread = ActivityThread.currentActivityThread(); */
+	ActivityThread = (*jni_env)->FindClass(jni_env, "android/app/ActivityThread");
+	ActivityThread__currentActivityThread =
+		(*jni_env)->GetStaticMethodID(jni_env,
+			ActivityThread, "currentActivityThread",
+			"()Landroid/app/ActivityThread;");
+	ActivityThread_getApplication =
+		(*jni_env)->GetMethodID(jni_env,
+			ActivityThread, "getApplication", "()Landroid/app/Application;");
+	activity_thread =
+		(*jni_env)->CallStaticObjectMethod(jni_env,
+			ActivityThread, ActivityThread__currentActivityThread);
+
+	/* Context local_context = activity_thread.getApplication(); */
+	local_context =
+		(*jni_env)->CallObjectMethod(jni_env,
+			activity_thread, ActivityThread_getApplication);
+
+	(*jni)->application_context = (*jni_env)->NewGlobalRef(jni_env, local_context);
+
+
+	Context = (*jni_env)->FindClass(jni_env, "android/content/Context");
+	Context__USB_SERVICE =
+		(*jni_env)->GetStaticObjectField(jni_env,
+			Context,
+			(*jni_env)->GetStaticFieldID(jni_env,
+				Context, "USB_SERVICE", "Ljava/lang/String;"));
+	Context_getPackageManager =
+		(*jni_env)->GetMethodID(jni_env,
+			Context, "getPackageManager",
+			"()Landroid/content/pm/PackageManager;");
+	Context_getSystemService =
+		(*jni_env)->GetMethodID(jni_env,
+			Context, "getSystemService",
+			"(Ljava/lang/String;)Ljava/lang/Object;");
+
+	/* PackageManager local_pkg_mgr = application_context.getPackageManager(); */
+	local_pkg_mgr =
+		(*jni_env)->CallObjectMethod(jni_env,
+			(*jni)->application_context, Context_getPackageManager);
+
+	(*jni)->package_manager = (*jni_env)->NewGlobalRef(jni_env, local_pkg_mgr);
+
+	/* UsbManager local_usb_mgr =
+		application_context.getSystemService(Context.USB_SERVICE); */
+	local_usb_mgr =
+		(*jni_env)->CallObjectMethod(jni_env,
+			(*jni)->application_context, Context_getSystemService,
+			Context__USB_SERVICE);
+
+	(*jni)->usb_manager = (*jni_env)->NewGlobalRef(jni_env, local_usb_mgr);
+
+	/* String local_perm_act = "libusb.android.USB_PERMISSION"; */
+	local_perm_act = (*jni_env)->NewStringUTF(jni_env,
+		"libusb.android.USB_PERMISSION");
+
+	(*jni)->permission_action = (*jni_env)->NewGlobalRef(jni_env, local_perm_act);
+
+	(*jni_env)->DeleteLocalRef(jni_env, local_perm_act);
+	(*jni_env)->DeleteLocalRef(jni_env, local_usb_mgr);
+	(*jni_env)->DeleteLocalRef(jni_env, local_pkg_mgr);
+	(*jni_env)->DeleteLocalRef(jni_env, local_context);
+
+	(*jni_env)->DeleteLocalRef(jni_env, Context);
+	(*jni_env)->DeleteLocalRef(jni_env, Context__USB_SERVICE);
+	(*jni_env)->DeleteLocalRef(jni_env, ActivityThread);
+
+	return LIBUSB_SUCCESS;
+}
+
+int android_jni_free(struct android_jni_context *jni)
+{
+	int r;
+	JNIEnv *jni_env;
+	JavaVM *javavm = jni->javavm;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS) {
+		goto out;
+	}
+
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->Intent);
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->PendingIntent);
+
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->PackageManager__FEATURE_USB_HOST);
+
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->application_context);
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->package_manager);
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->usb_manager);
+
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->permission_action);
+
+out:
+	if (pthread_getspecific(jni->detach_pthread_key) == javavm) {
+		(*javavm)->DetachCurrentThread(javavm);
+	}
+	pthread_key_delete(jni->detach_pthread_key);
+
+	free(jni);
+	return r;
+}
+
+int android_jni_detect_usbhost(struct android_jni_context *jni, int *has_usbhost)
+{
+	int r;
+	JNIEnv *jni_env;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* has_usbhost =
+		package_manager.hasSystemFeature(PackageManager.FEATURE_USB_HOST); */
+	*has_usbhost =
+		(*jni_env)->CallBooleanMethod(jni_env,
+			jni->package_manager, jni->PackageManager_hasSystemFeature,
+			jni->PackageManager__FEATURE_USB_HOST);
+
+	return LIBUSB_SUCCESS;
+}
+
+struct android_jni_devices
+{
+	struct android_jni_context *jni;
+	jobject iterator;
+};
+
+int android_jni_devices_alloc(struct android_jni_context *jni,
+	struct android_jni_devices **devices)
+{
+	int r;
+	JNIEnv *jni_env;
+	jobject deviceMap, deviceCollection, deviceIterator;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* HashMap<String, UsbDevice> deviceMap = usb_manager.getDeviceList(); */
+	deviceMap = (*jni_env)->CallObjectMethod(jni_env,
+		jni->usb_manager, jni->UsbManager_getDeviceList);
+
+	/* Collection<UsbDevice> deviceCollection = deviceMap.values(); */
+	deviceCollection = (*jni_env)->CallObjectMethod(jni_env,
+		deviceMap, jni->HashMap_values);
+
+	(*jni_env)->DeleteLocalRef(jni_env, deviceMap);
+
+	/* Iterator<UsbDevice> deviceIterator = deviceCollection.iterator(); */
+	deviceIterator = (*jni_env)->CallObjectMethod(jni_env,
+		deviceCollection, jni->Collection_iterator);
+
+	(*jni_env)->DeleteLocalRef(jni_env, deviceCollection);
+
+	*devices = malloc(sizeof(struct android_jni_devices));
+	if (*devices == NULL)
+		return LIBUSB_ERROR_NO_MEM;
+	(*devices)->jni = jni;
+	(*devices)->iterator = (*jni_env)->NewGlobalRef(jni_env, deviceIterator);
+
+	(*jni_env)->DeleteLocalRef(jni_env, deviceIterator);
+
+	return LIBUSB_SUCCESS;
+}
+
+int android_jni_devices_next(struct android_jni_devices *devices,
+	jobject *device, uint8_t *busnum, uint8_t *devaddr)
+{
+	struct android_jni_context *jni = devices->jni;
+	JNIEnv *jni_env;
+	jobject deviceIterator = devices->iterator;
+	jobject local_device;
+	uint16_t deviceid;
+	int r;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* if (!deviceIterator.hasNext()) */
+	if (!(*jni_env)->CallBooleanMethod(jni_env,
+			deviceIterator, jni->Iterator_hasNext))
+
+		return LIBUSB_ERROR_NOT_FOUND;
+
+	/* UsbDevice local_device = deviceIterator.next(); */
+	local_device =
+		(*jni_env)->CallObjectMethod(jni_env,
+			deviceIterator, jni->Iterator_next);
+
+	/* int deviceid = device.getDeviceId(); */
+	deviceid =
+		(*jni_env)->CallIntMethod(jni_env,
+			local_device, jni->UsbDevice_getDeviceId);
+
+	*device = (*jni_env)->NewGlobalRef(jni_env, local_device);
+	(*jni_env)->DeleteLocalRef(jni_env, local_device);
+
+	/* https://android.googlesource.com/platform/system/core/+/master/libusbhost/usbhost.c */
+	*busnum = deviceid / 1000;
+	*devaddr = deviceid % 1000;
+
+	return LIBUSB_SUCCESS;
+}
+
+void android_jni_devices_free(struct android_jni_devices *devices)
+{
+	struct android_jni_context *jni = devices->jni;
+	int r;
+	JNIEnv *jni_env;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r == LIBUSB_SUCCESS)
+		(*jni_env)->DeleteGlobalRef(jni_env, devices->iterator);
+
+	free(devices);
+}
+
+int android_jni_gen_descriptors(struct android_jni_context *jni, jobject device,
+	uint8_t **descriptors, size_t *descriptors_len)
+{
+	int r, num_configs, idx, version, tens, ones, tenths, hundredths;
+	JNIEnv *jni_env;
+	jobject config;
+	jstring jversion;
+	const char *sversion;
+	struct usbi_device_descriptor desc;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* int num_configs = device.getConfigurationCount(); */
+	num_configs =
+		(*jni_env)->CallIntMethod(jni_env,
+			device, jni->UsbDevice_getConfigurationCount);
+
+	*descriptors_len = LIBUSB_DT_DEVICE_SIZE;
+	*descriptors = malloc(*descriptors_len);
+	if (!*descriptors)
+		return LIBUSB_ERROR_NO_MEM;
+
+	/* parse binary coded decimal version */
+	/* String jversion = device.getVersion(); */
+	jversion =
+		(*jni_env)->CallObjectMethod(jni_env,
+			device, jni->UsbDevice_getVersion);
+
+	sversion = (*jni_env)->GetStringUTFChars(jni_env, jversion, NULL);
+	sscanf(sversion, "%d.%d", &ones, &hundredths);
+	(*jni_env)->ReleaseStringUTFChars(jni_env, jversion, sversion);
+
+	/* Android usb version bug was fixed in commit 608ec66d62647f60c3988922fead33fd7e07755e in Pie */
+	if (jni->Build__VERSION__SDK_INT < jni->Build__VERSION_CODES__P) {
+		/* undo pre-pie bug */
+		tenths = hundredths / 16;
+		hundredths = hundredths % 16;
+	} else {
+		/* bug has been fixed */
+		tenths = hundredths / 10;
+		hundredths = hundredths % 10;
+	}
+	tens = ones / 10;
+	ones = ones % 10;
+
+	version = hundredths | (tenths << 4) | (ones << 8) | (tens << 12);
+
+	desc = (struct usbi_device_descriptor){
+		.bLength = LIBUSB_DT_DEVICE_SIZE,
+		.bDescriptorType = LIBUSB_DT_DEVICE,
+		.bcdUSB = libusb_cpu_to_le16(version),
+		.bDeviceClass =
+			 (*jni_env)->CallIntMethod(jni_env,
+				device, jni->UsbDevice_getDeviceClass),
+		.bDeviceSubClass =
+			(*jni_env)->CallIntMethod(jni_env,
+				device, jni->UsbDevice_getDeviceSubclass),
+		.bDeviceProtocol =
+			(*jni_env)->CallIntMethod(jni_env,
+				device, jni->UsbDevice_getDeviceProtocol),
+		.bMaxPacketSize0 = 8,
+		.idVendor = libusb_cpu_to_le16(
+			(*jni_env)->CallIntMethod(jni_env,
+				device, jni->UsbDevice_getVendorId)),
+		.idProduct = libusb_cpu_to_le16(
+			(*jni_env)->CallIntMethod(jni_env,
+				device, jni->UsbDevice_getProductId)),
+		.bcdDevice = 0xFFFF,
+		.iManufacturer = 0,
+		.iProduct = 0,
+		.iSerialNumber = 0,
+		.bNumConfigurations = num_configs,
+	};
+	*(struct usbi_device_descriptor *)*descriptors = desc;
+
+	for (idx = 0; idx < num_configs; idx ++) {
+		/* UsbConfiguration config = device.getConfiguration(i); */
+		config =
+			(*jni_env)->CallObjectMethod(jni_env,
+				device, jni->UsbDevice_getConfiguration, idx);
+
+		r =
+			android_jni_gen_config(jni, jni_env, config,
+				descriptors, descriptors_len);
+
+		(*jni_env)->DeleteLocalRef(jni_env, config);
+
+		if (r != LIBUSB_SUCCESS)
+			return r;
+	}
+
+	return LIBUSB_SUCCESS;
+}
+
+int android_jni_detect_permission(struct android_jni_context *jni, jobject device,
+	int *has_permission)
+{
+	int r;
+	JNIEnv *jni_env;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* boolean has_permission = usb_manager.hasPermission(device); */
+	*has_permission =
+		(*jni_env)->CallBooleanMethod(jni_env,
+			jni->usb_manager, jni->UsbManager_hasPermission, device);
+
+	return LIBUSB_SUCCESS;
+}
+
+int android_jni_request_permission(struct android_jni_context *jni,
+	jobject device)
+{
+	int r;
+	JNIEnv *jni_env;
+	jobject intent, permission_intent;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* Intent intent = new Intent(permission_action); */
+	intent =
+		(*jni_env)->NewObject(jni_env,
+			jni->Intent, jni->Intent_init,
+			jni->permission_action);
+
+	/* PendingIntent permission_intent =
+		PendingIntent.getBroadcast(application_context, 0, intent, 0); */
+	permission_intent =
+		(*jni_env)->CallStaticObjectMethod(jni_env,
+			jni->PendingIntent, jni->PendingIntent__getBroadcast,
+			jni->application_context, 0, intent, 0);
+
+	(*jni_env)->DeleteLocalRef(jni_env, intent);
+
+	/* usb_manager.requestPermission(device, permission_intent); */
+	(*jni_env)->CallVoidMethod(jni_env,
+		jni->usb_manager, jni->UsbManager_requestPermission,
+		device, permission_intent);
+
+	(*jni_env)->DeleteLocalRef(jni_env, permission_intent);
+
+	return LIBUSB_SUCCESS;
+}
+
+int android_jni_connect(struct android_jni_context *jni,
+	jobject device, jobject *connection,
+	int *fd, uint8_t **descriptors, size_t *descriptors_len)
+{
+	int has_permission, r;
+	JNIEnv *jni_env;
+	jobject local_connection;
+	jbyteArray local_descriptors;
+	jbyte * local_descriptors_ptr;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	r = android_jni_detect_permission(jni, device, &has_permission);
+
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	if (!has_permission)
+		return LIBUSB_ERROR_ACCESS;
+
+	/* UsbDeviceConnection local_connection = usb_manager.openDevice(device); */
+	local_connection =
+		(*jni_env)->CallObjectMethod(jni_env,
+			jni->usb_manager, jni->UsbManager_openDevice, device);
+
+	/* fd output */
+	/* fd = local_connection.getFileDescriptor(); */
+	*fd =
+		(*jni_env)->CallIntMethod(jni_env,
+			local_connection, jni->UsbDeviceConnection_getFileDescriptor);
+
+	if (*fd == -1) {
+		(*jni_env)->DeleteLocalRef(jni_env, local_connection);
+		return LIBUSB_ERROR_IO;
+	}
+
+	/* byte[] local_descriptors = local_connection.getRawDescriptors(); */
+	local_descriptors =
+		(*jni_env)->CallObjectMethod(jni_env,
+			local_connection, jni->UsbDeviceConnection_getRawDescriptors);
+
+	/* descriptors buffer output */
+	*descriptors_len = (*jni_env)->GetArrayLength(jni_env, local_descriptors);
+	*descriptors = malloc(*descriptors_len);
+	if (!*descriptors) {
+		(*jni_env)->DeleteLocalRef(jni_env, local_descriptors);
+		(*jni_env)->DeleteLocalRef(jni_env, local_connection);
+		return LIBUSB_ERROR_NO_MEM;
+	}
+
+	/* jni buffer copy */
+	local_descriptors_ptr =
+		(*jni_env)->GetPrimitiveArrayCritical(jni_env, local_descriptors, NULL);
+	memcpy(*descriptors, local_descriptors_ptr, *descriptors_len);
+	(*jni_env)->ReleasePrimitiveArrayCritical(jni_env,
+		local_descriptors, local_descriptors_ptr, JNI_ABORT);
+
+	/* connection jobject output */
+	*connection = (*jni_env)->NewGlobalRef(jni_env, local_connection);
+
+	(*jni_env)->DeleteLocalRef(jni_env, local_descriptors);
+	(*jni_env)->DeleteLocalRef(jni_env, local_connection);
+
+	return LIBUSB_SUCCESS;
+}
+
+int android_jni_disconnect(struct android_jni_context *jni, jobject connection)
+{
+	int r;
+	JNIEnv *jni_env;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* connection.close(); */
+	(*jni_env)->CallVoidMethod(jni_env,
+		connection, jni->UsbDeviceConnection_close);
+
+	(*jni_env)->DeleteGlobalRef(jni_env, connection);
+	return LIBUSB_SUCCESS;
+}
+
+void android_jni_globalunref(struct android_jni_context *jni, jobject object)
+{
+	int r;
+	JNIEnv *jni_env;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return;
+
+	(*jni_env)->DeleteGlobalRef(jni_env, object);
+}
+
+static int android_jni_gen_config(struct android_jni_context *jni,
+	JNIEnv *jni_env, jobject config,
+	uint8_t **descriptors, size_t *descriptors_len)
+{
+	int r;
+	size_t num_ifaces, offset, idx;
+	struct usbi_configuration_descriptor desc;
+
+	/* int num_ifaces = config.getInterfaceCount(); */
+	num_ifaces =
+		(*jni_env)->CallIntMethod(jni_env,
+			config, jni->UsbConfiguration_getInterfaceCount);
+
+	offset = *descriptors_len;
+	*descriptors_len = offset + LIBUSB_DT_CONFIG_SIZE;
+	*descriptors = usbi_reallocf(*descriptors, *descriptors_len);
+	if (!*descriptors)
+		return LIBUSB_ERROR_NO_MEM;
+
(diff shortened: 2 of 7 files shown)


xloem:
I've gotten this code kind of together now.  Is anybody else using android and able to test it?

otavio:
This looks great. Do you have any public code making use of this?

xloem:
I started work on this for https://github.com/orgs/brainflow-dev/projects/1 which is just a casual project for me.  Not all the pieces are in a nice public format yet.

I'm working on a test app specifically for this PR when I find time to, don't know whether I'll get that going.

Thanks for your reply =)

laran002:
I am currently working on a QT wrapper for libusb. My solution is similar to what you have done, using jni from within Qt. I find your solution way more elegant with the direct integration.

I'll try it out and let you know how it turns out

mcuee:
To Android users: please try this out.

xloem:
> I had a quick test on an unrooted Android 10, and it appear to work. However, devices are only detected and prompted for request if plugged into the device before the application starts.

Thanks so much for your test.

That's basically the state of the code right now.  It enumerates devices when usb_init is called, but it doesn't change the set of cached devices afterwards.  Probably some changes to linux_usbfs.c are needed to make that.  I was planning to review the hotplug interface and see if there was something on android that was similar, but I was prioritising making a test app first.  A quick fix would be to just re-enumerate the devices when the user requests.

suriyo-cycron (approved):
@😆😆😆😆

themarpe:
+1 for freely rescannable device list :)

xloem:
The biggest thing in the way of me just implementing the device list rescanning is that I don't have an application to test it.  My brainflow app doesn't use the feature yet, and I'm not aware of other android libusb apps myself, so the first step to implement it is to write a test app.  I started that here: https://github.com/xloem/libusb/tree/android-test but if anybody knows of an existing, easy-to-build test app it would make it easier.

Please also send pull requests to this pull request.  The branch for this pull request is at https://github.com/xloem/libusb/tree/android .

themarpe:
We'll probably start experimenting with this for https://github.com/luxonis/XLink, but not yet in the pipeline. 
I'll make sure to report back any findings and/or fixes we do and open a PR to your branch.

Thanks for creating this implementation @xloem!

laran002:
I also have a test application that hopefully is ready to be published o github soon. In the meantime I'll keep pulling and testing whatever you commit inhouse.

jonngai:
Hi @xloem,

Thanks a ton for the android ndk work. It looks promising. 

Sorry, I know you mentioned that this app is a WIP but I ended up running your example here: https://github.com/xloem/libusb/tree/android-test/android/native-app-example

While running I had some issues getting a list of attached devices. Should the example app be able to make it past this point?
r = libusb_set_option(0, LIBUSB_OPTION_ANDROID_JNIENV, state->activity->env, 0);

I attached the logcat of the crash I’m running into.
[libusblog.txt](https://github.com/libusb/libusb/files/6266973/libusblog.txt)

For what it’s worth, I make it much further with LIBUSB_OPTION_WEAK_AUTHORITY.

Here's output filtered on libusb.so output. I can provide the full log if interested:
```
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: set libusb has weak authority: 0
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: created default context
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: libusb v1.0.24.11604
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: reported kernel version is 4.4.223-g52750b8f2138-ab6846512
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: found usbfs at /dev/bus/usb
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: max iso packet length is (likely) 49152 bytes
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example D/libusblogging: sysfs is available
2021-04-06 12:37:09.712 10174-10238/libusb.android.native_app_example E/libusblogging: opendir devices failed, errno=13
```

It may be related to these AVC:
```
2021-04-06 12:37:09.710 10174-10174/libusb.android.native_app_example W/ive_app_example: type=1400 audit(0.0:482): avc: denied { read } for name="usb" dev="tmpfs" ino=287020 scontext=u:r:untrusted_app_27:s0:c128,c257,c512,c768 tcontext=u:object_r:usb_device:s0 tclass=dir permissive=0 app=libusb.android.native_app_example
2021-04-06 12:37:09.710 10174-10174/libusb.android.native_app_example W/ive_app_example: type=1400 audit(0.0:483): avc: denied { read } for name="/" dev="tmpfs" ino=16956 scontext=u:r:untrusted_app_27:s0:c128,c257,c512,c768 tcontext=u:object_r:device:s0 tclass=dir permissive=0 app=libusb.android.native_app_example
scontext=u:r:untrusted_app_27:s0:c128,c257,c512,c768 tcontext=u:object_r:sysfs:s0 tclass=dir permissive=0 app=libusb.android.native_app_example
```

I'm testing on a pixel 2 running Android 11.

Thanks again for doing this.

xloem:
I've read your post but I am not set up to work on this these days.  It's inspiring you tried my work and I'll return and edit this post if I get set up again.  It's possible the untested prototype example code is passing an invalid jni pointer in; I'm not sure.  I would modify the libusb build to enable debug logging for more information, and use the ndk's gdb debugging support to break on the call and then inspect the stack frame to troubleshoot it.

jonngai:
Thanks, I'll give that a shot when I get a chance. The GetJavaVM call on that jni pointer being passed to libusb is what seems to be failing so it lines up.

jonngai:
Just as an update @xloem: It was the jnienv pointer. I switched away from using the purely native activity in that example and used the JNI interface from a java activity instead. It seems to work now. Thanks!

libusb/os/linux_android_jni.c
```diff
@@ -0,0 +1,1052 @@
...
+
+int android_jni_free(struct android_jni_context *jni)
+{
+	int r;
+	JNIEnv *jni_env;
+
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		free(jni);
+		return r;
```
Youw, line 210:
```suggestion
	if (r != LIBUSB_SUCCESS) {
		free(jni);
		return r;
	}
```
what can I say - I love python too

libusb/os/linux_android_jni.c
```diff
@@ -0,0 +1,1052 @@
...
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->PackageManager__FEATURE_USB_HOST);
+
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->application_context);
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->package_manager);
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->usb_manager);
+
+	(*jni_env)->DeleteGlobalRef(jni_env, jni->permission_action);
+
+	free(jni);
+}
```
Youw, line 224:
```suggestion
	free(jni);
	return LIBUSB_SUCCESS;
}
```

libusb/os/linux_android_jni.c
```diff
@@ -0,0 +1,1052 @@
...
+			(*jni_env)->CallObjectMethod(jni_env,
+				device, jni->UsbDevice_getManufacturerName)),
+		.iProduct = android_jni_gen_string(jni_env,
+			strings, strings_len,
+			(*jni_env)->CallObjectMethod(jni_env,
+				device, jni->UsbDevice_getProductName)),
+		.iSerialNumber = android_jni_gen_string(jni_env,
+			strings, strings_len,
+			(*jni_env)->CallObjectMethod(jni_env,
+				device, jni->UsbDevice_getSerialNumber)),
```
Youw, line 436: While this part is elegantly constructed, it doesn't work with all the devices.

In practice, `.iManufacturer`/`.iProduct`/`.iSerialNumber` is used in combination with `libusb_get_string_descriptor_ascii` call which sends the actual control request to the device (isn't it?) and if those string indexes doesn't happen to match the indexes in the deal descriptor (which is the case for some of my devices), the control request will return a wrong string descriptor.

As an idea, I was thinking if it is possible to cache `strings` and eventually use it by `libusb_get_string_descriptor**` functions, but at the end, the devices might define some custom string descriptors, except those three and `iInterface`, which should still be available when a USB control request is done, and not nesessary be mmentioned in the descriptor(s).

So as an alternative: is there an objection against asking a real descriptor from the device using a control request?

libusb/os/linux_usbfs.c
```diff
@@ -449,21 +486,160 @@ static void op_exit(struct libusb_context *ctx)
...
+		                         &iter_busnum, &iter_devaddr))
+	{
+		if (iter_busnum == busnum && iter_devaddr == devaddr) {
+			priv->android_jni_device = iter_device;
+			break;
+		} else {
+			android_jni_globalunref(cpriv->android_jni, iter_device);
+		}
+	}
+
```
Youw, line 623: I believe `devices` isn't being freed in this function
```suggestion

	android_jni_devices_free(devices);

```

libusb/os/linux_android_jni.c
xloem: thanks =) still have making gpt-neo function locally on my hobby queue (it's a popular python language model)

libusb/os/linux_android_jni.c
xloem: This is unfortunate as the raw descriptor is not at this time available without user-provided permission for the device.  The purpose of this code is to provide for very basic device enumeration without asking for permission yet.

It's been a long time since I've looked at this code, but I recall the intended behavior was to replace the descriptors with the real ones once a connection is opened.  I think your edge case may already be managed, but if it isn't perhaps a simple solution would be to fill the entries with token values and check for those token values later.

xloem: @Youw does the code path that starts with this comment address your concern sufficiently?  The precise raw descriptors are retrieved from the device when a handle is produced.  The function you express concern over requires a handle to be passed.
https://github.com/libusb/libusb/pull/874/files#diff-e1bd2bd456629177a570d54c2da4e03ff291bd6e3b7dd7f1c6cb320bd57341b4R561 I'm on mobile so the link may be imperfect ... it's line 561 of this same file.

libusb/os/linux_android_jni.c
```diff
@@ -0,0 +1,1054 @@
...
+
+	if (*fd == -1) {
+		(*jni_env)->DeleteLocalRef(jni_env, local_connection);
+		return LIBUSB_ERROR_IO;
+	}
+	
+	/* byte[] local_descriptors = local_connection.getRawDescriptors(); */
+	local_descriptors =
+		(*jni_env)->CallObjectMethod(jni_env,
+			local_connection, jni->UsbDeviceConnection_getRawDescriptors);
```
xloem, line 564: @Youw This is where correct descriptors are downloaded from the device.  The purpose of accessing strings before this was to leave people something to use if e.g. #875 happened.

libusb/os/linux_android_jni.c
Youw: I don't believe it is an edge case. Imagine simple device filtering by Vid, Pid and SerialNumber (common case for devices I work with):
```cpp
bool DeviceMatching(libusb_device* device) {
    libusb_device_descriptor desc;
    int err = libusb_get_device_descriptor(device, &desc);
    // handle err

    if (desc.idVendor != required_vendor_id || desc.idProduct != required_product_id)
        return false;

    libusb_device_handle* dev_handle;
    err = libusb_open(device, &dev_handle);
    // handle err

    unsigned char buffer[1024];
    // I'm getting a string other than Serial Number here
    err = libusb_get_string_descriptor_ascii(dev_handle, desc.iSerialNumber, buffer, 1024);
    // handle err

    bool result = strcmp(buffer, required_serial_number);
    libusb_close(device_handle);
    
    return result;
}
```

libusb/os/linux_android_jni.c
Youw: Ah, that's why `parse_config_descriptors` is done twise (once when enumerated, and secondly when actually opened). I've 
been reviewing the implementation closely, and didn't catch this moment.

Hm, so in theory, in my example above, I should be able to `libusb_get_device_descriptor` _after_ I open the device and get the correct result. I'll try.

libusb/os/linux_usbfs.c
```diff
@@ -449,21 +486,162 @@ static void op_exit(struct libusb_context *ctx)
...
+	r = android_jni_connect(cpriv->android_jni, priv->android_jni_device,
+		&hpriv->android_jni_connection, &fd, &descriptors, &descriptors_len);
+	if (r != LIBUSB_SUCCESS) {
+		if (r == LIBUSB_ERROR_ACCESS)
+			android_jni_request_permission(cpriv->android_jni, priv->android_jni_device);
+		return r;
+	}
+
+	free(priv->descriptors);
+	priv->descriptors = descriptors;
```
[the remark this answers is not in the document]
Youw: yeap, I figured, thanks

libusb/os/linux_android_jni.c
xloem: I see the issue now ... as you mentioned you requested the descriptor before opening the device.  Yes, that may fail =(

Would it be better to provide a dummy value that the get_string_descriptors could recognise and use to return an error?

Additionally the backend could detect if permission is already available, and get the correct descriptors if it is, but I feel like that could result in more errors leaking through developers' tests by only partially covering up the issue.  Making a map of generated to real ids has a similar concern.

libusb/os/linux_android_jni.c
```diff
@@ -0,0 +1,1054 @@
...
+	return LIBUSB_SUCCESS;
+}
+
+static int android_jni_gen_string(JNIEnv *jni_env,
+	char **strings, size_t *strings_len, jstring str)
+{
+	size_t str_len, offset, idx;
+	const uint16_t *str_ptr;
+	struct usbi_string_descriptor *desc;
+
```
Youw, line 797: this is possible due to subsequent calls to `android_jni_gen_string` without checks for reallocation failures that may happen a feww lines below
```suggestion

	if (!*strings)
		return LIBUSB_ERROR_NO_MEM;

```

xloem: Sorry, your comment starts mid-sentence to me; what is it that is possible?  That value is returned down around line 820 already for me over here, under the same set of conditions.  I'll add a comment that the function is only for #875 .  I was hoping I could help #875 happen and integrate it, but didn't end up being available to do so.

xloem: I understand now.  Thanks.

xloem:
I believe the `android_jni_gen_string` function and it's usage may have some further memory issues present in it.  This PR may need attention to remove the function or address the allocation inconsistencies around it, until I find time with a networked developer machine and test device.

I made the function hoping #875 would happen, but it hasn't had velocity, and I forgot I had left some of the code open.

libusb/os/linux_android_jni.h
```diff
@@ -0,0 +1,114 @@
...
+ * The intent will have action "libusb.android.USB_PERMISSION".
+ *
+ * For further information on this intent, see:
+ * https://developer.android.com/guide/topics/connectivity/usb/host#permission-d
+ */
+int android_jni_request_permission(struct android_jni_context *jni, jobject device);
+
+/* Connects to a device.
+ *
+ * The connection jobject should be freed with android_jni_globalunref().
```
Youw, line 99:
```suggestion
 * The connection jobject should be freed with android_jni_disconnect().
```

libusb/os/linux_android_jni.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * The descriptors buffer should be freed with free().
+ *
+ * If permission is needed, LIBUSB_ERROR_ACCESS is returned.
+ */
+int android_jni_connect(struct android_jni_context *jni, jobject device, jobject *connection,
+                        int *fd, uint8_t **descriptors, size_t *descriptors_len);
+
+/* Disconnects from a device. */
```
Youw, line 108:
```suggestion
/* Disconnects from a device and frees connection object. */
```

libusb/os/linux_android_jni.c
```diff
@@ -0,0 +1,1061 @@
...
+	r = android_jni_env(jni, &jni_env);
+	if (r != LIBUSB_SUCCESS)
+		return r;
+
+	/* connection.close(); */
+	(*jni_env)->CallVoidMethod(jni_env,
+		connection, jni->UsbDeviceConnection_close);
+
+	return LIBUSB_SUCCESS;
```
Youw, line 604:
```suggestion
	(*jni_env)->DeleteGlobalRef(jni_env, connection);
	return LIBUSB_SUCCESS;
```

libusb/os/linux_android_jni.c
Youw: Yeap, that works.
I've event prepared a good public example: https://github.com/libusb/hidapi/pull/273

libusb/os/linux_android_jni.c
Youw: > could result in more errors leaking through developers' tests by only partially covering up the issue

> Making a map of generated to real ids has a similar concern

I share both conserns.

Youw:
> I believe the unneeded `android_jni_gen_string` function and its usage may have some further memory issues present in it. This PR may need attention to remove the function or address the allocation inconsistencies around it, until I find time for this work with a networked developer machine and test device.
> 
> I made the function hoping #875 would happen, but it hasn't had velocity, and I forgot I had left some of the code open.
> 
> If anybody has time to tidy it, feel free to fork or modify my branch, optionally submit a PR to me or submit a new PR, anything you want.

I think for cases when something doesn't always work by default (and there is an explicit but "non-default" way to make it _always_ work) - let it always fail in cases where it may fail, i.e.: better let `iSerialNumber` and other fields be equal to `0` if those cannot be filled properly before `libusb_open`, and properly document it somewhere.
In that case all developers can catch and fix the issue early.

As for the #875 - I believe it can be easily implemented for Android JNI case even without `android_jni_gen_string`. The implementation would expicitly call `jni->UsbDevice_getSerialNumber` on demand.

Youw:
One more issue regarding this PR I see in general: this implementation calls to `javavm->AttachCurrentThread` on demand, but never calls to `DetachCurrentThread`. This will not make any issues only in one case: if libusb initialisation/device enumeration/creation and descruction is done _only in the main thread_.

If the client application will deside to use more than one thread for device enumeration/creation/destruction (which is not prohibited) - it will lead to the [whole application crash](https://stackoverflow.com/a/26534926/3697939) when exiting a thread where at least one of the JNI functions where called.

I was thinking how this can be worked-around on libusb side, and so far my best idea is to create jni-worker-thread, that is created/descructed with libusb context, and all JNI calls would be made in that thread (which shouldn't be a bottleneck, as it is ony needed for device enumeration/creation/destruction). `AttachCurrentThread`/`DetachCurrentThread` will be called on thread start/exit respectively, and that will ensure thread agnostic usage of the libusb even when used with JNI.

I think it worth adding to the TODO list in the first post.

xloem:
Thanks for catching that.  I tried to update the todo list.

Possibly a clear solution to `DetachCurrentThread` would be to use `GetEnv` to get the jni pointer, then set a flag if there is not one yet and attach to the thread to get one, then have a paired function that checks the flag and calls `DetachCurrentThread` if it is set.  This paired cleanup function could be called at the end of api calls in their own thread.  Note that the JNI is only used in a couple api calls.

Youw:
I thought of other strategies too.

The scenario that bothers me the most in that case:
- there is no current Env, so we AttachCurrentThread (and make a flag that we need to Detach);
- we open a Device and store a NewGlobalRef;
- we exit the function and have to DetachCurrentThread as per flag above, and here are some questions:
      - is it allowed to have a newly created GlobalRef and Detach the thread? Could it cause a crash?
      - is it allowed to use that object from a different (newly Attached thread)?
      - is it allowed to destroy that object from a different (newly Attached thread)?

I have some experience with JNI, but not enough to answer questions above.

And a separate question: how big the performance overhead would be to Attach/Detach for each API call? I understand it is only a few functions, but still. Lets imagine we want to enumerate and then open several devices in a row. I believe with that strategy there would be Attach/Detach once for enumeration and then at least once per each device. Would it be noticable?

xloem:
This states that global references are thread-agnostic: https://www.ibm.com/docs/en/sdk-java-technology/8?topic=collector-overview-jni-object-references
Android's docs recommended using a pthread destructor to detach from the thread: https://developer.android.com/training/articles/perf-jni#threads

Your worker thread idea is great too.

Youw:
Hm, maybe there is a third solution. As suggested by [JNI tips](https://developer.android.com/training/articles/perf-jni#threads) article: if no current Env - AttachCurrentThread and use `pthread_key_create()` to register a thread cleanup-function (where DetechCurrentThread will be called). I'm curios would it work for all kind of threads on Android, e.g. C++ `std::thread`.

Youw:
Based on [this](https://github.com/dropbox/djinni/issues/372) discussion, I tend to believe that `pthread_key_create()` solution should work on Android without issues regardless of what high-level API is used for threads (e.g. std::thread, boost::thread or other environment-specific libraries).

hjelmn:
Please squash this into a minimal number of commits.

libusb/os/linux_usbfs.c
```diff
@@ -449,21 +486,162 @@ static void op_exit(struct libusb_context *ctx)
 static int op_set_option(struct libusb_context *ctx, enum libusb_option option, va_list ap)
 {
 	UNUSED(ctx);
-	UNUSED(ap);
 
 #ifdef __ANDROID__
 	if (option == LIBUSB_OPTION_WEAK_AUTHORITY) {
-		usbi_dbg("set libusb has weak authority");
-		weak_authority = 1;
+		int parameter = va_arg(ap, int);
```
Youw, line 492: it used to be valid using `LIBUSB_OPTION_WEAK_AUTHORITY` without any parameters like `libusb_set_option(nullptr, LIBUSB_OPTION_WEAK_AUTHORITY)` and now we're expecting at least one (int) - is it still going to be valid for old code that doesn't pass any?

xloem: The examples I found passed a trailing parameter of `NULL`, which evaluates to zero.  If not everybody is passing the trailing `NULL` then this extra parameter should be removed.  Is there existing code as you describe?

xloem:
Thanks for all your reviews.  It looks like just a pretty small amount of work is needed to bring this up to community standards.  If there other people waiting for this PR, let me know and I'll try to find time to bring it into shape eventually.  And of course feel free to do so yourself.

libusb/os/linux_usbfs.c
Youw: ~~We can't know for sure - there could be private repositories that are using it.~~
~~I use a general rule for myself: if something is strictly valid, there will be someone who is using it.~~

A quick googling:
1) Perfectly valid usage:
https://github.com/IntelRealSense/RealSenseID/blob/4096914d860117ec2b3ab78fa2380ace5fae13e0/src/Capture/AndroidCapture.cc#L37
2) One of the examples you mentioned:
https://github.com/brainflow-dev/brainflow/blob/311375178ca2e88a9005ffea412f3f6cc4b0de84/src/utils/libftdi_serial.cpp#L264
3) An intuitive example, and personally I though to do exactly this:
https://github.com/IntelRealSense/RealSenseID/blob/4096914d860117ec2b3ab78fa2380ace5fae13e0/src/Capture/AndroidCapture.cc#L37

I'd prefer 3) to be correct one, but I guess to make it properly it is better to have a separate option, and probably out-of-scope of this PR.

xloem: The change I made to WEAK_AUTHORITY's arguments should be removed.

Everything we've discussed has been added to the todo list at the top, and the other items we haven't have been marked optional.

I'm unfortunately not using this PR myself any more, but if others need me to fix it up for merge so they can use it, I can probably eventually find some time to do so.

Youw: > Everything we've discussed has been added to the todo list at the top

This is as much as I'd expect.

I was considering using change in my project, but due to other reasons, I have to stick with wrap_sys_device.
But while my memory is fresh, I'm trying to mark here all the issues I find, just so the community (or maybe even myself, but later) will be able to finish it.

Thanks for being around!

normanr:
Note: For "shell-like" environments (like Termux and AnotherTerm) that launch sub-processes that want access to usb devices, there may not be a local java env in the process trying to access the usb device. They both have solutions/work-arounds that end up with the process (that wants to access the usb device) talking to a daemon over a unix socket to enumerate and open the desired usb device, returning the device's filehandle over the unix socket (using `SCM_RIGHTS`), and finally passing that to `libusb_wrap_sys_device`. It would be nice if that process was internalized to libusb so that pre-compiled applications that link against libusb "just-work".

libusb wouldn't need to necessarily implement the daemon side of the socket. A reference implementation would be nice, and would make it much easier to test, etc. It would be very simple to test on a non-mobile device by just changing the permissions for usb device, and run the daemon so that it has privileges, but the primary process does not. This is a little bit like sshd or wireshark's privilege separation, where the privileged process does very little (only the things privileges are needed for), and the unprivileged process does everything else.

xloem:
What normanr is describing appears to be at https://github.com/green-green-avk/libusb-1.0.23-android-helper-service-patch/blob/master/libusb/os/linux_android_helper.c .  It's unfortunate that alex initialized a new repository rather thank forking libusb to track changes, but it is not too hard to diff them.

adhiti33:
@xloem How do I get consolidated changes for libusb to work on Android ?

Should I just pick git from git from https://github.com/xloem/libusb/tree/android ?
    
I tried only with https://github.com/libusb/libusb/pull/830/commits/ff03b23022782dfaf4bfbac54c6251aea4e5e613 it did not work for me

Malinero:
@adhiti33 you can take a look at https://github.com/monero-project/monero-gui/pull/3581. It's minimalist.

There's basically 3 changes:

- [build libusb for android](https://github.com/monero-project/monero-gui/pull/3581/files#diff-095bbc6c91369f33542d44915e7d7457041583d5d657fcbe0439945d1e19526fR187)
- [Manifest changes](https://github.com/monero-project/monero-gui/pull/3581/commits/c6a681921e56da1e74560bc6d5b29b55e3098df8)
- [JNI_OnLoad](https://github.com/monero-project/monero-gui/pull/3581/files#diff-bc096c261807d048952d917f83c7d04b1f95d65ce1d8289b06ec20e5ae59c938R165) 

It works nicely. The other working approach (passing an opened file descriptor) would have required a loooot more changes to integrate. Many thanks @xloem (and @Youw for hidapi)

adhiti33:
@Malinero @xloem I took complete code from https://github.com/xloem/libusb/tree/android. 
Issue I observe is
1> libusb_get_device_list(ctx, &list); works only with permissive . If permissive is not set there is crash.

2> libusb_open further fails even with permissive with error -3.

Regarding 3 changes mentioned in previous note 1st are are taken care.

For 3rd change, Is it mandatory to pass vm or just set LIBUSB_OPTION_ANDROID_JAVAVM in native like libusb_set_option(NULL, LIBUSB_OPTION_ANDROID_JAVAVM);

xloem:
> For 3rd change, Is it mandatory to pass vm or just set LIBUSB_OPTION_ANDROID_JAVAVM in native like libusb_set_option(NULL, LIBUSB_OPTION_ANDROID_JAVAVM);

You must pass the actual vm or jni pointer for the code in this branch to run.  NULL will disable this android code.  If you're using JNA the special constant com.sun.jna.JNIEnv.CURRENT can be used to pass a JNIEnv pointer to your code.

If you still run into issues, it helps to see the initialization code that calls libusb and the logcat around the crash.

normanr:
> What normanr is describing appears to be at https://github.com/green-green-avk/libusb-1.0.23-android-helper-service-patch/blob/master/libusb/os/linux_android_helper.c . It's unfortunate that alex initialized a new repository rather than forking libusb to track changes, but it is not too hard to diff them.

There is a properly forked branch available at https://github.com/green-green-avk/libusb/tree/v1.0.23-android-libusbmanager (found via https://github.com/termux/termux-api/issues/349#issuecomment-701735271)

> Termux provides access to some of the java usb api via termux-api: https://github.com/termux/termux-api/blob/master/app/src/main/java/com/termux/api/UsbAPI.java .
> 
> So the present protocols differ between the terminal applications, but both already exist and could be added to the main repo with some work.

Correct. The termux ones seem to be very thin and only supports opening a usb device via an fd, where-as the helper-service properly supports enumeration and hot plug. In fact I wrote a bridge to shim the termux api as the helper-service api (https://github.com/normanr/termux-libusb-bridge).

I have a personal preference that the libusb library shouldn't require any java/jni things, and the java/jni support code could be refactored to work as a helper in the same way as termux/helper-service do, but I'm not a libusb maintainer, so please don't take my preference as a requirement for this PR.

adhiti33:
@xloem I am getting crash in below code on Android 10. Checking alternative.

Packagemanager call just before usb service call had succeeded.

int android_jni(JavaVM *javavm, struct android_jni_context **jni) {

	local_usb_mgr =
		(*jni_env)->CallObjectMethod(jni_env,
			(*jni)->application_context, Context_getSystemService,
			Context__USB_SERVICE);

adhiti33:
> @xloem I am getting crash in below code on Android 10. Checking alternative.
> 
> Packagemanager call just before usb service call had succeeded.
> 
> int android_jni(JavaVM *javavm, struct android_jni_context **jni) {
> 
> ```
> local_usb_mgr =
> 	(*jni_env)->CallObjectMethod(jni_env,
> 		(*jni)->application_context, Context_getSystemService,
> 		Context__USB_SERVICE);
> ```

@xloem Issue was due to forking to a separate process. Once I excecuted native command directly there was no issue.

Further I hit security exception at UsbDevice.getSerialNumber(). This is genuine bug: either we need to have try/catch or request user for permission if we get security exception.

		.iSerialNumber = android_jni_gen_string(jni_env,
			strings, strings_len,
			(*jni_env)->CallObjectMethod(jni_env,
				device, jni->UsbDevice_getSerialNumber)),

xloem:
> @xloem Issue was due to forking to a separate process. Once I excecuted native command directly there was no issue.

The JNIEnv/JavaVM pointer is only valid in the process it is given in.  It references local java state and doesn't persist across forks.  There are some spaces that could be probed here, like a spawning a new VM, but opportunities are not amazing given android's custom security model.

> Further I hit security exception at UsbDevice.getSerialNumber(). This is genuine bug: either we need to have try/catch or request user for permission if we get security exception.

Do you have information on how to catch the exception, or logcat output of it being produced?  Note that the current proposal is to remove that chunk of code: it's not used by libusb at this time.

normanr, thanks for your additional digging.

adhiti33:
> > @xloem Issue was due to forking to a separate process. Once I excecuted native command directly there was no issue.
> 
> The JNIEnv/JavaVM pointer is only valid in the process it is given in. It references local java state and doesn't persist across forks. There are some spaces that could be probed here, like a spawning a new VM, but opportunities are not amazing given android's custom security model.
> 
> > Further I hit security exception at UsbDevice.getSerialNumber(). This is genuine bug: either we need to have try/catch or request user for permission if we get security exception.
> 
> Do you have information on how to catch the exception, or logcat output of it being produced? Note that the current proposal is to remove that chunk of code: it's not used by libusb at this time.
> 
> normanr, thanks for your additional digging.

@xloem Here is exception log and fix to handle exception. 


Error logs:
=======

07-04 12:07:32.287  9220  9220 F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
07-04 12:07:32.287  9220  9220 F DEBUG   : Abort message: 'JNI DETECTED ERROR IN APPLICATION: JNI IsSameObject called with pending exception java.lang.SecurityException: User has not given 10120/com.test.printer permission to access device /dev/bus/usb/001/004
07-04 12:07:32.287  9220  9220 F DEBUG   :   at java.lang.Exception android.os.Parcel.createException(int, java.lang.String) (Parcel.java:2071)
07-04 12:07:32.287  9220  9220 F DEBUG   :   at void android.os.Parcel.readException(int, java.lang.String) (Parcel.java:2039)
07-04 12:07:32.287  9220  9220 F DEBUG   :   at void android.os.Parcel.readException() (Parcel.java:1987)
07-04 12:07:32.287  9220  9220 F DEBUG   :   at java.lang.String android.hardware.usb.IUsbSerialReader$Stub$Proxy.getSerial(java.lang.String) (IUsbSerialReader.java:123)
07-04 12:07:32.287  9220  9220 F DEBUG   :   at java.lang.String android.hardware.usb.UsbDevice.getSerialNumber() (UsbDevice.java:143)
07-04 12:07:32.287  9220  9220 F DEBUG   : Caused by: android.os.RemoteException: Remote stack trace:
07-04 12:07:32.287  9220  9220 F DEBUG   :      at com.android.server.usb.UsbUserSettingsManager.checkPermission(UsbUserSettingsManager.java:179)
07-04 12:07:32.287  9220  9220 F DEBUG   :      at com.android.server.usb.UsbSerialReader.getSerial(UsbSerialReader.java:96)
07-04 12:07:32.287  9220  9220 F DEBUG   :      at android.hardware.usb.IUsbSerialReader$Stub.onTransact(IUsbSerialReader.java:84)
07-04 12:07:32.287  9220  9220 F DEBUG   :      at android.os.Binder.execTransactInternal(Binder.java:1021)
07-04 12:07:32.287  9220  9220 F DEBUG   :      at android.os.Binder.execTransact(Binder.java:994)
07-04 12:07:32.287  9220  9220 F DEBUG   :
07-04 12:07:32.287  9220  9220 F DEBUG   : (Throwable with no stack trace)

Fix for exception:
=============

static int android_jni_get_serial_number(struct android_jni_context *jni, JNIEnv *jni_env, jobject device,
										 char **strings, size_t *strings_len) {
	ALOGE("android_jni_get_serial_number called");
	jobject serialNumber = (*jni_env)->CallObjectMethod(jni_env,
														device, jni->UsbDevice_getSerialNumber);
	ALOGE("android_jni_get_serial_number DONE");
	if ((*jni_env)->ExceptionCheck(jni_env)) {
                // customize as needed.
		ALOGE("<<< android_jni_get_serial_number -- Exception while getting serial number");
		(*jni_env)->ExceptionClear(jni_env);
		return LIBUSB_ERROR_NOT_SUPPORTED;
	}

	int serial_number = android_jni_gen_string(jni_env,
						   strings, strings_len,
						   serialNumber);
	(*jni_env)->DeleteLocalRef(jni_env, serialNumber);
	ALOGE("android_jni_get_serial_number END");
	return serial_number;


}

xloem:
@adhiti33 I added some of your code straight to the string_gen method.  Is this branch working for you now?

@Youw I think the memory leaks and other issues with android_jni_gen_string have been dealt with in 1521726e but it is still dead code pending #875.  Still, most libusb devs won't be familiar with jni.  Do you think it should be removed entirely?

adhiti33:
> @adhiti33 I added some of your code straight to the string_gen method. Is this branch working for you now?
> 
> @Youw I think the memory leaks and other issues with android_jni_gen_string have been dealt with in [1521726](https://github.com/libusb/libusb/commit/1521726ecb72d40d0a78d8990f9993abc3ab3287) but it is still dead code pending #875. Still, most libusb devs won't be familiar with jni. Do you think it should be removed entirely?

I  see one issue in new code LIBUSB_SUCCESS is returned even incase of failure. Is that fine  ?

	size_t idx = 0;

	if ((*jni_env)->ExceptionCheck(jni_env)) {
		(*jni_env)->ExceptionClear(jni_env);
		goto end;
	}

end:
	return idx;  // returns LIBUSB_SUCCESS even incase of failure ?

Youw:
> Do you think it should be removed entirely?

I think it better be.

xloem:
> return idx; // returns LIBUSB_SUCCESS even incase of failure ?

I believe this is correct.  This function returns a string index, not an error code.  Youw and I discussed earlier using a zero value here.  An error can be passed out by setting the strings pointer to zero if needed, but it looks like the code will be removed entirely.

> > Do you think it should be removed entirely?
> 
> I think it better be.

Can add to queue.  I can probably paste a link to a commit to revert to #875.

adhiti33:
> > return idx; // returns LIBUSB_SUCCESS even incase of failure ?
> 
> I believe this is correct. This function returns a string index, not an error code. Youw and I discussed earlier using a zero value here. An error can be passed out by setting the strings pointer to zero if needed, but it looks like the code will be removed entirely.

Agreed. My mistake I mistook error handling cases. Hope android_jni_gen_string() utility will comeback soon.

xloem:
@hjelmn I've now squashed this as you requested.
@Youw I believe I've addressed all the items we've discussed.

@adhiti33 when I was pulling the strings gen code out to be used later, I found I had forgotten that return values actually were processed as errors.  sorry for misleading you when changing it.

xloem:
My draft of rebasing this PR on to the latest mainline is at https://github.com/xloem/libusb/tree/android-rebase-2 .  It's unfortunately presently failing with no logcat output on my android device.  Hopefully I can find time to investigate it and update this branch.

xloem:
This has been rebased on latest master.  I briefly tested my changes, but I have not tested the new `no_device_discovery` system, in either the primary branch or in this one.

adisuciu:
Is it possible to add this fix to the PR ? https://github.com/libusb/libusb/issues/31  - or is this intended ? 

I've used this branch succesfully on a cross-platform project - Win/Linux/mac/Android , and I always have to manually rename libusb1.0.so to libusb-1.0.so for it to work.

This PR is great, hope you will merge it soon. Sorry if my post is offtopic.

Du-Jie:
> Is it possible to add this fix to the PR ? #31 - or is this intended ?
> 
> I've used this branch succesfully on a cross-platform project - Win/Linux/mac/Android , and I always have to manually rename libusb1.0.so to libusb-1.0.so for it to work.
> 
> This PR is great, hope you will merge it soon. Sorry if my post is offtopic.

hi，I want to know if you use JAVA API in Android? I am also learning about this problem recently, running libusb on a non-root phone.

xloem:
@Du-Jie The approach of this PR is to make the java api calls from c.  Elsewhere the more traditional solution is to make wrapping java that makes the calls.

Du-Jie:
> @Du-Jie The approach of this PR is to make the java api calls from c. Elsewhere the more traditional solution is to make wrapping java that makes the calls.

So, is there a example demo for testing your commit?  I think your commit is useful, but I don’t know how to use it... Thank you ^_^.

xloem:
"Build a test app" is still on the "todo" in the first post.  I think a couple plausible test apps have been linked in the comments by now, but I've been focusing on the core code changes needed to make it stable and mergeable.

libusb/libusb.h
```diff
@@ -2121,7 +2121,52 @@ enum libusb_option {
...
+	LIBUSB_OPTION_ANDROID_JNIENV = 4,
+
+	/** Like LIBUSB_OPTION_ANDROID_JNIENV, except a thread-agnostic JavaVM*
+	 * pointer is passed instead of a thread-specific JNIEnv* pointer.  If
+	 * you can pass a JavaVM* instead of a JNIEnv* pointer, the it does not
+	 * matter what thread it is passed from.
+	 *
+	 * Setting either option is equivalent.
+	 */
+	LIBUSB_OPTION_ANDROID_JAVAVM = 5,
```
xloem, line 2167: @Du-Jie this is basically what there is for documentation.  You set one of these options before libusb_init to engage the code.  How you get the java pointer passed differs depending on how you are building your app.

ankurverma85:
The current code causes seg fault on devices (Emulator) where the usb host is not present

android_jni_scan_devices returns LIBUSB_ERROR_NOT_SUPPORTED

which cascades down to failure of libusb_init() which makes it jump to error cleanup where it segfaults on usbi_hotplug_exit


```
err_io_exit:
....
usbi_hotplug_exit(_ctx);
```


It seems like an issue with libusb code (improper cleanup) , not necessarily the Pull-Request, but testing in android emulator without usb host is a much common occurence

xloem:
Do you have a backtrace?  The error is interesting because hotplug is presently disabled for android, but the code is left over from `LIBUSB_OPTION_WEAK_AUTHORITY` and I may have made errors migrating it.  Does the crash happen if you use the weak_authority approach instead of the jni approach?

If it's hard to look into it further please let me know so I can add investigating it to the queue here.

ankurverma85:
It basically because the `usbi_hotplug_init` doesnt get called so the list `ctx->hotplug_cbs` isnt initialized and `usbi_hotplug_exit` tries to clean up a non initialized list

Moving `usbi_hotplug_init` before `usbi_backend.init` fixes the issue so when init fails and it goes to `err_io_exit` it tries to cleanup a initialized list

ankurverma85:
There's also a pthread_key_create thats leaking keys on every libusb_init

I believe this should fix that
https://github.com/ankurverma85/libusb/commit/b25bc58902d14accc7299fd7892ccfc25aec0748

ankurverma85:
Forked out usbi_hotplug_init crash fixes as a separate PR
https://github.com/libusb/libusb/pull/989

xloem:
I've tentatively merged the new fixes in.

This branch has conflicts with the main branch, and also depends on #942, which is still open with solutions that have been changing.  It would be good to resolve #942 so this branch can stabilize.

If anybody is available to take over or co-maintain this JNI work, my capacity is waning and the situation is still in flux.

crroush:
@xloem it looks like everything form #942 and #830 have been added to master (which is great!).  I think it might make your changes here OBE or are there things here that are not part of the other fixes?  I also have a CMakeList.txt I made for the project (since android REALLY wants you to use cmake instead of ndk-build), not sure if you want it included in the android/jni directory.

xloem:
I've rebased this PR on the new main tree at https://github.com/xloem/libusb/tree/android-rebase but haven't tested it yet.

@crroush I think the cmake situation might be discussed at https://github.com/libusb/libusb/issues/596 which I think is still pending somebody implementing a resolution or discussing how to design a better plan.  Regarding OBE, I feel pain when I hear my hard and novel jni work described as obsoleted by a regression fix -- it is very hard for me to work and move forward on -- I may not understand right what OBE means.  The PR is a separate feature addition.

- This PR had 942 as a blocker so it's good news it's been merged.  I discovered I left some additional improvements here that could have been brought into that PR and sadly weren't.
- I think 830 is the readme one?  This PR adds a further way to implement android and hasn't had similar documentation added yet.

crroush:
@xloem I wasn't sure what was different on this branch vs the other.  Rereading what I wrote I can why you took it that way. I didn't mean to speak ill of your hard work.  I should have phrased things differently.  

Main thing I was looking to know now that 942 has been completed, what is on this branch that needs testing  and what new capability functionality is here that isn't on the main branch.  I have some cycles where I can contribute / test that is relevant to things I am working on so I was hoping to help in some way.

xloem:
Thanks, there's some info in the first post which I've updated a little.  There are missing features in this branch to add, and merging with #942 is the most pressing issue.  I might find time to address that today, fixing any bugs in the branch I listed, but if not or if you're faster there are a couple possible test apps listed in this thread or we also need one written.

In other news, android just closed https://github.com/android/ndk/issues/1320#issuecomment-956494889 so this PR can likely be made transparent for api >= 31 by autofinding the javavm pointer, which is incredible news if workable.

crroush:
@xloem I looked at your android-rebase branch, I found a few issues.  First [here](https://github.com/xloem/libusb/blob/android-rebase/libusb/core.c#L2235) and [here](https://github.com/xloem/libusb/blob/android-rebase/libusb/core.c#L2291)  ```usbi_mutex_static_lock```  is being called twice, so when you called the ```libusb_init```, we get the lock, then inside of ```libusb_set_option``` it is called again and will forever hang.  I removed the lock/unlock in ```libusb_set_option``` to "fix it", but I haven't traced through all the error conditions to make sure we unlock if there is an error.  

The second spot is [here](https://github.com/xloem/libusb/blob/android-rebase/libusb/os/linux_usbfs.c#L482-L484), since the context is no longer NULL, it returns return an error.  Removing this, everything seems to initialize correctly.  Using the unrooted_android call ```unrooted_usb_description``` in concert with #942 LibUsbAndroidIssue942 application, I get 
```
D/LibUsb: Dev (bus 2, device 2): 0781 - CFCA speed: 5G
D/LibUsb:   Manufacturer:              SANDISK
      Product:                   SDDR-409
D/LibUsb:   Serial Number:             B00213803580
      Configuration:
        wTotalLength:            44
        bNumInterfaces:          1
        bConfigurationValue:     1
        iConfiguration:          0
        bmAttributes:            80h
        MaxPower:                112
        Interface:
          bInterfaceNumber:      0
          bAlternateSetting:     0
          bNumEndpoints:         2
          bInterfaceClass:       8
          bInterfaceSubClass:    6
          bInterfaceProtocol:    80
          iInterface:            0
          Endpoint:
            bEndpointAddress:    81h
            bmAttributes:        02h
            wMaxPacketSize:      1024
            bInterval:           0
            bRefresh:            0
            bSynchAddress:       0
          USB 3.0 Endpoint Companion:
            bMaxBurst:           4
            bmAttributes:        00h
            wBytesPerInterval:   0
          Endpoint:
            bEndpointAddress:    02h
            bmAttributes:        02h
            wMaxPacketSize:      1024
            bInterval:           0
            bRefresh:            0
            bSynchAddress:       0
          USB 3.0 Endpoint Companion:
            bMaxBurst:           4
            bmAttributes:        00h
            wBytesPerInterval:   0
D/LibUsb:   Binary Object Store (BOS):
        wTotalLength:            22
        bNumDeviceCaps:          2
        USB 2.0 Extension Capabilities:
          bDevCapabilityType:    2
          bmAttributes:          00000f0eh
        USB 3.0 Capabilities:
          bDevCapabilityType:    3
          bmAttributes:          00h
          wSpeedSupported:       14
          bFunctionalitySupport: 1
          bU1devExitLat:         10
          bU2devExitLat:         2047
```

xloem:
@crroush it's great you addressed these things! are you able to push the code anywhere or is it easiest if I merge in the things you found by hand? if we can make the branch work with the main tree it will be more useful to people and mergeable.

ideally it should only accept the ctx pointer during libusb_init since changing the pointer after init is unimplemented, this seems small though

I also found there's some legacy option code that ended up in that rebase branch that should be removed (core.c around after line 2364, it sets the options again by accident)

crroush:
@xloem I am happy to push it wherever.  Should I fork the android_rebase or do you want to give me access to that branch or make me a branch to work off of?

xloem:
@crroush if you wanted you could fork either this branch or that one and open a PR on my repo.  i'm also looking at this more, maybe we can compare solutions.

i am thinking of rather than removing the NULL check entirely, which prevents the user from dangerously hopping settings after initialisation, to instead check ctx->list to ensure the call is happening prior to initialisation.

xloem:
crroush and I brought this into sync with master again.  the commit history is unfortunately still a little messy, as is the option handling code.  it's a big relief to have it mergeable with the main tree again.

themarpe:
@xloem quick question regarding the following:

> This PR is usable now, but not all possible features are implemented, such as device relisting or forking away from the process spawned by android. The hope is that the codebase is expandable by others.

Does this mean that only connected devices when libusb is initiialized are scanned and afterwards the returned list doesn't change anymore? What does device relisting entail and what challenges does it present to be able to be implemented?

Our usecase requires this and just want to know any details regarding this, before we begin implementing support for Android

xloem:
Yes, the lack of device relisting is a current limitation on android both in-tree and here.

The only thing in the way is implementing some more JNI and C to add the functionality.  I recall the biggest difficulty, after learning libusb's internals and the android jni calls, is deciding how to approach implementing hotplug, which for full support involves interoperating further with android in ways that don't appear all accessible via straight JNI calls, and may inspire some redesign of the code in this PR.

I have some existing work somewhere if it's helpful to dig it up, but it was based on an earlier tree and wasn't a complete outline yet.

I don't have the capacity to implement device relisting at this time myself alone.  It's possible that could change in the future.

themarpe:
So if we'd only require that we get an updated list of connected devices after `libusb_get_device_list` call, then we have to call `android_jni_scan_devices` again in that function to enable that kind of behavior? (Or only the underlying JNI call to get updated list of devices)

We don't use any hotplug libusb functionality as of right now.

And by the workaround, do you mean creating a new libusb context using `libusb_init` and use that to relist devices (can one connect to the newly found device using previous context?)

xloem:
I think that's basically it, although it's been a while since I looked and something more may end up looking like it makes sense.  Please share work back with the libusb project.

Yes, calling libusb_init is intended to get you a new device list with this code and the main tree code, although it's been a while since I've tried multiple contexts if there's some new bug.  I believe you can move devices over to the original context by using their filehandle as is done with the main tree code now, but also have not tried this myself.

themarpe:
Thanks for all the information @xloem and will share back any good results on topic of device relisting.

crroush:
@xloem one of the issues I am running into is not having a way to notify libusb that the permission is granted. On the java side you typically have a broadcast receiver that handles the requests and enumeration. So once permission is granted the app can continue doing what it needs to do without blocking the main thread.  What I am running to on the pure jni implementation is I am making the request, but it immediately fails, because it doesn't have permissions so then the user has to call it again once the permission is granted. 
so in ```linux_usbfs.c```:
image
Once request permission is sent it returns back and error, and then the user on the java/jni side has to call libusb_open a second time once permission is granted.  

I am not sure the best way here to handle this, but it certainly makes it a bit more cluttered when using libraries that depend on libusb with android.  Since you effectively need to open it twice, so it might make other interoperability with opensource libusb and android a bit more messy.  Or do all the interfacing the normal android way for permissions, then call libusb.  

Curious if you have ran into this and if you have any suggestions on what might be a "cleaner" way to internally handle the "waiting" on the user to respond  to the request.

xloem:
@crroush The reason that interface is not intuitive is because it is very difficult to respond to an intent without writing accessory java.  To prevent the experience entirely would likely mean embedding actual java code in libusb so as to derive from the abstract classes, (although it would be appropriate to open an issue with android to provide for an alternative solution, which would open avenues in many other android ports too).  The simplest solution is to specify [`android.hardware.usb.action.USB_DEVICE_ATTACHED`](https://developer.android.com/guide/topics/connectivity/usb/host#manifest) in your AndroidManifest.xml with an auxiliary xml [device filter](https://developer.android.com/guide/topics/connectivity/usb/host#manifest-example) so that the user provides permission before the device is discovered.  It can be a little confusing to put the xml files together but it makes it smoother sailing after the fact.

Another idea on the queue is to provide for a libusb option to accept an Intent object, so that devices can be naturally created in response to the permission dialog with developer wiring.

I haven't looked into re-rebasing this branch on master yet because hjelmn has further unmerged work changing the options subsystem.

crroush:
@xloem I agree, it would be nice if the NDK had a true usbmanager implementation that could be leveraged from the native side without the need of all the JNI or java.  
In the interim,  I will have to think about the best way to handle thing like the UHD-b2xx series SDRs.  You discover it, then load firmware, then discover it again to load the FPGA.  So you really can't discover the device until after you load the firmware.  So I will probably have to break it up some.  Good news is, I haven't ran into any show stoppers with this PR.

xloem:
That's definitely a project I'd be interested in supporting if it's free code (not today, but in general).

This JNI code just uses android's system enumeration, so the on-connection device filter should trigger if rediscovery works.  Note at present you need to reinitialise libusb to do rediscovery.

themarpe:
Hi @xloem 

I've created a PR against your branch https://github.com/xloem/libusb/pull/3  - it addresses TODO item of freely rescannable list of devices:

> optional: Add support for hotplug / device_list to recognise new devices as they change. (was on the android 8 platfrom, libusb_init is ok，but libusb_get_device_list return 0 -- 1.0.22 is okay, regression in 1.0.23 #584). It may not be reasonable to respond to device insertion without accessory java, but the list of what's connected should be freely rescannable.

Feel free to check it out and discuss any potential things to change/redo there:)

xloem:
themarpe, that's wonderful news.  If anybody has interest in testing or reviewing themarpe's device relisting PR, please do.

iiordanov:
Hi @xloem,

I am following up on a user request to retry implementing USB redirection for aSPICE and Opaque (https://github.com/iiordanov/remote-desktop-clients) on Android. (My previous implementation worked until Android 4.2.2 but stopped after Google made selinux enforcing on Android 5.0.)

I tried to use the code from this PR, but ran into an issue. Since my build system uses autogen.sh / configure script to build for Android, the additional android code is not picked up during the build resulting in missing symbols.

For reference:
make[2]: Entering directory `/home/iordan/software/bVNC/bVNC/remoteClientLib/jni/libs/deps/armeabi-v7a/build/libusb-c71d01c7b74ae26891288c7f0b72d399132d2455/libusb'
  CC       core.lo
  CC       descriptor.lo
  CC       hotplug.lo
  CC       io.lo
  CC       strerror.lo
  CC       sync.lo
  CC       os/threads_posix.lo
  CC       os/events_posix.lo
  CC       os/linux_usbfs.lo
  CC       os/linux_netlink.lo
  CCLD     libusb-1.0.la

Would it be possible to add support for Android builds with autogen.sh / configure?

CraigHutchinson:
First, this code is great, I am using the branch and it worked out the box.

Additional idea for the todo list:
- Automated acquire the JavaVM via use of `extern "C" jint JNI_OnLoad(JavaVM* vm, void* reserved)` entry point which could be included in the shared library builds. If anything would be worth just add this to documentation?
- Can we clarify this todo _"The way the application context is acquired may not work if the application is a service"_ ? I would like to work out how to run the native code as a service so I can test this if I can get pointers as to example/how to run NDK code as a service (sorry still getting used to android!)

djp952:
I wanted to chime in on this since I agree it's absolutely fantastic work and would really like to see it baselined even if it's "imperfect". I was able to get a Kodi PVR Radio addon I authored that can use a USB-attached RTL-SDR device for tuning to actually work on unrooted Android devices with a grand total of **two** lines of new code.  So awesome@xloem!

The only concern I noted was that a MAKE-based build of the code will not include the new **linux_android_jni.c** file. I haphazardly dealt with this here: https://github.com/djp952/libusb-android/commit/c7468f526272a6bd4ee3af3979da2eb696b12ba0

If I may, I would like to request the source material be squashed into a single commit if possible and rebased against at least libusb v1.0.25.  There is a merge commit in the source branch that makes this a tad difficult to work with, but it's not impossible and absolutely worth the effort.

Thank you for this @xloem.

mcuee:
I know next to nothing of Android development. But I am pretty okay with testing libusb under different OS, Android is a major exception which I have not touched. 
Ref: https://github.com/libusb/libusb/issues/1099#issuecomment-1080067439

Assume I want to carry out some simple tests of this branch, what I need to install to build and then carry out test of some generic USB devices (eg: using libusb example programs, or if possible using Termux). I am okay with different OS like Linux, macOS or Windows.

My previous trial with Android using Termux and pyusb was a failure. I believe this pul request will be more mature and hopefully I can get it working for some basic testing.
https://github.com/pyusb/pyusb/issues/285

mcuee:
Good comments from Tormod here.
https://github.com/libusb/libusb/issues/1094#issuecomment-1093845052
> If the people involved in 874 are maintaining their branches, I don't see enough utility in maintaining an intermediate branch, at least for what my own contribution is concerned. And if they are not... who is gonna maintain and develop the new branch? :) > ...
> If https://github.com/libusb/libusb/pull/1026 is done and 874 gets adapted to this and otherwise is finalized in time, I would be happy to see it included in 1.0.27.

CraigHutchinson:
I am experiencing a crash since merge of https://github.com/xloem/libusb/pull/3 (detail posted over on that PR). I'm happy to provide further diagnostic and I am currently investigating/learning what is going on!

themarpe:
@CraigHutchinson thanks for the feedback - we've decided to merge it here so it would be tested & reviewed by others as well and not go stale. I'm up for any discussion / changes to the approach taken in the PR.

CraigHutchinson:
@themarpe Agreed, without the merge I wouldn't have seen this improvement nor would I have been able to aid in finding issues.
I will provide further insights on this issue as I find them here.

CraigHutchinson:
From what I can gather the new merge has uncovered a hidden logic error in the existing code of this PR due to this line 
https://github.com/luxonis/libusb/blob/48e4a321de95fe88aacb31eba9da95e76f99f87a/libusb/os/linux_usbfs.c#L464

The logic of `libusb_exit(libusb_context *ctx)` is to call `usbi_backend.exit(_ctx)` prior to `usbi_hotplug_exit(_ctx)` where the former is freeing the `cpriv->android_jni` and the later call still uses it to release the device instance(s).

CraigHutchinson:
What is the status of this branch? I've been using it to great success. It needs merging to get fixes from the mainline code so that https://github.com/xloem/libusb/pull/3 can be merged in for hotplug support. It feel like its in a good state or does it need something else?

xloem:
All that's left here is for somebody to act as a maintainer of it enough to rebase and test things, keep the relevant changes alive, and engage the people with commit access, until it is merged. I'm sorry that I have not had the capacity to do this myself.

xloem:
> One thing which I think should be fixed pre merge is setting JavaVM pointer. It doesn't seem to work on doing it after libusb_init on a certain context. So it has to be applied to "global context". Is this intentional?

Most of my work here was done based on the old android interface which required setting a global value prior to calling libusb_init . There were changes happening in how libusb option handling was done, introducing for example the global context, and I haven't followed those to see what an appropriate implementation is now (were they all merged?).

> @CraigHutchinson I can bring it up to date sometime this week, not sure if I still have access to the branch though to be able to rebase, @xloem?

You're still listed as a contributer to my repository on github. Anybody else is welcome as well.

CraigHutchinson:
@xloem I'm not sure the best way to go about this but I rebased this change and merged  https://github.com/xloem/libusb/pull/3 Initial results are very positive but: https://github.com/libusb/libusb/pull/1164

mcuee:
I will close this one so that people will go to #1164 for further discussions.
* #1164

PR #875, opened 2021-02-15
Add support for querying serial without opening device

hjmallon:
I will definitely need some help testing this change and much reviewing. 

* Originally I made this change against 1.0.22 so some of it may be out of style/date (please shout and I will make changes).
* The darwin backend is reasonably well used, but the linux and windows backends are not!
* Linux backend requires iconv

Please feel free to ask about anything in here, comments etc.

## UPDATE:

* Linux backend had to go through some porting from my internal version, it builds *but I do not have a machine to test it on at the moment*
* Windows backend was so broken by the 1.0.22 -> master rebase that I have removed it from this PR. Again I do not have a PC to try it on.

### DIFF
--- a/examples/listdevs.c
+++ b/examples/listdevs.c
@@ -45,6 +45,13 @@ static void print_devs(libusb_device **devs)
 			for (j = 1; j < r; j++)
 				printf(".%d", path[j]);
 		}
+
+		unsigned char buf[255];
+		r = libusb_get_serial_string_descriptor_ascii(dev, buf, sizeof(buf));
+		if (r > 0) {
+			printf(" serial: \"%s\"", buf);
+		}
+
 		printf("\n");
 	}
 }
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -539,6 +539,49 @@ int API_EXPORTED libusb_get_device_descriptor(libusb_device *dev,
 }
 
 /** \ingroup libusb_desc
+ * Get the USB device serial for a given device. The string returned is Unicode,
+ * as detailed in the USB specifications.
+ *
+ * \param dev the device
+ * \param data output buffer for descriptor
+ * \param length size of data buffer
+ * \returns number of bytes returned in data, or LIBUSB_ERROR code on failure
+ * \see libusb_get_serial_string_descriptor_ascii()
+ */
+int API_EXPORTED libusb_get_serial_string_descriptor(libusb_device *dev,
+	unsigned char *data, int length)
+{
+	if (usbi_backend.get_serial_string_descriptor) {
+		return usbi_backend.get_serial_string_descriptor(dev, data, length);
+	}
+
+	return LIBUSB_ERROR_NOT_SUPPORTED;
+}
+
+/** \ingroup libusb_desc
+ * Get the USB device serial for a given device in C style ASCII.
+ *
+ * Wrapper around libusb_get_serial_string_descriptor().
+ *
+ * \param dev the device
+ * \param data output buffer for ASCII string descriptor
+ * \param length size of data buffer
+ * \returns number of bytes returned in data, or LIBUSB_ERROR code on failure
+ */
+int API_EXPORTED libusb_get_serial_string_descriptor_ascii(libusb_device *dev,
+	unsigned char *data, int length)
+{
+	union usbi_string_desc_buf str;
+	int r;
+
+	r = libusb_get_serial_string_descriptor(dev, str.buf, sizeof(str.buf));
+	if (r < 0)
+		return r;
+
+	return usbi_string_descriptor_to_ascii(&str, data, length);
+}
+
+/** \ingroup libusb_desc
  * Get the USB configuration descriptor for the currently active configuration.
  * This is a non-blocking function which does not involve any requests being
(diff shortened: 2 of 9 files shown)


dickens:
Why limit this feature to just the serial number?  It would be useful to query any of the three string descriptors that the device descriptor references (manufacturer, product, serial number).

mcuee:
Good point. It is good to have the three string descriptors.

hjmallon:
> Why limit this feature to just the serial number? It would be useful to query any of the three string descriptors that the device descriptor references (manufacturer, product, serial number).

Sure it should be easy enough to add these for linux and macOS. However there is a utility split. Getting the serial number was absolutely required for what I was doing. VID / PID and serial allow for a unique device to be identified without opening it, having the product strings is a nice-to-have.

I have a windows patch which I was using on 1.0.22 branch here https://github.com/hjmallon/libusb/commit/d0f9fd4852dfceaa4f8f55700ed9b74c4e5406dd but too much stuff has changed in that area in between and it doesn't build. I also do not currently have a windows machine to work with.

Youw:
> having the product strings is a nice-to-have

I'd call it quite required for other cases, common for the scenarios I have: enumerate all devices (maybe filtered out by Vid/Pid) and let the user select one explicitly using a UI or smth. In this case using Manufacturer+Product strings (sometime even +SerialNumber) makes it very easy for the user to recognize the device he needs.

> It is good to have the three string descriptors.

A question to libusb folks: do you think it is fine to have three different functions like `libusb_get_serial/manufacturer/product_string_descriptor` or would it be more desirable to have a more general `libusb_get_device_string_descriptor`?

mcuee:
To me it is more desirable to have a more general libusb_get_device_string_descriptor.

hjelmn:
I agree with @mcuee . A more general function should be implemented.

As for testing linux. You can use something like VirtualBox to test in ubuntu or similar.

mcuee:
For the latest Apple Silicon Macs, VirtualBox does not work, but I just found out that UTM works for Linux VMs under my Mac Mini M1 with USB support.

hjelmn:
@mcuee Ah yes. The VirtualBox developers have been very negative (politest way to put it) about supporting anything other than X86_64. It wasn't ever my first choice but it was the only free one I could think of. I use Parallels which adapted to AArch64 very quickly. Thanks for the pointer to another option.

xloem:
My commit removing early string generation from my nonrooted android jni work is at 809a9180d53178dde812ebb7cfec8766bd74d3d5 if ever needed to make something easier and quicker.

mcuee:
> My commit removing early string generation from my nonrooted android jni work is at [809a918](https://github.com/libusb/libusb/commit/809a9180d53178dde812ebb7cfec8766bd74d3d5) if ever needed to make something easier and quicker.

Sorry but I do not understand what you mean here. Are you going to create a new pull request based on this pull request, at least for Linux and Android?

xloem:
No, I expect I might not return to this project, so I wanted to leave my string code somewhere people could find it if needed.  Personally it took a lot of time to learn JNI and I hope the time I spent could save some time of others in the future.

mcuee:
> No, I expect I might not return to this project, so I wanted to leave the string code somewhere people could find it if needed. Personally it took a lot of time to learn JNI and I hope the time I spent could save some time of others in the future.

I see. No problem.

mcuee:
@hjmallon 

sonatique:
@mcuee : Think it would be a really nice feature to have. I probably could help for some platforms (Windows, maybe Linux/OSX)...

mcuee:
> @mcuee : Think it would be a really nice feature to have. I probably could help for some platforms (Windows, maybe Linux/OSX)...

That will be great! Thanks.

sonatique:
@mcuee :  I'll do my best but I have plenty of things do work on first, unfortunately.

mcuee:
> @mcuee : I'll do my best but I have plenty of things do work on first, unfortunately.

No problem. Take your time. Work is always more important. :-)

sonatique:
@mcuee : So to begin with I started from the commit mentionned by @hjmallon _"I have a windows patch which I was using on 1.0.22 branch here https://github.com/hjmallon/libusb/commit/d0f9fd4852dfceaa4f8f55700ed9b74c4e5406dd but too much stuff has changed in that area in between and it doesn't build."_ :  I rebased it on current master, solve compilation and other issues, and I made it build and work (as far as I can say) under Windows.

How shall I go on? Shall I make another PR? It is not clear to me how I can contribute code to this PR here.

Since there have been several requests to add similar support for manufacturer and product strings, my suggestion would be to:
- first having someone review this new state of the code, in order to validate the approach taken by @hjmallon .
- then, if lights are green, I could apply the same idea to the 2 other strings.

What do you think?

mcuee:
@sonatique 

Yes it is okay to create a new PR.

hjmallon:
Thanks @sonatique for taking this over. Sorry I don't have loads of time at the moment to rebase this stuff. If you need anything from me I can dig into the memory banks 🧠 and see if I remember how it worked.

mcuee:
I will close this one and we can continue in #1258.

PR #876, opened 2021-02-15
darwin: Don't drop partial data on timeout

zyp:
A timed out transfer might have been partially filled. Propagate `tpriv->size` to avoid dropping this data.


dickens:
This looks good to me.  @hjelmn?

PR #878, opened 2021-02-22
Minor fix to documentation

TahomaSoft:
Fixed a couple of errors that were revealed by running Doxygen. 

Changed two lines in doxygen.cfg.in to comment out deprecated setting and fixed paper size from invalid a4wide to a4.


tormodvolden:
You'll probably need to squash the commits or rebase, basically only leaving the last commit.

LudovicRousseau:
I fixed in issue by upgrading the config file in https://github.com/libusb/libusb/commit/90273023db02ed747f6d5f3658d43015d14f8d46

Thanks for the notice

TahomaSoft:
Thank you, and you are welcome!

tormodvolden:
The suggested change of PAPER_TYPE is still applicable. The current value of "a4wide" is not listed as a valid value.

LudovicRousseau:
@tormodvolden  exact. Fixed in https://github.com/libusb/libusb/commit/bda2344f5e7d460decd3968492faa7a5fb2797cd

PR #893, opened 2021-03-21
macOS: implement device capture to force unload of KEXT drivers

osy:
It's a well known issue on macOS that you cannot use libusb to attach to an interface already claimed by an KEXT for exclusive access. The common workaround is to do `sudo kextunload` or inject a codeless KEXT. Both require admin permissions, and does not work for all devices on latest macOS. Additionally, SIP has to be disabled to inject a codeless KEXT, making it a pain to work with USB devices.

This patch uses the `kUSBReEnumerateCaptureDeviceBit` introduced back in macOS 10.10 which will tell the kernel to unload all other drivers for you without needing any dirty hacks.

Unfortunately, a side effect is that it will reset (simulate unplug-replug) the device to do this and so I didn't want to make it the default behaviour of `libusb_open` as that may, for example, cause all USB devices to be reset if the application decides to read the product string of every device attached. I can't find a good way to change this without touching the API specs, so I introduced a new option `LIBUSB_OPTION_FORCE_CAPTURE_OPEN` that lets you toggle the "capture" behaviour.

Also, the way `libusb_close` works makes it a bit awkward to implement the "release" of the captured device because just closing the handle does not tell the KEXTs to start matching again and calling `USBDeviceReEnumerate` with `kUSBReEnumerateReleaseDeviceBit` will also cause the device to be reset which then makes the next call of `libusb_open` fetch a stale cached object and fail. However, calling `darwin_reset_device` to address the stale cache issue will cause the device to be opened again, which causes another "capture" and then the OS will never get the device back. So the workaround of this is as follows:

1. When `darwin_close` is called, if the device is captured, then `darwin_reset_device` will first be called.
2. `darwin_reset_device_capture` will call `USBDeviceReEnumerate` and release the capture.
4. `darwin_restore_state` calls `darwin_open` indicating that capture should NOT be used.
5. `darwin_open` succeeds and the device is re-opened without the capture bit.
6. Call returns to `darwin_close` who calls `USBDeviceClose` on the non-captured device handle.


osy:
Found an error in the patch, closing to address it.

Cryptiiiic:
This severely interferes with some of my projects so would appreciate if this pull doesn't get abandoned.

hjelmn:
If the author can't continue working on this I will pick it up. I was not aware of this feature but would like to make sure we can make use of it.

osy:
@hjelmn is there a way to contact you in private?

hjelmn:
@osy my moniker at mac dot com

Cryptiiiic:
@osy follow up just to check in after 7 days, this is still being looked into right?

osy:
Yes

mcuee:
Just create a new issue so that this will not get forgotten. #906

PR #896, opened 2021-03-23
Fix #895 - WinUSB DeviceInterfaceGUIDs handling of null-termination

### LINKED ISSUE #895: Should accept DeviceInterfaceGUID  REG_SZ with Unicode null-terminator 0x0000 i.e. 2-bytes
https://github.com/libusb/libusb/blob/bda2344f5e7d460decd3968492faa7a5fb2797cd/libusb/os/windows_winusb.c#L1589

A device implementing a DeviceInterfaceGUID  field which reports 40-bytes (19+null-terminated unicode characters) is reported as having "unexpected type/size of DeviceInterfaceGUID for '%s'". 

The current code only allows the worst case where only a single-byte null is provided. As per the documentation the user should cater for incorrect termination however the current logic is in error when handling the correct/expected null termination exists!

For example a well formed unicode string:
`using t = decltype(u"DeviceInterfaceGUID");`

where t becomes 40-bytes as each character including the null consumes 2-bytes:
`char16_t[20]`

The current logic only accepts an incorrectly terminated 39-byte string but should also accept 40-bytes as a valid size.

CraigHutchinson:
This patch fixes the REG_SZ logic which only accepts malformed 'partially' single-byte null-terminated unicode strings. It extends to support 2-byte unicode null (0x0000) and as per the docs also caters for missing null-termination.

Fixes #895

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1585,42 +1585,59 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 					s = pRegQueryValueExA(key, "DeviceInterfaceGUID", NULL, &reg_type,
 						(LPBYTE)guid_string, &size);
 				pRegCloseKey(key);
-				if ((s == ERROR_SUCCESS) &&
-				    (((reg_type == REG_SZ) && (size >= (sizeof(guid_string) - sizeof(uint16_t))) && (size <= sizeof(guid_string))) ||
-				     ((reg_type == REG_MULTI_SZ) && (size >= sizeof(guid_string)) && (size <= (sizeof(guid_string) - sizeof(uint16_t)))))) {
-					if (nb_guids == guid_size) {
-						new_guid_list = realloc((void *)guid_list, (guid_size + GUID_SIZE_STEP) * sizeof(void *));
-						if (new_guid_list == NULL) {
-							usbi_err(ctx, "failed to realloc guid list");
-							LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
-						}
-						guid_list = new_guid_list;
-						guid_size += GUID_SIZE_STEP;
-					}
-					if_guid = malloc(sizeof(*if_guid));
-					if (if_guid == NULL) {
-						usbi_err(ctx, "failed to alloc if_guid");
-						LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
-					}
-					if (!string_to_guid(guid_string, if_guid)) {
-						usbi_warn(ctx, "device '%s' has malformed DeviceInterfaceGUID string '%s', skipping", dev_id, guid_string);
-						free(if_guid);
-					} else {
-						// Check if we've already seen this GUID
-						for (j = EXT_PASS; j < nb_guids; j++) {
-							if (memcmp(guid_list[j], if_guid, sizeof(*if_guid)) == 0)
-								break;
-						}
-						if (j == nb_guids) {
-							usbi_dbg("extra GUID: %s", guid_string);
-							guid_list[nb_guids++] = if_guid;
-						} else {
-							// Duplicate, ignore
-							free(if_guid);
-						}
-					}
-				} else if (s == ERROR_SUCCESS) {
-					usbi_warn(ctx, "unexpected type/size of DeviceInterfaceGUID for '%s'", dev_id);
+				if ( s == ERROR_SUCCESS)
+				{
+                    /// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks
+                    /// - REG_SZ: "string may not have been stored with the proper terminating null characters."
+                    /// - As REG_SZ but also "Note that REG_MULTI_SZ strings could have two terminating null characters"
+                    unsigned int minREG_SZ = sizeof(guid_string) - sizeof(uint16_t);
+                    unsigned int maxREG_SZ = sizeof(guid_string);   
+                    
+                    /// As the reading code does not read enough for REG_MULTI_SZ we can handle it as REG_SZ
+				    if ( ((reg_type == REG_SZ) || (reg_type == REG_MULTI_SZ)) 
+                        && (size >= minREG_SZ) 
+                        && (size <= maxREG_SZ) ) 
+                    {
+                        if (nb_guids == guid_size) {
+                            new_guid_list = realloc((void *)guid_list, (guid_size + GUID_SIZE_STEP) * sizeof(void *));
+                            if (new_guid_list == NULL) {
+                                usbi_err(ctx, "failed to realloc guid list");
+                                LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
+                            }
+                            guid_list = new_guid_list;
+                            guid_size += GUID_SIZE_STEP;
+                        }
+                        if_guid = malloc(sizeof(*if_guid));
+                        if (if_guid == NULL) {
+                            usbi_err(ctx, "failed to alloc if_guid");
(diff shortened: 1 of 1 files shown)


mcuee:
This seems to be good to me.

libusb/os/windows_winusb.c
```diff
@@ -1586,8 +1586,8 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 						(LPBYTE)guid_string, &size);
 				pRegCloseKey(key);
 				if ((s == ERROR_SUCCESS) &&
-				    (((reg_type == REG_SZ) && (size == (sizeof(guid_string) - sizeof(char)))) ||
-				     ((reg_type == REG_MULTI_SZ) && (size == sizeof(guid_string))))) {
+				    (((reg_type == REG_SZ) && (size >= (sizeof(guid_string) - sizeof(uint16_t))) && (size <= sizeof(guid_string))) ||
+				     ((reg_type == REG_MULTI_SZ) && (size >= sizeof(guid_string)) && (size <= (sizeof(guid_string) - sizeof(uint16_t)))))) {
```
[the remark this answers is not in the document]
CraigHutchinson: Re parenthesis, they aren't 'necessary' as such but C-MISRA mandates them so its a habit for lots of developers as removes the often ambiguity and the odd edge cases of precedence.

Also re the  size will never be > sizeof(guid_string) which is true but if the prior code were changes then this if statement woudl not longer be correctly formed. The if-logic should be 'atomic' in its nature imho and its upto the optimising compiler to remove the redundant test as per your observation.

However, in respect to understandability of that code, its mostly not nice but I didn't want to do a dramatic refactor for a simple fix like this 👍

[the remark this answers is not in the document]
CraigHutchinson: I am going to do a minor cleanup anyhow as its nasty and I don't think the logic is right albeit was written a while back!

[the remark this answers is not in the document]
tormodvolden: I am side-tracking a bit but find it interesting that C-MISRA mandates "excessive" use of parentheses. Do you have a reference? I tried searching, but I think MISRA itself is pay-walled. IMO, `a == b &&` is not ambiguous.

OK, so I get the "atomic" point and I understand you are extending the original one-value size check to an interval check. I see the pattern `size >= a && size <= b` and I am tempted to translate this into: size must be in the interval [a, b]. However on line 1590 b is smaller than a...

libusb/os/windows_winusb.c
```diff
@@ -1585,42 +1585,59 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
+				if ( s == ERROR_SUCCESS)
+				{
+                    /// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks
+                    /// - REG_SZ: "string may not have been stored with the proper terminating null characters."
+                    /// - As REG_SZ but also "Note that REG_MULTI_SZ strings could have two terminating null characters"
+                    unsigned int minREG_SZ = sizeof(guid_string) - sizeof(uint16_t);
+                    unsigned int maxREG_SZ = sizeof(guid_string);   
+                    
+                    /// As the reading code does not read enough for REG_MULTI_SZ we can handle it as REG_SZ
+				    if ( ((reg_type == REG_SZ) || (reg_type == REG_MULTI_SZ))
```
tormodvolden, line 1597: Indentation looks broken (tabs).

tormodvolden:
Since you have nullified the first commit with the second, you should squash them and rework commit message in the end. But there is already the whitespace issues to take of.

Thanks for the link. It indeed says "_However, this does not mean that parentheses should be gratuitously added around every operation._"

CraigHutchinson (commented):
I am working in the web-browser atm, does this repo no have 'squash' on merge option? I see the whitespace issue but the web editor isn't so simple to work out what the correct indentation style was/is!

>  It indeed says "_However, this does not mean that parentheses should be gratuitously added around every operation._"

There are rules which enforce when the parenthesis should not be used also to sort this one out in the world of MISRA :) 
for example `return( a == b )` is a perfect example of the exception as it makes `return` look like a function call

libusb/os/windows_winusb.c
```diff
@@ -1585,42 +1585,59 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
+                    /// - REG_SZ: "string may not have been stored with the proper terminating null characters."
+                    /// - As REG_SZ but also "Note that REG_MULTI_SZ strings could have two terminating null characters"
+                    unsigned int minREG_SZ = sizeof(guid_string) - sizeof(uint16_t);
+                    unsigned int maxREG_SZ = sizeof(guid_string);   
+                    
+                    /// As the reading code does not read enough for REG_MULTI_SZ we can handle it as REG_SZ
+				    if ( ((reg_type == REG_SZ) || (reg_type == REG_MULTI_SZ)) 
+                        && (size >= minREG_SZ) 
+                        && (size <= maxREG_SZ) ) 
+                    {
```
tormodvolden, line 1600: Did you change anything below here except whitespace and the todo comment? It is hard to tell in the web interface.

CraigHutchinson: I pushed with the indentation fixed which is easier to see. The indentation wasn't correct in the orignal code so to see just what I changed you can check 'ignore whitespace changes' for the diff.
1. Simplified REG_SZ checking logic
2. Removed duplicated check for  `s == ERROR_SUCCESS`

CraigHutchinson: This is the diff I used to check the changes: https://github.com/libusb/libusb/pull/896/files?diff=split&w=1

tormodvolden: Ah thanks, I didn't know I could filter whitespace changes so easily here.

Alternatively, deal with other errors up front, and avoid any indentation changes:
```
+                               if (s != ERROR_SUCCESS) {
+                                       usbi_warn(ctx, "unexpected error from pRegQueryValueExA for '%s'", dev_id);
+                                       break;
+                               }
```
 
Now, to understand the whole change better, what is it what we get from pRegQueryValueExA()? Isn't it a string with the 38 characters (braces or parenthesis) format: {12345678-1234-1234-1234-123456789abc} ?

Written out (not binary encoded) there are 32 hex digits in a GUID plus the dashes and braces, so 38 string characters. I reread your bug report and it talks about 19 unicode characters? Does that make sense at all? If the GUID string was encoded in unicode it would be 76 bytes... I think unicode is not relevant here, and the double zero at the end would not be a unicode \0 but the double zero termination in a MS REG_MULTI_SZ.

38 bytes if it is a GUID without trailing zero and 39 if it is a GUID with one trailing zero (the only case supported up to now?) and 40 if it is a GUID with double zeroes (from REG_MULTI_SZ)? Should we accept 40 for a REG_SZ? Shouldn't we also check that these extra bytes are effectively zeroes? Or is it redundant because of string_to_guid()?

hjelmn: What is the status of this? Should it be taken as-is or are additional changes needed?

tormodvolden: If my above comment is correct, and the string_to_guid() takes the final check, I suggest the simpler change in pull request #927. I left Craig as author, but his e-mail address for sign-off should probably be changed, I only found the github-scrambled one. The difference is much a question of taste, but at least replacing sizeof(uint16_t) with 2*sizeof(char) is important for readability and correctness.

mcuee:
Close this in favor of https://github.com/libusb/libusb/pull/927

mcuee:
Ref: https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa

PR #898, opened 2021-03-25
make libusb.h usable with Watcom

sezero:


sezero:
Closing in favor of #904 which drops the problematic `ZERO_SIZED_ARRAY` change.

PR #899, opened 2021-03-26
WinUSB Compare ClassGuid in check for existing cached device

### LINKED ISSUE #897: windows_winusb not re-enumerating devices while ClassGUID has changes
The Dfu-Util project causes a reset of a device from DFU-Runtime to DFU-Mode. On windows the WinUSB device `libusb_get_device_list()` backend  re-uses the existing device reporting "found existing device for session [62]" descriptors which causes a bug in that changed/updated device descriptors are not accessible via the LibUsb api: [https://sourceforge.net/p/dfu-util/tickets/110](https://sourceforge.net/p/dfu-util/tickets/110)

# Current WinUSB implements only the following checks
1.  `DevInst`  is used to match session_id
https://github.com/libusb/libusb/blob/bda2344f5e7d460decd3968492faa7a5fb2797cd/libusb/os/windows_winusb.c#L1672
2. `priv->dev_id` path is used to match the device
https://github.com/libusb/libusb/blob/bda2344f5e7d460decd3968492faa7a5fb2797cd/libusb/os/windows_winusb.c#L1691

# The issue
When the device is re-enumerated it is the `ClassGuid` that actually changes and should also be checked

I am happy to try and make a patch if somebody can help provide some advice or best approach as my assumption here may not be correct with limited knowledge i.e:
-  I am not sure if the `ClassGuid` is cached in priv for the device at present to perform the extra comparison in the same manner as at L1691 already?

CraigHutchinson:
Fixed #897

### DIFF
--- a/libusb/os/windows_common.h
+++ b/libusb/os/windows_common.h
@@ -259,7 +259,7 @@ struct winusb_device_priv {
 	} usb_interface[USB_MAXINTERFACES];
 	struct hid_device_priv *hid;
 	PUSB_CONFIGURATION_DESCRIPTOR *config_descriptor; // list of pointers to the cached config descriptors
-+	GUID class_guid; // ClassGUID is cheched for change during re-enumeration @note cached descriptors invalidated if the class changes
+	GUID class_guid; // ClassGUID is cheched for change during re-enumeration @note cached descriptors invalidated if the class changes
 };
 
 struct usbdk_device_handle_priv {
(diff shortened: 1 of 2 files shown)


libusb/os/windows_common.h
```diff
@@ -259,6 +259,7 @@ struct winusb_device_priv {
 	} usb_interface[USB_MAXINTERFACES];
 	struct hid_device_priv *hid;
 	PUSB_CONFIGURATION_DESCRIPTOR *config_descriptor; // list of pointers to the cached config descriptors
++	GUID class_guid; // ClassGUID is cheched for change during re-enumeration @note cached descriptors invalidated if the class changes
```
sezero, line 262: stray '+' here

CraigHutchinson: Thanks

mcuee:
Close this and please refer to #900.

PR #900, opened 2021-03-26
winusb: Check for ClassGuid changes on re-enumeration

tormodvolden:
Previously only the dev_id (VID/PID/serial) was checked, but if the
bcdDevice has changed, Windows will have re-enumerated the device and
possibly attributed another ClassGuid to it.

Fixes issue #897.

Thanks to Craig Hutchinson for reporting, suggestion and testing.


mcuee:
This seems to be good to me.

mcuee:
Looks good to me.

PR #904, opened 2021-04-05
make libusb.h usable with Watcom

sezero:


mcuee:
I do not think we will support Watcom. But you can always have the private version for Watcom.

mcuee:
I tend to believe you need many more changes to use Watcom to build libusb from source codes. So I believe this is just to help Watcom user to import libusb-1.0.dll and use libusb with Watcom project, right? 

We do not want give people the false assumption that libusb project supports Watcom.

sezero:
>  So I believe this is just to help Watcom user to import libusb-1.0.dll and use libusb with Watcom project, right?

Yes.

mcuee:
I see. In this case, we will not be able to take the patch and create a false impression that we support Watcom. Watcom can always add a local patch.

mcuee:
Since Nathan has merged the pull request, I am also okay with that.

PR #907, opened 2021-04-13
autogen: Support being called from external build dir

3v1n0:
Building libusb from another build directory doesn't properly work right
now, as calling autogen.sh from there won't work.

An example is when using jhbuild to build it.

So always use absolute paths to call configure and bootstrap scripts

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -2,7 +2,9 @@
 
 set -e
 
-./bootstrap.sh
+srcdir="$(dirname "$0")"
+
+"$srcdir"/bootstrap.sh
 if [ -z "$NOCONFIGURE" ]; then
-    exec ./configure --enable-examples-build --enable-tests-build "$@"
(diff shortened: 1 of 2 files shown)


LudovicRousseau:
I understand you may need to call `configure` from a build directory.
But why do you also need to call `autogen.sh` from another directory?

Call `./bootstrap.sh` to generate the scripts, Makefile.in, etc. Do not use `autogen.sh` in your case.
And then call `configure` from another directory.

3v1n0:
Well, that's what jhbuild does by default, it can be configured to build inside the builddir, but that's unconvenient.

So in general, all the `autogen.sh` definitions support being called directly from the build dir not having to deal with it.
And assuming the change isn't destructive, I don't see why not having it in :)

LudovicRousseau:
Merged in https://github.com/libusb/libusb/commit/1001cb5558cf6679af7bce3114bba1d3bb7b6f7f
Thanks

PR #914, opened 2021-05-17
doc: Add since version tag to libusb_wrap_sys_device()

Youw:


mcuee:
This looks good to me.

tormodvolden:
BTW, a related question: Why is some of the API documentation in libusb.h and some in core.c? Should it ideally be in the former?

PR #916, opened 2021-05-21
doc: correction of LIBUSB_OPTION_WEAK_AUTHORITY availability

Youw:
Even though its implementation is in linux/usbfs,
in fact it is under `ifdef` and only used by Android builds.


mcuee:
This looks good to me.

PR #935, opened 2021-06-15
New NO_ENUMERATION option to replace WEAK_AUTHORITY option

tormodvolden:
All that the old option was doing, was skipping the device enumeration.
Since this has a much broader usage scope than "weak authority", give
the option a name reflecting exactly what it does.

For now, keep the old option name working as an alias.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1205,8 +1205,10 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
  * handle for the underlying device. The handle allows you to use libusb to
  * perform I/O on the device in question.
  *
- * Must call libusb_set_option(NULL, LIBUSB_OPTION_WEAK_AUTHORITY)
- * before libusb_init if don't have authority to access the usb device directly.
+ * Call libusb_set_option(NULL, LIBUSB_OPTION_NO_DEVICE_DISCOVERY) before
+ * libusb_init() if you want to skip enumeration of USB devices. In particular,
+ * this might be needed on Android if you don't have authority to access USB
+ * devices in general.
  *
  * On Linux, the system device handle must be a valid file descriptor opened
  * on the device node.
@@ -2219,6 +2221,7 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
 
 		/* Handle all backend-specific options here */
 	case LIBUSB_OPTION_USE_USBDK:
+	case LIBUSB_OPTION_NO_DEVICE_DISCOVERY:
 	case LIBUSB_OPTION_WEAK_AUTHORITY:
 		if (usbi_backend.set_option)
 			return usbi_backend.set_option(ctx, option, ap);
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -2098,20 +2098,30 @@ enum libusb_option {
 	 */
 	LIBUSB_OPTION_USE_USBDK = 1,
 
-	/** Flag that libusb has weak authority.
+	/** Do not scan for devices
 	 *
 	 * With this option set, libusb will skip scanning devices in
-	 * libusb_init().
+	 * libusb_init(). Must be set before calling libusb_init().
 	 *
-	 * This option should be set before calling libusb_init(), otherwise
-	 * libusb_init() might fail. The option is typically needed on Android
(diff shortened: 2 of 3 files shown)


tormodvolden:
It would be good if those who introduced the original WEAK_AUTHORITY could comment on this, and in particular its deprecation. Maybe I missed something? Any plans for extending WEAK_AUTHORITY to do more than NO_ENUMERATION?

mcuee:
Commit https://github.com/libusb/libusb/commit/89b810ec9bde2689b06fdd2b23ccbb3f926d7575 adds the option LIBUSB_OPTION_WEAK_AUTHORITY to Android.

Original pull-request
https://github.com/libusb/libusb/pull/760

mcuee:
@chrisju @Peter-St @xolem Please review this if you got the time. Thanks.

Peter-St:
For my part I approve with tormodvolden. If there were no negations, we could merge this PR too.

xloem:
It looks like most of the changes happened in #925 .  It's notable that there could be race conditions since a static flag was used here, but that was already the case.  #874 will likely be further delay to be changed for these but the improved code organisation is good.

tormodvolden:
Yes, it prevents the hotplug event monitoring from starting. I think this makes sense, since the hotplug add event handling depends on enumeration of the device, AFAICS. We could also call the new option NO_DEVICE_DISCOVERY - it would be less specific to the mechanism of enumeration, clearly cover hotplug deactivation, and maybe be more explanatory for the user?

LudovicRousseau:
I vote for `NO_DEVICE_DISCOVERY`.
I suggest to add in the documentation that this option will also disable the hotplug mechanism.

One question @tormodvolden: why do you change the value of `LIBUSB_OPTION_WEAK_AUTHORITY` from 2 to 3 instead of having `LIBUSB_OPTION_WEAK_AUTHORITY` and `LIBUSB_OPTION_NO_ENUMERATION` both defined as 2?

tormodvolden:
@LudovicRousseau I was thinking in case we wanted to display a deprecation message for WEAK_AUTHORITY later.

tormodvolden:
As soon as there is agreement on the naming, I will change that too. Would you like me to squash and repush in the end, or do you do that anyway with the NANO bump process?

mcuee:
I am fine with NO_DEVICE_DISCOVERY as well.

Peter-St:
Yep,
NO_DEVICE_DISCOVERY is preferable to LIBUSB_OPTION_NO_ENUMERATION.

tormodvolden:
I see the potential issue if someone is using 1.0.25 and WEAK_AUTHORITY, the application will set option 3. If the same binary is later run-time linked against a newer libusb where we use 3 for something different after having dropped WEAK_AUTHORITY, unexpected things will happen. I don't know if we guarantee back-compatibility for minor releases? Or we will have to keep number 3 reserved until next major release.

Alternatively we use 2 for both now, and we won't be able to give a run-time warning. A build-time warning would be better anyway but I don't see how to do that.

Or we do it the brutal way and drop it now, so compile will fail. The option hasn't been in much use yet, only on the experimental(?) Android platform. If an application must be able to compile smoothly against both 1.0.24 and 1.0.25 the preprocessor will have to check libusb version and pick name accordingly.

xloem:
If this is merged with additional option 3, I'll try to make time to merge with #874 (which currently uses options 3 and 4) so anybody using my other pr will discover the new numbering sooner rather than later.

xloem:
I reviewed this code more while thinking about #942 and noticed that, although the docs of this option have been generalised, it still only exists for the linux backends.  Probably good to document that unless it changes.

chrisju:
LIBUSB_OPTION_WEAK_AUTHORITY is used when there is no permission to directly access the usb port, such as cannot scan the usb port and hot plug detect.

tormodvolden:
I just wonder if it would be better to leave the deprecated LIBUSB_OPTION_WEAK_AUTHORITY at value 2 as well. We wouldn't be able to issue a deprecation warning then, but it prevents some possible future issues with binary compatibility and technical debt. And as soon as we drop LIBUSB_OPTION_WEAK_AUTHORITY altogether, the word should be well out already, and a broken compile can be fixed promptly. Comments?

Issue #912, opened 2021-04-29
Segfault with libusbk driver


kulkin-fish:
analysis-dbg.exe caused an Access Violation at location 000000006B608500 in module libusb-1.0.dll Reading from location 000000000000054C.

AddrPC           Params
000000006B608500 000000002568F930 00007FFD00000004 000000006B62FA50  libusb-1.0.dll!usbi_signal_transfer_completion  [D:/msys64/tmp/libusb-1.0.24/libusb/io.c @ 1734]
  1732: 
  1733: if (dev_handle) {
> 1734: struct libusb_context *ctx = HANDLE_CTX(dev_handle);
  1735: unsigned int event_flags;
  1736: 
000000006B60BD99 00000000222EDC80 0000000000000000 00000000222CC7D0  libusb-1.0.dll!windows_iocp_thread  [D:/msys64/tmp/libusb-1.0.24/libusb/os/windows_common.c @ 442]
   440: usbi_dbg("transfer %p completed, length %lu",
   441:  USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer), ULONG_CAST(num_bytes));
>  442: usbi_signal_transfer_completion(itransfer);
   443: }
   444: 
00007FFDD06BAF5A 00007FFDD07106D0 00000000222CC7D0 0000000000000000  msvcrt.dll!_beginthreadex
00007FFDD06BB02C 0000000000000000 0000000000000000 0000000000000000  msvcrt.dll!_endthreadex
00007FFDD0737034 0000000000000000 0000000000000000 0000000000000000  KERNEL32.DLL!BaseThreadInitThunk
00007FFDD0922651 0000000000000000 0000000000000000 0000000000000000  ntdll.dll!RtlUserThreadStart

kulkin-fish:
[libusb-bug.txt](https://github.com/libusb/libusb/files/6397880/libusb-bug.txt)

mcuee:
Why do you think this is a libusb bug and not the application "analysis-dbg.exe"? 

Are you the author of the application? If yes please post the relevant codes.

kulkin-fish:
#include <libusb-1.0/libusb.h>

int main(int argc, char* argv[]){
  libusb_context* libusbContext = nullptr;
  if(libusb_init(&libusbContext) != LIBUSB_SUCCESS)
    return -1;

  libusb_device** devList = nullptr;
  const int devCount = libusb_get_device_list(libusbContext, &devList);
  if(devCount <= 0){
[...]

PR #943, opened 2021-06-30
Access violation fixed for libusbk backend at the moment of device close

dmitry-zakablukov:
This pr fixes issue #912.
The reason of access violation is that the background thread may access a device handle memory after the handle has been closed.
To avoid that it is necessary to ensure that the device handle is not closed. This requires a lock on opened device handles in a context.


mcuee:
Tested pass. Thanks.

mcuee (approved):
Looks good to me. But please review as well.

mcuee:
@tormodvolden @dickens Please take a look if you got time. Thanks.

PR #927, opened 2021-06-11
windows: Allow GUID with and without trailing zeroes

tormodvolden:
Improve the GUID length check to accommodate the cases where the GUID
string fetched from registry does not have a trailing zero, or if a
multi-string entry, one or two trailing zeroes.

Signed-off-by: Craig Hutchinson <54269136+CraigHutchinson@users.noreply.github.com>
[Tormod: Use correct types and simplify comparisons]
Signed-off-by: Tormod Volden <debian.tormod@gmail.com>

PR #948, opened 2021-07-11
winusb: Ignore missing DeviceInterfaceGUID

tormodvolden:
I assume not all devices have a DeviceInterfaceGUID in the registry. Only warn on other errors.

Prior to commit a0305872 all errors were ignored.


mcuee (approved):
Tested okay. Thanks.
https://github.com/libusb/libusb/pull/927#issuecomment-877800858

PR #952, opened 2021-07-14
Fix noticeable performance degradation after commit 32a22069428cda9d6…

### LINKED ISSUE #951: Performance regression after commit "core: Refactor initialization and how the default context is handled"
Commit 32a22069428cda9d63aa666e92fb8882a83d4515 introduces a regression when `usbi_default_context` is not initialized in case non-NULL context is passed to the `libusb_init()` - [link to error location](https://github.com/libusb/libusb/commit/32a22069428cda9d63aa666e92fb8882a83d4515#diff-c1f9bc250077d41456a3e580fca0ddf5d8c25b741bff6d9b9505990a8b70b254R2277). This causes unwanted performance degradation (at least on Windows platform) because after the commit the `usbi_get_context()` always returns NULL and `log_v()` function is now forced to call `get_env_debug_level()` [for every single debug message](https://github.com/libusb/libusb/blob/339d24c8235476ceee43f7d704415ccab85e2583/libusb/core.c#L2575).
With USB2.0 FS BULK device I am working on performance drop is about 35% (355 KiB/s -> 260 KiB/s).

bohdan-tymkiv:
Fixes #951

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2308,7 +2308,7 @@ int API_EXPORTED libusb_init(libusb_context **ctx)
 #endif
 
 	/* default context should be initialized before calling usbi_dbg */
-	if (!ctx) {
+	if (!usbi_default_context) {
 		usbi_default_context = _ctx;
 		default_context_refcnt = 1;
 		usbi_dbg("created default context");


mcuee (approved):
Tested okay.

libusb/core.c
```diff
@@ -2308,7 +2308,7 @@ int API_EXPORTED libusb_init(libusb_context **ctx)
 #endif
 
 	/* default context should be initialized before calling usbi_dbg */
-	if (!ctx) {
+	if (!usbi_default_context) {
```
dmitry-zakablukov, line 2311: Maybe this can fix performance, but this breaks a logic: if `libusb_init(ctx)` is called with `ctx != null`, then `usbi_default_context` will be the same as `ctx`. And after `libusb_exit(ctx)` is called, `usbi_default_context` will point to the garbage.

mcuee: Good point. Thanks for pointing this out.

mcuee (changes requested):
Based on the comment, this needs update.

dmitry-zakablukov:
@bohdan-tymkiv , please consider this approach:
```
int API_EXPORTED libusb_init(libusb_context **ctx)
{
    /* default context should be initialized */
	if (ctx && !usbi_default_context) {
        libusb_init(NULL);
    }
   ...
```

dmitry-zakablukov:
Looks like now we should explicitly create default context by our own. I.e. on startup of your program just call `libusb_init(NULL)` and you are done.
Without that call there will be no default context and log function will check the environment for log level.

mcuee:
> Looks like now we should explicitly create default context by our own. I.e. on startup of your program just call `libusb_init(NULL)` and you are done.
> Without that call there will be no default context and log function will check the environment for log level.

Hmm, I think that may break the existing codes.

mcuee:
@hjelmn @dickens @tormodvolden Please also take a look at this issue. Thanks.

mcuee:
@jonasmalacofilho Please comment as well since it also affect pyusb based on my test results in #951.

mcuee:
> > Looks like now we should explicitly create default context by our own. I.e. on startup of your program just call `libusb_init(NULL)` and you are done.
> > Without that call there will be no default context and log function will check the environment for log level.

> Hmm, I think that may break the existing codes.

Comments from Nathan in #855
> Generally speaking most software I have worked on has either intentionally crashed or not avoided crashing on misuse. I would have preferred libusb had returned that status in libusb_exit but that would require and backwards-incompatible API change. I am with @mcuee in that crashing is probably fine here since you asked libusb to delete something that you never created.
> 
> I also agree with @dickens on that libusb is handling the default context incorrectly. There should be no default context unless you explicitly call `libusb_init(NULL)`. We should focus on changing that. It may break some code but I don't think we ever claimed that `libusb_init(&ctx); libusb_exit(NULL);` would work. I always assumed it was clear that is not the correct usage.

mcuee:
> Looks like now we should explicitly create default context by our own. I.e. on startup of your program just call `libusb_init(NULL)` and you are done.
> Without that call there will be no default context and log function will check the environment for log level.

@jonasmalacofilho If the above is correct, I wonder why pyusb is affected.

https://github.com/pyusb/pyusb/blob/master/usb/backend/libusb1.py#L710

```
# implementation of libusb 1.0 backend
class _LibUSB(usb.backend.IBackend):
    @methodtrace(_logger)
    def __init__(self, lib):
        usb.backend.IBackend.__init__(self)
        self.lib = lib
        self.ctx = c_void_p()
        _check(self.lib.libusb_init(byref(self.ctx)))```

**Comment by jonasmalacofilho** [comment]
> > Looks like now we should explicitly create default context by our own. I.e. on startup of your program just call `libusb_init(NULL)` and you are done.
> > Without that call there will be no default context and log function will check the environment for log level.
> 
> @jonasmalacofilho If the above is correct, I wonder why pyusb is affected.
> 
> https://github.com/pyusb/pyusb/blob/master/usb/backend/libusb1.py#L710
> 
> ```py
> # implementation of libusb 1.0 backend
> class _LibUSB(usb.backend.IBackend):
>     @methodtrace(_logger)
>     def __init__(self, lib):
>         usb.backend.IBackend.__init__(self)
>         self.lib = lib
>         self.ctx = c_void_p()
>         _check(self.lib.libusb_init(byref(self.ctx)))```
> ```

@mcuee,

If I understand correctly (it took me a couple of tries), we are doing exactly what has been reported on #951, that is, calling `libusb_init` with a non-NULL pointer.  The PyUSB call is equivalent to:

```c
void *ctx;
libusb_init(&ctx);
```

Which, according to #951, now triggers a performance regression due to the lack of default context.

mcuee:
> If I understand correctly (it took me a couple of tries), we are doing exactly what has been reported on #951, that is, calling `libusb_init` with a non-NULL pointer. The PyUSB call is equivalent to:
> 
> ```c
> void *ctx;
> libusb_init(&ctx);
> ```
> 
> Which, according to #951, now triggers a performance regression due to the lack of default context.

@jonasmalacofilho Thanks for the explanation. I thought 
``` self.ctx = c_void_p() ``` means NULL pointer but it actually just means ```void *```.

jonasmalacofilho:
@mcuee,

I haven't done my own investigation into the problem as far as libusb goes, but from the comments in this PR and on the corresponding issue, it's something entirely confined in libusb.

PyUSB usage of `libusb_init()` is within the spec, and changing it to no longer get a context out could break downstream projects that link to libusb in more than one way (for example, by simultaneously using PyUSB and hidapi on some platforms/configurations).

---

I'm curious: do you know why in libusb only the default context is used when logging?

jonasmalacofilho:
Actually, let me clarify what I meant.

PyUSB (AFAIK) does `libusb_init(&ctx); ...; libusb_exit(ctx)`, which is valid.

However, it seems that libusb is considering no longer providing a default context unless something calls `libusb_init(NULL)`.  While that may not violate any explicit past contract, it's fairly weird: it would mean that for PyUSB to be able to rely on the default context (which apparently is always necessary for libusb logging), we would actually need to call init and exit twice: `libusb_init(NULL); libusb_init(&ctx); ...; libusb_exit(ctx); libusb_exit(NULL);`.

Is this really the intended design?

Or is there a way for the logging functions in libusb to use the exported/non-default context instead?

mcuee:
@jonasmalacofilho Thanks for the analysis.

@hjelmn @LudovicRousseau 
It seems to me there are two ways out.
1) revert the first part of the changes in https://github.com/libusb/libusb/commit/32a22069428cda9d63aa666e92fb8882a83d4515
- usbi_default_context is only set if libusb_init() is called with NULL.

That part of the change is not strictly necessary in the first place as per the discussion https://github.com/libusb/libusb/issues/855.

2) keep the above part of the change, then fix #951 and potential other issues, say change the logging functions in libusb to use the exported/non-default context instead as per Jonas.

mcuee:
BTW, https://github.com/libusb/libusb/issues/928 openocd issue is related to the change as well but in that case Openocd code is not correct so they fixed it in openocd.

mcuee:
@bohdan-tymkiv Please share your view as well. Thanks.

mcuee:
@dmitry-zakablukov Please share your opinion as well. Thanks.

dmitry-zakablukov:
@mcuee , in our project we initialize default context every time we create a libusb context. Hence we have default context for logging and have no performance drop.

mcuee:
Close this in favor of #957.

PR #961, opened 2021-07-22
darwin: only attempt to auto-detach driver is one is active

### LINKED ISSUE #960: macOS kernel driver detach code improvement
Commit https://github.com/libusb/libusb/commit/b0fd4d873e224bbb1985971834d31d956080c7a4 is good that it downgrades the warning/error message created by the kernel driver detach codes.

```
darwin: downgrade error message on failure to open interface to info
This resolves an issue introduced by the implementation of the kernel detach code. This
code used claim_interface to detect if a kernel driver is attached. The issue is that if
a driver is attached or the interface is not available and an error message was printed
out. Since this is not an error the info level is more appropriate.

Also downgrading a number of warnings to clean up the output.

Closes #955
```

I think we can improve the code -- claim_interface() should not be called if libusb_kernel_driver_active() shows that there is no driver is attached to the interface.

hjelmn:
This commit improves the auto-detach code so that it doesn't attempt to detach a kernel
driver if none exists. This will avoid extra info/warning output on the failed detach.

Closes #960

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -2490,7 +2490,7 @@ static int darwin_attach_kernel_driver (struct libusb_device_handle *dev_handle,
 
 static int darwin_capture_claim_interface(struct libusb_device_handle *dev_handle, uint8_t iface) {
   enum libusb_error ret;
-  if (dev_handle->auto_detach_kernel_driver) {
+  if (dev_handle->auto_detach_kernel_driver && darwin_kernel_driver_active(dev_handle, iface)) {
     ret = darwin_detach_kernel_driver (dev_handle, iface);
     if (ret != LIBUSB_SUCCESS) {
       usbi_info (HANDLE_CTX (dev_handle), "failed to auto-detach the kernel driver for this device, ret=%d", ret);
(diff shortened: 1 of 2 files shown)


hjelmn:
@mcuee Should do the trick to remove the messages.

mcuee:
Great. This works very well. Thanks.
```

libusb_testn on  improve_autodetach ❯ ./examples/xusb 04b4:00f3
Using libusb v1.0.24.11644

Opening device 04B4:00F3...

Reading device descriptor:
            length: 18
      device class: 0
               S/N: 3
           VID:PID: 04B4:00F3
         bcdDevice: 0100
   iMan:iProd:iSer: 1:2:3
          nb confs: 1

Reading BOS descriptor: no descriptor

Reading first configuration descriptor:
             nb interfaces: 1
              interface[0]: id = 0
interface[0].altsetting[0]: num endpoints = 0
   Class.SubClass.Protocol: FF.00.00

Kernel driver attached for interface 0: 0

Claiming interface 0...

Reading string descriptors:
   String (0x01): "Cypress"
   String (0x02): "WestBridge "
   String (0x03): "0000000004BE"

Releasing interface 0...
Closing device...
```

PR #963, opened 2021-07-23
Windows: Fix timeouts if only endpoint 0 exists

julianscheel:
The default timeouts of 5 seconds used by WinUSB are cleared in
configure_endpoints(). This was not called if not at least one extra
endpoint apart from endpoint 0 existed, so that in those cases the
default timeout of 5 seconds remained in place.
Fix this by calling configure_endpoints, even if no interface specific
endpoints were found, so that it can configure endpoint 0.

Signed-off-by: Julian Scheel <julian@jusst.de>

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -530,6 +530,10 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 		usbi_dbg(HANDLE_CTX(dev_handle), "no endpoints found for interface %u", iface);
 		libusb_free_config_descriptor(conf_desc);
 		priv->usb_interface[iface].current_altsetting = altsetting;
+
+		// Extra init may be required to configure endpoint 0
+		if (priv->apib->configure_endpoints)
+			r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
 		return LIBUSB_SUCCESS;
 	}
 


mcuee:
@julianscheel Do you encounter this issue is real use case, like DFU (device FW upgrade)? What is the symptom?

julianscheel:
@mcuee Yes, faced this in a DFU implementation for an XMOS based device, where the flash erase command takes >5 seconds to execute. Thus the belonging request on endpoint 0 is blocked for more than 5 seconds and did timeout on Windows, due to the WinUSB default timeout of 5 seconds on the control endpoint.

mcuee:
@dickens @tormodvolden Please help to check if there are any side effects of this change. Thanks.

tormodvolden:
There must be something wrong with that DFU implementation. The device should not block a request while doing the flash erase, it should respond to the triggering DFU_GET_STATUS request with a _poll timeout_ value larger than the time that it expects to be busy erasing, then the host will wait for the specified time and pick up communication afterwards.

julianscheel:
@tormodvolden Sure, but even then, the 5.0 second timeout is only existent on Windows, while on macOS and Linux there is no 5.0 second timeout. So for consistency between platforms using libusb I actually think it would make sense to clear that timeout. 

And furthermore whenever you have a second endpoint activated it actually is cleared already. The only case where that timeout remains in place is when only endpoint 0 is used.

tormodvolden:
Yes, I agree to consistency. But you should still "complain" to the device bootloader authors/distributors and try get them fix it.

mcuee:
@tormodvolden Since you have lots of experiences with DFU as the maintainer of dfu-util, just wondering how often you see reports of such type of dfu devices.

mcuee:
@tormodvolden It seems to me the patch is safe to be merged. Do you agree?

tormodvolden:
I don't know much details of the Windows API, but this change seems to make sense and should be safe AFAICS. You are however not checking the return value `r` like is done in the case of several endpoints. Instead of duplicating this it might make sense to put the existing multiple endpoint code in an "else" clause and let it finish off the same in both cases.

tormodvolden:
@mcuee Often DFU implementations are broken one way or another, and this example of wrongly suspending a USB transaction while flashing or erasing is not unique. Further, DFU devices (in DFU mode) should only use the control endpoint and not have other endpoints in the interfaces (they often have multiple alternate settings), so this endpoint initialisation would go missed everywhere FWIW. But in the case of dfu-util it specifically sets the timeout to 5 seconds on all platforms, so this change wouldn't make any difference for the timeout in the end.

mcuee:
@tormodvolden Thanks for the explanation for the common pitfalls of DFU devices.

tormodvolden (changes requested):
You are not checking the return value r like is done in the case of several endpoints. Instead of duplicating this it might make sense to put the existing multiple endpoint code in an "else" clause and let it finish off the same in both cases.

tormodvolden (approved):

tormodvolden:
I have addressed my previously requested changes in my rework at https://github.com/tormodvolden/libusb/commit/9db6ff8f6032e60e22254b6429ae7e62a832152f but I'd like additional reviews and approvals of this.

Issue #944, opened 2021-07-02
C++ symbol for libusb_set_option not exported correctly


tckillian:
The libusb_set_option() function is not being properly exported for use with C++ in MSVC.

I believe the export was briefly specified correctly in @dickens original commit, here:
https://github.com/libusb/libusb/blob/v1.0.22-rc1/libusb/libusb-1.0.def#L156

It looks like it's changed several times since then, but the export style of this function does not match other exports in the master branch.

mcuee:
Please help to create a pull request if you can. Thanks.

mcuee:
@tckillian @tormodvolden  Please check if #949 is correct or not. Thanks.

mcuee:
First time learning to create a pull request. :-)

tormodvolden:
From going through the git log, it was only changed once, in commit 0aac390, maybe inadvertently, @dickens?

BTW shouldn't the ordinal numbers be unique for each function?
https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=msvc-160

It looks like people regenerate this file from a DLL instead of maintaining it by hand.

tormodvolden:
Note the original commit was commit 539f22e2, not in the RC that you both are referring to. It was actually modified twice:
- commit 46874013 removed the "@8" ordinal number.
- commit 0aac390 added the underscore

tormodvolden:
@hjelmn what do you think about the ordinal numbers?

mcuee:
@tormodvolden It seems to me the ordinal number is not unique. I believe it is there to support legacy code (32bit only). But now we need to see if 8 is correct or not based on the following.

Ref: https://docs.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-160
>Because the MSVC compiler uses name decoration for C++ functions, you must either use the decorated name internal_name or define the exported functions by using extern "C" in the source code. The compiler also decorates C functions that use the __stdcall calling convention with an underscore (_) prefix and a suffix composed of the at sign (@) followed by the number of bytes (in decimal) in the argument list.

mcuee:
@tormodvolden From libusb-1.0.def history file, it seems that this should be manually generated.
https://github.com/libusb/libusb/commit/b88ce7d2450a2d77b84cca5ffe6bcb449a5ae05b
```
* Generated from a MinGW32 generated DLL using the command:
[...]

PR #964, opened 2021-07-25
Update libusb-1.0.def

mcuee:
This fixes issue #944 C++ symbol for libusb_set_option not exported correctly.
As per the report in issue #944, original commit is correct but subsequent edits cause the regression.
https://github.com/libusb/libusb/blob/v1.0.22-rc1/libusb/libusb-1.0.def#L156

Signed-off-by: Xiaofan Chen xiaofanc@gmail.com

### DIFF
--- a/libusb/libusb-1.0.def
+++ b/libusb/libusb-1.0.def
@@ -155,7 +155,7 @@ EXPORTS
   libusb_set_interface_alt_setting
   libusb_set_interface_alt_setting@12 = libusb_set_interface_alt_setting
   libusb_set_option
-  _libusb_set_option = libusb_set_option
+  libusb_set_option@8 = libusb_set_option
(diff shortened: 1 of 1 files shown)


mcuee:
@tormodvolden Please take a look, hopefully this time it is done correctly (using Github Desktop client under Windows). I need to learn how to do it with git client directly under my Mac Mini M1.

@hjelmn This should be safe to take. Thanks.

mcuee:
I am still learning how to use git properly. 

Another attempt is here: https://github.com/libusb/hidapi/pull/311

tormodvolden:
Never mind, but the reference in the message is wrong, that was not the original commit, see https://github.com/libusb/libusb/issues/944#issuecomment-877810383

Issue #966, opened 2021-07-28
 Windows: Possible data race


mcuee:
From the mailing list.
https://marc.info/?t=162740011400005&r=1&w=4

On Tue, Jul 27, 2021 at 11:36 PM Bohdan Tymkiv  wrote:
>
> Hello dear libusb maintainers,
>
> I am facing an interesting issue with pyOCD and the DAPLink device. From time to time libusb fails with the following errors:
> (A) libusb: error [composite_copy_transfer_data] program assertion failed - no function to copy transfer data
> (B) libusb: error [windows_submit_transfer] program assertion failed - transfer HANDLE is NULL after transfer was submitted
>
> While issue (B) is non-critical, issue (A) completely breaks the communication with the device.
>
> pyOCD uses pyUSB under the hood (thus synchronous libusb API) and spawns two threads to communicate with the DAPLink:
> Main Thread - writes requests to DAPLink so it submits only OUT transfers.
> New Thread - reads responses from DAPLink so it submits only IN transfers.
> Both threads call libusb_handle_events() and there is a chance that transfer submitted by thread (A) will be handled by a call to libusb_handle_events in thread (B).
>
> It is quite difficult to reproduce the issue but adding a Sleep(10) at the line 750 of windows_common.c
> https://github.com/libusb/libusb/blob/e2edecd3a3f3ff79f58c992862f15b00e4eb3c49/libusb/os/windows_common.c#L750
> causes the issue (B) to appear immediately. That's because during Sleep() the USB transfer completes in the second thread and transfer_priv->handle is now NULL.
>
> To easily reproduce the issue (A), which is critical, Sleep(10) can be added at line 3122 of windows_winusb.c
> https://github.com/libusb/libusb/blob/e2edecd3a3f3ff79f58c992862f15b00e4eb3c49/libusb/os/windows_winusb.c#L3122
>
> DAPLink is a composite device with 5 interfaces.
> I am communication with interface #5 but because transfer_priv->interface_number is not initialized yet when transfer completes, the 'current_interface' here:
> https://github.com/libusb/libusb/blob/e2edecd3a3f3ff79f58c992862f15b00e4eb3c49/libusb/os/windows_winusb.c#L4530
> is not valid (zero).
>
> The rough way of fixing this is to protect the call to priv->backend->submit_transfer() here:
> https://github.com/libusb/libusb/blob/e2edecd3a3f3ff79f58c992862f15b00e4eb3c49/libusb/os/windows_common.c#L738
[...]

PR #967, opened 2021-07-28
Windows: Fix for data race which causes invalid current_interface in copy_transfer_data

bohdan-tymkiv:
This only fixes issue (A) described in #966 :
libusb: error [composite_copy_transfer_data] program assertion failed - no function to copy transfer data


timrprobocom:
For all that my two cents are worth, to me this looks like the correct fix for a simple ordering oversight.

mcuee:
@hjelmn 
I will trust Tim Roberts and Orin Eman. Please take a look and merge this if you are okay as well.

mcuee (approved):

PR #981, opened 2021-08-20
Darwin: After last CFRelease of struct field, set it to NULL

seanm:


seanm:
@hjelmn care to review?  I'm not 100% sure of this change, but I think it's correct.  I realize that just because one decrements an object's reference count doesn't mean the object is dead, but I think in this case this release is indicating that libusb is done with that object, and it seems best to null it from that structure field, especially as other code checks the field for null.

tormodvolden:
Ideally the trivial typo fixes (which are also not confined to macOS) should be in a separate commit and not have to await the more difficult evaluation of the code change.

Issue #902, opened 2021-03-30
Pyusb with new version of libusb 1.0 on HID mangling one byte of data on Windows 10


EnochMakenp:
Using the latest PyUSB, in conjunction with the latest libusb1.0 on Windows 10 (2020-12-11) with an access card reader that behave as a HID device is missing the last byte of data in the ctrl_transfer communication. I'm asking for 8 bytes of data and only 7 are read, which is annoying as I want to read the card UID, coded on 8 bytes...

The code I use in PyUSB is:
```
    device = usb.core.find(idVendor=vendorID, idProduct=productID)
    ret = device.ctrl_transfer(0x21, 0x09, 0x0300, 0, (0x8f,0x00,0x00,0x00,0x00,0x00,0x00,0x00),8) #makes the request
    response = device.ctrl_transfer(0xa1, 1, 0x0300, 0, 8) # read the response, ask for 8 bytes
```
and after this the content of response is :
array('B', [186, 137, 24, 128, 0, 0, 229])
Note that I have only 7 bytes instead of the 8 requested 
(if I ask for 7 bytes, the command fail, but if I request 9 or more bytes , I still get the 7 bytes...)

I used Wireshark (with USBPcap) to spy on the USB transaction and the 8 bytes of data are available on the USB packet, as can be see in the following transcript:
```
Frame 8: 36 bytes on wire (288 bits), 36 bytes captured (288 bits) on interface wireshark_extcap2040, id 0
USB URB
    [Source: 5.11.0]
    [Destination: host]
    USBPcap pseudoheader length: 28
    IRP ID: 0xffff960ef084e060
    IRP USBD_STATUS: USBD_STATUS_SUCCESS (0x00000000)
    URB Function: URB_FUNCTION_CONTROL_TRANSFER (0x0008)
    IRP information: 0x01, Direction: PDO -> FDO
    URB bus id: 5
    Device address: 11
    Endpoint: 0x80, Direction: IN
    URB transfer type: URB_CONTROL (0x02)
    Packet Data Length: 8
    [Request in: 7]
    [Time from request: 0.000164000 seconds]
    Control transfer stage: Complete (3)
    [bInterfaceClass: Unknown (0xffff)]
    CONTROL response data: ba8918800000e5c9
```
The control data (ba8918800000e5c9) has 8 bytes and in the response above I'm missing the last byte "c9", and the collected bytes in python stop at "e5" (=229).

[...]

PR #986, opened 2021-08-27
Fix Windows HID backend missing byte

mcuee:
https://github.com/libusb/libusb/issues/902#issuecomment-810726897
Jonas' comment is correct, when the report id is zero, we should indeed skip the first byte of transfer_priv->hid_buffer, since it is not part of the actual report. But the report size should not change.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -4205,8 +4205,6 @@ static enum libusb_transfer_status hid_copy_transfer_data(int sub_api, struct us
 				}
 
 				if (transfer_priv->hid_buffer[0] == 0) {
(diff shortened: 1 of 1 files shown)


tormodvolden:
Style suggestion: You could maybe include more of Jonas' explanation in the commit message, so that the git log can be read and understood without having to look up a github tracker. Or if there is already enough explained, just thank \<full name\> for the investigation and put the link less prominently below.

mcuee:
I think the explanation is enough. The others are not really relevant in this case.

mcuee:
@tormodvolden Please check the updated commit message to see if that is okay.

tormodvolden:
Perfect.

PR #989, opened 2021-09-02
Fix crash (segv) in libusb_init if backend.init failes

ankurverma85:
If backend.init fails. the control goes to err_io_exit which tries to clean up hotplug related lists that arent initialized.

Moving hotplug_init before makes sure the lists are valid so if backend.init fails they get cleaned up without errors.

Also adding mutx_unlock to a few more error scenarios

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2342,13 +2342,14 @@ int API_EXPORTED libusb_init(libusb_context **ctx)
 	list_add(&_ctx->list, &active_contexts_list);
 	usbi_mutex_static_unlock(&active_contexts_lock);
 
+	usbi_hotplug_init(_ctx);
+
 	if (usbi_backend.init) {
 		r = usbi_backend.init(_ctx);
 		if (r)
 			goto err_io_exit;
 	}
 
-	usbi_hotplug_init(_ctx);
 
 	if (ctx)
 		*ctx = _ctx;


libusb/core.c
```diff
@@ -2342,13 +2343,15 @@ int API_EXPORTED libusb_init(libusb_context **ctx)
 	list_add(&_ctx->list, &active_contexts_list);
 	usbi_mutex_static_unlock(&active_contexts_lock);
 
+	usbi_hotplug_init(_ctx);
+
 	if (usbi_backend.init) {
 		r = usbi_backend.init(_ctx);
 		if (r)
+			usbi_mutex_static_unlock(&default_context_lock);
```
Youw, line 2351: this doesn't look right: `err_io_exit` does unlock `default_context_lock` a few lines below

tormodvolden:
Note that this pull request conversation and title doesn't get included in the git log, so please add an explanation to the commit message as well.

ankurverma85:
> Note that this pull request conversation and title doesn't get included in the git log, so please add an explanation to the commit message as well.

done. squashed with capturing the title as a commit message

ankurverma85:
@mcuee , @tormodvolden 

Anything else needed before this can be merged ?

tormodvolden:
It would be good if one of core developers who actually touched this code would chime in and say, yes, this was an oversight, or, no, this should be dealt with differently. But from the look of it, the suggested patch and explanation, thumbs up from me.

You still have an unfinished commit message though, I would suggest having "Fix segmentation fault in libusb_init() if backend.init fails" on the first line, a second empty line, then the already existing explanation.

tormodvolden:
git blame points to commit 32a22069. @dickens, can you please review this fixup?

mcuee:
Chris seems to be busy and not able to review as of now.
@hjelmn -- please help here when you got time. Thanks.

photron:
@ankurverma85 are you sure that the init order is wrong here? Maybe the exit/cleanup order is wrong instead. Because your move of the usbi_hotplug_init call before the usbi_backend.init call might not be allowed. Maybe usbi_hotplug_init assumes/requires an initialized backend.

tormodvolden:
The suggested change is at least consistent with how it was before the above-mentioned commit. What's inside usbi_hotplug_init() used to be called before usbi_backend.init(), see https://github.com/libusb/libusb/commit/32a22069428cda9d63aa666e92fb8882a83d4515#diff-c1f9bc250077d41456a3e580fca0ddf5d8c25b741bff6d9b9505990a8b70b254L2296
However, the hotplug_ready flag is new.

photron:
@tormodvolden You're correct. I take my question back and agree with the suggested change.

PR #995, opened 2021-09-22
core: Unlock and clear default ctx in all error paths in libusb_init

photron:
Commit 32a22069428cda9d63aa666e92fb8882a83d4515 reordered and refactored `libusb_init`. This resulted in moving code outside the default ctx lock that was unrelated to it. But this resulted in broken error path cleanup logic that leaves the default ctx as locked, half initialized and freed in case a `libusb_set_option` or the `usbi_io_init` call fails.

Undo part of the previous reordering to unlock and clear the default ctx in all error paths in `libusb_init`.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2333,10 +2333,8 @@ int API_EXPORTED libusb_init(libusb_context **ctx)
 		libusb_version_internal.micro, libusb_version_internal.nano, libusb_version_internal.rc);
 
 	r = usbi_io_init(_ctx);
-	if (r < 0) {
-		usbi_mutex_static_unlock(&default_context_lock);
+	if (r < 0)
 		goto err_free_ctx;
-	}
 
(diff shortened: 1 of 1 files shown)


tormodvolden:
Could possibly the unlock on line 2372 be moved down to err_free_ctx instead?

photron:
Yes, that would be the correct solution. That matches how it was before commit 32a22069428cda9d63aa666e92fb8882a83d4515. I'll rework this patch.

photron:
The problem is even worse. If the call to `usbi_io_init` failed (which is unlikely) then `libusb_init` could leave the default ctx in a broken state as freed memory. The next `libusb_init` call would then assume the default ctx is already initialized and access freed memory resulting in a crash.

hjelmn (approved):

hjelmn:
LGTM

PR #998, opened 2021-09-22
git hooks: Only apply to master branch

tormodvolden:
Consider this an RFC. I know there is LIBUSB_SKIP_NANO but it is easy to forget. For me it would be natural to develop and prepare stuff on separate branches, and only use master to stage things to push.

### DIFF
--- a/.private/post-rewrite.sh
+++ b/.private/post-rewrite.sh
@@ -18,6 +18,10 @@ if [ -n "$LIBUSB_SKIP_NANO" ]; then
   exit 0
 fi
 
+if [ "$(git branch --show-current)" != "master" ]; then
+  exit 0
+fi
+
 case "$1" in
   amend)
     # Check if a .amend exists. If none, create one and warn user to re-commit.
--- a/.private/pre-commit.sh
+++ b/.private/pre-commit.sh
@@ -38,6 +38,10 @@ fi
 
 eval $TYPE_CMD || { echo "git command not found. Aborting." >&2; exit 1; }
 
+if [ "$(git branch --show-current)" != "master" ]; then
+  exit 0
+fi
+
 NANO=`git log --oneline | wc -l`
 NANO=`expr $NANO + $BRANCH_OFFSET`
 # Amended commits need to have the nano corrected. Current versions of git hooks


mcuee (approved):

mcuee:
I think this is a good idea.

tormodvolden:
Hmm, this disables it during rebasing, when there is no real branch. How can we know we are rebasing the master branch?

tormodvolden:
This is starting to look ugly but this would work:
`if [ "$(git branch --show-current)" != "master" ] && ! git branch --list | grep -q '^\* (no branch, rebasing master)$'; then`

tormodvolden:
Independent of this patch, I got bitten earlier today by the -1 +1 toggle for each time I do an amend. There must be a better way. A bit weird also that the nano version is not simply the length of the git log, but minus 1.

pbatard:
@tormodvolden, you may be interested in the updated version of these scripts that I'm using in Rufus. For one thing (at least for Windows and *NIX) they should be able to detect an amend by checking the process list and whether a `--amend` argument was provided to the `git` command.

We are also using `git rev-list HEAD --count` rather than `git log --oneline | wc -l` to compute the nano.

See https://github.com/pbatard/rufus/blob/master/_pre-commit.sh and https://github.com/pbatard/rufus/blob/master/_detect-amend.sh if you are interested.

tormodvolden:
@pbatard thanks! I'll look into those.

Jslaton1 (approved):

hjelmn:
@tormodvolden Any progress on bringing back @pbatard's changes? Would be nice to get this done so I can skip the extra ```git commit --amend --no-edit```. One of these days I will forget it and screw up the nano :).

tormodvolden:
No, grepping the process table seemed a bit too much for me. However, it seems impossible to do this "cleanly" with the current hooks available in the git command.

I have a partial fix I can suggest, so that an already correct nano is not changed, at least breaking the infinite +1/-1 toggle.

mcuee:
Shall we drop the nano number totally and use `git commit id` or `date+git commit id`?

libusb seems to be the only project which is using the nano number.

mcuee:
Close this old PR in favour of #1799.
* #1799

Or something else -- to remove VERSION_NANO completely.
* https://github.com/libusb/libusb/issues/1727

mcuee:
@tormodvolden 

I am closing old PRs and I think this can be closed. Please re-open if you think it is worth keeping open. Thanks.

PR #1004, opened 2021-10-02
Keep LIBUSB_OPTION_WEAK_AUTHORITY as a macro with same value

tormodvolden:
and bump LIBUSB_API_VERSION.

Alternatively we could omit the macro and get rid of LIBUSB_OPTION_WEAK_AUTHORITY for good, the hard way.


mcuee:
Looks okay to me.

mcuee (approved):

PR #1007, opened 2021-10-06
Windows: Fix ssize_t unconditionally defined

pcercuei:
The standard procedure to define ssize_t on Windows is to wrap it around
a check for the _SSIZE_T_DEFINED macro.

If not done, it makes it impossible to use the libusb.h header along
with other headers (from other libraries) that also attempt to define
ssize_t.

Signed-off-by: Paul Cercueil <paul@crapouillou.net>

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -31,8 +31,11 @@
 #define inline __inline
 #endif
 /* ssize_t is also not available */
+#ifndef _SSIZE_T_DEFINED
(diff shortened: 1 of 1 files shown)


pcercuei:
Yep, I noticed that too. Do you want a `Fixes:` tag?

tormodvolden:
These are tricky to include in the initial merge request, aren't they :) No, thanks, I already amended it in my tree. We always merge locally anyway.

PR #1023, opened 2021-11-06
Fix compilation on macOS 10.9 and earlier

evanmiller:
`kUSBReEnumerateCaptureDeviceMask` was introduced in the 10.10 SDK.

Downstream report: https://trac.macports.org/ticket/63668


tormodvolden:
Thanks a lot!

Issue #1003, opened 2021-10-02
no_enumeration should be per context, not global


tormodvolden:
This was brought up in a discussion somewhere, so I open this ticket so it is not forgotten. Likely related to #942.

tormodvolden:
Carried over from #942:
Now, if you want to have one context where you want to skip enumeration (because you use libusb_wrap_sys() anyway) and another context with enumeration, you will have to initialize the latter first, then set LIBUSB_OPTION_NO_DEVICE_DISCOVERY and initialize the former. If these are in independent plug-ins (like the example in the Contexts documentation) this is not ideal. I guess we could make libusb_set_option(NULL, LIBUSB_OPTION_NO_DEVICE_DISCOVERY, 0) work to turn it off, or we need something more sophisticated to manipulate the default options in general. If we were to start from scratch, I think there should have been a flag option argument to libusb_init(), and the options being bit masks...

Also, libusb_set_option(NULL, ...) does not affect an existing, default context. Existing applications may depend on this, if this was the behavior earlier(?), consistent to the use of NULL ctx in all other functions.

Maybe we should add a new libusb_set_default_options() instead?

xloem:
additional possible approaches:
- passing options in a list structure to libusb_init
- separating allocation from initialisation so early options can be set with data
- refactoring linux_usbfs to initialise on first use

tormodvolden:
By "initialisation" you mean also the device discovery?

tormodvolden:
To remain backwards compatible, maybe we can add something like a new libusb_create_context(&ctx, options, ...) and if you call libusb(ctx) with an already created context it can skip allocation and apply these settings for "initialisation" including whether to do device discovery. Well, I guess that's what you had in mind for your second approach above.

xloem:
yeah.  and I guess by initialisation I did mean initial device discovery and likely hotplug setup I suppose; I don't remember the internal design well.

tormodvolden:
@xloem did you do some work on this as part of some other merge request? There will be 1.0.25 RC soon, so last chance to get something in.

xloem:
58ac16bd1e0f5f5f01a26c40720501d5f2974a30 is part of #874 .  I have trouble thinking about the PR, and am unlikely to resolve that before the release.  There may be other changes in the branch that are relevant: or that commit may simply be mergeable.  Needs testing or a glance through the code at least.  The change is pretty simple.

hjelmn:
[...]

PR #1026, opened 2021-11-20
core: add libusb_init_context to allow setting options during context creation

hjelmn:
The new initialization function addresses some shortcomings of the libusb_set_option()
interface. Namely, it allows setting the no-enumeration option (and others) on only the
contexts where it is requested. The old initialization function (libusb_init()) is
deprecated and will be removed in a future release. For now it translates to a call to
libusb_init_context() with no options specified.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1,7 +1,7 @@
 /* -*- Mode: C; indent-tabs-mode:t ; c-basic-offset:8 -*- */
 /*
  * Core functions for libusb
- * Copyright © 2012-2013 Nathan Hjelm <hjelmn@cs.unm.edu>
+ * Copyright © 2012-2021 Nathan Hjelm <hjelmn@cs.unm.edu>
  * Copyright © 2007-2008 Daniel Drake <dsd@gentoo.org>
  * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
  *
@@ -111,17 +111,18 @@ struct list_head active_contexts_list;
  * libusb uses stderr for all logging. By default, logging is set to NONE,
  * which means that no output will be produced. However, unless the library
  * has been compiled with logging disabled, then any application calls to
- * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level), or the setting of the
- * environmental variable LIBUSB_DEBUG outside of the application, can result
- * in logging being produced. Your application should therefore not close
- * stderr, but instead direct it to the null device if its output is
- * undesirable.
+ * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level),
+ * libusb_init_context, or the setting of the environmental variable
+ * LIBUSB_DEBUG outside of the application, can result in logging being
+ * produced. Your application should therefore not close stderr, but instead
+ * direct it to the null device if its output is undesirable.
  *
- * The libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) function can be
- * used to enable logging of certain messages. Under standard configuration,
- * libusb doesn't really log much so you are advised to use this function
- * to enable all error/warning/ informational messages. It will help debug
- * problems with your software.
+ * The libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) or
+ * libusb_init_context functions can be used to enable logging of certain
+ * messages. With the default configuration, libusb will not log much so if
+ * you are advised to use one of these functions to enable all
+ * error/warning/informational messages. It will help debug problems with your
+ * software.
  *
  * The logged messages are unstructured. There is no one-to-one correspondence
  * between messages being logged and success or failure return codes from
@@ -137,19 +138,19 @@ struct list_head active_contexts_list;
  * The LIBUSB_DEBUG environment variable can be used to enable message logging
  * at run-time. This environment variable should be set to a log level number,
  * which is interpreted the same as the
- * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) parameter. When this
- * environment variable is set, the message logging verbosity level is fixed
- * and libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) effectively does
- * nothing.
+ * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level), or
+ * libusb_init_context(&ctx, &(struct libusb_init_option){.option = LIBUSB_OPTION_LOG_LEVEL, .value = {.ival = level}}, 0).
+ * When the environment variable is set, the message logging verbosity level is
+ * fixed and setting the LIBUSB_OPTION_LOG_LEVEL option has no effect.
  *
  * libusb can be compiled without any logging functions, useful for embedded
- * systems. In this case, libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level)
- * and the LIBUSB_DEBUG environment variable have no effects.
+ * systems. In this case, neither the LIBUSB_OPTION_LOG_LEVEL option, nor the
+ * LIBUSB_DEBUG environment variable will have any effect.
  *
  * libusb can also be compiled with verbose debugging messages always. When
  * the library is compiled in this way, all messages of all verbosities are
- * always logged. libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) and
- * the LIBUSB_DEBUG environment variable have no effects.
+ * always logged. Again, in this case, neither the LIBUSB_OPTION_LOG_LEVEL
+ * option, nor the LIBUSB_DEBUG environment variable will have any effect.
  *
  * \section remarks Other remarks
  *
@@ -327,9 +328,9 @@ if (cfg != desired)
  * developed modules may both use libusb.
  *
  * libusb is written to allow for these multiple user scenarios. The two
- * "instances" of libusb will not interfere: libusb_set_option() calls
- * from one user will not affect the same settings for other users, other
- * users can continue using libusb after one of them calls libusb_exit(), etc.
+ * "instances" of libusb will not interfere: an option set by one user will have
+ * no effect the same option for other users, other users can continue using
+ * libusb after one of them calls libusb_exit(), etc.
  *
  * This is made possible through libusb's <em>context</em> concept. When you
  * call libusb_init(), you are (optionally) given a context. You can then pass
@@ -436,6 +437,7 @@ if (cfg != desired)
   * - libusb_hotplug_deregister_callback()
   * - libusb_hotplug_register_callback()
   * - libusb_init()
+  * - libusb_init_context()
   * - libusb_interrupt_event_handler()
   * - libusb_interrupt_transfer()
   * - libusb_kernel_driver_active()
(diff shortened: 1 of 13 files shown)


tormodvolden:
Is this also changing the options to only have one argument, instead of the va_args of libusb_set_option() ?

hjelmn:
Indeed. Though the varargs was originally intended to cover log level. It does give some extensibility. We could add padding to the end of the struct to allow larger values in the future.

xloem:
I looked through this first commit and it looks harmless and organised to my error-prone eye.  It appears backwards-compatible regarding varargs use of options and use of no_device_discovery on android.

Thanks for your design work on this.

mcuee:
Does this replace -#995

hjelmn:
@mcuee No. This is separate from that one. We should merge #995, make a decision on this one and plan for a new RC. I am still trying to figure out how to test DFU.

hjelmn:
Updated the documentation. One or two small things I need to fix up then this is ready to merge if we get a consensus.

hjelmn:
Note that before an RC we need to increment both ```lt_current``` and ```lt_age```.

xloem:
@harbulot do you have interest in verifying that this doesn't break older android code?

hjelmn:
Any additional comments?

xloem:
- i just tested this branch on android with https://github.com/harbulot/LibUsbAndroidIssue942 and it passed great
- note that no_enumeration is still a static global in this branch, so it doesn't alone make it safe to set the flag differently on different contexts, but it lays the groundwork to do so cleanly
- i might personally replace default options with this entirely, just to simplify things

lgtm

libusb/libusb.h
```diff
@@ -1342,6 +1342,71 @@ enum libusb_log_cb_mode {
...
+	 * does nothing: you'll always get messages from all levels.
+	 */
+	LIBUSB_OPTION_LOG_LEVEL = 0,
+
+	/** Use the UsbDk backend for a specific context, if available.
+	 *
+	 * This option should be set immediately after calling libusb_init(), otherwise
+	 * unspecified behavior may occur.
+	 *
+	 * Only valid on Windows.
```
Mindavi, line 1377: Will it cause an error initializing or will it be ignored? May be good to document here.

libusb/libusb.h
```diff
@@ -1342,6 +1342,71 @@ enum libusb_log_cb_mode {
...
+	 *
+	 * Hotplug functionality will also be deactivated.
+	 *
+	 * The option is useful in combination with libusb_wrap_sys_device(),
+	 * which can access a device directly without prior device scanning.
+	 *
+	 * This is typically needed on Android, where access to USB devices
+	 * is limited.
+	 *
+	 * Only valid on Linux.
```
Mindavi, line 1394: Same as above, will this be ignored on WIndows or will it return an error during init?

Mindavi (commented):
Don't know the codebase, but some of the documentation gave me more questions, might be nice to be more clear there.

libusb/core.c
```diff
@@ -2277,6 +2279,26 @@ static enum libusb_log_level get_env_debug_level(void)
...
+ * be reused (and nothing will be initialized/reinitialized).
+ *
+ * \param ctx Optional output location for context pointer.
+ * Only valid on return code 0.
+ * \param options Optional array of options to set on the new context.
+ * \param num_options Number of elements in the options array.
+ * \returns 0 on success, or a LIBUSB_ERROR code on failure
+ * \see libusb_contexts
+ */
+int API_EXPORTED libusb_init_with_options(libusb_context **ctx, struct libusb_init_option options[], int num_options)
```
seanm, line 2301: Wouldn't making the array `const` be better?

hjelmn: Done.

libusb/libusb.h
hjelmn: Hmm, yeah. Good call, this text was missing from the description when it was added. Ignoring is the reasonable behavior.

libusb/libusb.h
hjelmn: Hmm, yeah. Good call, this text was missing from the description when it was added. Ignoring is the reasonable behavior.

hjelmn:
Any last comments? I plan to merge this before the new year if there are no objections.

hjelmn:
Deprecated ```libusb_init()``` in the latest push.

hjelmn:
Updated the tests and examples.

tormodvolden:
If we are going to deprecate libusb_init(), I suppose libusb_init_with_options() will be the default (and later, the only) way of initializing the library. Should we therefore give it a simpler (more "default"-sounding) name? Having "_with_options" in the name kind of suggests there will be another alternative without options. libusb_initialize() or libusb_init_ctx() ?

tormodvolden:
I'd suggest the less cryptic libusb_init_context() too. It is instructive in that it underlines that a context is initialized, not "the library".

hjelmn:
@tormodvolden Good point. Will make that change.

hjelmn:
Done. Picked ```libusb_init_context()``` as the new function name. With this change we may want to call the new release libusb 1.1.0.

libusb/core.c
```diff
@@ -136,19 +137,19 @@ struct list_head active_contexts_list;
  * The LIBUSB_DEBUG environment variable can be used to enable message logging
  * at run-time. This environment variable should be set to a log level number,
  * which is interpreted the same as the
- * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) parameter. When this
- * environment variable is set, the message logging verbosity level is fixed
- * and libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level) effectively does
- * nothing.
+ * libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, level), or
+ * libusb_init_context(&ctx, &(struct libusb_init_option){.option = LIBUSB_OPTION_LOG_LEVEL, .value = {.ival = level}}).
```
tormodvolden, line 141: Isn't the number of options missing here?

libusb/core.c
```diff
@@ -2269,14 +2271,36 @@ static enum libusb_log_level get_env_debug_level(void)
  *
  * If you do not provide an output location for a context pointer, a default
  * context will be created. If there was already a default context, it will
- * be reused (and nothing will be initialized/reinitialized).
+ * be reused (and nothing will be initialized/reinitialized). Deprecated in
+ * favor of \ref liusb_init_with_options().
```
tormodvolden, line 2275: Typo: liusb_

tormodvolden: Wrong function name too :)

libusb/core.c
```diff
@@ -2269,14 +2271,36 @@ static enum libusb_log_level get_env_debug_level(void)
...
+	return libusb_init_context(ctx, NULL, 0);
+}
+
+/** \ingroup libusb_lib
+ * Initialize libusb. This function must be called before calling any other
+ * libusb function.
+ *
+ * If you do not provide an output location for a context pointer, a default
+ * context will be created. If there was already a default context, it will
+ * be reused (and nothing will be initialized/reinitialized). If num_options
```
tormodvolden, line 2293: Any specified options will be ignored as well, right?

tormodvolden:
> Done. Picked `libusb_init_context()` as the new function name. With this change we may want to call the new release libusb 1.1.0.

In general, has it already been discussed whether to follow "standard" semantic versioning?

libusb/core.c
hjelmn: Done.

libusb/core.c
hjelmn: Done.

libusb/core.c
hjelmn: Correct. Added.

hjelmn:
should be good to go now

libusb/libusb.h
```diff
@@ -1342,6 +1342,72 @@ enum libusb_log_cb_mode {
 	LIBUSB_LOG_CB_CONTEXT = (1 << 1)
 };
 
+/** \ingroup libusb_lib
+ * Available option values for libusb_set_option().
```
tormodvolden, line 1346: and libusb_init_context()

hjelmn: Done.

tormodvolden:
I would have liked to see more feedback from others on such an important API change, but I guess the other core developers are MIA at the moment. An alternative to squeezing this into 1.0.25 would be to release 1.0.25 now with all its regression fixes, and rather soon follow up with 1.0.26 with the API change. I think the current option brokenness in git is not worse than in 1.0.24 so it wouldn't be a regression?

Do you plan to follow up with a fix for #1003?

hjelmn:
@tormodvolden I agree. Delaying this until 1.0.25 makes sense given the level of feedback,

tormodvolden:
@xloem, do you agree that although the situation for NO_DISCOVERY and other options is not satisfactory in current git without this API change (and #1003), it is not worse than 1.0.24?

CraigHutchinson:
@xloem @tormodvolden I noticed this development has gone stale, it was mentioned as being a blocker to a few major PRs:
https://github.com/libusb/libusb/issues/1094#issuecomment-1093788721

My primary interest is if we can get movement on #874 . I would be willing to help out, will try and read up on the convos when I get some time...

xloem:
> @xloem, do you agree that although the situation for NO_DISCOVERY and other options is not satisfactory in current git without this API change (and #1003), it is not worse than 1.0.24?

tormod, I'm afraid I don't remember these well enough to have an opinion.

However, my opinion is that it is important to keep APIs backwards compatible in this library, because many, many people use libusb and very few people develop libusb, so changes will likely break things the project never learns about.

> @xloem @tormodvolden I noticed this development has gone stale, it was mentioned as being a blocker to a few major PRs: [#1094 (comment)](https://github.com/libusb/libusb/issues/1094#issuecomment-1093788721)
> 
> My primary interest is if we can get movement on #874 . I would be willing to help out, will try and read up on the convos when I get some time...

It looks like this PR simply hasn't been merged. It could probably be rebased and merged immediately. I would personally ensure that all api changes are backwards compatible before doing so; they may be so already.

jefdriesen:
I guess the main disadvantage of the proposed libusb_init_context API is that it's limited to only integer options. Extending the structure in the future will break backwards compatibility, unless the size remains the same (e.g. a 64bit integer). It also doesn't allow to set a log callback function.

An alternative solution could be something like is done for the pthread_create function. This function takes a pthread_attr_t which acts as a placeholder to store all the options. Something similar could be done for libusb:

```
libusb_attr *attr = NULL;
libusb_context *ctx = NULL;

libusb_attr_init(&attr);
libusb_attr_set_loglevel(attr, ...);
libusb_attr_set_logfunction(attr, ...);
libusb_attr_set_option(attr, ...);

libusb_init_context(&ctx, attr);

libusb_attr_free(attr);
```
It's a lot more verbose, but it can easily be extended with new functionality if necessary.

mcuee:
@jwrdegoede and @LudovicRousseau 
Just wondering if you could comment here as well since @tormodvolden thinks this is gating for 1.0.27 release.
* https://github.com/libusb/libusb/issues/1094#issuecomment-1313277220

hjelmn:
I think having something as verbose as setting attributes (like pthreads) is overkill for libusb. It is probably sufficient to add a void * to the union to ensure that a pointer can be passed. That is sufficient to set the a log callback function if desired.

hjelmn:
Anyway, now that I have some cycles to finish this up are there any other reviews? This should go in sooner rather than later.

mcuee:
@hjelmn 

github action MSYS2 mingw build failed to build the examples. Strange.

tormodvolden:
I'll squash the libusb-1.0.def update into the first commit and reformat the commit messages to <80 width before I push this.

tormodvolden:
The final piece here will be #1003. As for version minor bumping, I'd suggest we let this bake in 1.0.x while still having libusb_init() around, then bump to 1.1.0 (at the latest) when we dump libusb_init().

tormodvolden:
Merged. Thanks for the work @hjelmn, this should solve a long-standing headache in option handling.

tormodvolden:
https://github.com/libusb/libusb/commit/6622f386f52807dac76c8a260c98aa02c311bc93#r98159441
> There is really no need to ratatoille this code.
> struct libusb_init_option can just be declared as a prototype before the function declaration using it, and then the definition can
> come later.

@hselasky You mean, to reduce the file changes to a minimum?

hselasky:
@tormodvolden : yes, but it is not a big deal.

PR #1031, opened 2021-11-30
darwin: fix behavior of libusb_set_interface_alt_setting when it stal…

hjelmn:
…ls the pipe

In some versions of macOS a pipe stall returned when setting the alternate interface
causes the interface to become unusable. To handle this case the backend was always
re-claiming the interface before clearing the pipe stall. In macOS Monterey
unconditionally re-claiming the interface leads to an error due to the process already
having exclusive access. To resolve this issue we attempt to clear the halt and only
re-claim the interface if clearing the pipe stall returns kIOUSBUnknownPipeErr. Tested
with 12.0.1 and 10.13 and get the expected results in both cases with a custom USB
device that has this behavior.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1,8 +1,8 @@
 /* -*- Mode: C; indent-tabs-mode:nil -*- */
 /*
  * darwin backend for libusb 1.0
- * Copyright © 2008-2020 Nathan Hjelm <hjelmn@cs.unm.edu>
- * Copyright © 2019-2020 Google LLC. All rights reserved.
+ * Copyright © 2008-2021 Nathan Hjelm <hjelmn@cs.unm.edu>
+ * Copyright © 2019-2021 Google LLC. All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


libusb/os/darwin_usb.c
```diff
@@ -1586,12 +1587,34 @@ static int darwin_release_interface(struct libusb_device_handle *dev_handle, uin
   return darwin_to_libusb (kresult);
 }
 
+static int darwin_clear_check_alt_setting_and_clear_halt(struct libusb_device_handle *dev_handle, uint8_t altsetting, struct darwin_interface *cInterface) {
```
tormodvolden, line 1590: Naming: Is the first _clear_ intentional?
And do we need darwin_ on all static functions?

libusb/os/darwin_usb.c
```diff
@@ -1623,23 +1646,18 @@ static int darwin_set_interface_altsetting(struct libusb_device_handle *dev_hand
...
-
-  if (ret) {
-    darwin_release_interface (dev_handle, iface);
-    usbi_err (HANDLE_CTX (dev_handle), "could not reclaim interface");
+  ret = darwin_clear_check_alt_setting_and_clear_halt(dev_handle, altsetting, cInterface);
+  if (LIBUSB_ERROR_NOT_FOUND == ret) {
+    /* For some reason we need to reclaim the interface after the pipe error with some versions of macOS */
+    ret = darwin_claim_interface (dev_handle, iface);
+    if (LIBUSB_SUCCESS != ret) {
+      darwin_release_interface (dev_handle, iface);
```
tormodvolden, line 1654: Do we need to release if claim failed?

hjelmn: Ideally, yes. It was there before and it ensures we clean up the internal structures. May leak without it.

libusb/os/darwin_usb.c
hjelmn: Fixed. Was not matching my own naming conventions. The clear was not intentional

mcuee:
Test shows it fixed the issue under my Mac Mini M1 (macOS 12.0.1).

tormodvolden (approved):

PR #1056, opened 2022-02-03
(darwin) Use GetPipePropertiesV3

tormodvolden:
I cannot test the InterfaceVersion < 550 path (OSX Lion and earlier) but I compile-tested by overriding InterfaceVersion to 549.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1389,12 +1389,16 @@ static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle,
 
   /* current interface */
   struct darwin_interface *cInterface = &priv->interfaces[iface];
+#if InterfaceVersion >= 550
+  IOUSBEndpointProperties pipeProperties = {.bVersion = kUSBEndpointPropertiesVersion3};
+#else
+  UInt8 dont_care1, dont_care3;
+  UInt16 dont_care2;
+#endif
 
   IOReturn kresult;
 
   UInt8 numep, direction, number;
-  UInt8 dont_care1, dont_care3;
-  UInt16 dont_care2;
   int rc;
   struct libusb_context *ctx = HANDLE_CTX (dev_handle);
 
@@ -1410,9 +1414,14 @@ static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle,
 
   /* iterate through pipe references */
   for (UInt8 i = 1 ; i <= numep ; i++) {
+#if InterfaceVersion >= 550
+    kresult = (*(cInterface->interface))->GetPipePropertiesV3 (cInterface->interface, i, &pipeProperties);
+    number = pipeProperties.bEndpointNumber;
+    direction = pipeProperties.bDirection;
+#else
     kresult = (*(cInterface->interface))->GetPipeProperties(cInterface->interface, i, &direction, &number, &dont_care1,
                                                             &dont_care2, &dont_care3);
-
+#endif
     if (kresult != kIOReturnSuccess) {
       /* probably a buggy device. try to get the endpoint address from the descriptors */
       struct libusb_config_descriptor *config;
@@ -2024,11 +2033,17 @@ static int submit_iso_transfer(struct usbi_transfer *itransfer) {
   struct darwin_transfer_priv *tpriv = usbi_get_transfer_priv(itransfer);
 
   IOReturn kresult;
-  uint8_t direction, number, interval, pipeRef, transferType;
-  uint16_t maxPacketSize;
+  uint8_t pipeRef, interval;
   UInt64 frame;
   AbsoluteTime atTime;
   int i;
+#if InterfaceVersion >= 550
+  IOUSBEndpointProperties pipeProperties = {.bVersion = kUSBEndpointPropertiesVersion3};
+#else
+  /* None of the values below are used in libusb for iso transfers */
+  uint8_t direction, number, transferType;
+  uint16_t maxPacketSize;
+#endif
 
   struct darwin_interface *cInterface;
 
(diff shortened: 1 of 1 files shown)


mcuee:
Just wondering if we should really care about OS X Lion or ealier version or not.
I remember my first Mac Mini (2011 version, bought late 2011, dead in early 2019) ran OS X Lion from the beginning. 

Ref: https://en.wikipedia.org/wiki/OS_X_Lion
Released to manufacturing	July 20, 2011
Latest release	10.7.5 (Build 11G63) / October 4, 2012

tormodvolden:
Indeed, unlike Apple I think we should, within reason, support 10-15 year old hardware, but not necessarily 10 year old operating systems when said hardware can have their OS upgraded. What is interesting with Lion though, it is a free download (since last year).

Anyway, it is fine as long as it doesn't gives us any extra work nor limits functionality. It is more the newer versions that give us pain :)

mcuee:
> Indeed, unlike Apple I think we should, within reason, support 10-15 year old hardware, but not necessarily 10 year old operating systems when said hardware can have their OS upgraded. What is interesting with Lion though, it is a free download (since last year).

> Anyway, it is fine as long as it doesn't gives us any extra work nor limits functionality. It is more the newer versions that give us pain :)

Good point. I am more concerning on the testing side. 

I think old version can be supported as long as those who find issues are willing to come out with the patch (or at least support the testing) which does not cause too much troubles.

Two more older version of interests -- I am not so sure if libusb 1.0.25 works on them or not, especially the 32bit Intel and PowerPC.
1) Mac OS X 10.6 Snow Leopard -- last version to support x86 32bit
2) Mac OS X 10.5 Leopard -- last version to support PowerPC

tormodvolden:
BTW, macports does a great job of making things build on older OS versions back to 10.6 (and PPC) using compatibility libraries and extra compilers. For building libusb the first obstacle on older systems is that we require C11, this means at least gcc 4.7 or later (or clang). On macOS 10.4 and 10.5 the standard compiler is gcc 4.0.

Out of curiosity I pulled out an old G4 MacBook with 10.4.11 and it built libusb 1.0.23 (and latest dfu-util) fine*. That's not bad.

*) just needed to remove the MAC_OS_X_VERSION_MIN_REQUIRED references, seems that macro is not defined.
Curiosly it also says during configuration:
`checking whether OS X target version is 10.12 or later... yes`

mcuee:
From here:
https://ports.macports.org/port/libusb/details/

[Monterey (arm64)](https://build.macports.org/builders/ports-12_arm64-builder/builds/41899)
[ Monterey (x86_64)](https://build.macports.org/builders/ports-12_x86_64-builder/builds/20717)
[ Big Sur (arm64)](https://build.macports.org/builders/ports-11_arm64-builder/builds/46045)
[ Big Sur (x86_64)](https://build.macports.org/builders/ports-11_x86_64-builder/builds/63355)
[ Catalina](https://build.macports.org/builders/ports-10.15_x86_64-builder/builds/90283)
[ Mojave](https://build.macports.org/builders/ports-10.14_x86_64-builder/builds/117756)
[ High Sierra](https://build.macports.org/builders/ports-10.13_x86_64-builder/builds/141265)
[ Sierra](https://build.macports.org/builders/ports-10.12_x86_64-builder/builds/177045)
[ El Capitan](https://build.macports.org/builders/ports-10.11_x86_64-builder/builds/172465)
[ Yosemite](https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/173746)
[ Mavericks](https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/178522)
[ Mountain Lion](https://build.macports.org/builders/ports-10.8_x86_64-builder/builds/79586)
[ Lion](https://build.macports.org/builders/ports-10.7_x86_64-builder/builds/88197)
[ Snow Leopard (x86_64)](https://build.macports.org/builders/ports-10.6_x86_64-builder/builds/92188)
[ Snow Leopard (i386)](https://build.macports.org/builders/ports-10.6_i386-builder/builds/58073)
 Leopard (ppc legacy)

hjelmn:
Let's just completely remove support for the old interface. GetPipePropertiesV3 was added in 10.9 which was released in 2013. Older versions of macOS can use older versions of libusb.

tormodvolden:
I don't see why actively remove support for older OS versions just because they are old, unless it is in the way for something. As mcuee showed above macports are building for many old platforms for whatever reasons so I am happy letting them do so.

mcuee:
I am more towards actively removing old OS support because of testing concerns. But in this case, probably we can still keep the support as of now since macports project is still able to support the old versions -- so they may have the means to carry out the test.

If the test fails somehow for a specific old version, and there is no easy way out, we can drop the support with a specific pull request for easy tracking in the future.

hjelmn:
@mcuee Exactly my thinking. I can not test 10.9 even at this point so it is up to packagers to ensure it works. Since they end up patching a lot anyway they can patch this support back in if they want.

Let's go ahead and take this as is but we need to think about making it clear what we (the maintainers) can actually test.  That way packagers can do the due diligence if they want to support anything outside that.

seanm:
My app supports macOS 10.9 and newer, and I can easily test that version and newer.  I too would like libusb to keep supporting old macOS versions.

I can also test even older macOS versions, with a little more work.  The oldest I still have working is 10.5 on PowerPC.  I could try it when next in the office (currently forced WFH due to government covid rules).

>just needed to remove the `MAC_OS_X_VERSION_MIN_REQUIRED` references, seems that macro is not defined

It should pretty much always be available.  Was Availability.h somehow not included?

tormodvolden:
> It should pretty much always be available. Was Availability.h somehow not included?

IIRC, I just replaced `#if MAC_OS_X_VERSION_MIN_REQUIRED > x` with `#if 0` and `#include <AvailabilityMacros.h>` was still there. Can check again, but since this was 1.0.23 on 10.4 PPC it is less important.

We might want to prefix #if MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_12 with `#if defined(MAC_OS_X_VERSION_10_12) &&` though.

seanm:
Oh I recommend never using those specific version defines, because they don't exist in old SDKs.  So instead of:

`#if MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_12`

just do:

`#if MAC_OS_X_VERSION_MIN_REQUIRED < 101200`

tormodvolden:
> just do:
> #if MAC_OS_X_VERSION_MIN_REQUIRED < 101200

Only in one location we had used MAC_OS_X_VERSION_10_12 instead of 101200. Fix added to this PR.

seanm:
The oldest I can Screen Share to right now is macOS 10.10 with Xcode 7.2.1.  This PR builds.

I confirm the compiler goes ***into*** these branches:

`#if !defined(HAVE_CLOCK_GETTIME)`

and

`#if InterfaceVersion >= 550`

Which is correct I believe.

tormodvolden:
Merged the "fix type in elapsed time calc" part. The change to newer macOS API is borderline for 1.0.26.

tormodvolden:
I hope I can assume that the duplicate GetPipeProperties() call that was introduced in commit bae927f was an oversight, and is not needed for some obscure reason? From the arguments involved, it looks like GetBusFrameNumber() and GetPipeProperties() doesn't influence each other. @hjelmn?

tormodvolden:
I will include this in 1.0.26 on the grounds that we already use this API, and this change only makes it consistently used.

Issue #1021, opened 2021-11-05
Better detect Windows version in libusb debug log


mcuee:
Right now the Windows version detection in libusb debug logs seems to have some limitations.

For my testing under Windows 10, it will only detect as Windows 8.

It is better to differentiate Windows 8, 8.1, 10 and 11 at least. Not so sure about server versions of Windows.

a-v-s:
Windows 10 needs a manifest to be able to detect the version, otherwise it will report windows 8.
However, when being called from a library, does the library or the application require the manifest?

mcuee:
It is said you have to go through WMI objects.
https://docs.microsoft.com/en-us/answers/questions/547050/win32-api-to-detect-windows-11.html

mcuee:
There are some codes here from libwdi, we can probably reuse it.
https://github.com/pbatard/libwdi/blob/master/libwdi/libwdi.c#L184

tormodvolden:
Note that we already use the same code as libwdi, I think it is broken in the same way there if there is no "application manifest".
I think just printing "8 (or later)" is as much as we can reasonably do, and should do. Doing all the WMI trickery is too much, also since we only use the version for display purpose.

If libusb is used in an application with manifest (stating support for the actual version), I believe it will show the exact version (10 or 11).

mcuee:
It seems to me Zadig works fine without the clear application manifest file.

Ref: zadig source code:
https://github.com/pbatard/libwdi/tree/master/examples

Zadig debug log example under my Windows 11 laptop.
```
Zadig 2.7.765
Windows 11 Home, 64-bit (Build 22000.434)
ini file 'zadig.ini' not found in 'C:\work\libusb' - default parameters will be used
default driver set to 'WinUSB'
0 devices found.
7 devices found.
...
``1

mcuee:
But I think pull request #1060 is correct.

Ref: https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version
> * For applications that have been manifested for Windows 8.1 or Windows 10. Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to [Targeting your application for Windows](https://docs.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1).

tormodvolden:
> It seems to me Zadig works fine without an obvious application manifest file.
> 
[...]

Issue #1049, opened 2022-01-25
MinGW 32-bit build requires -static-libgcc to avoid DLL migraine


tormodvolden:
When cross-building MinGW 32-bit (gcc 9.3 on Ubuntu 20.04) the resulting DLL depends on various gcc DLLs, so if just installing the libusb.dll the program using it will fail to execute (_"libgcc_s_sjlj-1.dll was not found"_).

Adding the extra linker option `-static-libgcc` fixes this. However, this is not needed for 64-bit builds, so it is possible that there is a bug somewhere. This is not very new, but wasn't like this some years ago either.

This would better be rechecked with a newer MinGW toolchain.

mcuee:
Ah, I see the same problem with MSYS2 with 32bit build. And yes 64bit build is not affected.

mcuee:
> This is not very new, but wasn't like this some years ago either.

It seems to me this has been there for a while. Just windering which version you check does not cause this issue. 

My GCC version is pretty new under MSYS2.

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=C:\msys64\mingw32\bin\gcc.exe
COLLECT_LTO_WRAPPER=C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/11.2.0/lto-wrapper.exe
Target: i686-w64-mingw32
[...]

PR #1060, opened 2022-02-05
windows: Add "or later" to debug print of Windows version 8 (because we can't know)

tormodvolden:
See also discussion in #1021.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,6 +42,8 @@ for:
       - cmd: msbuild "%APPVEYOR_BUILD_FOLDER%\msvc\libusb_2015.sln" /m /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
       - cmd: C:\msys64\usr\bin\bash -l "%APPVEYOR_BUILD_FOLDER%\.private\appveyor_build.sh" MinGW
       - cmd: C:\cygwin\bin\bash -l "%APPVEYOR_BUILD_FOLDER%\.private\appveyor_build.sh" cygwin
+    after_build:
+      - cmd: 7z a "libusb-build_%APPVEYOR_BUILD_WORKER_IMAGE%_%PLATFORM%_%CONFIGURATION%.7z" tag_* README-build.txt %PLATFORM%\%CONFIGURATION%\dll\*.* %PLATFORM%\%CONFIGURATION%\lib\*.* %PLATFORM%\%CONFIGURATION%\examples\*.exe %PLATFORM%\%CONFIGURATION%\tests\*.exe C:\msys64\home\appveyor\libusb-MinGW-Win32 C:\cygwin\home\appveyor\libusb-cygwin-Win32
 
   -
     matrix:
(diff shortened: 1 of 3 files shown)


mcuee (approved):
The change looks to be an improvement over the current codes. I am fine with the changes.

mcuee:
Ref: https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version
* For applications that have been manifested for Windows 8.1 or Windows 10. Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to [Targeting your application for Windows](https://docs.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1).

mcuee (approved):

tormodvolden:
@mcuee Does it display "11" for you now?

(Never mind the appveyor stuff, I just tested it in the same CI run)

mcuee:
> @mcuee Does it display "11" for you now?

Not yet. How do I add the manifest file? Without the manifest file it still shows Windows 8.

tormodvolden:
I guess the manifest must be declared in the .vcxproj as is done in libwdi. But maybe you can take an application that already has a Windows 11 manifest, and run-time link with your new libusb.dll?

I have now updated the build version threshold to be 20000, like in libwdi. I also added a candidate for #1049 on top because then I can check in the appveyor artifacts that it worked correctly there, and we can test everything with one build.

mcuee:
Adding the manifest file seems to be simpler than what I think.
I just create a simple manifest file from the libwdi example and include it in the project (libusb-1.0.dll, listdevs and xusb). And now it can print the Windows versions.

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1"> 
    <!-- If you don't have those in your app, Windows 10's GetVersionEx() -->
    <!-- DELIBERATELY reports the WRONG version... which brings us to     -->
    <!-- tonight's spelling bee: "How do you spell 'spineless morons'?"   -->
    <!-- The answer: "M-I-C-R-O-S-O-F-T; 'spineless morons'!"             -->
    <application>
      <!-- Windows 10 -->
      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
      <!-- Windows 8.1 -->
      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
      <!-- Windows 8 -->
      <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
      <!-- Windows 7 -->
      <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
      <!-- Windows Vista -->
      <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/> 
    </application> 
  </compatibility>
</assembly>

```

mcuee:
I think the pull request is now good to be merged.

The only question now is whether we want to include the manifest file in the libusb-1.0.dll build or not. It is trivial for the VS build. It seems to be possible with the MinGW as well.

Ref: https://stackoverflow.com/questions/33000158/embed-manifest-file-to-require-administrator-execution-level-with-mingw32

Ref: git has similar manifest file.
https://github.com/git/git/blob/master/compat/win32/git.manifest

tormodvolden:
I think the manifest belongs to the application but it would be an option to add one to our examples/ builds.

mcuee:
> I think the manifest belongs to the application but it would be an option to add one to our examples/ builds.

I agree with you. I think you can merge this pull request. Then we can consider the option to add the manifest to probably the xusb example in the future.

tormodvolden:
Merged.

PR #1067, opened 2022-02-10
Various static analysis fixes

benzea:
Most of it are just changes to shut up static analysers. The only functional change should be the removal of a sleep workaround if a device file cannot be opened. It seems possible for this to happen with `netlink` monitoring, but shouldn't be the case when using `udev`.


benzea:
> Did you do a full Coverity analysis and it didn't find any real issues? And a whole lot of false positives? It does seem like Coverity needs fixing instead :)

Not quite sure what settings we throw at it by default. There were some real issues, but nothing major AFAICT:
 * The `nanosleep` really is wrong, and I just disabled it for `udev` which is what we care about
 * One of the assertions in the examples is actually a real bug. But, this is "just" a buffer overflow when reading a firmware file

> But when it goes as far as where you write that this will make the code less efficient, and we'll willingly allow this just to satisfy someone's third-party static analyzer, isn't this going too far?

I don't think we are talking about any relevant overhead in the patches. The biggest slowdown is going to be 1d576b41cfe229cbf98a3fc9aeb819562cccaaae. But, it just removes a fast-path in the case that the user is seriously misusing the API. Scenarios where this could even happen are rather obscure, e.g. when an too old `libusb.so is loaded at runtime after `enum libusb_option` was extended.

tormodvolden:
Alright, I didn't look too close at the code in https://github.com/libusb/libusb/commit/1d576b41cfe229cbf98a3fc9aeb819562cccaaae It does indeed require some effort and run-time linking to run into that case. But it seems natural to fend off any wrong input values before accessing `default_context_options[option]`. What about removing
`case LIBUSB_OPTION_MAX:`
instead, does that please Coverity? Coverity is certainly right there is something unfinished there, and one of the two should be removed.

The nanosleep with udev was introduced in commit c59d574b with reference to a race with the kernel. Are you sure this cannot happen? It reminds me of an issue I have heard of* where a device node is created and udev rules being applied later, although that wouldn't be the case here, I suppose, with ENOENT.

*) https://stackoverflow.com/questions/39524234/bug-with-writing-to-file-in-linux-sys-class-gpio

benzea:
> Alright, I didn't look too close at the code in [1d576b4](https://github.com/libusb/libusb/commit/1d576b41cfe229cbf98a3fc9aeb819562cccaaae) It does indeed require some effort and run-time linking to run into that case. But it seems natural to fend off any wrong input values before accessing `default_context_options[option]`. What about removing `case LIBUSB_OPTION_MAX:` instead, does that please Coverity? Coverity is certainly right there is something unfinished there, and one of the two should be removed.

I am sure it'll please coverity. I was concerned about it producing other warnings about switch statements that don't test all enum values or don't have a default clause, so I didn't try that.

> The nanosleep with udev was introduced in commit [c59d574](https://github.com/libusb/libusb/commit/c59d574b211bedcab951b2f19de58bb03a04d671) with reference to a race with the kernel. Are you sure this cannot happen? It reminds me of an issue I have heard of* where a device node is created and udev rules being applied later, although that wouldn't be the case here, I suppose, with ENOENT.
> 
> *) https://stackoverflow.com/questions/39524234/bug-with-writing-to-file-in-linux-sys-class-gpio

So, what should happen is that:
 * kernel sends netlink event
 * udev creates device
 * udev sends its event
 * libusb processes the event

This is indeed race free. Now, if libusb is using the kernel `netlink` notification directly, then it is obviously racing with `udev`.

> Oh, and to your comment about "fixing" it by removing netlink: It seems netlink is still useful because udev is too slow on some systems #237.

udev might be doing a lot of things before reporting the device to userspace, so maybe it is just being slow, a log from `udevadm monitor` would show whether that is the case. I suppose that could be caused by a broken udev rule possibly (or complicated things like usb_modeswitch actually doing something, but I doubt that here).

libusb/core.c
```diff
@@ -2641,16 +2641,13 @@ static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
 		header_len = 0;
 	}
 
-	text_len = vsnprintf(buf + header_len, sizeof(buf) - (size_t)header_len,
+	text_len = vsnprintf(buf + header_len,
+		sizeof(buf) - (size_t)header_len - (int)sizeof(USBI_LOG_LINE_END),
 		format, args);
-	if (text_len < 0 || text_len + header_len >= (int)sizeof(buf)) {
+	if (text_len < 0 || text_len + header_len + (int)sizeof(USBI_LOG_LINE_END) >= (int)sizeof(buf)) {
```
windfine, line 2647: No need to reassign text_len when text_len + header_len + (int)sizeof(USBI_LOG_LINE_END) == (int)sizeof(buf).

tormodvolden:
> I am sure it'll please coverity. I was concerned about it producing other warnings about switch statements that don't test all enum values or don't have a default clause, so I didn't try that.

You were right about that, of course. I tried removing the LIBUSB_OPTION_MAX case and now GCC complains
`warning: enumeration value ‘LIBUSB_OPTION_MAX’ not handled in switch [-Wswitch-enum]`

libusb/libusbi.h
```diff
@@ -807,7 +812,8 @@ struct usbi_event_source {
 
 int usbi_add_event_source(struct libusb_context *ctx, usbi_os_handle_t os_handle,
 	short poll_events);
-void usbi_remove_event_source(struct libusb_context *ctx, usbi_os_handle_t os_handle);
+void usbi_remove_event_source(struct libusb_context *ctx, usbi_os_handle_t os_handle)
```
seanm, line 826: This commit lgtm, but it seems you sometimes put the annotation before and sometimes after. Does it indeed accept either way?  Might be nice for consistency to always do it one way or the other.

libusb/libusbi.h
```diff
@@ -222,6 +222,10 @@ static inline void list_add_tail(struct list_head *entry,
 LIBUSB_NONNULL(1)
 static inline void list_del(struct list_head *entry)
 {
+#ifndef NDEBUG
```
seanm, line 225: The NDEBUG check is not needed, the `assert` macro does that.

libusb/libusbi.h
```diff
@@ -257,7 +257,14 @@ static inline void list_splice_front(struct list_head *list, struct list_head *h
 
 static inline void *usbi_reallocf(void *ptr, size_t size)
 {
-	void *ret = realloc(ptr, size);
+	void *ret;
+
+	if (size == 0) {
```
seanm, line 262: Hmmm, I'm not convinced about the semantics you describe.  It doesn't match FreeBSD for example:

https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdlib/reallocf.c

would doing exactly what FreeBSD does pacify your tool?

libusb/core.c
```diff
@@ -2202,10 +2202,6 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
 		return r;
 	}
 
-	if (option >= LIBUSB_OPTION_MAX) {
```
seanm, line 2204: This removal changes behaviour though... if you don't get into `if (NULL == ctx)` then `return LIBUSB_ERROR_INVALID_PARAM` no longer happens.

If you remove `case LIBUSB_OPTION_MAX:` but leave the `default` would that shut it up?

libusb/core.c
```diff
@@ -2641,16 +2641,13 @@ static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
...
-		 * that the output was truncated. */
-		text_len = (int)sizeof(buf) - header_len;
-	}
-	if (header_len + text_len + (int)sizeof(USBI_LOG_LINE_END) >= (int)sizeof(buf)) {
-		/* Need to truncate the text slightly to fit on the terminator. */
-		text_len -= (header_len + text_len + (int)sizeof(USBI_LOG_LINE_END)) - (int)sizeof(buf);
+		 * that the output was truncated (e.g. glibc < 2.1). */
+		text_len = (int)sizeof(buf) - header_len - (int)sizeof(USBI_LOG_LINE_END);
 	}
 	strcpy(buf + header_len + text_len, USBI_LOG_LINE_END);
```
seanm, line 2652: While this is being refactored, could this final strcpy get folded into the vsnprintf too?  Although the code is probably correct, strcpy is evil since it does not know the size of the buffer, vsnprintf is thus much safer.

libusb/core.c
tormodvolden: That's what I did in commit cd8078d but now we have a GCC warning instead...

seanm:
@benzea there are a lot of great fixes here, are you going to continue working on this PR?

libusb/libusbi.h
benzea: Well, I put it after function declarations, but in front for definitions.

Obviously this worked fine for GCC at least, but I personally don't really care as to the position of these as long as the annotations work.

libusb/libusbi.h
benzea: Likely yes. If you prefer that, then I would say pull in the code and see if anyone complains about it in the future.

libusb/core.c
benzea: Exactly. You either get complaints about not covering all enum values in the switch, or you have dead code.

The only behaviour change is that the context is resolved though. I do think that this is reasonable, in particular for something that is obviously a programming/dependency version error. Like, it would almost be reasonable to just assert() here.

benzea:
As said elsewhere, I am not working at RH anymore, and as such I need to do these updates in my spare time. So, I am OK with doing some updates, but I am not going to be able to do e.g. verification whether alternative approaches fix the warnings (though, some poor person in RH will need to drop these patches from the RH package once merged, and will then see whether it works ...).

libusb/libusbi.h
benzea: Hmm, I just looked at it, and was about to change it. And then, I had this distinct feeling that the freebsd code would still trigger that warning.

Not sure, I don't really care. My take is, please just drop the patch if you don't like it as-is and put in the alternative version.

benzea:
Oh, since when do I need to sign commits here? Well, that requires me to do some digging for my key and stuff :-/

mcuee:
@benzea 

Thanks a lot for the updates.
No need to sign the commits.

libusb/libusbi.h
seanm: OK I just tried clang and it seems to *require* it in front for definitions, as you did.  And for declarations/prototypes, it accepts before and after.

So this is ok as is.

libusb/libusbi.h
seanm: I also note that C23 has now clarified that passing 0 as size to `realloc` is undefined behaviour, see https://en.cppreference.com/w/c/memory/realloc

But of course `reallocf` is not part of C, it's from FreeBSD (and also in macOS). Which is why I think we'd be much better off exactly copying FreeBSD's implementation.

I can do that in a separate PR if you want to just drop this commit from this PR (which already has merge conflicts apparently).

seanm: Hey, wait, `usbi_reallocf` is not public API, and nothing in libusb even uses it.  Maybe we can just remove it...

libusb/os/linux_usbfs.c
```diff
@@ -1438,7 +1438,7 @@ static void op_close(struct libusb_device_handle *dev_handle)
 
 	/* fd may have already been removed by POLLERR condition in op_handle_events() */
 	if (!hpriv->fd_removed)
-		usbi_remove_event_source(HANDLE_CTX(dev_handle), hpriv->fd);
+		usbi_remove_event_source(DEVICE_CTX(dev_handle->dev), hpriv->fd);
```
seanm, line 1441: Personally, I'm not a fan of this change.  The old formulation is more clear in its intent.

You said an explicit null check did not pacify the tool, but what about a `LIBUSB_NONNULL(1)` attribute on op_close?

libusb/core.c
seanm: `option >= LIBUSB_OPTION_MAX` is possible if some old libusb dynamic library is installed, and some new application uses some new future enum value. You'd want to `return LIBUSB_ERROR_INVALID_PARAM` in that case I think, which in master is done both by the early check and the later switch statement's default case.

If you remove the early check, as in this commit, then like I said there is a behaviour change (some additional code is run), but I'm not sure if it matters or not.

libusb/libusbi.h
benzea: Hah, even better then, I suppose :-)

libusb/os/linux_usbfs.c
benzea: Hmm, yes, that might work. Maybe just do that and see if anyone complains in the future? Or just drop this :-)

I am not in a position to rerun the checker anymore. I was doing this work for the RHEL libusb package but I am not with Red Hat anymore. As such, the work I can do on this is limited, but I am happy to do a quick rebase to drop the commits that don't seem quite right.

Though, also feel free to just cherry-pick whatever you like, either push to this PR or open a new one.

seanm:
@tormodvolden I could pick this up if @benzea doesn't want to, but only after all my other patches are merged. I have too much WIP libusb stuff lying around currently.

PR #1068, opened 2022-02-10
windows: Link gcc helpers statically on MinGW

tormodvolden:
This has been tested to do what it should, only add -static-libgcc to the linker flags when building the final libusb.dll on MinGW. There are however an infinite number of ways to do this with autoconf, so style and consistency could be reviewed. I am adding EXTRA_LDFLAGS  instead of using (abusing?) LT_LDFLAGS which I suspect is used for non-libtool flags some other places. OTOH for instance EXTRA_CFLAGS is only used locally in configure.ac and put into AM_CFLAGS directly. I am not using AM_LDFLAGS if that is a thing, and I think that is fine since the extra flag should only be applied to the DLL linking. Maybe a more specific variable name could be used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -124,6 +124,7 @@ case $host in
 	platform=windows
 	test "x$enable_shared" = xyes && create_import_lib=yes
 	EXTRA_CFLAGS="-mwin32 -fno-omit-frame-pointer"
+	EXTRA_LDFLAGS="-static-libgcc"
 	;;
 *)
 	AC_MSG_RESULT([Null])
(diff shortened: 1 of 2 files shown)


mcuee (approved):

mcuee:
Test results with MSYS2 32bit compiler:
```

MINGW32 /mingw32/bin
$ objdump -x libusb-1.0.dll | grep DLL
        DLL
 vma:            Hint    Time      Forward  DLL       First
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
```

mcuee:
I think this is safe to be merged.

tormodvolden:
Yes, I was not worried about it not working, but I hoped a autoconf stylist could chime in.

tormodvolden:
Merged in commit f9162d3

Issue #490, opened 2018-11-05
WinUSB backend stutters on bulk IN requests and RAW_IO pipe policy


bsccara:
I'm working on a Java app that uses LibUsb 1.0.22 through Usb4Java 1.3.0. I need to get data through bulk transfers from a device that implements the WinUSB spec. Since the data is streamed from the device at the maximum possible speed I keep 4 IN transfers in-flight at all times. The PC (same for both OS) has a A4-5000 CPU, so its no powerhouse, and the device uses Full Speed, so no speed demon either.
When running on Kubuntu 18.04 I can sustain the maximum throughoutput, 12~13 64-byte packets per frame, continuously. But on Windows 8.1 I'm getting periods of around 1 frame with no USB traffic, apart from SOFs, after the end of most transfers (each 128K). This was checked on the wire, not on the PC.
Reading the WinUSB spec I noticed that SetPipePolicy can accept a RAW_IO policy. From my understanding if that policy is not set WinUSB will queue all IN requests and feed them one at a time to the lower layers, therefore negating the purpose of having multiple requests in-flight. The docs claim that setting that policy makes it pass the requests along immediately, prevent that issue.
Since LibUSB's WinUSB backend does not set that policy does that mean that the code cannot support RAW_IO or can I enable it some other way on my app (through JNA probably) ?

mcuee:
You may want to ask in libusb-devel mailing list to get better help.

mcuee:
https://marc.info/?t=145424689900001&r=1&w=2

>From Tim
I think a good argument could be made that libusb should set RAW_IO unconditionally.  Libusb itself does much of the buffer management that gets triggered in WinUSB when RAW_IO is turned off

mcuee:
Still it will make libusb Windows which is already complex even more complex.
>From Chris
Using RAW_IO introduces certain complexities for libusb. In
particular, if the provided transfer buffer is not a multiple of the
endpoint max packet size, we have two options:

1) Round the buffer size down to a max packet size multiple. If the
buffer is smaller than the endpoint max packet size, libusb has to
allocate a buffer.
2) Round the buffer size up to the next max packet size multiple and
allocate a buffer of that size.

Now if we have to allocate and the transfer completes with more data
than the user's buffer will allow, we have to keep the excess data
until the next transfer.

Then there is the case where the user's buffer is greater than
MAXIMUM_TRANSFER_SIZE. Now the Windows backend is getting into the
business of breaking up the transfer into multiple smaller ones.

[...]

Issue #149, opened 2016-02-25
Windows: set RAW_IO WinUSB Pipe Policy unconditionally


mcuee:
Ref: http://marc.info/?l=libusb-devel&m=145528312815832&w=2
Ref: https://msdn.microsoft.com/en-us/library/windows/hardware/ff728833(v=vs.85).aspx

On Mon, Feb 1, 2016 at 10:01 AM, Tim Roberts timr@probo.com wrote:

> I think a good argument could be made that libusb should
> set RAW_IO unconditionally.  Libusb itself does much of the buffer
> management that gets triggered in WinUSB when
> RAW_IO is turned off.

mcuee:
From further discussion this may not be a good idea. Close this for now.

PR #1069, opened 2022-02-11
Improve WinUSB performance by enabling RAW_IO when possible

PetteriAimonen:
Related issues:
* https://github.com/libusb/libusb/issues/149
* https://github.com/libusb/libusb/issues/490

Without RAW_IO, WinUSB backend only submits one transfer request at a time. The duration to switch transfers causes lost data when capturing from devices with small internal buffer, such as FX2 based devices.

<s>The problem with enabling RAW_IO unconditionally is that it requires transfer size to be divisible by maximum packet size. This patch enables it only if the transfer size is divisible by 512, which is divisible by all USB maximum size values (16, 32, 64, 512).</s>

Updated version of patch enables RAW_IO only if option `LIBUSB_OPTION_WINUSB_RAW_IO` is enabled.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2230,6 +2230,7 @@ int API_EXPORTED libusb_set_option(libusb_context *ctx,
 		/* Handle all backend-specific options here */
 	case LIBUSB_OPTION_USE_USBDK:
 	case LIBUSB_OPTION_NO_DEVICE_DISCOVERY:
+	case LIBUSB_OPTION_WINUSB_RAW_IO:
 		if (usbi_backend.set_option)
 			return usbi_backend.set_option(ctx, option, ap);
 
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -2123,10 +2123,16 @@ enum libusb_option {
 	 * Only valid on Linux.
 	 */
 	LIBUSB_OPTION_NO_DEVICE_DISCOVERY = 2,
-
 #define LIBUSB_OPTION_WEAK_AUTHORITY LIBUSB_OPTION_NO_DEVICE_DISCOVERY
 
-	LIBUSB_OPTION_MAX = 3
+	/** Use WinUSB RAW_IO policy on Windows.
+	 * Improves performance by allowing the backend to queue multiple transfer
+	 * requests, the same way it works by default on other platforms.
+	 * This requires transfer sizes to be a multiple of maximum packet size.
+	 */
+	LIBUSB_OPTION_WINUSB_RAW_IO = 3,
+
+	LIBUSB_OPTION_MAX = 4
 };
 
 int LIBUSB_CALL libusb_set_option(libusb_context *ctx, enum libusb_option option, ...);
--- a/libusb/os/windows_common.c
+++ b/libusb/os/windows_common.c
@@ -614,6 +614,10 @@ static int windows_set_option(struct libusb_context *ctx, enum libusb_option opt
 		usbi_dbg(ctx, "switching context %p to use UsbDk backend", ctx);
 		priv->backend = &usbdk_backend;
 		return LIBUSB_SUCCESS;
+	} else if (option == LIBUSB_OPTION_WINUSB_RAW_IO) {
+		usbi_dbg(ctx, "enabling RAW_IO for context %p", ctx);
+		priv->use_raw_io = true;
+		return LIBUSB_SUCCESS;
 	}
 
 	return LIBUSB_ERROR_NOT_SUPPORTED;
--- a/libusb/os/windows_common.h
+++ b/libusb/os/windows_common.h
@@ -340,6 +340,7 @@ struct windows_context_priv {
 	const struct windows_backend *backend;
 	HANDLE completion_port;
 	HANDLE completion_port_thread;
+	BOOL use_raw_io;
 };
 
 union windows_device_priv {
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2505,6 +2505,8 @@ static void winusbx_close(int sub_api, struct libusb_device_handle *dev_handle)
 
 static int winusbx_configure_endpoints(int sub_api, struct libusb_device_handle *dev_handle, uint8_t iface)
 {
+	struct libusb_context *ctx = HANDLE_CTX(dev_handle);
+	struct windows_context_priv *ctx_priv = usbi_get_context_priv(ctx);
 	struct winusb_device_handle_priv *handle_priv = get_winusb_device_handle_priv(dev_handle);
 	struct winusb_device_priv *priv = usbi_get_device_priv(dev_handle->dev);
 	HANDLE winusb_handle = handle_priv->interface_handle[iface].api_handle;
@@ -2546,6 +2548,12 @@ static int winusbx_configure_endpoints(int sub_api, struct libusb_device_handle 
 		if (!WinUSBX[sub_api].SetPipePolicy(winusb_handle, endpoint_address,
 			AUTO_CLEAR_STALL, sizeof(UCHAR), &policy))
 			usbi_dbg(HANDLE_CTX(dev_handle), "failed to enable AUTO_CLEAR_STALL for endpoint %02X", endpoint_address);
+
+		if (ctx_priv->use_raw_io) {
+			if (!WinUSBX[sub_api].SetPipePolicy(winusb_handle, endpoint_address,
+				RAW_IO, sizeof(UCHAR), &policy))
+				usbi_dbg(HANDLE_CTX(dev_handle), "failed to enable RAW_IO for endpoint %02X", endpoint_address);
+		}
 
 		if (sub_api == SUB_API_LIBUSBK) {
 			if (!WinUSBX[sub_api].SetPipePolicy(winusb_handle, endpoint_address,


diabolo38:
It's  good that this PR make winusb back end faster and  behave more like other OS one. But doing it blindly may create regression by changing legacy behavior. It could brake existing application mixing synchronous and or  asynchronous transfer from multiple thread where order of completion might be important .

mcuee:
Please help to split the Windows github action to another pull request. Thanks.

mcuee:
I actually think it is better to use libusb_set_option to set up the RAW_IO pipe policy when needed.

mcuee:
> @mcuee Does this look reasonable? [devEmbedded@1f10388](https://github.com/devEmbedded/libusb/commit/1f10388a04bb17c63848f3ca8ffa8338cc50b428)

This looks better. 

@tormodvolden Please take a look at the above proposal as well. There are multiple issues with WinUSB isoc implementations. Hopefully this will be a positive move.

tormodvolden:
Thanks, this is interesting. What does the Windows documentation say about enabling RAW_IO? For instance why isn't it used by default in the OS (if transfer size allows)? A new option seems like the safe alternative, but if this is already how it works on other platforms, it can be further discussed.

About the github actions, can it also build with MinGW? Our current Appveyor CI builds both VS and MinGW, alternatively we could strip the Appveyor job to only build MinGW.

PetteriAimonen:
@tormodvolden The relevant documentation is [here](https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification) and old libusb discussions [here](https://marc.info/?t=145424689900001&r=1&w=2).

I also think that it should be safe to enable by default if transfer size allows, and that wouldn't require Windows-specific hacks in applications. But it will change timing of USB packets, so it is difficult to rule out regressions with some applications like @diabolo38 mentions.

I'm not sure if there is a way to set per-endpoint options in libusb currently?

Regarding GitHub action, I didn't actually notice the Appveyor CI because that doesn't work on forked repositories. The [GitHub Win2019 runner](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md) does have MinGW installed so building with it should be possible also. But the Appveyor CI builds with many different Visual Studio versions and I'm not sure if that is possible with the GitHub runner.

tormodvolden:
Thanks for the links! Can this change break applications though, if the same code works on Linux and macOS (where multiple transfers already are allowed)?

And no, there is no per-endpoint option mechanism in libusb, hence "ideally"...

mcuee:
> Thanks for the links! Can this change break applications though, if the same code works on Linux and macOS (where multiple transfers already are allowed)?

The change request does not affect Linux and macOS as the codes changed are only for Windows.

If we look at the patches here, it will also not affect existing codes since it is only enabled when  LIBUSB_OPTION_WINUSB_RAW_IO is set.

Ref: 
https://github.com/devEmbedded/libusb/commit/1f10388a04bb17c63848f3ca8ffa8338cc50b428

tormodvolden:
That was not what I meant. I mean the same application code, if it works with Linux and macOS, it probably would work with the modified libusb Windows code as well (if I understand correctly that the order of completion is not guaranteed in Linux and macOS). So any regression would only happen on code that worked on Windows but was broken on Linux and macOS.

PetteriAimonen:
I'm not aware of any case where enabling RAW_IO should cause problems for properly written applications. But it is hard to prove a negative. And of course the timings may vary, so a buggy application might work with luck on Linux but fail with RAW_IO on Windows.

mcuee:
Maybe this is good to have for the future 1.0.27 release if there are more reviewers jumping in. Not so sure when @dickens will have the time to come back.

mcuee:
I am think this will be good for isoc transfer as well -- just need to port the code to isoc transfer.

martinling (changes requested):
I don't think this approach is safe.

I don't get any impression from the [documentation](https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification) on this topic that [SetPipePolicy](https://docs.microsoft.com/en-us/windows/win32/api/winusb/nf-winusb-winusb_setpipepolicy) is intended to be called whilst transfers are in flight. Usually a program will call `SetPipePolicy` to set up an endpoint at the start, before submitting any transfers. In libusb, this happens in `winusbx_configure_endpoints`.

With the change in this PR, libusb would be changing the pipe behaviour whilst existing transfers are already in flight. What happens when a transfer started with RAW_IO is ongoing and the policy is changed to disable RAW_IO, or vice versa? There is nothing in the documentation that says what happens, or that suggests this usage is safe.

There is already one [call](https://github.com/libusb/libusb/blob/f341e4a086edd4f45c2db4df892a52ebfd52ac5b/libusb/os/windows_winusb.c#L3108) to  `SetPipePolicy` in `winusbx_submit_bulk_transfer`. This is used to apply the `LIBUSB_TRANSFER_ADD_ZERO_PACKET` transfer flag by changing the `SHORT_PACKET_TERMINATE` flag on the pipe policy, at submission time. But that code includes the comment:
```
// Changing this can be a problem with packets in flight, so only allow on the first transfer.
```
If the `LIBUSB_TRANSFER_ADD_ZERO_PACKET` differs on a subsequent transfer, the WinUSB backend will return `LIBUSB_ERROR_NOT_SUPPORTED` on submission, rather than risk changing the pipe policy on the fly.

In the case of that flag, libusb can at least make a reasonable guess which policy is required based on the first transfer. But it's not safe to infer whether RAW_IO can be applied based on the size of the first transfer. For programs that always submit appropriately sized transfers, it will work fine, but for a program that submits e.g. 512 bytes followed by 123 bytes, libusb would guess wrong, and then be stuck needing to change the policy with a transfer already in flight.

So I don't think any version of this automatic approach can work. I think the best way to integrate RAW_IO is going to be the `libusb_set_option` approach in https://github.com/devEmbedded/libusb/commit/1f10388a04bb17c63848f3ca8ffa8338cc50b428.

mcuee:
Glad that there are more discussions here. Hopefully a better solution can be proposed from that discussion.
* https://github.com/libusb/libusb/issues/490

mcuee:
So it seems that this pull request is not working.
The approach in https://github.com/devEmbedded/libusb/tree/dev_raw_io is the way to go.

@PetteriAimonen Please help to change the pull request. Thanks. Please also help to fix the github warning of unsigned commits.

PetteriAimonen:
@mcuee Can you be more specific on the changes needed? Just a pull request of the other branch, or the more elaborate option mechanism with context & all?

mcuee:
I am thinking just the other branch with libusb_set_option().

mcuee:
@tormodvolden How do you like this one to progress?

tormodvolden:
I think what is meant is that you push the one commit from the other branch here (replacing the old branch contents), ideally rebased on latest master while we're at it.

tormodvolden:
Personally I also liked the variant (discussed somewhere else?) to have this as a transfer flag. Yes, it will leave it to the user to keep the flag consistently on the same pipe, but it will also allow different flags for different pipes within the same libusb context. But I understand we would need to call SetPipePolicy() once from libusb_submit_transfer() and keep track of this which could get ugly too.

PetteriAimonen:
I have updated the pull request as requested. I also took out the [Windows MSVC GitHub workflow](https://github.com/devEmbedded/libusb/blob/dev_raw_io_2_old/.github/workflows/windows.yml) from this pull request, if wanted I can submit it separately.

I think it could make sense to add this simple global option first, and if someone has an actual usecase for per-transfer flags, it can be expanded in the future. In my use case the global option would be easier to use through Python bindings.

tormodvolden:
Great. Please file the MSVC GitHub workflow separately. Maybe with some of the elements from #1116 it could replace the appveyor build altogether.

As a library, we just have to be very careful introducing any new feature, because it is a pain to transition to an improved feature. I must admit I am also slightly biased against using the global context options because there are so many issues with our mechanism for them. It might get better with #1026.

PetteriAimonen:
One point that speaks in favor of global context option is that sigrok, a library that interfaces with a lot of different USB devices, has been running with libusb fork that has RAW_IO unconditionally enabled for years now. In my opinion that shows in the vast majority of cases, there is no need to have endpoint granularity.

I haven't yet heard of any real-world application needing per-endpoint setting of RAW_IO.
Forcing users to a more complex mechanism for the common case seems unwise, especially as varargs are not compile-time checked and could easily lead to crashes if wrong parameters are passed.

tormodvolden:
I question a bit "a lot of different USB devices" because when sigrok came up in discussion lately it turned out all these "different devices" were using the same USB chip.

> especially as varargs are not compile-time checked and could easily lead to crashes if wrong parameters are passed.

This with varargs speaks in favour of which solution?

PetteriAimonen:
> when sigrok came up in discussion lately it turned out all these "different devices" were using the same USB chip.

This is a misunderstanding.
It is true that most (but AFAIK not all) devices that absolutely required `RAW_IO` are based on the Cypress FX2 chipset.
But sigrok supports many other devices also and uses `RAW_IO` for all devices, whether they would work without it or not.

> This with varargs speaks in favour of which solution?

AFAIK it was suggested that the per-endpoint granularity would be implemented by passing the pointers to the device and endpoint as a varargs argument to `libusb_set_option`. If there is a mistake when doing this, it will crash only on Windows, only when using WinUSB backend. If one e.g. forgets the arguments from the function call, it will read random pointers from stack. It is true that they would be application bugs, but as a libusb user, I would prefer to have a simple option for my needs, and a `LIBUSB_ENABLE_RAWIO_PER_ENDPOINT` for the extended one, if anyone actually needs it.

In my opinion, we could just wait for someone that has the need to both be able to enable RAWIO for some devices **and** not enable it for other devices. So far it seems that for many devices it doesn't matter whether it is on or off.

tormodvolden:
> This is a misunderstanding. It is true that most (but AFAIK not all) devices that absolutely required `RAW_IO` are based on the Cypress FX2 chipset. But sigrok supports many other devices also and uses `RAW_IO` for all devices, whether they would work without it or not.

I see. That is reassuring.

> AFAIK it was suggested that the per-endpoint granularity would be implemented by passing the pointers to the device and endpoint as a varargs argument to `libusb_set_option`.

Aha. Yes, the libusb_set_option use of vararg is also something I don't like about it. This could be one thing in favour of using a transfer flag instead, which is just a bit mask.

PetteriAimonen:
Yes, I'm totally fine with the `libusb_transfer_flags` approach also. I'm not going to pre-emptively implement it, but I can implement it if we can reach a decision that it would be acceptable for merging.

martinling:
> One point that speaks in favor of global context option is that sigrok, a library that interfaces with a lot of different USB devices, has been running with libusb fork that has RAW_IO unconditionally enabled for years now. In my opinion that shows in the vast majority of cases, there is no need to have endpoint granularity.

I disagree with this.

I think it's pretty likely that we have drivers in libsigrok which don't actually satisfy the RAW_IO buffer requirements, but nobody has noticed because testing is so patchy. There are hundreds of different devices supported, many of them old or rare. Nobody has all of them available to test, and most of the developers and users don't use Windows.

In fact just at a glance I see several drivers that clearly don't meet the RAW_IO requirements: `kecheng-kc-330b`, `hantek-dso`, `lecroy-logicstudio`, `sysclk-lwla`, `saleae-logic-pro`... that's as far as I got down the list, I haven't done an exhaustive audit. At the moment, using any of those on the sigrok Windows builds should fail, but either nobody has tested, or nobody has submitted a bug. Actually in some cases I think they have: https://sigrok.org/bugzilla/show_bug.cgi?id=1428 could very well be a result of this and there are probably more.

What should happen is that those drivers should just have RAW_IO disabled, but that can't happen right now because sigrok just patches libusb to unconditionally enable it on the Windows builds so that other popular drivers go fast.

If libusb adds an option for RAW_IO but only at the context level, then sigrok will still have the same problem, because it uses a shared context for the whole library. So drivers that aren't RAW_IO safe still won't be able to be fixed, without breaking performance for other drivers.

If the option were at least per-device then it could be set in each driver. But in many devices there are separate endpoints for configuration and data, and it makes sense to use small odd-size transfers on the configuration endpoints whilst using large, RAW_IO-safe ones on the data endpoints.

So I maintain my position that the option should be per-endpoint - which corresponds to its place in the Windows API.

I don't accept the argument that we should exclude that solution just because users might call a variadic function incorrectly. The library already accepted that risk when adding a variadic function to the API in the first place. If libusb does not trust callers to use that function correctly then it could add a `libusb_set_endpoint_option` function. Either way, this API detail shouldn't be what decides what the granularity of the setting should be.

The transfer flags approach could work, but it would be the same messy situation as `LIBUSB_TRANSFER_ADD_ZERO_PACKET`, where it can be used only if it is set on the first transfer on an endpoint and then kept the same on all subsequent transfers, with `LIBUSB_ERROR_NOT_SUPPORTED` being returned on submission if it differs for any later transfer. But at least in that case it makes sense to use a flag, because on Linux the flag *can* be set per-transfer. The RAW_IO option cannot.

mcuee:
@tormodvolden What is your opinion here? I am more leaning towards Option 4 or Option 6.

benzea:
OK, just a side comment from someone who doesn't really care about windows. I see two restrictions:
 1. The buffer length must be a multiple of the maximum endpoint packet size.
 2. The length must be less than or equal to the value of MAXIMUM_TRANSFER_SIZE retrieved by WinUsb_GetPipePolicy.

For restriction 1. there are two possibilities:
 1. The user is providing an appropriate buffer. This can be hard to tell (maybe `libusb_dev_mem_alloc` can help?).
 2. On mismatch, libusb can provide its own buffer and copy to/from the user buffer.

I am not sure about restriction 2. I guess if you submit a (BULK) transfer that is larger than the MAXIMUM_TRANSFER_SIZE, then the request would be split by the kernel. It seems like it would be legal for libusb to either fail with `LIBUSB_ERROR_INVALID_PARAM` or handle the splitting into multiple URBs itself in the OS layer.

PetteriAimonen:
@benzea I think we are already past the point of considering option 2 ways to automatically detect if the application will stick to RAW_IO constraints. Because RAW_IO cannot be changed after first transfer is issued, that approach fails if the application then submits a transfer that does not match the conditions.

benzea:
> I think we are already past the point of considering option 2 ways to automatically detect if the application will stick to RAW_IO constraints.

Yes, you can't predict what the application will do. But it is possible to ensure the constraints are met. And, I believe it is easier than https://github.com/libusb/libusb/issues/490#issuecomment-761554688 is claiming.

As I see it, the requirement that windows makes is that the allocated buffer length is a multiple of the maximum packet size (tbh. I wonder if it fails to mention that the memory should also be aligned to some boundary, but maybe MS allocation APIs will automatically align memory of certain sizes). The important thing to note is that the "buffer_length" passed to `WinUsb_ReadPipe` can be any value smaller than `MAX_TRANSFER_SIZE`.

As such, the following approach may work:
 1. Emit as many (concurrent) `WinUsb_ReadPipe` requests with length `MAX_TRANSFER_SIZE` as possible
 2. Emit a single smaller `WinUsb_ReadPipe` transfer for the tail, when needed.
     For this, there are two ways of dealing with the buffer:
     * Use a private buffer and do memcpy. This is entirely safe and simple to do.
     * If anyone really cares about performance in this case (which I doubt), then one could ensure the buffer is of size `MAX_TRANSFER_SIZE` by:
       * adding new API for transfer/buffer allocation
       * using `libusb_dev_mem_alloc` and some bookkeeping to detect that it is a safe buffer.

martinling:
@benzea, it's not possible for libusb to just enable RAW_IO and take over management of the sizes internally.

Consider the case where libusb has set RAW_IO and the max packet size is 512 bytes.

What happens when the application submits transfer to read 100 bytes? Then libusb has to submit a `WinUsb_ReadPipe` request for 512 bytes, because it must be a multiple of the packet size. Your approach is that it should read 512 bytes from the device, return the first 100 bytes to the application, and keep the next 412 for next time.

But the device may not 512 bytes to give you! It may NAK the host's IN packets indefinitely, or even STALL. It wasn't expecting to be asked for more than the 100 bytes that the application wanted. So the application gets an error, or a timeout, or hangs.

Even if the device can give you more bytes, the following ones may not be the right ones, by the time the application requests them. Perhaps the application reads 100 bytes, then makes a control transfer that changes what will be read next from that endpoint. Then it asks for 200 bytes, and you give it garbage.

It can't be done. The application has to opt-in and promise that its transfer sizes will meet the requirements.

benzea:
> Then libusb has to submit a WinUsb_ReadPipe request for 512 bytes, because it must be a multiple of the packet size.

I read this differently. i.e. looking at the API:

```
BOOL WinUsb_ReadPipe(
  [in]            WINUSB_INTERFACE_HANDLE InterfaceHandle,
  [in]            UCHAR                   PipeID,
  [out]           PUCHAR                  Buffer,
  [in]            ULONG                   BufferLength,
  [out, optional] PULONG                  LengthTransferred,
  [in, optional]  LPOVERLAPPED            Overlapped
);
```

In my view, the `BufferLength` here is misnamed and should be `RequestLength`. The actual length of the allocated buffer can be different and longer though. With that, I read the constraints as meaning:

* The "buffer length" (i.e. allocation size) of the buffer must be a multiple of the maximum endpoint packet size.
* The "length" (i.e. request length passed to WinUsb_ReadPipe) must be less than or equal to the value of MAXIMUM_TRANSFER_SIZE retrieved by WinUsb_GetPipePolicy.

Said differently, I was assuming the following works perfectly with `RAW_IO` set:
```
  char *buf = malloc(512);
  WinUsb_ReadPipe (interface, pipe_id, buf, 100, NULL, overlapped);
```

martinling:
With regard to the options that @PetteriAimonen set out: I would be happy with either option 4 (`libusb_set_option` with varargs) or 6 (new function).

I'm opposed to option 5 (transfer flags). Firstly because it's confusing/misleading - it gives the impression this feature can be enabled per-transfer, when it can't. But more importantly, I don't like it because it's more intrusive to existing code.

It's quite common to have helper functions that create and submit transfers, which in a larger codebase might be reused over many different devices and scenario.

If I want to enable RAW_IO in a particular scenario, it should suffice to just turn it on for the endpoint in question, after making sure that all the transfer size requirements are met in practice. With option 4 or 6 I can do that.

But the transfer flag option requires digging down into those helper functions and adding the flag to the transfers they submit - but only in the scenarios I've ensured are RAW_IO-safe. If the same functions are also used for non-RAW_IO endpoints, then the code has to be refactored to pass that distinction around. All of which is unnecessary because in reality the underlying setting is on an endpoint, not a transfer.

martinling:
@benzea:
> With that, I read the constraints as meaning:
>   * The "buffer length" (i.e. allocation size) of the buffer must be a multiple of the maximum endpoint packet size.

I don't think that interpretation can be correct, because the documentation for RAW_IO says:

> WinUSB handles read requests as follows:
>   * A request that is not a multiple of the maximum endpoint packet size fails.

`WinUsb_ReadPipe` can only check the request length passed to it, not the allocation size of the buffer, so it must be the request length that it is checking to be a multiple of the max packet size, not the buffer length. Otherwise it could not fail a request that didn't meet this requirement.

benzea:
Puh, OK, seems like I must have completely misinterpreted that.

PetteriAimonen:
I suggest implementing option 4 with this interface:

     libusb_context *context = ...;
     libusb_device_handle *handle = ...;
     uint8_t endpoint = ....;
     unsigned int max_transfer_size = 0;
     libusb_set_option(context, LIBUSB_OPTION_WINUSB_RAW_IO, handle, endpoint, &max_transfer_size);

And internally behaving like this:

1. On platforms or backends other than WinUSB, does nothing.
2. With WinUSB backend, first checks that `HANDLE_CTX(handle) == context` and that the given endpoint number exists in the device. If not, returns `LIBUSB_ERROR_INVALID_PARAM`.
3. If parameters are ok, immediately calls `SetPipePolicy` on the endpoint. If `SetPipePolicy` returns `false`, `LIBUSB_ERROR_OTHER` is returned.
4. If `max_transfer_size` pointer argument is not `NULL`, `GetPipePolicy(MAXIMUM_TRANSFER_SIZE)` is called and its result stored there for information to the application.
5. It is up to the application to ensure that no transfers are ongoing when `RAW_IO` is enabled and that any future transfers on the endpoint will conform to `RAW_IO` rules.
6. There is no disable function, because the policy apparently cannot be changed on the fly anyway. If application needs to disable the policy, it must close and re-open the USB device.

martinling:
@PetteriAimonen

I like that idea, in particular because of the following features:
- Validation of the varargs to catch calls made without the required arguments. 
- Returning `MAX_TRANSFER_SIZE` as part of the call so that a separate function isn't required.

I have two issues with it as proposed, though:

1. The arguments should include a boolean, so that RAW_IO can be disabled without having to close and reopen the device.
    - There are scenarios where this will be appropriate. Consider a device that implements a serial protocol on an endpoint. To drive it you need to make a lot of small, odd-size transfers containing commands and responses. But some commands may result in large blocks of data being sent back, so when reading these, it will make sense to enable RAW_IO and then queue up multiple transfers for speed. After the large data block is finished, you then need to return to sending small, odd-size transfers for the next commands. It's necessary to turn RAW_IO off again at that point, and safe to do so - provided that no transfers are in flight on that endpoint, which is the same requirement as when enabling the option.
2. What should be written at the `max_transfer_size` pointer when this call is made on a non-WinUSB backend?
    - If the function truly "does nothing" on non-WinUSB, then it leaves this value uninitialised. The application would need to know which backend is in use to decide whether it should look at the value and limit its transfers. I think a better choice is that on non-WinUSB, a zero should be written there, and a value of zero should be documented to mean there is no limit in effect.

PetteriAimonen:
@martinling 

1. I guess that can make sense for some purposes. The enable-argument should probably be `int`, considering libusb API doesn't currently use `bool` anywhere.
2. Leaving the argument uninitialized and documenting it to be so would allow applications to do  `unsigned int max_transfer_size = some_sane_default;` instead of doing `if (max_transfer_size == 0) max_transfer_size = some_sane_default;` afterwards.

martinling:
@PetteriAimonen 
> 1. I guess that can make sense for some purposes. The enable-argument should probably be `int`, considering libusb API doesn't currently use `bool` anywhere.

Yeah, that makes sense.

> 2. Leaving the argument uninitialized and documenting it to be so would allow applications to do  `unsigned int max_transfer_size = some_sane_default;` instead of doing `if (max_transfer_size == 0) max_transfer_size = some_sane_default;` afterwards.

That's true, but it forces the application code to be structured in that way. The application may not want to apply a limit at all unless the backend imposes one. And this may be getting called from some language binding or intermediate layer which does not control the transfer sizes itself. It should be able to tell whether something was written to that output parameter or not.

From Rust for instance I would like to be able to call some method like `device.enable_winusb_raw_io(endpoint)` and get an `Option<NonZeroUsize>` with the max transfer size, which may be either `Some(size)` or `None`. I shouldn't have to pass it a size to modify.

PetteriAimonen:
For 2., the binding layer could do `unsigned int max_transfer_size = 0;` beforehand to get the behavior of "0 when unlimited". This would also work for any old version of libusb without RAW_IO support.

Which brings to question what `libusb_set_option()` should return on other platforms? Any old version will return `LIBUSB_ERROR_INVALID_PARAM`. Currently platform-specific options return `LIBUSB_ERROR_NOT_SUPPORTED` on platforms that do not support them. In my opinion if an error is returned, it shouldn't touch the arguments either. Returning `LIBUSB_SUCCESS` on other platforms feels weird also.

martinling:
Yeah, I guess it doesn't matter so much what it writes there, if we can distinguish what's happened by the return code.

I don't think it's necessary for the call to return `LIBUSB_SUCCESS` on other platforms. I think it's fine if it returns `LIBUSB_ERROR_NOT_SUPPORTED`. Application code can handle that error by proceeding without a limit. That still results in code that will run on all platforms without `#ifdef`, and makes the intent very clear.

Then the situation for the caller is simple - `max_transfer_size` is valid iff the call returns `LIBUSB_SUCCESS`.

patstew:
FWIW I've been using this varargs option for a while: https://github.com/patstew/libusb/commit/f65c74528c6b56777e7bc2d6f5851a790f382da7

martinling:
> FWIW I've been using this varargs option for a while: [patstew@f65c745](https://github.com/patstew/libusb/commit/f65c74528c6b56777e7bc2d6f5851a790f382da7)

I think this looks like a good starting point. Per the discussion above, I think the following changes are still needed:

1. It should take two further arguments: `int enable` and `int *max_transfer_size`.

   1. The `enable` argument allows disabling RAW_IO as well as enabling it.

   2. The `max_transfer_size` pointer gives the caller a way to find out what the buffer requirements are.
      - If the pointer is not NULL, the backend should call `GetPipePolicy` to retrieve the `MAXIMUM_TRANSFER_SIZE` policy. If NULL, this step can be skipped.
      - The `GetPipePolicy` call should be made before the `SetPipePolicy` call, and an error should be returned if it fails. Otherwise we could have an ambiguous situation, where the `libusb_set_option` call fails and it's unclear whether the RAW_IO setting was changed or not.
      - Before a successful return, the backend should place the retrieved value in `*max_transfer_size`. This should be done only after the `SetPipePolicy` call succeeds. This is so that a caller can safely point at a default size setting and only have it updated if required.

2. The error codes need some adjustment:
   1. If the backend is not WinUSB, then `LIBUSB_ERROR_NOT_SUPPORTED` should be returned, to ensure consistency across operating systems and backends. Portable code will want to specifically ignore a `LIBUSB_ERROR_NOT_SUPPORTED` error, as this indicates there is no RAW_IO setting needed, and that arbitrary transfer sizes can therefore be used.

   1. If the `SetPipePolicy` call fails then `LIBUSB_ERROR_IO` seems like a more appropriate return code than `LIBUSB_ERROR_ACCESS`, since there's no indication this is a permission issue.

Would folks be happy with those changes?

mcuee:
@PetteriAimonen 
Please help to review the proposal from @martinling. Thanks.

PetteriAimonen:
@mcuee I think Martin's suggestions are good. I'm not sure when I'll find time to work on this again, or whether @patstew would be interested in implementing those changes.

martinling:
I have some time available to work on implementing and testing this next week. We're going to want this feature to support an upcoming product so I'm keen to try to get it into a release.

martinling:
I've completed the revised version and submitted it as #1208.

Reviews and testing would be very much appreciated, particularly from @PetteriAimonen, @patstew, and anyone else who's been using alternative versions of this feature.

PetteriAimonen:
I'm closing this in favor of PR #1208.

PR #1071, opened 2022-02-12
Fix typo and syntax 

yegorich:
Fix a typo and add missing commas.

Also, delete a blank line at the end of the file and
use /**/ for comments consistently.

Signed-off-by: Yegor Yefremov <yegorslists@googlemail.com>

### DIFF
--- a/android/examples/unrooted_android.c
+++ b/android/examples/unrooted_android.c
@@ -24,12 +24,12 @@
 
 /*
  * This example creates a shared object which can be accessed over JNA or JNI from Java or Kotlin in Android.
- * Hint: If you are using Android Studio set the "Debug type" to "Java Only" to receive debug messages.
+ * Hint: If you are using Android Studio, set the "Debug type" to "Java Only" to receive debug messages.
  */
 
 /*
  * Usage:
- * First you have to connect your USB device from the Java side.
+ * First, you have to connect your USB device from the Java side.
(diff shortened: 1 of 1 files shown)


mcuee (approved):

tormodvolden:
Thanks. But if we start on this file, maybe also include other low-hanging fruit, like "Example" with capital E in the same phrase? A full review of the whole file would be welcome.

yegorich:
@tormodvolden done.

tormodvolden:
Great, thanks. It is useful in this case to add "in comments" in the commit message to clarify this is "cosmetic" without affecting code. But we can add that when we merge it.

tormodvolden (approved):

mcuee:
I think this can be safely merged now.

android/examples/unrooted_android.c
```diff
@@ -270,7 +270,7 @@ static void print_device(libusb_device *dev, libusb_device_handle *handle)
 }
 
 
-// fileDescriptor = is the native File Descriptor obtained in Java and transfered to native over JNA for Example.
+// fileDescriptor = is the native File Descriptor obtained in Java and transferred to native over JNA, for example.
```
tormodvolden, line 273: We should probably use /* */ instead of // for consistency.

yegorich: Done.

PR #1073, opened 2022-02-15
Fix late transfer free's

benzea:
The log handler in libusb_free_transfer now tries to get the context. However, that requires a pointer to the device, which may either be NULL already (`do_close` in `core.c` does this), or may currently even be an invalid pointer if the device for a non in-flight transfer was closed.

This PR fixes both issues.

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -329,10 +329,11 @@ void usbi_log(struct libusb_context *ctx, enum libusb_log_level level,
 #endif /* ENABLE_LOGGING */
 
 #define DEVICE_CTX(dev)		((dev)->ctx)
-#define HANDLE_CTX(handle)	(DEVICE_CTX((handle)->dev))
-#define TRANSFER_CTX(transfer)	(HANDLE_CTX((transfer)->dev_handle))
+#define HANDLE_CTX(handle)	((handle) ? DEVICE_CTX((handle)->dev) : NULL)
 #define ITRANSFER_CTX(itransfer) \
-	(TRANSFER_CTX(USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer)))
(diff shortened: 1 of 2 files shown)


tormodvolden:
We'd sure love to have some automatic testing with device emulation.

mcuee:
> Now, I was wondering how big the interest is in having umockdev based tests inside libusb proper. Note that there is the caveat that the `poll` behaviour is not properly emulated.

Yes, for sure that will be very good.

benzea:
Eeks. This breaks the synchronous API. It uses `transfer->dev_handle == NULL` to decide whether the device was closed in the meantime, rather than the transfer completing successfully.

Not sure what to do about that right now. The straight forward but ugly thing would to track all transfers in dev_handle that were not free'ed yet. Then one can clear dev_handle in device close.

benzea:
Or, well, we change the check in `sync_transfer_wait_for_completion`. But I am not sure in how far this is an API break.

benzea:
OK, here is the fancy patch, which ensures that `libusb_free_transfer` has the device while and therefore the context when running.

The simple version is to just replace `TRANSFER_CTX(transfer)` with `NULL`.

Another alternative could be to keep the previous patch but change the logic in `sync_transfer_wait_for_completion `. But, I am not sure what assumptions other API users might make about `dev_handle`.

libusb/libusbi.h
```diff
@@ -562,6 +563,9 @@ struct usbi_transfer {
 	uint32_t state_flags;   /* Protected by usbi_transfer->lock */
 	uint32_t timeout_flags; /* Protected by the flying_stransfers_lock */
 
+	/* This is used for logging mostly. As long as it is set, the */
```
hadess, line 566: Is this missing the end of the sentence?

benzea: Oh, that comment was from an earlier iteration of the patch.

PR #1080, opened 2022-02-22
Simplify clock_gettime() conditionalization and fallback on macOS

seanm:
Removed the logic in configure.ac (for Darwin only) that detected if the clock_gettime() function exists.  Instead, just use the preprocessor to introspect the macOS SDK and deployment target and call clock_gettime() if they are new enough.

Also replaced the fallback code for older macOS with mach_absolute_time() and gettimeofday(), which are simplier because they do not need setup and teardown carefully balanced and their usage is in one place instead of spread in multiple places in the file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -211,29 +211,9 @@ esac
 dnl headers not available on all platforms but required on others
 AC_CHECK_HEADERS([sys/time.h])
 
-if test "x$platform" = xposix; then
-	dnl check availability of clock_gettime()
-	if test "x$backend" = xdarwin; then
-		dnl need to verify that OS X target is 10.12 or later for clock_gettime()
-		AC_MSG_CHECKING([whether OS X target version is 10.12 or later])
-		AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
-				#include <AvailabilityMacros.h>
-				#if MAC_OS_X_VERSION_MIN_REQUIRED < 101200
-				# error "Target OS X version is too old"
-				#endif
-			], [])],
-			[AC_MSG_RESULT([yes])
-			 osx_10_12_or_later=yes],
-			[AC_MSG_RESULT([no])
-			 osx_10_12_or_later=])
-		if test "x$osx_10_12_or_later" = xyes; then
-			AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [have_clock_gettime=])
-		else
-			AC_MSG_NOTICE([clock_gettime() is not available on target OS X version])
-		fi
-	else
-		AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [AC_MSG_ERROR([clock_gettime() is required on this platform])])
-	fi
+dnl check availability of clock_gettime(), except don't bother on Darwin, because the result is not used.
+if test "x$platform" = xposix && test "x$backend" != xdarwin; then
+	AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [AC_MSG_ERROR([clock_gettime() is required on this platform])])
 
 	if test "x$have_clock_gettime" = xyes; then
 		dnl the clock_gettime() function needs certain clock IDs defined
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -32,10 +32,7 @@
 #include <fcntl.h>
 #include <sys/sysctl.h>
 
-#include <mach/clock.h>
-#include <mach/clock_types.h>
-#include <mach/mach_host.h>
-#include <mach/mach_port.h>
+#include <mach/mach_time.h>
 
 /* Suppress warnings about the use of the deprecated objc_registerThreadWithCollector
  * function. Its use is also conditionalized to only older deployment targets. */
@@ -62,11 +59,6 @@ static const mach_port_t darwin_default_master_port = 0;
    darwin_cached_devices_mutex must be acquired first. */
 static pthread_mutex_t libusb_darwin_at_mutex = PTHREAD_MUTEX_INITIALIZER;
 static pthread_cond_t  libusb_darwin_at_cond = PTHREAD_COND_INITIALIZER;
-
-#if !defined(HAVE_CLOCK_GETTIME)
-static clock_serv_t clock_realtime;
-static clock_serv_t clock_monotonic;
-#endif
 
 #define LIBUSB_DARWIN_STARTUP_FAILURE ((CFRunLoopRef) -1)
 
@@ -608,16 +600,6 @@ static int darwin_first_time_init(void) {
     return LIBUSB_ERROR_OTHER;
   }
 
-#if !defined(HAVE_CLOCK_GETTIME)
-  /* create the clocks that will be used if clock_gettime() is not available */
-  host_name_port_t host_self;
-
-  host_self = mach_host_self();
-  host_get_clock_service(host_self, CALENDAR_CLOCK, &clock_realtime);
-  host_get_clock_service(host_self, SYSTEM_CLOCK, &clock_monotonic);
-  mach_port_deallocate(mach_task_self(), host_self);
-#endif
-
   int rc = pthread_create (&libusb_darwin_at, NULL, darwin_event_thread_main, NULL);
   if (0 != rc) {
     usbi_err (NULL, "could not create event thread, error %d", rc);
@@ -680,11 +662,6 @@ static void darwin_exit (struct libusb_context *ctx) {
     pthread_join (libusb_darwin_at, NULL);
 
     darwin_cleanup_devices ();
-
-#if !defined(HAVE_CLOCK_GETTIME)
-    mach_port_deallocate(mach_task_self(), clock_realtime);
-    mach_port_deallocate(mach_task_self(), clock_monotonic);
-#endif
   }
   usbi_mutex_unlock(&darwin_cached_devices_mutex);
 }
@@ -2402,27 +2379,36 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   return usbi_handle_transfer_completion (itransfer, darwin_transfer_status (itransfer, tpriv->result));
 }
 
-#if !defined(HAVE_CLOCK_GETTIME)
 void usbi_get_monotonic_time(struct timespec *tp) {
-  mach_timespec_t sys_time;
+/* Check if the SDK is new enough to declare clock_gettime(), and the deployment target is at least 10.12. */
+#if ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
+  clock_gettime(CLOCK_MONOTONIC, tp);
+#else
+  mach_timebase_info_data_t machTimeBaseInfo;
+  mach_timebase_info(&machTimeBaseInfo);
 
-  /* use system boot time as reference for the monotonic clock */
-  clock_get_time (clock_monotonic, &sys_time);
+  uint64_t uptime = mach_absolute_time();
+  uint64_t uptimeNano = uptime * machTimeBaseInfo.numer / machTimeBaseInfo.denom;
 
-  tp->tv_sec  = sys_time.tv_sec;
-  tp->tv_nsec = sys_time.tv_nsec;
+  uint64_t uptimeSeconds = uptimeNano / NSEC_PER_SEC;
+  uint64_t uptimeNanoRemainder = uptimeNano - (uptimeSeconds * NSEC_PER_SEC);
+
+  tp->tv_sec = uptimeSeconds;
+  tp->tv_nsec = uptimeNanoRemainder;
+#endif
 }
 
 void usbi_get_real_time(struct timespec *tp) {
-  mach_timespec_t sys_time;
-
-  /* CLOCK_REALTIME represents time since the epoch */
-  clock_get_time (clock_realtime, &sys_time);
-
-  tp->tv_sec  = sys_time.tv_sec;
-  tp->tv_nsec = sys_time.tv_nsec;
-}
+/* Check if the SDK is new enough to declare clock_gettime(), and the deployment target is at least 10.12. */
+#if ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
+  clock_gettime(CLOCK_REALTIME, tp);
+#else
+  struct timeval tv;
+  gettimeofday(&tv, NULL);
+  tp->tv_sec = tv.tv_sec;
+  tp->tv_nsec = tv.tv_usec * NSEC_PER_USEC;
 #endif
+}
 
 #if InterfaceVersion >= 550
 static int darwin_alloc_streams (struct libusb_device_handle *dev_handle, uint32_t num_streams, unsigned char *endpoints,
--- a/Xcode/config.h
+++ b/Xcode/config.h
@@ -8,12 +8,6 @@
 /* Define to 1 to enable message logging. */
 #define ENABLE_LOGGING 1
 
-/* On 10.12 and later, use newly available clock_*() functions */
-#if MAC_OS_X_VERSION_MIN_REQUIRED >= 101200
-/* Define to 1 if you have the `clock_gettime' function. */
-#define HAVE_CLOCK_GETTIME 1
-#endif
-
 /* On 10.6 and later, use newly available pthread_threadid_np() function */
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
 /* Define to 1 if you have the 'pthread_threadid_np' function. */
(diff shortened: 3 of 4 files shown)


configure.ac
```diff
@@ -201,27 +201,7 @@ AC_CHECK_HEADERS([sys/time.h])
...
-			 osx_10_12_or_later=])
-		if test "x$osx_10_12_or_later" = xyes; then
-			AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [have_clock_gettime=])
-		else
-			AC_MSG_NOTICE([clock_gettime() is not available on target OS X version])
-		fi
-	else
-		AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [AC_MSG_ERROR([clock_gettime() is required on this platform])])
-	fi
+	AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [AC_MSG_ERROR([clock_gettime() is required on this platform])])
```
tormodvolden, line 204: Wouldn't this show an error when building on older macOS?

libusb/os/darwin_usb.c
```diff
@@ -2332,27 +2306,36 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   return usbi_handle_transfer_completion (itransfer, darwin_transfer_status (itransfer, tpriv->result));
 }
 
-#if !defined(HAVE_CLOCK_GETTIME)
 void usbi_get_monotonic_time(struct timespec *tp) {
-  mach_timespec_t sys_time;
+// If the SDK is new enough to have the clock_gettime() declaration, and the deplopment target is at least 10.12...
+#if 0 && ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
```
tormodvolden, line 2311: #if 0 ?

configure.ac
seanm: Oh maybe.  I'm not familiar with this `.ac` language... from the name, I guess `AC_MSG_ERROR` would pop an error.  I guess I could restore the Darwin check and just not perform this check at all.

libusb/os/darwin_usb.c
```diff
@@ -2332,27 +2306,36 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
...
-  mach_timespec_t sys_time;
-
-  /* CLOCK_REALTIME represents time since the epoch */
-  clock_get_time (clock_realtime, &sys_time);
-
-  tp->tv_sec  = sys_time.tv_sec;
-  tp->tv_nsec = sys_time.tv_nsec;
-}
+// If the SDK is new enough to have the clock_gettime() declaration, and the deplopment target is at least 10.12...
+#if 0 && ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
```
tormodvolden, line 2330: #if 0 here also.

tormodvolden:
It looks like you are doing two potentially independent things here: 1) moving the logic out of configure.ac and into darwin_usb.c and 2) replace the old fallback with mach_absolute_time(). If not two commits, it could at least be stated in the commit message.

configure.ac
tormodvolden: It will print "clock_gettime() is required on this platform" which is false :)

libusb/os/darwin_usb.c
```diff
@@ -2332,27 +2306,36 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   return usbi_handle_transfer_completion (itransfer, darwin_transfer_status (itransfer, tpriv->result));
 }
 
-#if !defined(HAVE_CLOCK_GETTIME)
 void usbi_get_monotonic_time(struct timespec *tp) {
-  mach_timespec_t sys_time;
+// If the SDK is new enough to have the clock_gettime() declaration, and the deplopment target is at least 10.12...
```
tormodvolden, line 2310: > If the SDK is new enough to have the clock_gettime()
- isn't that what HAVE_CLOCK_GETTIME will tell us?

seanm: They are similar, yes. But compile-time checks are generally preferable to configure-time checks. The latter are an anti-pattern IMNSHO, they are very cross-compile unfriendly.  Why go to the trouble of building a test program when you can just adjust your build with the preprocessor?  Especially when we need a solution that works both with autotools and Xcode.

tormodvolden:
Also bonus points if you explain why gettimeofday() and mach_absolute_time() are better than clock_get_time().

seanm:
> It looks like you are doing two potentially independent things here: 1) moving the logic out of configure.ac and into darwin_usb.c and 2) replace the old fallback with mach_absolute_time(). If not two commits, it could at least be stated in the commit message.

Correct.  My commit message was also WIP, but I've expanded it now.

configure.ac
seanm: Indeed. I've taken another stab. Does it look more correct now?

seanm:
> Also bonus points if you explain why gettimeofday() and mach_absolute_time() are better than clock_get_time().

Done.

libusb/os/darwin_usb.c
tormodvolden: And feature checks are preferable to version checks :) but I agree this is not obvious when "cross-compiling" for a different macOS version. I guess configure is great for setting all the overall flags for a big build, but fine-navigation between versions subtleties are easier with the preprocessor.

Theoretically feature checks would be more robust e.g. in case this gets built in a pimped-up build environment where e.g. clock_gettime() could show up supported through compatibility add-ons (à la macports). But we can leave all that thinking to macports :)

configure.ac
```diff
@@ -201,25 +201,7 @@ AC_CHECK_HEADERS([sys/time.h])
...
-			 osx_10_12_or_later=yes],
-			[AC_MSG_RESULT([no])
-			 osx_10_12_or_later=])
-		if test "x$osx_10_12_or_later" = xyes; then
-			AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [have_clock_gettime=])
-		else
-			AC_MSG_NOTICE([clock_gettime() is not available on target OS X version])
-		fi
-	else
+	if test "x$backend" != xdarwin; then
```
tormodvolden, line 204: I appreciate the minimum invasiveness, but these two tests could be &&'d for readability.

seanm: Which two tests?

tormodvolden:
Great.

Typo "wiith" in commit message.

configure.ac
tormodvolden:
```
if test "x$platform" = xposix; then
	dnl check availability of clock_gettime()
	if test "x$backend" != xdarwin; then
```

seanm: But doesn't the `dnl` statement need to run for the first test but not the second?

libusb/os/darwin_usb.c
```diff
@@ -2332,27 +2306,36 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
...
+#if ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
```
tormodvolden, line 2311: Isn't MAC_OS_X_VERSION_MAX_ALLOWED always >=  MAC_OS_X_VERSION_MIN_REQUIRED? So that only the latter needs checking?

configure.ac
seanm: Oh damn is that a comment?!

tormodvolden: I think it might get something printed, but if you are not checking for the availability it shouldn't display either.

seanm: OK, I've taken another stab.

libusb/os/darwin_usb.c
seanm: >And feature checks are preferable to version checks :)

The version check is the feature check.  clock_gettime() is documented to have been added in 10.12, thus checking the version is checking for the feature.

>but I agree this is not obvious when "cross-compiling" for a different macOS version

Or any cross-compilation.  Imagine trying to build libusb for iOS from macOS.  Or building for ARM on Intel.  Or building for FreeBSD from macOS. Any test you do on the machine that's building can give the wrong result for the machine you're building *for*.

But I don't want to start a religious argument. :)

configure.ac
tormodvolden: stackoverflow to the rescue: dnl stands for “Discard to Next Line”  (so pure comment)

configure.ac
```diff
@@ -199,29 +199,9 @@ esac
...
-		fi
-	else
-		AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [AC_MSG_ERROR([clock_gettime() is required on this platform])])
-	fi
+dnl check availability of clock_gettime(), except don't bother on Darwin, becuause the result is not used.
+if test "x$platform" = xposix && test "x$backend" != xdarwin; then
+	AC_CHECK_FUNCS([clock_gettime], [have_clock_gettime=yes], [AC_MSG_ERROR([clock_gettime() is required on this platform])])
 
 	if test "x$have_clock_gettime" = xyes; then
 		dnl the clock_gettime() function needs certain clock IDs defined
```
tormodvolden, line 207: Will these certain ID's be needed on newer macOS also? Because now we skipped that test above and $have_clock_gettime is unset.

configure.ac
seanm: Yeah, found similar. What a weird comment indicator.

libusb/os/darwin_usb.c
tormodvolden: I am regularly cross-building libusb, e.g. for Windows on Linux. autoconf does the right thing if you've got the details right.

configure.ac
seanm: It looks to be doing a compile-test for `CLOCK_MONOTONIC` and `CLOCK_REALTIME`. Not necessary to perform such a test, the usage of those two constants is also guarded in the `#if ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))` test.

tormodvolden: Sounds good. Good you added "becuause the result is not used." but typo in bacuause.

seanm: Man I suck at typing!

libusb/os/darwin_usb.c
seanm: Hmm, it generally would be, but I'm not 100% sure it *must* be, so I'd argue to keep it like this.

tormodvolden: OK. Just another nitpick: `/* comments */` are standard except inside Windows files.

tormodvolden: "deplopment" :) And the ellipses are unneeded. Maybe rephrase to something shorter and less redundant, e.g. "If the SDK declares clock_gettime(), and the deployment target supports it"

From what I understand MAC_OS_X_VERSION_MAX_ALLOWED is the highest version that the SDK knows about, and the target cannot be set higher. But I am fine with keeping both checks because I don't know for sure. LOL, I googled that macro now and this came on top :) https://lists.apple.com/archives/xcode-users/2005/Aug/msg00399.html

seanm:
I think/hope this should be ready to merge now... @tormodvolden ?

tormodvolden:
@seanm Yes, thanks, I think this is good. But I'll leave it for 1.0.27 unless it fixes some observed bug or regression.

seanm:
>@seanm Yes, thanks, I think this is good. But I'll leave it for 1.0.27 unless it fixes some observed bug or regression.

@tormodvolden can we merge this now?

mcuee:
Please help to fix the github warning.

Merging is blocked
The base branch requires all commits to be signed.

tormodvolden:
This PR is on the top of the list in #1094 so it is not forgotten :) I have just waited with pushing anything after the last release while watching to see if something serious comes up and justifies a fixup release. A few important changes were pushed to the repo. As it stands now, I don't think we'll make a fixup release -> discussion in #1094.

mcuee, I don't think we have to worry about the GitHub warning. It only applies if using GitHubs own workflows. We generally pull and rebase patches manually.

mcuee:
> mcuee, I don't think we have to worry about the GitHub warning. It only applies if using GitHubs own workflows. We generally pull and rebase patches manually.

I see. Thanks. Somehow this pops up recently. Can we disable it?

tormodvolden:
I don't know if it can be disabled, but I don't think it is any problem either since we don't want to use GitHub merge anyway.

mcuee:
> I don't know if it can be disabled, but I don't think it is any problem either since we don't want to use GitHub merge anyway.

Not a real problem, just an eye sore to me...
But it is okay if this can not be disabled. I will get used to this.

seanm:
@tormodvolden friendly ping... I think this is ready to merge..

neheb:
Doesn't this fallback to the mach code unconditionally?

seanm:
>Doesn't this fallback to the mach code unconditionally?

I don't believe so.  Can you elaborate on what you think is wrong?

neheb:
I would think it should be:

```
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -534,7 +534,7 @@ static inline void usbi_localize_device_descriptor(struct libusb_device_descript
        desc->bcdDevice = libusb_le16_to_cpu(desc->bcdDevice);
 }
 
-#if defined(HAVE_CLOCK_GETTIME) && !defined(__APPLE__)
+#if defined(HAVE_CLOCK_GETTIME) || ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
 static inline void usbi_get_monotonic_time(struct timespec *tp)
 {
        ASSERT_EQ(clock_gettime(CLOCK_MONOTONIC, tp), 0);
diff --git a/libusb/os/darwin_usb.c b/libusb/os/darwin_usb.c
index 3f3a35af..f49b99ca 100644
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -304,7 +304,7 @@ static usb_device_t **darwin_device_from_service (struct libusb_context *ctx, io
   const int max_retries = 5;
 
   /* The IOCreatePlugInInterfaceForService function might consistently return
-     an "out of resources" error with certain USB devices the first time we run 
+     an "out of resources" error with certain USB devices the first time we run
      it. The reason is still unclear, but retrying fixes the problem */
   for (int count = 0; count < max_retries; count++) {
     kresult = IOCreatePlugInInterfaceForService(service, kIOUSBDeviceUserClientTypeID,
@@ -2382,11 +2382,8 @@ static int darwin_handle_transfer_completion (struct usbi_transfer *itransfer) {
   return usbi_handle_transfer_completion (itransfer, darwin_transfer_status (itransfer, tpriv->result));
 }
 
+#if ((MAC_OS_X_VERSION_MAX_ALLOWED < 101200) || (MAC_OS_X_VERSION_MIN_REQUIRED < 101200))
 void usbi_get_monotonic_time(struct timespec *tp) {
-/* Check if the SDK is new enough to declare clock_gettime(), and the deployment target is at least 10.12. */
-#if ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
-  clock_gettime(CLOCK_MONOTONIC, tp);
-#else
   mach_timebase_info_data_t machTimeBaseInfo;
   mach_timebase_info(&machTimeBaseInfo);
 
@@ -2398,20 +2395,15 @@ void usbi_get_monotonic_time(struct timespec *tp) {
 
   tp->tv_sec = uptimeSeconds;
   tp->tv_nsec = uptimeNanoRemainder;
-#endif
 }
 
 void usbi_get_real_time(struct timespec *tp) {
-/* Check if the SDK is new enough to declare clock_gettime(), and the deployment target is at least 10.12. */
-#if ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
-  clock_gettime(CLOCK_REALTIME, tp);
-#else
   struct timeval tv;
   gettimeofday(&tv, NULL);
   tp->tv_sec = tv.tv_sec;
   tp->tv_nsec = tv.tv_usec * NSEC_PER_USEC;
-#endif
 }
+#endif
 
 #if InterfaceVersion >= 550
 static int darwin_alloc_streams (struct libusb_device_handle *dev_handle, uint32_t num_streams, unsigned char *endpoints,
```

I think that's where my confusion came from.

seanm:
> --- a/libusb/libusbi.h
> +++ b/libusb/libusbi.h
> @@ -534,7 +534,7 @@ static inline void usbi_localize_device_descriptor(struct libusb_device_descript
>         desc->bcdDevice = libusb_le16_to_cpu(desc->bcdDevice);
>  }
>  
> -#if defined(HAVE_CLOCK_GETTIME) && !defined(__APPLE__)
> +#if defined(HAVE_CLOCK_GETTIME) || ((MAC_OS_X_VERSION_MAX_ALLOWED >= 101200) && (MAC_OS_X_VERSION_MIN_REQUIRED >= 101200))
>  static inline void usbi_get_monotonic_time(struct timespec *tp)

No, because on macOS `usbi_get_monotonic_time` is not declared here, it's in darwin_usb.c.

PR #1084, opened 2022-02-26
Put extra data under parent interface instead of endpoint

bariscelik:
The subdescriptors that come after the first endpoint was being saved as extra data of **endpoint**. I guess this was because endpoint descriptors always at last of sub-descriptors.

But in my case, I'm using a UVC camera and it sends endpoint descriptors with mixed order. 

As a solution, extra data comes after the endpoint descriptors, have been added as sub-descriptors of related **interface**.

Note: I'm not sure the solution meets USB standards, maybe the way my camera works is wrong

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -81,7 +81,8 @@ static void clear_endpoint(struct libusb_endpoint_descriptor *endpoint)
 }
 
 static int parse_endpoint(struct libusb_context *ctx,
-	struct libusb_endpoint_descriptor *endpoint, const uint8_t *buffer, int size)
+	struct libusb_endpoint_descriptor *endpoint,
+	struct libusb_interface_descriptor *interface, const uint8_t *buffer, int size)
 {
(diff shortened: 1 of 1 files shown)


mcuee:
Please post the USB descriptors of your USB device, for example, using "lsusb -vvv" and Windows USBView or macOS USBProber app.

It could be your device is faulty as per the USB specifications.

mcuee:
I tend to believe the device is non compliant and thus the patch is not valid.

mcuee:
But I will leave this issue open and maybe others can comment further.

tormodvolden:
What does your patch do if the extra data actually belongs to the endpoint?

bariscelik:
> What does your patch do if the extra data actually belongs to the endpoint?

As far as I've seen so far, there is no standard for extra data for endpoints. Also, endpoint descriptors are usually under the regarded interface in a hierarchical order.

mcuee:
Closing this one as it is not valid. Please reopen if you have new updates.

PR #1085, opened 2022-02-27
msvc: rework msbuild files

shuffle2:
refactors the msbuild files to simplify the files and allow properly building all possible targets without trampling over each other. I've tested that the `msvc/build_all.ps1` script works on a machine with vs2022 installed alongside the vs2013, vs2015, vs22017, and vs2019 toolsets for x86, x64, arm32, and arm64.

I haven't bothered with the CI stuff because 1) i haven't dealt with that on github before, and 2) i want to see if this change will be accepted first and 3) maybe someone else knows how to do it :)

note: I've kept most actual compile/link/etc settings the same as they were previously or their defaults. I also kept `ClCompile.RuntimeLibrary` being set for different target types in separate files, because I've seen that vcpkg is regex replacing them to modify the build - so these changes shouldn't negatively impact vcpkg.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -3245,7 +3245,13 @@ static enum libusb_transfer_status winusbx_copy_transfer_data(int sub_api, struc
 			// iso only supported on libusbk-based backends for now
 			PKISO_CONTEXT iso_context = transfer_priv->iso_context;
 			for (i = 0; i < transfer->num_iso_packets; i++) {
-				transfer->iso_packet_desc[i].actual_length = iso_context->IsoPackets[i].actual_length;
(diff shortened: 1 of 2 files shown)


mcuee:
Have you tried this on a machine with only VS2019 or VS2017?

shuffle2:
I haven’t. It can be made to work as such but is really against how the build system is designed to work nicely, and may cause reintroduction of duplication for no gain. See this https://devblogs.microsoft.com/cppblog/stuck-on-an-older-toolset-version-move-to-visual-studio-2015-without-upgrading-your-toolset/ or other documentation about “native multi targeting” - the idea is that developers can only maintain a single set of solution and project files, be able to rely on latest build system and ide features, and still produce binaries with old compiler tool chains.

PR #911, opened 2021-04-24
macOS: implement kernel driver detach

osy:
It's a well known issue on macOS that you cannot use libusb to attach to an interface already claimed by an KEXT for exclusive access. The common workaround is to do `sudo kextunload` or inject a codeless KEXT. Both require admin permissions, and does not work for all devices on latest macOS. Additionally, SIP has to be disabled to inject a codeless KEXT, making it a pain to work with USB devices.

The original implementation relied on a new option but that has been superseded by a new implementation which uses the `libusb_detach_kernel_driver()` and `libusb_set_auto_detach_kernel_driver()` APIs. However there are a few caveats.

First, the `libusb_detach_kernel_driver` APIs expect a single USB interface to be detachable. On macOS, we can only "capture" an entire USB device (and force all interfaces to be released by any existing driver). To keep compatibility with existing code, we count the number of (successful) calls to `libusb_detach_kernel_driver` and when that number is balanced by `libusb_attach_kernel_driver`, the entire device is re-enumerated and the original kernel driver is allowed to re-attach.

Second, `libusb_detach_kernel_driver` is expected to be called after `libusb_open` (this doesn't seem to be documented but I see that is the usage from existing code). This is problematic because the effect of `USBDeviceReEnumerate` with `kUSBReEnumerateCaptureDeviceMask` doesn't seem to take place until after `USBDeviceOpen`. So we call `darwin_restore_state` to close and re-open the device.

Third, `libusb_reset_device` uses `USBDeviceReEnumerate` which will release the capture flag and allow the driver to re-claim the interfaces. I'm not sure what the reasoning for avoiding `ResetDevice` (I see a comment "from macOS 10.11 ResetDevice no longer does anything so just use USBDeviceReEnumerate" but it seems to work fine in 10.15...). If we are currently capturing a USB device, then `ResetDevice` is used instead of `USBDeviceReEnumerate`.

`kUSBReEnumerateCaptureDeviceMask` requires either the application to be running as root OR the application has the entitlement `com.apple.vm.device-access` AND the user has accepted the authorization for the capture with `IOServiceAuthorize`. To get the second part to work, we detect when the entitlement `com.apple.vm.device-access` is available and then call `IOServiceAuthorize` in `darwin_detach_kernel_driver` before `darwin_reenumerate_device`. However, as the authorization decision is cached in the plugin's start() call, it is necessary to stop and start the USB plugin again.

[...]

PR #1088, opened 2022-03-06
darwin: reset by re-enumerate on iOS platforms

osy:
On non-macOS platforms, ResetDevice() does nothing so we use the "old" way of
calling re-enumerate. If the device is captured, we have to re-enumerate, then
re-capture, re-authorize, and finally restore the state.

I realized today that this old commit never made it upstream in #911 because I committed it the day after the PR was closed. Since it only applies to iOS, it doesn't change much but for completion sake here it is.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -95,6 +95,8 @@ static enum libusb_error process_new_device (struct libusb_context *ctx, struct 
 static enum libusb_error darwin_get_cached_device(struct libusb_context *ctx, io_service_t service, struct darwin_cached_device **cached_out,
                                                   UInt64 *old_session_id);
 
+static int darwin_detach_kernel_driver (struct libusb_device_handle *dev_handle, uint8_t interface);
+
 #if defined(ENABLE_LOGGING)
 static const char *darwin_error_str (IOReturn result) {
   static char string_buffer[50];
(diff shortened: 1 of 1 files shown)


mcuee:
Does libusb even work on iOS or iPadOS?

osy:
Yes but you must be jailbroken.

mcuee:
> Yes but you must be jailbroken.

Interesting to know this. I myself will not jailbreak any iOS device but I understand there are still people who are doing that.

mcuee:
Just to be clear, I have no issues if this is merged. I am just saying I am not into things like Jailbreak myself.

hjelmn:
Yes. This is worth merging. Does no harm.

libusb/os/darwin_usb.c
```diff
@@ -1842,15 +1844,37 @@ static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, b
 
 static int darwin_reset_device (struct libusb_device_handle *dev_handle) {
   struct darwin_cached_device *dpriv = DARWIN_CACHED_DEVICE(dev_handle->dev);
+  unsigned long claimed_interfaces = dev_handle->claimed_interfaces;
+  int8_t active_config = dpriv->active_config;
+  int capture_count;
   IOReturn kresult;
+  enum libusb_error ret;
 
+#if TARGET_OS_OSX
```
tormodvolden, line 1853: From the discussion in #1097 I am unsure if this is the best conditional to use here.

tormodvolden:
It seems to me this also affects the macOS (non-iOS) path, if dpriv->capture_count == 0 ?

hjelmn:
@tormodvolden How so? If ```capture_count == 0``` on macOS it will skip the ResetDevice then call ```darwin_reenumerate_device``` which is the same. The conditional after is false so it then returns. Same as before. What am I missing?

tormodvolden:
For macOS the error checking would be new, and possibly call darwin_detach_kernel_driver() and darwin_restore_state(). So there is a behaviour change. Not saying that it is not a code improvement, but this is not what the PR or commit message says.

I am extra careful about this now because I am considering including it in the upcoming RC: A purely iOS-only change I would include in the upcoming RC without much consideration. Whereas for any macOS change I will need to understand it well. OTOH if this fixes something on macOS, it is more important than the iOS part, but what does it fix?

In the existing code, darwin_reset_device() is a no-op when dpriv->capture_count == 0. What does this mean?

tormodvolden:
Ah, I missed that "conditional after is false so it then returns".

tormodvolden:
Please let me know what you think about my slight rearrangement in https://github.com/tormodvolden/libusb/commit/17ea686ecf3a47956afd0bd22676810c7e054935 It also deals with undefined TARGET_OS_OSX

osy:
LGTM

tormodvolden:
Just something related, a comment in darwin_reenumerate_device() says `from macOS 10.11 ResetDevice no longer does anything`. Does that mean this should be extended to 10.11+ ? @osy also mentioned this comment in the #911 OP.

Would the new "iOS" code fully take care of the "authorization for capture" also for macOS - which was the motivation for using ResetDevice() according to the comment `we have to use ResetDevice as USBDeviceReEnumerate() loses the authorization for capture` ?

PR #1091, opened 2022-03-17
Use HidD_GetIndexedString() instead of HidD_GetProductString() to get…

mhjartstrom:
… the correct iProduct string on Windows

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -3860,7 +3860,10 @@ static int hid_open(int sub_api, struct libusb_device_handle *dev_handle)
 
 		priv->hid->string_index[1] = dev->device_descriptor.iProduct;
 		if (priv->hid->string_index[1] != 0)
-			HidD_GetProductString(hid_handle, priv->hid->string[1], sizeof(priv->hid->string[1]));
+			// Using HidD_GetIndexedString() instead of HidD_GetProductString(), as the latter would otherwise return the name
+			// of the interface instead of the iProduct string whenever the iInterface member of the USB_INTERFACE_DESCRIPTOR
+			// structure for the interface is nonzero (see Remarks section in the documentation of the HID API routines)
+			HidD_GetIndexedString(hid_handle, priv->hid->string_index[1], priv->hid->string[1], sizeof(priv->hid->string[1]));
 		else
 			priv->hid->string[1][0] = 0;
 


libusb/os/windows_winusb.c
```diff
@@ -3860,7 +3860,9 @@ static int hid_open(int sub_api, struct libusb_device_handle *dev_handle)
 
 		priv->hid->string_index[1] = dev->device_descriptor.iProduct;
 		if (priv->hid->string_index[1] != 0)
-			HidD_GetProductString(hid_handle, priv->hid->string[1], sizeof(priv->hid->string[1]));
+			// Use HidD_GetIndexedString() instead of HidD_GetProductString() to get the correct iProduct string whenever
```
tormodvolden, line 3863: This sounds like a comment for the commit message, but is likely confusing in the code itself since it refers to previous code that is not present any longer.

mcuee:
@mhjartstrom What is the issue you are encountering?

mhjartstrom:
> @mhjartstrom What is the issue you are encountering?

I was using a HID composite device with three interfaces (e.g. keyboard and touch screen built into one) and while the three interfaces all referenced their own names in the interface descriptor, the iProduct string was set to the name of the last interface instead of the actual product's name (e.g. "TOUCH" repeated twice instead of "PRODUCT" and "TOUCH").

This behavior differ from what for instance Microsoft USB Device Viewer will report for the same device. This fix will make them report the same thing.

mhjartstrom:
It's actually described in the documentation of both routines, see Remarks under [https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/hidsdi/nf-hidsdi-hidd_getproductstring].

libusb/os/windows_winusb.c
mhjartstrom: I expanded the comment somewhat. Better?

Don't know what to do about the commit message? Will this be a squash merge, where the message can be edited? Suggestions?

mcuee:
This seems to be a good fix to be merged.

libusb/os/windows_winusb.c
tormodvolden: Thanks, that seems better to me. Yes, we will always squash such commits, and you can also squash and repush yourself.

Since this is non-obvious, you can also add the description of the observed issue to the commit message.

mhjartstrom: OK, would you like me to do the squash? Not sure about the repush... I guess I would have to force push in order to overwrite the existing branch, would GitHub handle the reference to this issue anyway?

About the commit message; by "add the description" do you mean my updated code comment, or what I wrote in [my issue below](https://github.com/libusb/libusb/pull/1091#issuecomment-1072091703)?

tormodvolden: Yes, you can simply force push on the same branch. I meant the PR comment below where you describe the observed bug.

tormodvolden:
The commit message needs some work, it seems like some phrases were copy-pasted together and don't make all sense any longer. But I think it overall has the right contents.

mhjartstrom:
> The commit message needs some work, it seems like some phrases were copy-pasted together and don't make all sense any longer. But I think it overall has the right contents.

Well, it _was_ copy-pasted together as per your request. I think it's good, informative and I haven't seen anything wrong with it. Look, either you accept this as-is or you say what you want it to say... I don't want to push different suggestions until you are satisfied.

tormodvolden:
> Fix an issue where a HID composite device with three interfaces (e.g. keyboard and touch screen built into one),

I think I misread it, I thought above was meant to be a complete phrase, but "where" didn't lead anywhere, and the next phrase also seemed strange. That's why I thought it was hastily copy-pasted.

> while all three interfaces having referenced their own names in the interface descriptor, was reported to have
an iProduct string that was set to the name of the last interface instead of the actual product's name (e.g.
"TOUCH" repeated twice instead of "PRODUCT" and "TOUCH").

Now that I read them together, it makes sense (it is just very long). However, conventionally a commit message has a short summary line (which shows up in git summaries), an empty line, and then whatever long explanations are needed. That's why I misread it. We just need to add a summary line. But I can also come up with something.

> Use HidD_GetIndexedString() instead of HidD_GetProductString(), as the latter would otherwise return the name
of the interface instead of the iProduct string whenever the iInterface member of the USB_INTERFACE_DESCRIPTOR
structure for the interface is nonzero (see Remarks section in the documentation of the HID API routines).

I don't understand the logic behind Microsoft's behaviour, but that's how it is :)

> This behavior differ from what for instance Microsoft USB Device Viewer will report for the same device. This
fix will make them report the same thing.

A question: Does this only happen when there are 3 or more interfaces, or would it also be seen with 2 interfaces?

tormodvolden:
Has this been tested on many devices? @mcuee Are you able to confirm the issue and the fix? I am considering sneaking this into 1.0.26 before the RC door closes.

mcuee:
@tormodvolden Sorry I do not have such USB HID device to carry out the test. But I think it is fine.

Issue #1039, opened 2022-01-02
USB devices with incorrect interface descriptors cause read access violations


Eekle:
I have a pair of USB speakers that seem to have an invalid USB descriptor.

Looking via USB Device Tree Viewer:

```     ------------------ Configuration Descriptor -------------------
bLength                  : 0x09 (9 bytes)
bDescriptorType          : 0x02 (Configuration Descriptor)
wTotalLength             : 0x0087 (135 bytes)
bNumInterfaces           : 0x03 (3 Interfaces)
bConfigurationValue      : 0x01 (Configuration 1)
iConfiguration           : 0x00 (No String Descriptor)
<snip>
```
and then later on...

```
        ---------------- Interface Descriptor -----------------
bLength                  : 0x09 (9 bytes)
bDescriptorType          : 0x04 (Interface Descriptor)
bInterfaceNumber         : 0x03 (Interface 3)
bAlternateSetting        : 0x00
bNumEndpoints            : 0x01 (1 Endpoint)
bInterfaceClass          : 0x03 (HID - Human Interface Device)
bInterfaceSubClass       : 0x00 (None)
bInterfaceProtocol       : 0x00 (None)
iInterface               : 0x00 (No String Descriptor)
Data (HexDump)           : 09 04 03 00 01 03 00 00 00  
```

This Interface Number is incorrect, i believe. It can't possibly be interface 3, when the configuration says there are only 3 interfaces total.

However libusb does not check this index is in bounds.

When doing a control transfer, we eventually end up in `auto_claim` where it iterates over the array of 32 possible descriptors. It finds a matching descriptor at interface index 3, and then passes that index down.

In `windows_assign_endpoints`, we then use that to index into the config description interface struct. This is an out of bounds access, and the value of `if_desc` is nonsense.

https://github.com/libusb/libusb/blob/f2b218b61867f27568ba74fa38e156e5f55ed825/libusb/os/windows_winusb.c#L606-L611

When `if_desc` is then accessed, the stack dies with a read access violation.

https://github.com/libusb/libusb/blob/f2b218b61867f27568ba74fa38e156e5f55ed825/libusb/os/windows_winusb.c#L526-L530

I appreciate that this is really a hardware issue, but it seems egregious to me that it blows up with an access violation rather than erroring out or anything that might be caught higher up.

I initially encountered this issue trying to work with `probe-rs`, which was silently failing because of this error.

---

OS: Win 11
LibUSB version: 1.0.24.0 (as in version.h)

[...]

PR #1093, opened 2022-03-18
add interface bounds checking

lunixbochs:
See https://github.com/libusb/libusb/issues/1039

This adds bounds checks to fields such as `priv->interfaces[]`, `config->interface[]`, `config->altsetting[]`

In each case, I attempted to use the field's actual size for the bound. For config, thats `bNumInterfaces`, and otherwise it's generally either `num_altsettings` or `USB_MAXINTERFACES`

I return `LIBUSB_ERROR_NOT_FOUND` in most cases if the iface index exceeds known bounds. For `sunos_usb.c` I just silently don't set `hpriv->altsetting`, because device releasing doesn't appear to do anything there anyway.

I have done basic tests on Darwin.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1430,6 +1430,10 @@ static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle,
         return rc;
       }
 
+      if (iface >= config->bNumInterfaces) {
+        usbi_err (HANDLE_CTX (dev_handle), "interface %d out of range for device", iface);
+        return LIBUSB_ERROR_NOT_FOUND;
+      }
       endpoint_desc = config->interface[iface].altsetting[alt_setting].endpoint + i - 1;
 
       cInterface->endpoint_addrs[i - 1] = endpoint_desc->bEndpointAddress;
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -523,6 +523,10 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 		return r;
 	}
 
+	if (iface >= conf_desc->bNumInterfaces) {
+		usbi_err(HANDLE_CTX(dev_handle), "interface %d out of range for device", iface);
+		return LIBUSB_ERROR_NOT_FOUND;
+	}
 	if_desc = &conf_desc->interface[iface].altsetting[altsetting];
 	safe_free(priv->usb_interface[iface].endpoint);
 


tormodvolden:
This would add redundant checks. E.g. darwin_set_interface_altsetting() and winusbx_set_interface_altsetting() are only called from libusb_set_interface_alt_setting() which already has done the same check.

lunixbochs:
I've reduced it to two checks. One is the windows crash, the other looks like a similar potential crash on Darwin.

I would like to register that I would prefer cheap defensive checks over making undocumented assumptions about the future behavior of callers, but it's not my project, so I will defer to maintainers.

libusb/os/darwin_usb.c
```diff
@@ -1430,6 +1430,10 @@ static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle,
         return rc;
       }
 
+      if (iface >= config->bNumInterfaces) {
+        usbi_err (HANDLE_CTX (dev_handle), "interface out of range for device");
+        return LIBUSB_ERROR_NOT_FOUND;
```
tormodvolden, line 1435: Shouldnt we use LIBUSB_ERROR_INVALID_PARAM like the checks in core.c?

tormodvolden:
The only possible future issue I can see is these functions being called with unchecked values from other places in the backend but AFAICS the checks have been done in core.c for all existing cases. Such issues would be an internal error IMO and I would prefer asserts, in contrast to the user value check in core.c.

libusb/os/darwin_usb.c
tormodvolden: Yes, I can agree with INVALID_PARAM  when > max valid interface number (32), and NOT_FOUND when the device just doesn't have that value. Maybe libusb_detach_kernel_driver() should be changed?

lunixbochs:
As a user of the library, an assert means what would otherwise be a recoverable mistake (returning an error specific to the broken device) would crash the parent app, which is ultimately the same end user impact as the issue I'm trying to fix here. It would make more sense to me to log and return an error. "USB doesn't work" is noticeable enough that exiting the app should probably be used sparingly.

tormodvolden:
I meant to use assert just for libusb developers to detect an internal error (if somehow the checks in core are bypassed or not sufficient in future code revisions). I am talking about the redundant USB_MAXINTERFACES checks here. Absolutely, any broken devices thrown at us should ideally return errors and not assert/crash. And if the application uses wrong values we should return errors.

libusb/os/windows_winusb.c
```diff
@@ -523,6 +523,9 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 		return r;
 	}
 
+	if (iface >= conf_desc->bNumInterfaces) {
+		return LIBUSB_ERROR_NOT_FOUND;
```
tormodvolden, line 527: Should we maybe print an error message here as well?

lunixbochs: Feel free to suggest a line of code, I don't know the debugging convention in libusb. I only did it for the darwin error because there was an example right next to my patch.

tormodvolden: I think this should be fine:
`usbi_err(HANDLE_CTX(dev_handle), "interface out of range for device");`
or even:
`usbi_err(HANDLE_CTX(dev_handle), "interface %d out of range for device", iface);`
I don't know if we could hint stronger to this being a broken device but it would be better to check for and report about that during enumeration, so I'd leave it like this.

lunixbochs: ok, I added the interface number to darwin and added the same message to windows

tormodvolden: Excellent.

tormodvolden:
Should this in principle prevent #1039 from crashing?

lunixbochs:
Yes. Someone can probably test this before/after the patch on windows without needing buggy hardware, by modifying [auto_claim()](https://github.com/libusb/libusb/blob/f2b218b61867f27568ba74fa38e156e5f55ed825/libusb/os/windows_winusb.c#L630) to return a large interface number.

tormodvolden:
Great. I'll just add bug references to the commit. Thanks for your contribution!

Eekle:
Reporter of #1039 here - Is this closed intentionally?

PR #749, opened 2020-07-02
WinUSB isochronous transfer fixes

Seneral:
See reasoning for implenentation and example on how to make use of these changes in #747

Fixed missing CHECK_WINUSBX_AVAILABLE removed in 9c28ad2
Fixed #747 by correcting frame calculation for full-speed devices
Fixed instability of isochronous stream due to handling of the ContinueStream parameter
	- Added signalling to user callback using stall event to allow user to control resubmission behaviour
	- Allows user callback to start more transfers only after first transfer is submitted with ContinueStream=False
	- Results in a stable, self-regulating continuous isochronous stream (see #747)

PR #1095, opened 2022-03-20
Fix ISOCH by returning actual subapi in winusb_copy_transfer_data rat…

kahluwalia:
…her than SUB_API_NOTSET

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -3222,12 +3222,12 @@ static enum libusb_transfer_status winusbx_copy_transfer_data(int sub_api, struc
 {
 	struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
 	struct winusb_transfer_priv *transfer_priv = get_winusb_transfer_priv(itransfer);
-	struct winusb_device_priv *priv = usbi_get_device_priv(transfer->dev_handle->dev);
 	int i;
 
-	CHECK_WINUSBX_AVAILABLE(sub_api);
-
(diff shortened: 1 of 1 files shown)


tormodvolden:
Thanks for the patch submission. It would be great to improve Windows ISOCH. This needs to be reviewed carefully. I do remember from going through subapi stuff (but not ISOCH, and all this some time ago) and that "NOTSET" was used to defer the selection to another place. Do you have more evidence, like debug log before and after?

tormodvolden:
This was introduced in commit 9c28ad21 and that commit did so many things that it is hard to understand what was intended. It does look like winusbx_copy_transfer_data() expects the actual sub_api to be passed.

kahluwalia:
I've changed the way the sub_api is determined - it is now done inside winusbx_copy_transfer_data(). It seems this is more in line with all the other winusbx_XXX functions and I hope it reduces the chances this change can affect any of the other implementations.

@tormodvolden I haven't got any logs, but a change was required to determine the sub_api to get ISOCH mode working on an IN endpoint with the WinUSB driver.

tormodvolden:
[I mentioned #749 above but it looks like GitHub doesn't display it]

Thanks a lot, now it looks pretty close to the fix suggested in #749, which maybe is slightly more efficient in the non-isoch case.

kahluwalia:
@tormodvolden aligned with https://github.com/libusb/libusb/pull/749

tormodvolden:
Great. Do you have a full name for the AUTHORS list (also for the commit log it wouldn't be bad)?

kahluwalia:
@tormodvolden I've tried to add full name to commit - not sure if it worked

tormodvolden:
It worked, thank you.

Issue #1100, opened 2022-03-28
Wrong implicit cast in winusbx_copy_transfer_data


tormodvolden:
1.0.26-rc1 on gcc 11:  
```
CC       os/windows_winusb.lo
../../libusb/os/windows_winusb.c: In function 'winusbx_copy_transfer_data':
../../libusb/os/windows_winusb.c:97:32: warning: implicit conversion from 'enum libusb_error' to 'enum libusb_transfer_status' [-Wenum-conversion]
   97 |                         return LIBUSB_ERROR_ACCESS;     \
      |                                ^~~~~~~~~~~~~~~~~~~
../../libusb/os/windows_winusb.c:97:32: note: in definition of macro 'CHECK_WINUSBX_AVAILABLE'
   97 |                         return LIBUSB_ERROR_ACCESS;     \
      |                                ^~~~~~~~~~~~~~~~~~~
```

Issue #1045, opened 2022-01-21
github actions for Window


mcuee:
On Fri, Jan 21, 2022 at 5:23 PM Jef Driesen wrote:
On 20/01/2022 22:32, Ludovic Rousseau wrote:
> It would be nice to also have an automatic build for Windows. But I do
> not know Windows.
> Volunteers are welcome.

Have a look at the github action for my project:

https://github.com/libdivecomputer/libdivecomputer/blob/master/.github/workflows/build.yml

It's a very similar project compared to libusb and has libusb and hidapi as
dependencies. It supports Windows builds using both mingw and msvc, so you can
probably re-use parts for creating a libusb action. The mingw build even builds
libusb and hidapi from source.

Jef

tormodvolden:
In the meantime we have Windows builds through Appveyor though: https://ci.appveyor.com/project/dickens/libusb/builds/42292441

mcuee:
Close this now that #1353 has been merged.

PR #1102, opened 2022-03-28
Add CI build on MSYS2

tormodvolden:


tormodvolden:
The warnings provided by recent GCC can be useful for the windows backend too, as seen in #1100.

The GitHub Action provided by MSYS2 uses an updated MSYS2 (https://www.msys2.org/docs/ci/) currently with GCC 11.2. Notes about autoreconf: https://www.msys2.org/docs/autotools/

mcuee (approved):

mcuee:
This will partially fix the following as well.
* https://github.com/libusb/libusb/issues/1045

mcuee:
I think this should be safe to be merged.

Issue #1105, opened 2022-04-03
WinUSB isoch OUT doesn't report packet actual_length


tormodvolden:
Running isochronous OUT transfers on WinUSB, the log shows
`[windows_handle_transfer_completion] handling transfer 00000000009B9D70 completion with errcode 0, length 0
`
and the application receives 0 in the `libusb_iso_packet_descriptor` `actual_length` member.

Interestingly, the winusbx_copy_transfer_data() WinUSB code updates `actual_length` for IN but not for OUT.

[iso-test-winusb.log](https://github.com/libusb/libusb/files/8404454/iso-test-winusb.log)
[iso-test-linux.log](https://github.com/libusb/libusb/files/8404455/iso-test-linux.log)

mcuee:
Same here. This time I am using USB PIC Benchmark FW from libusbk project. Isoc transfer will report transfer of 0 bytes, for both libusbk and winusb driver.
Host program: https://github.com/walac/isochronous with some debug printf.

```
Successfully found pyusb benchmark firmware device.
Successfully set configuration 1.
Successfully claimed Interface 0.
Successfully claimed Interface 0 alt setting 1.
packet size is 64 .
n= 64 .
Packet 0 requested to transfer 64 bytes, transfered 0 bytes.
packet size is 64 .
Packet 0 requested to transfer 64 bytes, transfered 64 bytes.
```
[winusb_debug.zip](https://github.com/libusb/libusb/files/8404640/winusb_debug.zip)
[libusbk_debug.zip](https://github.com/libusb/libusb/files/8404641/libusbk_debug.zip)

tormodvolden:
Thanks for confirming it happens on other devices and for libusbK as well.

It has been like this in winusbx_copy_transfer_data() since WinUSB isoch was introduced in #284. I think it is because Windows usbd doesn't report the actual length for OUT transfers (it is maybe pointless anyway). However Linux and macOS seem to fill in these values, although I don't know if they have any significance or are just copied from the requested length. I started preparing a PR for this (WinUSB only) in #1107, it would be great if you can test it. But it needs extending to libusbK before it should be merged.

tormodvolden:
[...]

PR #1107, opened 2022-04-03
windows: Update isochronous OUT packet actual length

tormodvolden:
On Windows, unlike other platforms, the isochronous packet actual_length
value is not set on completion of OUT transfers. Our API requires the
user to check this value for isochronous transfers instead of the transfer
actual_length, if the transferred length is of interest.

The usbd Length field is not used for isochronous OUT transfers:
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/usb/ns-usb-_usbd_iso_packet_descriptor

To make it consistent with other platforms, just return the requested length.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -3241,7 +3241,13 @@ static enum libusb_transfer_status winusbx_copy_transfer_data(int sub_api, struc
 			// iso only supported on libusbk-based backends for now
 			PKISO_CONTEXT iso_context = transfer_priv->iso_context;
 			for (i = 0; i < transfer->num_iso_packets; i++) {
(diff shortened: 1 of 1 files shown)


tormodvolden:
I tested this with my STM32 custom firmware and it worked (WinUSB).

mcuee:
Yes this works for both libusbk and WinUSB.
https://github.com/libusb/libusb/issues/1105#issuecomment-1087053412

mcuee (approved):
Tested by Xiaofan Chen.

tormodvolden:
Thanks for testing! I think this is safe enough and good to merge now, since it makes things consistent across platforms and avoids confusion.

PR #1109, opened 2022-04-04
windows: Pass on returned isochronous packet status on libusbK

tormodvolden:
There was a pending TODO for libusbK, and AFAICS the desired solution was already applied for WinUSB.

The second commit should not be part of this but  included for now just for testing. If you see "Offset 0" in the log (or huge/negative packet lengths before applying this) you are hitting the same issue as me. This could indicate a serious issue we have to deal with later.


tormodvolden:
Note this PR includes #1107.

PR #1111, opened 2022-04-05
whitespace fix

shuffle2:


tormodvolden (approved):

tormodvolden:
Note that the continuation of the `if` clause on the second line is aligned to the opening parenthesis, and should not be fully indented like the "then" clause.

PR #1112, opened 2022-04-05
msvc: silence warning C4200 in libusb.h

shuffle2:
this is a patch I have to apply to libusb on every project I use it with, would be great to get it fixed in mainline after all this time :)

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -26,6 +26,9 @@
 #define LIBUSB_H
 
 #if defined(_MSC_VER)
+#pragma warning(push)
+/* Disable: warning C4200: nonstandard extension used : zero-sized array in struct/union */
+#pragma warning(disable:4200)
 /* on MS environments, the inline keyword is available in C++ only */
 #if !defined(__cplusplus)
 #define inline __inline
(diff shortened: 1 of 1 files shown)


tormodvolden:
What is the warning?

shuffle2:
The warning is because of libusb's `ZERO_SIZED_ARRAY`, which msvc can accept but always warns on.
When included by c++ compilation unit, it's a level-2 warning https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-levels-2-and-4-c4200?view=msvc-170

tormodvolden:
You can also include the error message verbatim in the commit message.

Is C++ relevant here?

shuffle2:
A libusb user including the header from a c++ file is more likely to be annoyed by the warning because it is activated at level 2 as opposed to being included from a C file, where it's not activated until level 4. (on msvc, level 4 is the last step before -Wall).

This causes problems when you also enable -Werror *for your own source*, since the libusb header just gets merged into the compilation unit.

On recent msvc this can be worked around by marking libusb.h as an "external header" and changing the warning level for external headers (https://docs.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170) but IMO it's nicer to just be warning-clean to begin with.

tormodvolden:
Good. I would maybe swap the text a bit and mention "zero-sized array" in the summary but I can also amend it when picking it. In general, feel free to paste more of the warning showing where it happens.

PR #1115, opened 2022-04-05
windows: use sdk-defined GUIDs where possible

shuffle2:
changes the symbols to be extern declspec(selectany)

this is a followup to https://github.com/libusb/libusb/commit/9cb2e5c111a7fc1eec13dcf73667187b06f68628 (which thankfully made the variables static!)

I think this is "the windows way", but I can understand prefering `static` over `extern __declspec(selectany)`, but just wanted to make the PR in case you weren't aware that these GUIDs are defined by the sdk. Feel free to close if there's a problem using those headers...

### DIFF
--- a/libusb/os/windows_winusb.h
+++ b/libusb/os/windows_winusb.h
@@ -24,7 +24,8 @@
 #define LIBUSB_WINDOWS_WINUSB_H
 
 #include <devioctl.h>
-#include <guiddef.h>
+#include <initguid.h>
+#include <usbiodef.h>
 
 #include "windows_common.h"
 
@@ -42,13 +43,8 @@
 // Additional return code for HID operations that completed synchronously
 #define LIBUSB_COMPLETED	(LIBUSB_SUCCESS + 1)
 
-// http://msdn.microsoft.com/en-us/library/ff545978.aspx
-// http://msdn.microsoft.com/en-us/library/ff545972.aspx
-// http://msdn.microsoft.com/en-us/library/ff545982.aspx
-static const GUID GUID_DEVINTERFACE_USB_HOST_CONTROLLER = {0x3ABF6F2D, 0x71C4, 0x462A, {0x8A, 0x92, 0x1E, 0x68, 0x61, 0xE6, 0xAF, 0x27}};
-static const GUID GUID_DEVINTERFACE_USB_HUB = {0xF18A0E88, 0xC30C, 0x11D0, {0x88, 0x15, 0x00, 0xA0, 0xC9, 0x06, 0xBE, 0xD8}};
-static const GUID GUID_DEVINTERFACE_USB_DEVICE = {0xA5DCBF10, 0x6530, 0x11D2, {0x90, 0x1F, 0x00, 0xC0, 0x4F, 0xB9, 0x51, 0xED}};
-static const GUID GUID_DEVINTERFACE_LIBUSB0_FILTER = {0xF9F3FF14, 0xAE21, 0x48A0, {0x8A, 0x25, 0x80, 0x11, 0xA7, 0xA9, 0x31, 0xD9}};
+// libusb0 Filter Device Interface GUID
+DEFINE_GUID(GUID_DEVINTERFACE_LIBUSB0_FILTER, 0xF9F3FF14, 0xAE21, 0x48A0, 0x8A, 0x25, 0x80, 0x11, 0xA7, 0xA9, 0x31, 0xD9);
 
 // The following define MUST be == sizeof(USB_DESCRIPTOR_REQUEST)
 #define USB_DESCRIPTOR_REQUEST_SIZE	12U


tormodvolden:
LGTM, not being a Win dev. Ideally we would comment on where the remaining GUID_DEVINTERFACE_LIBUSB0_FILTER is coming from (libusb-win32 ?).

shuffle2:
hrm, another thing i've thought about is that you probably want this to work with non-msvc compilers targeting windows as well? In that case, `DEFINE_GUID` might not do the right thing, because it's defined like:

```c
#ifndef DECLSPEC_SELECTANY
#if (_MSC_VER >= 1100)
#define DECLSPEC_SELECTANY  __declspec(selectany)
#else
#define DECLSPEC_SELECTANY
#endif
#endif

#define DEFINE_GUID(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8) \
        EXTERN_C const GUID DECLSPEC_SELECTANY name \
                = { l, w1, w2, { b1, b2,  b3,  b4,  b5,  b6,  b7,  b8 } }
```

I suppose mingw (or whatever) doesn't define _MSC_VER, right? Having either `selectany` or `static` is important here.

Maybe instead of the current commit in this PR, I will just add a comment about why these need to be defined by libusb (to support non-msvc) instead.

the LIBUSB0_FILTER seems to be from libusbK: https://github.com/huzqujm/usb-travis/blob/4e0fda969f31ecf3a475882b968c6d3acb5eec6a/libusbK/src/lusb_defdi_guids.h#L35

tormodvolden:
Yes, we definitely want to support MinGW (native or cross-building). Our current CI flow includes a MinGW build.

tormodvolden:
> because it's defined like

Where is it defined like this? MinGW also has DEFINE_GUID and probably has it defined in a way that compiles correctly.

tormodvolden:
The MinGW build of this PR is here https://github.com/libusb/libusb/actions/runs/2095628955

shuffle2:
it will compile on mingw but `DECLSPEC_SELECTANY` will be defined to empty value, right?
The problem with that is the resulting GUID becomes a public symbol and causes link-time name collisions if you have the same name defined by other objects, which is (un)surprisingly common. I'm not sure how to download the build artifacts from the github action run to check the binary.

For instance I hit this on my project because it was using a libusb version before https://github.com/libusb/libusb/commit/9cb2e5c111a7fc1eec13dcf73667187b06f68628 , and also linking against SDL2, which happens to also define the GUID with the same symbol name. On msvc, this can work if *all* the symbols have the `selectany` attribute, which is the way the windows sdk headers intend to solve the problem. Without `selectany`, you can just make them `static`.

I've updated this PR with a commit that hopefully will prevent well-meaning people like myself from causing regression in the future :)

shuffle2:
> Where is it defined like this? MinGW also has DEFINE_GUID and probably has it defined in a way that compiles correctly.

fwiw, it's defined like that in the windows sdk headers. entirety of  `initguid.h` is just
```c
#if _MSC_VER > 1000
#pragma once
#endif

#define INITGUID
#include <guiddef.h>
```
and `guiddef.h` then:
```c
#ifndef DECLSPEC_SELECTANY
#if (_MSC_VER >= 1100)
#define DECLSPEC_SELECTANY  __declspec(selectany)
#else
#define DECLSPEC_SELECTANY
#endif
#endif

#ifndef EXTERN_C
#ifdef __cplusplus
#define EXTERN_C    extern "C"
#else
#define EXTERN_C    extern
#endif
#endif

#ifdef DEFINE_GUID
#undef DEFINE_GUID
#endif

#ifdef INITGUID
#define DEFINE_GUID(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8) \
        EXTERN_C const GUID DECLSPEC_SELECTANY name \
                = { l, w1, w2, { b1, b2,  b3,  b4,  b5,  b6,  b7,  b8 } }
#else
#define DEFINE_GUID(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8) \
    EXTERN_C const GUID FAR name
#endif // INITGUID
```

if the version of `DEFINE_GUID` without the actual `__declspec(selectany)` inserted is used, you wind up with the linker issue.

libusb/os/windows_winusb.h
```diff
@@ -45,9 +45,13 @@
 // http://msdn.microsoft.com/en-us/library/ff545978.aspx
 // http://msdn.microsoft.com/en-us/library/ff545972.aspx
 // http://msdn.microsoft.com/en-us/library/ff545982.aspx
+// These GUIDs are provided by usbiodef.h from the Windows SDK, but use of that header by non-msvc
```
tormodvolden, line 48: Maybe instead of "the Windows SDK" say "the SDK" or system libraries, since "the Windows SDK" sounds like excluding MinGW, or do I get that wrong?

shuffle2: I refer to this: https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/ if mingw provides a different version of these same header + lib (+ other..) files, I'm not familiar with it.

tormodvolden: Exactly, what I suspected, you meant the Microsoft "Windows SDK"(TM). Which you don't use if you build with MinGW.

libusb/os/windows_winusb.h
```diff
@@ -45,9 +45,13 @@
...
+// compilers would result in the symbols being public, possibly causing a name collision during link.
```
tormodvolden, line 49: Hmm isn't it rather "the usbiodef.h in MinGW declares them as public symbols, which may clash with symbols in other objects during linking"?

But does it actually work to mix and match objects from different compilers/SDKs anyway? I thought this was not recommended and the reason for us shipping release binary libraries and objects built with different compilers. But my all means, if this mixing works, we should try keeping it working.

tormodvolden: Never mind, I guess you now found out that it won't be public on MinGW either.

shuffle2: I think I've caused a lot of confusion here because I didn't realize mingw supported `__declspec(selectany)`, however it is supported. Additionally, even though mingw provides it's own version of the headers which define `DEFINE_GUID` and those 3 GUID values, it is correct - so my original version of this PR is OK and won't cause breaks on mingw.

tormodvolden:
So if I understand correctly, the issue you had with SDL2 was fixed in libusb commit 9cb2e5c and your original patch would be fine.

shuffle2:
changed it back and added comment about the libusb0 GUID. sorry for the churn!

tormodvolden:
Tested on older MinGW also (cross-build on Ubuntu 20.04).

tormodvolden (approved):

mcuee (approved):

Issue #1110, opened 2022-04-04
Different behaviour of `libusb_cancel_transfer` on MacOS / Darwin is not documented


martinling:
On Darwin, the kernel does not make it possible to cancel a transfer on an endpoint except by aborting all transfers on that endpoint, so `libusb_cancel_transfer` does just that. This OS-specific behaviour is not mentioned in the [API documentation](https://libusb.sourceforge.io/api-1.0/group__libusb__asyncio.html#ga685eb7731f9a0593f75beb99727bbe54) for that function. Please document this.

This was raised previously in #549, where it was suggested that this is "documented somewhere", but it doesn't seem to be. The only mentions I found of this libusb behaviour were in that issue, and in [this mailing list thread](https://libusb-devel.narkive.com/wCCiU2eB/libusb-for-the-macintosh), from 13 years ago when Darwin support was first being added to libusb-0.1.

Something else that would be worth documenting here is that `libusb_cancel_transfer` on Darwin calls `ClearPipeStallBothEnds` after aborting transfers, which causes a `CLEAR_FEATURE` / `ENDPOINT_HALT` request to be sent to the device. This does not happen on other operating systems, and doesn't seem to be a common request, so some devices fail to implement it correctly.

In both #549 and our own [recent issue on HackRF](https://github.com/greatscottgadgets/hackrf/issues/1052), this caused a problem due to missing or broken handling of that request. The correct result is to reset the data toggle bit for the specified endpoint and ACK the request, as we [fixed here](https://github.com/greatscottgadgets/hackrf/pull/1081). If the device does not reset the data toggle bit, the next data packet on that endpoint can be silently dropped, leading to data loss. If it fails to ACK the request, things will hang while the host retries it.

hjelmn:
That was intended to be documented but looks like it got lost. Probably don't even have the branch with that change anymore.

As for the usage of ```ClearPipeStallBothEnds```. It is recommended by Apple after the abort to clear the data toggle. That seemed desirable at the time and this is the first I have heard of any issue (13 years later).

From the documentation:

```
[...]

PR #1117, opened 2022-04-06
darwin: do not clear device data toggle for newer versions of macOS

hjelmn:
Historically Mac OS X always cleared the data toggle on the host side. For
consistency, libusb has been calling ClearPipeStallBothEnds to also clear the
device side toggle. Newer versions of the IOUSBLib do not clear the host side
toggle so there is no need to make this call. Additionally, some buggy devices
may fail to correctly implement clearing the data toggle.

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -3,8 +3,8 @@
  * I/O functions for libusb
  * Copyright © 2007-2009 Daniel Drake <dsd@gentoo.org>
  * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
- * Copyright © 2019 Nathan Hjelm <hjelmn@cs.umm.edu>
- * Copyright © 2019 Google LLC. All rights reserved.
+ * Copyright © 2019-2022 Nathan Hjelm <hjelmn@cs.unm.edu>
+ * Copyright © 2019-2022 Google LLC. All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -311,7 +311,11 @@ if (r == 0 && actual_length == sizeof(data)) {
  * libusb_cancel_transfer() is asynchronous/non-blocking in itself. When the
  * cancellation actually completes, the transfer's callback function will
  * be invoked, and the callback function should check the transfer status to
- * determine that it was cancelled.
+ * determine that it was cancelled. On macOS and iOS it is not possible to
+ * cancel a single transfer. In this case cancelling one tranfer on an endpoint
+ * will cause all transfers on that endpoint to be cancelled. In some cases
+ * the call may cause the endpoint to stall. A call to \ref libusb_clear_halt
+ * may be needed.
  *
  * Freeing the transfer after it has been cancelled but before cancellation
  * has completed will result in undefined behaviour.
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -311,11 +311,11 @@ if (r == 0 && actual_length == sizeof(data)) {
  * libusb_cancel_transfer() is asynchronous/non-blocking in itself. When the
  * cancellation actually completes, the transfer's callback function will
  * be invoked, and the callback function should check the transfer status to
- * determine that it was cancelled. On macOS and iOS it is not possible to
- * cancel a single transfer. In this case cancelling one tranfer on an endpoint
- * will cause all transfers on that endpoint to be cancelled. In some cases
- * the call may cause the endpoint to stall. A call to \ref libusb_clear_halt
- * may be needed.
+ * determine that it was cancelled.
+ *
+ * On macOS and iOS it is not possible to cancel a single transfer. In this
+ * case cancelling one transfer on an endpoint will cause all transfers on
+ * that endpoint to be cancelled.
  *
  * Freeing the transfer after it has been cancelled but before cancellation
  * has completed will result in undefined behaviour.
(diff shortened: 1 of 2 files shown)


tormodvolden:
So on newer macOS, the behaviour will now be more consistent with other platforms? If so, it will be a large motivation factor and should be mentioned. If in some cases a libusb_clear_halt() will be need to be added, it would already have been the case for other platforms?

tormodvolden:
```
../../libusb/os/darwin_usb.c:2285:28: error: variable 'kresult' is uninitialized when used here [-Werror,-Wuninitialized]
  return darwin_to_libusb (kresult);
```

libusb/os/darwin_usb.c
```diff
@@ -2276,8 +2276,11 @@ static int darwin_abort_transfers (struct usbi_transfer *itransfer) {
 
   usbi_dbg (ctx, "calling clear pipe stall to clear the data toggle bit");
```
tormodvolden, line 2277: Maybe the usbi_dbg should be inside the #if clause.

libusb/io.c
```diff
@@ -311,7 +311,11 @@ if (r == 0 && actual_length == sizeof(data)) {
  * libusb_cancel_transfer() is asynchronous/non-blocking in itself. When the
  * cancellation actually completes, the transfer's callback function will
  * be invoked, and the callback function should check the transfer status to
- * determine that it was cancelled.
+ * determine that it was cancelled. On macOS and iOS it is not possible to
+ * cancel a single transfer. In this case cancelling one tranfer on an endpoint
+ * will cause all transfers on that endpoint to be cancelled. In some cases
+ * the call may cause the endpoint to stall. A call to \ref libusb_clear_halt
+ * may be needed.
```
martinling, line 318: I don't think this description is accurate:

> In some cases the call may cause the endpoint to stall. A call to `libusb_clear_halt` may be needed.

It's not clear to me what you mean by this call causing the endpoint to stall.

From my reading of the standard, the endpoint is only stalled/halted as a result of a `STALL` packet from the device, a bus error, or a `SetFeature(ENDPOINT_HALT)` request.

I don't think `libusb_cancel_transfer` / `AbortPipe` can cause an endpoint to be stalled, unless the call involved sending a `SetFeature(ENDPOINT_HALT)` request to the device, and I see no indication that `AbortPipe` does this.

As far as I can see, a call to `libusb_clear_halt` should only be needed if the endpoint was already known to be stalled. Calling `libusb_cancel_transfer` cannot cause it to become stalled as far as I can see.

libusb/os/darwin_usb.c
```diff
@@ -2276,8 +2276,11 @@ static int darwin_abort_transfers (struct usbi_transfer *itransfer) {
...
 
-  /* newer versions of darwin support clearing additional bits on the device's endpoint */
+#if InterfaceVersion <= 245
+  /* with older releases of IOUSBFamily the OS always clears the host side data toggle. for
+     consistency also clear the data toggle on the device. */
```
martinling, line 2281: I think this difference should be documented in the description of `libusb_cancel_transfer`, with some text along the lines of:

> On Mac OS versions prior to 10.5, this function has the side effect of clearing the host-side data toggle bit, and will send a `ClearFeature(ENDPOINT_HALT)` request to clear the device-side data toggle. If the device does not implement this request correctly, the data toggle bits on the host and device may be left out of sync.

martinling (changes requested):
I think the functional change here is correct, but I don't think the documentation is accurate. See inline comments.

libusb/os/darwin_usb.c
```diff
@@ -2276,8 +2276,11 @@ static int darwin_abort_transfers (struct usbi_transfer *itransfer) {
...
   kresult = (*(cInterface->interface))->ClearPipeStallBothEnds(cInterface->interface, pipeRef);
+#endif
 
   return darwin_to_libusb (kresult);
```
tormodvolden, line 2285: The uninitialized 'kresult' needs fixing.

tormodvolden:
@hjelmn if it is OK with you, I'll squash my kresult fixup into your commit. The docs update is still missing, I can take a look at it.

libusb/io.c
tormodvolden: Maybe this has been observed on some devices, even if it is against the standard? But I agree "In some cases" is too vague. It is tempting to remove this part if we cannot be more accurate. Something like "For some devices" would be better - if it is true.

tormodvolden:
Hmm, the text was already changed in #1121. Disregard my last "docs" commit for now.

tormodvolden:
So I basically added "On macOS versions prior to 10.5" and "on particular devices" and the rest is text rewrap.

libusb/io.c
tormodvolden: Now after reading https://github.com/libusb/libusb/issues/1110#issuecomment-1090621211 I believe the part about endpoint stalls was based on a misunderstanding and should be removed?

tormodvolden:
Corrected docs to more accurately say "When built for" macOS < 10.5. Removed comment about possible endpoint halt, which seems not correct, as explained in https://github.com/libusb/libusb/issues/1110#issuecomment-1090621211

libusb/io.c
martinling: Yes that's right.

libusb/os/darwin_usb.c
martinling: This was done in #1121.

libusb/io.c
```diff
@@ -1584,13 +1588,15 @@ int API_EXPORTED libusb_submit_transfer(struct libusb_transfer *transfer)
...
+ * - When built for macOS versions prior to 10.5, this function sends a
+ *   \c ClearFeature(ENDPOINT_HALT) request for the transfer's endpoint.
+ *   (Prior to libusb 1.0.27, this request was sent on all Darwin systems.)
+ *   If the device does not handle this request correctly, the data toggle
+ *   bits for the endpoint can be left out of sync between host and device,
+ *   which can have unpredictable results when the next data is sent on
+ *   the endpoint, including data being silently lost.  A call to
+ *   \ref libusb_clear_halt will not resolve this situation, since that
+ *   function uses the same request. Therefore, if your program runs on
  *   Darwin and uses a device that does not correctly implement
```
martinling, line 1600: Perhaps now we're adding the fix we should be more specific here and say:

> If your program runs on MacOS < 10.5 or libusb < 1.0.27, and uses a device that does not correctly implement `ClearFeature(ENDPOINT_HALT) requests, it may only be safe to cancel transfers when followed by a device reset using `libusb_reset_device`.

tormodvolden: Excellent. I made you the author as well.

martinling (approved):

tormodvolden:
I squashed the kreturn fixup into the first commit.

tormodvolden:
Fixed whitespace error of mine in text.

mcuee:
@tormodvolden 
I think this can be merged.

libusb/os/darwin_usb.c
```diff
@@ -2263,15 +2263,17 @@ static int darwin_abort_transfers (struct usbi_transfer *itransfer) {
...
+    kresult = (*(cInterface->interface))->AbortStreamsPipe (cInterface->interface, pipeRef, itransfer->stream_id);
   else
 #endif
-    (*(cInterface->interface))->AbortPipe (cInterface->interface, pipeRef);
+    kresult = (*(cInterface->interface))->AbortPipe (cInterface->interface, pipeRef);
 
-  usbi_dbg (ctx, "calling clear pipe stall to clear the data toggle bit");
-
-  /* newer versions of darwin support clearing additional bits on the device's endpoint */
+#if InterfaceVersion <= 245
```
seanm, line 2271: Is this is right test?  Shouldn't it be `MAC_OS_X_VERSION_MIN_REQUIRED < 1050`?

If you build with the 10.6 SDK, InterfaceVersion will be >= 245, but you may still have a deployment target of 10.5.

martinling: I don't think either is actually the right test, because what we really want is the version in use at runtime.

seanm: Given that we're talking about very old macOS versions, I think just altering the compile-time check is better. A runtime check would incur a tiny performance hit, and just about everyone is using >= 10.5 these days.

mcuee: Do we really care about 10.5 at all?
macports is supporting old version but only down to 10.6 Snow Leopard.
https://ports.macports.org/port/libusb/details/

seanm: I think it's nice to keep 10.5 support, it's the last version to support PowerPC. The codebase already supports it. All this PR needs is a one line change:

`#if InterfaceVersion <= 245`

to

`MAC_OS_X_VERSION_MIN_REQUIRED < 1050`

mcuee: I see. In that case, no problem from my side.

Just wondering if anyone has really tested libusb on macOS 10.5 on Power PC though.

tormodvolden: I think Nathan's reasoning is that building for older releases with newer SDK is difficult, or impossible, to get right, and the old SDKs should be used when building for older releases.

Anyway I think we can merge this as-is for a start, and then fix up the version checks systematically in all places.

Issue #1108, opened 2022-04-03
segmentation fault: 11 libusb @1.0.25_0 under macOS 


chriskuku:
When doing an `st-info --probe`  I'm occasionally getting 

```
  dev-type:   unknown
libusb: debug [darwin_event_thread_main] darwin event thread exiting
Segmentation fault: 11

```
Why?

tormodvolden:
Can you please try with libusb 1.0.26rc-1 ? It is possible that the application code is doing something wrong, but still it shouldn't segfault in libusb. Do you have the source code to your st-info?

tormodvolden:
Once you have libusb 1.0.26-rc1 e.g. with homebrew
```
brew unlink libusb
brew install libusb --HEAD
```

run with libusb debugging and attach the output:
`LIBUSB_DEBUG=4 st-info --probe`
or
`LIBUSB_DEBUG=4 st-info --probe > st-info-debug.log 2>&`1

chriskuku:
Thanks. I’m not using homebrew. Can you lead me to git checkout  or how to do that using macports?
—
Christoph

chriskuku:
> Can you please try with libusb 1.0.26rc-1 ? It is possible that the application code is doing something wrong, but still it shouldn't segfault in libusb. Do you have the source code to your st-info?

I checked it out yesterday or before yesterday from https://github.com/stlink-org/stlink
Do you want me to send you a tar ball from my checked out directory? Otherwise you can check out the source from github.
--
Christoph

tormodvolden:
You can download libusb binaries from Code -> Releases here on GitHub, or build yourself from a git checkout. macports has a libusb-devel package but it is unfortunately not up to date.

Thanks, if your st-info is latest/recent git from the official stlink repo then it's fine.

chriskuku:
I followed the homebrew advice above but it seems I'm getting into troubled waters with it:

```
$ brew install libusb --HEAD
Running 'brew update --preinstall'...
==> Auto-updated Homebrew!
Updated 2 taps (homebrew/core and homebrew/cask).
==> Updated Formulae
Updated 3 formulae.
==> Updated Casks
Updated 6 casks.

==> Downloading https://ghcr.io/v2/homebrew/core/automake/manifests/1.16.5
[...]

PR #1119, opened 2022-04-07
darwin: Fix crash in log handler when stopping event thread

benzea:
The darwin event thread (in contrast to other OS implementations) tries
to log to the context that created this. However, this context is only
guaranteed to be valid until the thread has started. It may be that
another context is the last one to be destroyed, causing the event
thread to log using an already destroyed context.

Fix this by setting ctx to NULL and explicitly logging to a NULL
context.

Fixes: #1108

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -494,6 +494,7 @@ static void *darwin_event_thread_main (void *arg0) {
   io_iterator_t          libusb_rem_device_iterator;
   io_iterator_t          libusb_add_device_iterator;
 
+  /* ctx must only be used for logging during thread startup */
   usbi_dbg (ctx, "creating hotplug event source");
 
   runloop = CFRunLoopGetCurrent ();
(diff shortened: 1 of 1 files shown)


tormodvolden:
Actually the "setting ctx to NULL" seems not fixing anything, since it is not referenced later. I guess it is just a signal to the human code reader? Or is it confusing?

I see further up in the function that ctx is passed to some hotplug registration functions. I guess it is fine, if that is just being passed as a context pointer argument to the darwin_devices_detached() callback, since the latter is not using this argument. But if it is like this it we could use NULL there also.

benzea:
> Actually the "setting ctx to NULL" seems not fixing anything, since it is not referenced later. I guess it is just a signal to the human code reader? Or is it confusing?

Yeah, the variable is unused after that. I just thought I would make it explicit that the pointer must not be dereferenced at that point.

Anyway, I just changed it to add a comment at the start of the function.

> I see further up in the function that ctx is passed to some hotplug registration functions. I guess it is fine, if that is just being passed as a context pointer argument to the darwin_devices_detached() callback, since the latter is not using this argument. But if it is like this it we could use NULL there also.

I changed it to NULL. I do not know what that parameter means though. That said, the ptr parameter to the handler functions is unused in all cases. :-)

tormodvolden (approved):

tormodvolden:
Yes, it seems the ptr parameter was unused ever since the introduction of darwin_devices_attached() in 9b76386 and ctx was used by habit.

tormodvolden:
Merged. Thanks for fixing yet another crash!

PR #1121, opened 2022-04-09
darwin: Document OS-specific behaviour of libusb_cancel_transfer

### LINKED ISSUE #1110: Different behaviour of `libusb_cancel_transfer` on MacOS / Darwin is not documented

martinling:
There are two differences in the behaviour of `libusb_cancel_transfer` on Darwin:

- Cancellation cancels all transfers on the same endpoint.

- A `ClearFeature(ENDPOINT_HALT)` request is sent after cancellation.

This documents both differences and their implications, and closes #1110.

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -1576,6 +1576,26 @@ int API_EXPORTED libusb_submit_transfer(struct libusb_transfer *transfer)
  * \ref libusb_transfer_status::LIBUSB_TRANSFER_CANCELLED
  * "LIBUSB_TRANSFER_CANCELLED."
  *
+ * This function behaves differently on Darwin-based systems (macOS and iOS):
+ *
+ * - Calling this function for one transfer will cause all transfers on the
+ *   same endpoint to be cancelled. Your callback function will be invoked with
+ *   a transfer status of
+ *   \ref libusb_transfer_status::LIBUSB_TRANSFER_CANCELLED
+ *   "LIBUSB_TRANSFER_CANCELLED" for each transfer that was cancelled.
+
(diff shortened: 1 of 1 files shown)


hjelmn:
LGTM

PR #1125, opened 2022-04-12
darwin: add missing locking around cached device list cleanup

### LINKED ISSUE #1124: stress test assert on darwin_init
In rust wrapper darwin asserted on simple test.
https://github.com/a1ien/rusb/blob/master/libusb1-sys/tests/test.rs#L20
Assertion failed: (list_empty(&darwin_cached_devices)), function darwin_init, file os/darwin_usb.c, line 603. 
https://github.com/a1ien/rusb/runs/5970785517?check_suite_focus=true#step:7:32

In libusb we have same test https://github.com/libusb/libusb/blob/master/tests/stress.c#L29 but looks like we didn't run test on mac os.

I am also didn't have macos so I can't reproduce.

P.S. I found that @mcuee run stress test.. So this is strange that it' fail on CI

hjelmn:
The cleanup funtion darwin_cleanup_devices was intented to be called only once
at program exit. When the initialization/finalization code was changed to
destroy the cached device list on last exit this function should have been
modified to take a lock on darwin_cached_devices_lock. This commit updates the
cleanup to protect it with the cached devices lock and updates darwin_init to
print out an error and return if a reference leak is detected.

Fixes #1124


tormodvolden:
Note there a "funtion" typo in the commit message, and the nano version bump.

libusb/os/darwin_usb.c
```diff
@@ -58,6 +58,8 @@ static int init_count = 0;
 static const mach_port_t darwin_default_master_port = 0;
 
 /* async event thread */
+/* if both this mutex and darwin_cached_devices_mutex are to be aquired then
```
tormodvolden, line 61: acquired

tormodvolden:
> Also using this opportunity to

Don't be shy making extra commits however trivial they may be :) We have historically been a little bit burned on commits doing too many changes at once, although this one is rather overseeable.

PR #1129, opened 2022-04-15
config: allow configure to continue without pkg-config installed

hjelmn:
libusb only uses pkg-config macros in configure to check for umockdev. Since
this is only required for testing it makes sense to protect the usage of these
macros to only be used when pkg-config is installed. This will be the case for
maintainers and testers.


hjelmn:
Merged in ba698478afc3d3a72644eef9fc4cd24ce8383a4c

PR #1133, opened 2022-04-19
descriptor: Add support for BOS platform descriptor parsing

Tiwalun:
Add support for parsing platform descriptors, as defined in USB 3.2, and show them in the xusb output.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -305,6 +305,7 @@ enum libusb_descriptor_type {
 #define LIBUSB_BT_USB_2_0_EXTENSION_SIZE	7
 #define LIBUSB_BT_SS_USB_DEVICE_CAPABILITY_SIZE	10
 #define LIBUSB_BT_CONTAINER_ID_SIZE		20
+#define LIBUSB_BT_PLATFORM_DESCRIPTOR_MIN_SIZE		20
 
 /* We unwrap the BOS => define its max size */
 #define LIBUSB_DT_BOS_MAX_SIZE				\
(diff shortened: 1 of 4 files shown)


Tiwalun:
I don't know how to update the .def file, is there a guide somewhere?

Tiwalun:
Should be fixed now, I've only set up commit signing on Linux and not yet on Windows, so only one of the commits was signed.

mcuee:
Yes this has been fixed. I think this is a useful enancement to libusb. 
@tormodvolden Please take a look as well.

tormodvolden:
I know too little about these platform descriptors to judge whether this is needed. Are there some good references on this?

In any case, this would be an extension of our API so it should be carefully reviewed, not only that it is correct code-wise, but also that it is a useful feature and that this is the best way to offer the functionality.

hjelmn:
Probably better to call this the platform capability descriptor. I see no reason to take this with that minor modification.

libusb/libusb.h
```diff
@@ -523,7 +524,10 @@ enum libusb_bos_type {
 	LIBUSB_BT_SS_USB_DEVICE_CAPABILITY = 0x03,
 
 	/** Container ID type */
-	LIBUSB_BT_CONTAINER_ID = 0x04
+	LIBUSB_BT_CONTAINER_ID = 0x04,
+
+	/** Platform descriptor */
+	LIBUSB_BT_PLATFORM_DESCRIPTOR = 0x05,
```
hjelmn, line 530: As much as I would love to take the trailing comma there is a reason we don't have these. Microsoft for a long time neglected their C compiler and it remained only ISO-C89 compliant for a long time. The trailing comma is a ISO-C99 feature and would break older Microsoft compilers. I don't care at all but we should run this by someone who cares about windows before we move libusb.h to ISO-C99.

tormodvolden: Thanks for pointing out the ISO-C89 compliance, I don't think we have that mentioned anywhere. The comma would also be inconsistent with the other enumerations in that file. I don't see it as useful to break compatibility for an unnecessary comma.

tormodvolden:
> Probably better to call this the platform capability descriptor. I see no reason to take this with that minor modification.

I checked the USB standard, it actually calls it "platform descriptor", and in the same way we have e.g. "container_id_descriptor" and that everything inside a BOS are capabilities. So it is consistent. But I know macOS calls it "platform capability descriptor" :)

tormodvolden:

PR #1149, opened 2022-06-08
Improve Doxygen generated documentation

LudovicRousseau:
Links are now created to other definitions
- `libusb_set_option()` and `LIBUSB_OPTION_NO_DEVICE_DISCOVERY` in `libusb_wrap_sys_device()`
- `LIBUSB_ERROR_*` and `LIBUSB_SUCCESS` in return values

With no negative comments I plan to merge PR


mcuee:
Not an expert of doxygen but the patches look fine to me.

mcuee (approved):

libusb/core.c
```diff
@@ -1209,10 +1209,11 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
...
- * Call libusb_set_option(NULL, LIBUSB_OPTION_NO_DEVICE_DISCOVERY) before
- * libusb_init() if you want to skip enumeration of USB devices. In particular,
- * this might be needed on Android if you don't have authority to access USB
- * devices in general.
+ * Call libusb_set_option() with the parameters (NULL,
+ * \ref LIBUSB_OPTION_NO_DEVICE_DISCOVERY) before libusb_init() if you want
+ * to skip enumeration of USB devices. In particular, this might be
+ * needed on Android if you don't have authority to access USB devices
+ * in general.
  *
```
tormodvolden, line 1217: Are you actually creating a link to libusb_set_option() here?

tormodvolden:
Why is the bump to 1.9.4 required? It seems to me that many changes in the first commit can be done without any version bump. The 2 other commits look good (apart from my comment above).

libusb/core.c
tormodvolden: Ah now I get it, it is recognized and linked because it now has no parameters.

tormodvolden:
I pushed the other 2 for now. With 1.9.0 (what is on Ubuntu 20.04) I would get warnings with the new config. I guess with 1.9.4 you'd get warnings with the old config. I don't know if there is a way to please anyone. But shouldn't 1.9.4 see the "1.9.0" version in the config and run in some compatibility mode?

LudovicRousseau:
I have no problem with the NON-upgrade of `doxygen.cfg.in` file. We can do that later when it is no more problematic with 'old' systems.

I regenerated & updated the online API documentation https://libusb.sourceforge.io/api-1.0/group__libusb__dev.html#ga98f783e115ceff4eaf88a60e6439563c
Thanks

PR #1151, opened 2022-06-13
windows_common: add clock_gettime check

neheb:
Under MSYS2, the windows usbi_get_monotonic_time interferes with the static inline function
under libusbi.h

### DIFF
--- a/libusb/os/windows_common.c
+++ b/libusb/os/windows_common.c
@@ -848,6 +848,7 @@ static int windows_handle_transfer_completion(struct usbi_transfer *itransfer)
 		return usbi_handle_transfer_completion(itransfer, status);
 }
 
+#ifndef HAVE_CLOCK_GETTIME
 void usbi_get_monotonic_time(struct timespec *tp)
 {
 	static LONG hires_counter_init;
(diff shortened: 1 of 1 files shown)


tormodvolden:
LGTM, thanks!

tormodvolden:
Looking again at this, I cannot understand how HAVE_CLOCK_GETTIME can have been set for a Windows build. Can you please try reverting this to reproduce the issue, and attach the output from running configure as well as config.log?

neheb:
I got this error from my meson build file, not the autotools one. It checks for clock_gettime no matter the OS: https://github.com/neheb/libusb/commit/c8268a6d0cef1baf909e366d1ceb271caeb0a120#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R20

I don't see a benefit in only doing so for non windows platforms.

tormodvolden:
Ah, that explains it. I was wondering if there was something wrong with our autotools configuration.

Don't be afraid to add a bit of context next time. It is useful to know how issues show up and why they need fixing.

The only benefit I see is that a Windows build with clock_gettime available will still build with the same code as other Windows builds so it is more consistent. Of course from other points of view it can be interesting that a running code is more consistent with other platforms. So it depends. It is always possible to override it by editing config.h in any case.

The commit in question here is either way good, it makes it consistent whatever be the value of HAVE_CLOCK_GETTIME.

PR #1158, opened 2022-06-24
Add VS 2022 solution

thefat32:


thefat32:
It is certainly a better approach. Closing this!

PR #1160, opened 2022-06-26
mark libusb_set_option as WINAPIV

neheb:
WINAPI evaluates to __stdcall, which does not work with variadic
functions. MSVC and MinGW implicitly demote this to __cdecl. Instead of
having this implicit, make it explicit.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2183,7 +2183,7 @@ void API_EXPORTED libusb_set_log_cb(libusb_context *ctx, libusb_log_cb cb,
  * on this platform
  * \returns \ref LIBUSB_ERROR_NOT_FOUND if LIBUSB_OPTION_USE_USBDK is valid on this platform but UsbDk is not available
  */
-int API_EXPORTED libusb_set_option(libusb_context *ctx,
+int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
 	enum libusb_option option, ...)
 {
 	int arg = 0, r = LIBUSB_SUCCESS;
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -120,8 +120,10 @@ typedef SSIZE_T ssize_t;
  */
 #if defined(_WIN32) || defined(__CYGWIN__)
 #define LIBUSB_CALL WINAPI
+#define LIBUSB_CALLV WINAPIV
 #else
 #define LIBUSB_CALL
+#define LIBUSB_CALLV
 #endif /* _WIN32 || __CYGWIN__ */
 
 /** \def LIBUSB_API_VERSION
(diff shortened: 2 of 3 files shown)


mcuee:
I think this makes sense.

mcuee:
Hopefully more people can come here and review this change.

seanm:
Looks reasonable.

mcuee:
@tormodvolden 
How do you like this?

tormodvolden:
This looks correct, from the looks of it :) But I haven't been able to find a reference confirming this is the exact thing to do. E.g. https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-wsprintfa shows Microsoft using WINAPIV but where is it written when to use WINAPIV? I follow the logic of CDECL needed for varargs, and WINAPIV ending up being __cdecl, and the "V" in the end could very well be something about variadic though... https://docs.microsoft.com/en-us/office/client-developer/excel/developing-dlls mentions WINAPIV without really defining its proper use.

tormodvolden:
No one can find the semantic definition of WINAPIV? :)

mcuee:
> No one can find the semantic definition of WINAPIV? :)

Not so sure if the following helps.
https://azrael.digipen.edu/~mmead/www/Courses/CS220/TipsAndTechniques-1.html
```
cdecl - Function names are decorated with a leading underscore ( _ ) and the case is preserved. This is the default and is what enables the vararg macros to work correctly.

#elif (_MSC_VER >= 800) || defined(_STDCALL_SUPPORTED)
#define CALLBACK    __stdcall
#define WINAPI      __stdcall
#define WINAPIV     __cdecl
#define APIENTRY    WINAPI
#define APIPRIVATE  __stdcall
#define PASCAL      __stdcall
```

mcuee:
The following may help as well.
https://docs.microsoft.com/en-us/cpp/cpp/cdecl?source=recommendations&view=msvc-170
__cdecl is the default calling convention for C and C++ programs. Because the stack is cleaned up by the caller, it can do vararg functions. The __cdecl calling convention creates larger executables than [__stdcall](https://docs.microsoft.com/en-us/cpp/cpp/stdcall?view=msvc-170), because it requires each function call to include stack cleanup code. The following list shows the implementation of this calling convention. The __cdecl modifier is Microsoft-specific.

tormodvolden:
Thanks, that confirms it will technically do the right thing, as my comment above, but not what the intention is. It seems pretty clear that WINAPIV is like WINAPI but for varargs, but it strange that this isn't spelled out somewhere.

tormodvolden:
So do I understand correctly that clang doesn't implicitly demote this to __cdecl ?

neheb:
It does. This is a warning fix.

tormodvolden:
Alright. Yes, I could see it was building (but warning) once I got the clang CI running in the meantime.

tormodvolden:

PR #1162, opened 2022-06-28
libusb-1.0.def: remove LIBRARY

neheb:
Under MSYS2 and WrapDB, LIBRARY must be removed as otherwise proper linking
cannot take place.


mcuee:
I do not think this is good as it may break other tools.

neheb:
It turns out the issue was different. Fixed in another way.

Issue #1154, opened 2022-06-17
libusb sample hotplugtest.c not worked in android libusb v1.0.24


Nithu7:
Hi Team,

I have tried to run libusb sample **hotplugtest.c**  on android platform (OS v12) with my usb camera device and it is NOT working.
I checked the AOSP libusb version and it is **v1.0.24**

I replaced this with libusb version **v.1.0.21** & **v1.0.22** and it is working fine and I am not albe to see notification during device connect/disconnect on same android platform. I replaced libusb version **v1.0.23** and this also NOT working.

so v1.0.21 & 1.0.22 is working and v1.0.23 & v1.0.24 NOT working.
But In linux environment (ubntu 18.04) I did not see this kind of issue with **v1.0.24** as well.

So I Suspected this is something related with android + libusb version > v1.0.22
`
I just want to quickly check is this known issue,  is there any fix or patch is there?
Could you suggest?

Regards
Muthu

mcuee:
Can you try the following on top of 1.0.26 release? We do not support AOSP version -- you have to ask Google.
* #874

Nithu7:
Hi mcuee,

Sure I will try it,

Between AOSP libusb version I meant below. (v1.0.24 for android-12)
https://cs.android.com/android/platform/superproject/+/android-12.1.0_r8:external/libusb/

Nithu7:
Hi mcuee,

Updating information:

**v1.0.22 from https://github.com/libusb/libusb/releases**   **[OK]**
1. Download & Compile v1.0.22 from the link https://github.com/libusb/libusb/releases on my android environment.
2. Then take hotplugtest.c  from below path then generate executable by linking with v1.0.22  // Native compilation not JNI
    https://github.com/libusb/libusb/tree/master/examples
3. Confirmed hotplugtest.c Send event properly during device disconnect/reconnect.

**v1.0.23 from https://github.com/libusb/libusb/releases**  **[ FAIL]**
1. Download & Compile v1.0.23 from the link https://github.com/libusb/libusb/releases on my android environment.
2. Then take hotplugtest.c  from below path then generate executable by linking with v1.0.23  // Native compilation not JNI
    https://github.com/libusb/libusb/tree/master/examples
3. Confirmed hotplugtest.c NOT sending event during device disconnect/reconnect.

We compared  v1.0.22 & v1.0.23 libusb source and randomly applied following patch, after that v1.0.23 seems to be working for hotplugtest.c in android.

```
[...]

PR #1168, opened 2022-07-31
linux_usbfs: Enable libusb hotplug events for Android native services

MuthuSanjeevi:
Android native services with proper permission still need an support for hotplug events from libusb instead of android platform API. 

Reference:
https://github.com/libusb/libusb/issues/1154

### DIFF
--- a/README
+++ b/README
@@ -4,6 +4,7 @@
 [![Build Status](https://ci.appveyor.com/api/projects/status/xvrfam94jii4a6lw?svg=true)](https://ci.appveyor.com/project/LudovicRousseau/libusb)
 [![Coverity Scan Build Status](https://scan.coverity.com/projects/2180/badge.svg)](https://scan.coverity.com/projects/libusb-libusb)
 
+#
 libusb is a library for USB device access from Linux, macOS,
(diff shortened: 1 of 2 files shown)


tormodvolden:
If this is a simple revert of commit 2f3bc98 (#242) it should be presented as such as well. But what about the issue solved by that commit? Would it pop up again?

aiuto:
Hi all.  I'm one of the people who supports libusb inside Google. I also do a lot of work with broad cross OS portability.

From that point of view, and @tormodvolden's question above, this feels like a situation where the Android NDK version needs to be taken into account.  So the testing for this needs to include both the NDK one would have had in 2017, AND the default NDK one would get today. @Nithu7 Have you tried this patch on real hardware with both an older and newer NDK?

CraigHutchinson:
@Nithu7 Could be related to https://github.com/xloem/libusb/pull/3

mcuee:
@MuthuSanjeevi 

Closing this very old PR. Please reopen if you want to further proceed the PR.

mcuee:
Better to keep this PR open.

PR #1170, opened 2022-08-08
Fix config.h to define __FUNCTION__ for vs2015

amot-dev:
I wasn't able to build libusb using vs2015 until making this fix. All other versions worked fine. 
image

What ended up fixing this was changing config.h to also define \_\_func__ as \_\_FUNCTION__ for vs2015 (currently only done on vs2013).

This seems like a weird fix, since \_\_func__ should be supported on vs2015 since it supports the C++11 standard. Thoughts?

### DIFF
--- a/msvc/config.h
+++ b/msvc/config.h
@@ -10,8 +10,8 @@
 #error "Visual Studio 2013 or later is required."
 #endif
 
-/* Visual Studio 2013 does not support __func__ */
-#if (_MSC_VER < 1900)
+/* Visual Studio 2013 and 2015 do not support __func__ */
+#if (_MSC_VER <= 1900)
 #define __func__ __FUNCTION__
 #endif
 


Youw:
> should be supported on vs2015 since it supports the C++11 standard

Maybe you have to explicitly enable the C++ support by `/std:c++11` flag?
If that's the case, then we need to check here also for `_MSC_VER < 1900 || __cplusplus < 201103L` instead

tormodvolden:
Why don't why see this in our Appveyor CI builds, where vs2015 is included?

amot-dev:
Adding the flag worked, but now I'm wondering if that's mainly because of the `_MSC_VER < 1900 || __cplusplus < 201103L` line. I'm using v140 build tools but the VS 2022 IDE. Would that affect `_MSC_VER`?

Youw:
> Why don't we see this in our Appveyor CI builds, where vs2015 is included?

My guess in that environment the C++11 is enabled by default, so macro is available.

> I'm using v140 build tools but the VS 2022 IDE. Would that affect _MSC_VER?

IDE version doesn't affect the `_MSC_VER`. Only version of the compiler (as provided by build tools) and compilation flags.

Youw:
> This seems like a weird fix, since __func__ should be supported on vs2015 since it supports the C++11 standard. Thoughts?

> Adding the flag worked, but now I'm wondering if that's mainly because of the _MSC_VER < 1900 || __cplusplus < 201103L line.

> My guess in that environment the C++11 is enabled by default, so macro is available.

I just realized how silly all of these statements are, since we're talking about compiling a `C` source, and not `C++`.

mcuee:
> I just realized how silly all of these statements are, since we're talking about compiling a `C` source, and not `C++`.

Sorry I know very little about compiler thingy and even C language in general, why are the statements silly? Is it that the following codes in `libusbi.h` makes it needs to be compiled as C++ under Visual C++, since it use `__func__`.
https://github.com/libusb/libusb/blob/c060e9ce30ac2e3ffb49d94209c4dae77b6642f7/libusb/libusbi.h#L315
`#define _usbi_log(ctx, level, ...) usbi_log(ctx, level, __func__, __VA_ARGS__)`

Ref: https://docs.microsoft.com/en-us/cpp/cpp/func?view=msvc-140
> (C++11) The predefined identifier __func__ is implicitly defined as a string that contains the unqualified and unadorned name of the enclosing function. __func__ is mandated by the C++ standard and is not a Microsoft extension.

Youw:
> why are the statements silly?

C and C++ - are different languages. Both have its own standard/documentation and what is most important - they have *different compilers*.

It just happened to be that C++ compiler *can* compile *many of* the C sources (but not even _most of them_).

The fact that `cl.exe` (MSVC compilers) compiles both `.c` and `.cpp` files - doesn't mean it threats those equaly or that it has to.

---
> Ref: https://docs.microsoft.com/en-us/cpp/cpp/func?view=msvc-140

That piece of documentation is explicitly written for C++ compiler, which *technically* means that it has nothing to do to compiling `.c` source files.

I say "technically", because I take into account that this is MSVC, and any number of bugs or misconceptions may be present there.

> the following codes in `libusbi.h` makes it needs to be compiled

`.h` files are not being compiler by the compiler. Those are textually included into a temporary `.c` (or `.cpp`) where it was `#include`d (e.g. `libusb/core.c`) and compiled as part of that source, *with the compiler for that source*.

Youw:
The `__func__` variable is defined by the C99 standard (but not C++99/03), and the fact that MSVC2013 and earlied don't have it when compiling `.c` sources - only means it is a MSVC bug.

We just need to figure out what is the different anout of bugs included in MSVC2015 and VS2015 Build Tools for MSVC2022, for this case.

PR #1174, opened 2022-08-17
Some miscellaneous fixes to the Solaris backend

jclulow:
We're using libusb pretty heavily at [Oxide](https://oxide.computer) to drive a variety of debugger and programming devices.  We're running an [illumos](https://illumos.org)-based OS, which is a descendant of Solaris.

This PR comprises a few small fixes to the existing SunOS (really Solaris) backend that were easy enough to test, and which I made before splitting out a new illumos backend.  I wanted to get them up into the original Solaris backend in case anybody is still using that.

The illumos backend bits are in [our `illumos-v1.0.25` branch](https://github.com/libusb/libusb/compare/master...oxidecomputer:libusb:illumos-v1.0.25) and I will be following up with pull requests to integrate that backend.

Please let me know how I can help get this integrated!  Thank you.

### DIFF
--- a/libusb/os/sunos_usb.c
+++ b/libusb/os/sunos_usb.c
@@ -1,6 +1,6 @@
 /*
- *
  * Copyright (c) 2016, Oracle and/or its affiliates.
+ * Copyright 2023 Oxide Computer Company
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -831,7 +831,7 @@ sunos_check_device_and_status_open(struct libusb_device_handle *hdl,
 {
 	char	filename[PATH_MAX + 1], statfilename[PATH_MAX + 1];
 	char	cfg_num[16], alt_num[16];
-	int	fd, fdstat, mode;
+	int	fd, fdstat, mode, e;
 	uint8_t	ifc = 0;
 	uint8_t	ep_index;
 	sunos_dev_handle_priv_t *hpriv;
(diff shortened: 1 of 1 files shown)


tormodvolden:
I am glad to see someone is looking at the sunos/solaris backend, and I am also curious on how much it is used. There was a contributed bug fix just two years ago though.

Yes, this looks fine. Question related to first commit: Are some errno macros positive on solaris or illumos?

About the second commit: The commit message could be more specific and mention that this prevents possible overflow, rather than just say "improvements". And the "if" statement now, already more than before, reminds me of a C obfuscation contest. Maybe a separate "if" clause for statfilename would be tolerable?

libusb/os/sunos_usb.c
```diff
@@ -831,7 +831,7 @@ sunos_check_device_and_status_open(struct libusb_device_handle *hdl,
 {
 	char	filename[PATH_MAX + 1], statfilename[PATH_MAX + 1];
 	char	cfg_num[16], alt_num[16];
-	int	fd, fdstat, mode;
+	int	fd, fdstat, mode, e;
```
tormodvolden, line 834: Looks like the other lines are tab aligned, and this one could be fixed up at the same time.

libusb/os/sunos_usb.c
```diff
@@ -870,11 +870,17 @@ sunos_check_device_and_status_open(struct libusb_device_handle *hdl,
...
-	    "out", (ep_addr & LIBUSB_ENDPOINT_ADDRESS_MASK));
-	(void) snprintf(statfilename, PATH_MAX, "%sstat", filename);
+	    "out", (ep_addr & LIBUSB_ENDPOINT_ADDRESS_MASK))) < 0 ||
+	    e >= (int)sizeof (filename) ||
+	    (e = snprintf(statfilename, sizeof (statfilename), "%sstat",
+	    filename)) < 0 || e >= (int)sizeof (statfilename)) {
+		usbi_dbg(HANDLE_CTX(hdl),
+		    "path buffer overflow for endpoint 0x%02x", ep_addr);
+		return (EINVAL);
+	}
```
seanm, line 883: The readability of this could be improved I think. Break it in into smaller more digestible components.

mcuee:
@jclulow 

Please help to address the comments. Thanks.

jclulow:
With my apologies, I am definitely trying to make time to get it -- I've been really busy with a lot of higher priority stuff at work this quarter.

FWIW, though it's not perfect, I think it's actually good to integrate as-is, as in incremental improvement.  I'll try to get to the comments soon though.

libusb/os/sunos_usb.c
tormodvolden: Never mind, I must have seen it wrongly.

libusb/os/sunos_usb.c
tormodvolden: I added a commit changing the whitespace to make it more readable. I was not able to add it to this PR (maybe blocked for contributions?) but please see it here: https://github.com/tormodvolden/libusb/commits/pr1174_sunos

jclulow: Hi, sorry for being absent.  I've taken another swing at it by breaking it up even more into two separate blocks, and making the checks separate from the call.

Is that better?

tormodvolden: Yes, thanks, that is better. I had made the multi-line arguments to sprintf vertically aligned (to the first), but it is not consistent in this file anyway and if SunOS (or what it is called nowadays) coding style is different, that's fine.

Maybe it would be helpful to have slightly different error messages for the two cases.

tormodvolden: > Maybe it would be helpful to have slightly different error messages for the two cases.

Ah, they are :) Thanks.

PR #1183, opened 2022-08-30
fix memory leak - configuration descriptor wan't released

Youw:
- Darwin: config descriptor never released in case if GetPipePropertiesV3 fails;
- Windows: config descriptor not released for some broken config descriptors;

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -525,7 +525,7 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 
 	if (iface >= conf_desc->bNumInterfaces) {
 		usbi_err(HANDLE_CTX(dev_handle), "interface %d out of range for device", iface);
-		r = LIBUSB_ERROR_NO_MEM;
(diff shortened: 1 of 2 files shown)


libusb/os/windows_winusb.c
```diff
@@ -525,7 +525,8 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 
 	if (iface >= conf_desc->bNumInterfaces) {
 		usbi_err(HANDLE_CTX(dev_handle), "interface %d out of range for device", iface);
-		return LIBUSB_ERROR_NOT_FOUND;
+		r = LIBUSB_ERROR_NO_MEM;
```
tormodvolden, line 528: You change it from LIBUSB_ERROR_NOT_FOUND to LIBUSB_ERROR_NO_MEM, why?

Youw:
```suggestion
		r = LIBUSB_ERROR_NOT_FOUND;
```

tormodvolden:
Thanks! Good catch, by the way.

PR #1187, opened 2022-08-31
core: Avoid mutex lock while printing error and warning

tormodvolden:
AFAICS the mutex was only ensuring consistency across the two "if" comparisons. So instead use a temporary variable for this.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1398,20 +1398,22 @@ static void do_close(struct libusb_context *ctx,
 	for_each_transfer_safe(ctx, itransfer, tmp) {
 		struct libusb_transfer *transfer =
 			USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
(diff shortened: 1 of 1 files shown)


tormodvolden:
The original plan was to use atomic_load(), but that only works with atomic_t and thus failed on macOS. I guess the mutex lock will then still be needed to ensure atomic access. But limit it as much as possible.

tormodvolden:
@dmitry-zakablukov Maybe you can help to review this?

dmitry-zakablukov:
@tormodvolden , I have looked on changes, looks good to me. But I can't add myself as a reviewer to approve this pr.

tormodvolden:
Thanks! Yes, I tried to add you through the GitHub Reviewers function too, it didn't work, but it is not so important either. I can add a Reviewed-By: to the commit message, which I find better anyway.

PR #1188, opened 2022-09-01
msvc: add configuration for specifying /MT compilation flag in order to statically link vc++ runtime dependencies

sonatique:
and adapt .def file to allow different dll output file names without conflicts (there is a priori no need to specify output name twice, once in the "$target" properties and once in the .def.

The idea here is that by using /MT instead of /MD, the resulting Windows dll do include all vc++ runtime dependencies, avoiding the need to provide them separately in the well-known dll.

In my opinion this could become the default build for Windows since the size of the dll remains reasonable (around 280KB) and  this avoid the hassle of having to distribute the vc++ runtime dll when it is otherwise not needed.


tormodvolden:
> and adapt .def file to allow different dll output file names without conflicts (there is a priori no need to specify output name twice, once in the "$target" properties and once in the .def.

Sounds good, this was likely also the issue in  #1162.

> The idea here is that by using /MT instead of /MD, the resulting Windows dll do include all vc++ runtime dependencies, avoiding > the need to provide them separately in the well-known dll.

Can also be added to the commit message, for clarity. Maybe also what /MT is defined as, or a link to https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library :)

tormodvolden:
The failed CI jobs must also be sorted out.

sonatique:
> The failed CI jobs must also be sorted out.
Yes indeed, I'll have a look asap

sonatique:
> > The failed CI jobs must also be sorted out.
> 
> Yes indeed, I'll have a look asap

Still trying to understand the cause of the failure

sonatique:
> > > The failed CI jobs must also be sorted out.
> > 
> > 
> > Yes indeed, I'll have a look asap
> 
> Still trying to understand the cause of the failure

Seems OK now, was a problem with how I initialy modified .def file

sonatique:
Hello: is there a way to restarted failed appveyor builds? The failure we have here doesn't seem to be related to my actual modifications:
image
Thanks!

tormodvolden:
I don't know if anyone other than the owner of the job can rekick it. Anyway, what you can do is simply force repush your branch. E.g. refresh it with `git commit --amend --reset-author` and then `git push -f` _your-remote_

tormodvolden:
It seems you now removed "LIBRARY" from the top of the .def file.

mcuee:
> It seems you now removed "LIBRARY" from the top of the .def file.

I believe removing "LIBRARY" is not correct.
* https://github.com/libusb/libusb/pull/1162

sonatique:
> > It seems you now removed "LIBRARY" from the top of the .def file.
> 
> I believe removing "LIBRARY" is not correct.
> 
> * [libusb-1.0.def: remove LIBRARY #1162](https://github.com/libusb/libusb/pull/1162)

Oh, yes, OK I just read this other issue. It seems however that all build chain tested by appveyor are OK with it.
I have to admit my knowledge are limited here. Under MSCV, just removing the name after LIBRARY is fine, but then some build failed here. 

I'll do research about it. You seemed to have a strongly based opinion in the other thread, do you have any reference link for me to start with? Thanks

sonatique:
> 

I repushed, thanks.

sonatique:
Sorry to give you all this additionnal work, but I am now facing the "base branch requires all commits to be signed". I signed my commit with `git commit -S` using my public SSH key, but it doesn't seem to be sufficient for the system to accept my commit. I alread followed related github "how to" but to no avail.
Any pointers where I could look at to solve this issue?

tormodvolden:
The signing warning is a non-issue here, you can ignore it.

sonatique:
> > It seems you now removed "LIBRARY" from the top of the .def file.
> 
> I believe removing "LIBRARY" is not correct.
> 
> * [libusb-1.0.def: remove LIBRARY #1162](https://github.com/libusb/libusb/pull/1162)

@mcuee : I tried a completly different approach, not touching the .def file. What do you thing?

mcuee:
> @mcuee : I tried a completly different approach, not touching the .def file. What do you think?

Thanks for the updates. I am okay with the changes now.

sonatique:
> The changes to the vcxproj files should be placed in `ProjectConfigurations.Base.props` instead, and dropped from the vcxproj files. otherwise looks fine to me

Thank you @shuffle2 , and sorry that I initially missed that. I just pushed a new commit to do as you wrote.

tormodvolden:
@shuffle2 Thanks a lot for your review (also in other places)!

This looks fine to me now.

tormodvolden:
@sonatique Merged, thanks for your contribution!

I massaged the commit message to hopefully be more readable.

PR #1191, opened 2022-09-03
Find interface descriptor by bInterfaceNumber

Youw:
In some cases, the interface number doesn't match its index in the list of interfaces in configuration descriptor;
that is the case for some (non USB-compliant) devices, e.g.:
    - when the device (descriptor) is explicitly non USB-compliant, and its interface numbers simply doesn't match its indexes (e.g. if[3]{0, 3, 4} or if[4]{0,3,1,2});
    - other cases when the decice descriptor is broken;

In such cases need to find the descriptor by iteration over all of them, to find the matching `bInterfaceNumber`.
Otherwise we might try to access data outside of the interfaces array, or access incorrect interfaces descriptor which `bInterfaceNumber` doesn't match the required one.

Related: #1093

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1379,13 +1379,14 @@ static IOReturn darwin_get_interface (usb_device_t **darwin_device, uint8_t ifc,
 static const struct libusb_interface_descriptor *get_interface_descriptor_by_number(struct libusb_device_handle *dev_handle, struct libusb_config_descriptor *conf_desc, int iface, uint8_t altsetting)
 {
   int i;
+
   for (i = 0; i < conf_desc->bNumInterfaces; i++) {
     if (altsetting < conf_desc->interface[i].num_altsetting && conf_desc->interface[i].altsetting[altsetting].bInterfaceNumber == iface) {
       return &conf_desc->interface[i].altsetting[altsetting];
     }
   }
 
-  usbi_err(HANDLE_CTX(dev_handle), "interface %d not found for device", iface);
+  usbi_err(HANDLE_CTX(dev_handle), "interface %d with altsetting %d not found for device", iface, (int)altsetting);
   return NULL;
 }
 
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1376,6 +1376,19 @@ static IOReturn darwin_get_interface (usb_device_t **darwin_device, uint8_t ifc,
   return kIOReturnSuccess;
 }
 
+static const struct libusb_interface_descriptor *get_interface_descriptor_by_number(struct libusb_device_handle *dev_handle, struct libusb_config_descriptor *conf_desc, int iface, uint8_t altsetting)
+{
+  int i;
+  for (i = 0; i < conf_desc->bNumInterfaces; i++) {
+    if (altsetting < conf_desc->interface[i].num_altsetting && conf_desc->interface[i].altsetting[altsetting].bInterfaceNumber == iface) {
+      return &conf_desc->interface[i].altsetting[altsetting];
+    }
+  }
+
+  usbi_err(HANDLE_CTX(dev_handle), "interface %d not found for device", iface);
+  return NULL;
+}
+
 static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle, uint8_t iface) {
   struct darwin_device_handle_priv *priv = usbi_get_device_handle_priv(dev_handle);
 
@@ -1417,6 +1430,7 @@ static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle,
     if (kresult != kIOReturnSuccess) {
       /* probably a buggy device. try to get the endpoint address from the descriptors */
       struct libusb_config_descriptor *config;
+      const struct libusb_interface_descriptor *if_desc;
       const struct libusb_endpoint_descriptor *endpoint_desc;
       UInt8 alt_setting;
 
@@ -1431,13 +1445,13 @@ static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle,
         return rc;
       }
 
-      if (iface >= config->bNumInterfaces) {
-        usbi_err (HANDLE_CTX (dev_handle), "interface %d out of range for device", iface);
+      if_desc = get_interface_descriptor_by_number (dev_handle, config, iface, alt_setting);
+      if (if_desc == NULL) {
         libusb_free_config_descriptor (config);
         return LIBUSB_ERROR_NOT_FOUND;
       }
 
-      endpoint_desc = config->interface[iface].altsetting[alt_setting].endpoint + i - 1;
+      endpoint_desc = if_desc->endpoint + i - 1;
 
       cInterface->endpoint_addrs[i - 1] = endpoint_desc->bEndpointAddress;
       libusb_free_config_descriptor (config);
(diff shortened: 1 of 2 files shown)


tormodvolden:
For Darwin:
- I understand the change affects only (and is only needed for) devices where GetPipeProperties fails because of broken endpoint descriptors (workaround introduced in commit aa1d76c). Have you tested this on such a device?
- The second line of get_endpoints() is:
  `struct darwin_interface *cInterface = &priv->interfaces[iface];`
  Doesn't it already go wrong here, since iface is used as an index?
- Also darwin_claim_interface() starts with the same line.

tormodvolden:
Alright, I guess points 2 and 3 above are fine if the indexing is consistent.

libusb/os/windows_winusb.c
```diff
@@ -485,6 +485,19 @@ static int get_interface_by_endpoint(struct libusb_config_descriptor *conf_desc,
...
+static const struct libusb_interface_descriptor *get_interface_descriptor_by_number(struct libusb_device_handle *dev_handle, struct libusb_config_descriptor *conf_desc, int iface, uint8_t altsetting)
+{
+	int i;
+	for (i = 0; i < conf_desc->bNumInterfaces; i++) {
+		if (altsetting < conf_desc->interface[i].num_altsetting && conf_desc->interface[i].altsetting[altsetting].bInterfaceNumber == iface) {
+			return &conf_desc->interface[i].altsetting[altsetting];
+		}
+	}
+
+	usbi_err(HANDLE_CTX(dev_handle), "interface %d not found for device", iface);
```
tormodvolden, line 497: Maybe also include altsetting in error message?

libusb/os/windows_winusb.c
```diff
@@ -485,6 +485,19 @@ static int get_interface_by_endpoint(struct libusb_config_descriptor *conf_desc,
 	return LIBUSB_ERROR_NOT_FOUND;
 }
 
+static const struct libusb_interface_descriptor *get_interface_descriptor_by_number(struct libusb_device_handle *dev_handle, struct libusb_config_descriptor *conf_desc, int iface, uint8_t altsetting)
+{
+	int i;
```
tormodvolden, line 490: To be strict we always have an empty line after declarations.

tormodvolden:
Can we also describe what problem we are solving, how it would manifest? On a broken device it would show "interface x out of range for device" for a high enough interface number and it would be impossible to use the interface, or potentially try to access the wrong interface for lower interface numbers?

libusb/os/darwin_usb.c
```diff
@@ -1376,6 +1376,19 @@ static IOReturn darwin_get_interface (usb_device_t **darwin_device, uint8_t ifc,
   return kIOReturnSuccess;
 }
 
+static const struct libusb_interface_descriptor *get_interface_descriptor_by_number(struct libusb_device_handle *dev_handle, struct libusb_config_descriptor *conf_desc, int iface, uint8_t altsetting)
+{
+  int i;
```
Youw, line 1381:
```suggestion
  int i;

```

libusb/os/darwin_usb.c
```diff
@@ -1376,6 +1376,19 @@ static IOReturn darwin_get_interface (usb_device_t **darwin_device, uint8_t ifc,
...
+  for (i = 0; i < conf_desc->bNumInterfaces; i++) {
+    if (altsetting < conf_desc->interface[i].num_altsetting && conf_desc->interface[i].altsetting[altsetting].bInterfaceNumber == iface) {
+      return &conf_desc->interface[i].altsetting[altsetting];
+    }
+  }
+
+  usbi_err(HANDLE_CTX(dev_handle), "interface %d not found for device", iface);
```
Youw, line 1388:
```suggestion
  usbi_err(HANDLE_CTX(dev_handle), "interface %d with altsetting %d not found for device", iface, (int)altsetting);
```

libusb/os/windows_winusb.c
Youw:
```suggestion
	int i;

```

libusb/os/windows_winusb.c
Youw:
```suggestion
	usbi_err(HANDLE_CTX(dev_handle), "interface %d with altsetting %d not found for device", iface, (int)altsetting);
```

tormodvolden:
@Eekle Can you please also try this PR, with your device from #1039?

Youw:
> For Darwin:

1. I've explicitly temporary commented out the `GetPipeProperties` code path to make sure the modified code branch is used, and checked it that way. Checked before and after the fix to make sure it is needed.
2. `priv->interfaces` is an array of fixed size of MAX_NUM_NTERFACES elements, which is always indexed by the `bInterfaceNumber` and not the interface index in the descriptor.
3. See above - not an issue.

Youw:
> describe what problem we are solving

Updated PR description. Doesn't look like what you where thinking?

Eekle:
> @Eekle Can you please also try this PR, with your device from #1039?

Could you give me a minimal example I can build to test this? My original bug report was from a downstream program, I'm not familiar with running libusb on its own.

Eekle:
I have tested v1.0.25 of libusb and this PR, by building and running `xusb.exe VID:PID` (with the VID and PID of the offending device.

v1.0.25 produces a read access violation. This PR correctly reports the device state:

```Reading first configuration descriptor:
              total length: 135
         descriptor length: 9
             nb interfaces: 3
              interface[0]: id = 0
interface[0].altsetting[0]: num endpoints = 0
   Class.SubClass.Protocol: 01.01.00
              interface[1]: id = 1
interface[1].altsetting[0]: num endpoints = 0
   Class.SubClass.Protocol: 01.02.00
interface[1].altsetting[1]: num endpoints = 1
   Class.SubClass.Protocol: 01.02.00
       endpoint[0].address: 01
           max packet size: 00C0
          polling interval: 01
              interface[2]: id = 3
interface[2].altsetting[0]: num endpoints = 1
   Class.SubClass.Protocol: 03.00.00
       endpoint[0].address: 82
           max packet size: 0004
          polling interval: 01

tormodvolden:
@Eekle Thanks a lot! Can you also test current master (without this PR), if it is not too much trouble. Current master shouldn't crash like v1.0.25 but we may see interfaces being mixed up.

Eekle:
Current master produces the same output:

``` Reading first configuration descriptor:
              total length: 135
         descriptor length: 9
             nb interfaces: 3
              interface[0]: id = 0
interface[0].altsetting[0]: num endpoints = 0
   Class.SubClass.Protocol: 01.01.00
              interface[1]: id = 1
interface[1].altsetting[0]: num endpoints = 0
   Class.SubClass.Protocol: 01.02.00
interface[1].altsetting[1]: num endpoints = 1
   Class.SubClass.Protocol: 01.02.00
       endpoint[0].address: 01
           max packet size: 00C0
          polling interval: 01
              interface[2]: id = 3
interface[2].altsetting[0]: num endpoints = 1
   Class.SubClass.Protocol: 03.00.00
       endpoint[0].address: 82
           max packet size: 0004
          polling interval: 01

tormodvolden:
Thanks! OK, I guess your device is not broken enough to validate this PR :)

Youw:
Anything I can do to help progress with this?

mcuee:
@tormodvolden 
I think this is safe to merge. Please check if you are okay to merge it or not. Thanks.

tormodvolden:
Thanks @Youw for fixing this, and @Eekle for your testing.

PR #1193, opened 2022-09-04
Allow Windows build on clang - Issue1152

tormodvolden:


neheb:
In MSYS2, there's also this patch:

https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libusb/fix-dll-exports.patch

with this description: https://github.com/msys2/MINGW-packages/commit/41f09e3a46104f5292af6ce3bb7ac05e86144315

tormodvolden:
Just adding that MSYS2 patch, I get such warnings on 32-bit build, but not on 64-bit build:
```
/usr/bin/i686-w64-mingw32-ld: warning: resolving _libusb_alloc_streams by linking to _libusb_alloc_streams@16
Use --enable-stdcall-fixup to disable these warnings
Use --disable-stdcall-fixup to disable these fixups
/usr/bin/i686-w64-mingw32-ld: warning: resolving _libusb_alloc_transfer by linking to _libusb_alloc_transfer@4
```

neheb:
hmm I would have thought that's fixed by https://github.com/libusb/libusb/pull/1193/commits/b7d4da82b5ef5d188de84b081f73604d36e06b34

tormodvolden:
Well I am building with GCC so it is not affected by that commit. I haven't tested the MSYS2 patch with clang.

tormodvolden:
Has MSYS been building libusb with clang all this time? Where do they filter out --add-stdcall-alias and -mwin32?

tormodvolden:
Ah, now I saw it in their PKGBUILD file. A bit sad they are doing this but never approach us to get it fixed properly upstream.

mcuee:
> Ah, now I saw it in their PKGBUILD file. A bit sad they are doing this but never approach us to get it fixed properly upstream.

Indeed. I have not checked out their PKGBUILD file for quite some time, if not I should be able to catch this earlier.

mcuee:
> Just adding that MSYS2 patch, I get such warnings on 32-bit build, but not on 64-bit build:
> 
> ```
> /usr/bin/i686-w64-mingw32-ld: warning: resolving _libusb_alloc_streams by linking to _libusb_alloc_streams@16
> Use --enable-stdcall-fixup to disable these warnings
> Use --disable-stdcall-fixup to disable these fixups
> /usr/bin/i686-w64-mingw32-ld: warning: resolving _libusb_alloc_transfer by linking to _libusb_alloc_transfer@4
> ```

So the remaining thing is to see if this `[fix-dll-exports.patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libusb/fix-dll-exports.patch)` is needed or not (or maybe needed by clang32/clang64 only), right?

The other changes have been incorporated.
https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-libusb

neheb:
Maybe https://github.com/libusb/libusb/blob/master/configure.ac#L196 should be fixed too. clock_gettime on MinGW is implemented in winpthreads instead of librt. That's the reason for this fix: https://github.com/libusb/libusb/commit/5691f303ad272a2514f68a96d6c18aeff62eb33d

I use a meson based build that checks for it properly.

mcuee:
> > Just adding that MSYS2 patch, I get such warnings on 32-bit build, but not on 64-bit build:
> > ```
> > /usr/bin/i686-w64-mingw32-ld: warning: resolving _libusb_alloc_streams by linking to _libusb_alloc_streams@16
> > Use --enable-stdcall-fixup to disable these warnings
> > Use --disable-stdcall-fixup to disable these fixups
> > /usr/bin/i686-w64-mingw32-ld: warning: resolving _libusb_alloc_transfer by linking to _libusb_alloc_transfer@4
> > ```
> 
> So the remaining thing is to see if `fix-dll-exports.patch` is needed or not (or maybe needed by clang32/clang64 only), right?

I remember that `libusb-1.0.def` is needed for MSVC build dll to be compatible with MinGW GCC build (at least for 32bit). Maybe it is also needed for clang as well to make the DLL compatible with MinGW GCC build. I am not so sure though.

Maybe @pbatard can help a bit here.

tormodvolden:
> Maybe https://github.com/libusb/libusb/blob/master/configure.ac#L196 should be fixed too. clock_gettime on MinGW is implemented in winpthreads instead of librt. That's the reason for this fix: [5691f30](https://github.com/libusb/libusb/commit/5691f303ad272a2514f68a96d6c18aeff62eb33d)

That line 196 is in a Linux-only section though. Can you explain a bit more? Maybe we should open a separate ticket for it.

mcuee:
@tormodvolden 
I think this pull requst is good to go now. Let's worried about the `libusb-1.0.def` issue later.

mcuee (approved):
Reviewed and tested to be good.

neheb:
@tormodvolden reading the description for enable-stdcall-fixup is interesting:

```
If the link finds a symbol that it cannot resolve, it will attempt to do “fuzzy linking” by looking for another defined symbol that differs only in the format of the symbol name (cdecl vs stdcall)
```

Another note: there's no warning for libusb_set_option.

The way I'm understanding this is that the def file is specifying all of these functions as _cdecl (aka WINAPIV) in addition to _stdcall (WINAPI).

neheb:
Yeah. Getting rtid of libusb_set_option from the def file and keeping libusb_set_option@8 results in a linking error.

neheb:
btw does --add-stdcall-alias even do anything? I tried to see if there's a difference under MSYS2/objtool and couldn't find a difference.

mcuee:
> btw does --add-stdcall-alias even do anything? I tried to see if there's a difference under MSYS2/objtool and couldn't find a difference.

Last time (before mingw-w64) it is important to make the old mingw.org 32bit compiler DLL to be compatible with MSVC generated 32bit DLL through the libusb-1.0.def file.
Ref: https://github.com/libusb/libusb/commit/bd623f70d99fad8b975f8d4ee62ea4b1abf1272f

Now we do not really support mingw.org 32bit compiler any more, so I am not so sure if this is still required or not.

BTW, mingw.org compiler project seems to be still alive.
https://osdn.net/projects/mingw/

neheb:
Ah I have no idea about that. MSYS2 uses MinGW-w64.

mcuee:
> Ah I have no idea about that. MSYS2 uses MinGW-w64.

Now the question is whether it matter or not with MSYS2 mingw32 (MinGW-w64 32bit). If it does not matter, then we can drop both the `--add-stdcall-alias` linker option in `configure.ac` and `--kill-at` DLL tool option in `Makefile.am`.

We do not really need to worry about the old `mingw.org` 32bit compiler now.

mcuee:
> So the remaining thing is to see if `fix-dll-exports.patch` is needed or not (or maybe needed by clang32/clang64 only), right?
> 
> The other changes have been incorporated. https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-libusb

@tormodvolden 

From my test results here, I think the patch is needed for clang32/clang64 only. It should not be applied to MSYS2 mingw32 (MSYS2 is not right in this aspect).
* https://github.com/libusb/libusb/issues/944#issuecomment-1245112054

mcuee:
> Now the question is whether it matters or not with MSYS2 mingw32 (MinGW-w64 32bit). If it does not matter, then we can drop both the `--add-stdcall-alias` linker option in `configure.ac` and `--kill-at` DLL tool option in `Makefile.am`.
> 
> We do not really need to worry about the old `mingw.org` 32bit compiler now.

I am thinking that we can drop both the `--add-stdcall-alias` linker option in `configure.ac` and `--kill-at` DLL tool option in `Makefile.am`, then add the MSYS2 clang patch for mingw32/mingw64/clang32/clang64. Need to test this out.

I am thinking we can now drop out

mcuee:
@tormodvolden 

Any update to remove the WIP status and then merge this one?

tormodvolden:
Maybe -mwin32 support also should be probed for in configure. But our testing (in the CI and on my Linux system) shows it not to be needed in mingw. The probing for --add-stdcall-alias cannot do any harm other than adding more cruft in configure.ac. It is a question if we would need this option if we instead specified libusb-1.0.def while building, but this can be looked at later.

Note the MSYS patch does indeed add -export-symbols libusb-1.0.def. This option is not supported by GCC, here the .def should be added to the list of object files instead. But doing this for GCC generates new warnings, that should go away by adding --enable-stdcall-fixup or --disable-stdcall-fixup (sic!) according to the GCC man page. But they don't. And so it goes on.

So yes, I think we can merge this, allowing clang builds, and improve any other things later.

tormodvolden:
Merged with commit c2525a03.

PR #1194, opened 2022-09-05
use regular check for clock_gettime on macOS

neheb:

let's see what happens.


neheb:
eh forget it.

mcuee:
> eh forget it.

It is good to state the reason why you decide to close the pull request. Thanks. 

If the pull request is kind of WIP, you can always mention that.

neheb:
it seems the original code checks for the version of macOS that is targeted.

PR #1197, opened 2022-09-09
def: remove __stdcall libusb_set_option

neheb:
libusb_set_option uses variable arguments and therefore cannot use __stdcall. In practice, this does not get exported. Verified under MinGW's objdump tool with 32 and 64-bit combinations.


mcuee (approved):

tormodvolden:
> In practice, this does not get exported.

What does this mean exactly? That libusb_set_option() is not exported? Does it then have to be in the .def file at all?

neheb:
The @ version is not exported. The regular on is.

PR #1200, opened 2022-09-21
Remove unpaired usbi_mutex_unlock call in libusb_set_interface_alt_setting

### LINKED ISSUE #1199: `libusb_set_interface_alt_setting` unlocks device handle lock without having it locked first if device has been disconnected
Dear maintainers,

TLDR: I believe [this change](https://github.com/libusb/libusb/commit/1a08aa84d96397a3840a75abe66051f5360c2c84#diff-c1f9bc250077d41456a3e580fca0ddf5d8c25b741bff6d9b9505990a8b70b254L1766-L1772) has introduced a regression that causes [`usbi_mutex_destroy(&dev_handle->lock)`](https://github.com/libusb/libusb/blob/master/libusb/core.c#L1440) to fail with EBUSY if `libusb_set_interface_alt_setting` and `libusb_close` have been called after the device has been disconnected.

Context: I am a libuvc user (libuvc linking against the current libusb main) who
1. calls [`uvc_close`](https://github.com/libuvc/libuvc/blob/master/src/device.c#L1721-L1749) once the camera has failed to provide frames for a while, e.g. due to a disconnect.
2. `uvc_close` will attempt to release the interface which will call [`libusb_set_interface_alt_setting`](https://github.com/libuvc/libuvc/blob/master/src/device.c#L1023).
3. If the device had been disconnected, [this code block will be executed](https://github.com/libusb/libusb/blob/master/libusb/core.c#L1765-L1768), unlocking the `dev_handle->lock` without having it locked first.
4. Later, in `libusb_close`/`do_close`, [`usbi_mutex_destroy(&dev_handle->lock)`](https://github.com/libusb/libusb/blob/master/libusb/core.c#L1440) will fail because the lock is in an unexpected state

https://github.com/libusb/libusb/commit/1a08aa84d96397a3840a75abe66051f5360c2c84#diff-c1f9bc250077d41456a3e580fca0ddf5d8c25b741bff6d9b9505990a8b70b254L1766-L1772 moved the `usbi_mutex_lock` call without removing the corresponding `usbi_mutex_unlock(&dev_handle->lock)` call.

papr:
Fixes #1199

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1763,7 +1763,6 @@ int API_EXPORTED libusb_set_interface_alt_setting(libusb_device_handle *dev_hand
 		return LIBUSB_ERROR_INVALID_PARAM;
 
 	if (!usbi_atomic_load(&dev_handle->dev->attached)) {
-		usbi_mutex_unlock(&dev_handle->lock);
 		return LIBUSB_ERROR_NO_DEVICE;
 	}
 


seanm:
This code indeed looks wrong, and I believe your fix is correct.

@tormodvolden please take a look?

tormodvolden:
Out of interest, does this manifest as a warning or error, or was it simply discovered by code review?

papr:
I ran into an issue caused by this bug. :) And found it while debugging it.

tormodvolden:
Alright, I hadn't seen the associated bug report. Good catch, thanks for your contribution!

PR #1203, opened 2022-10-05
git ignore "build" folder

sonatique:


CraigHutchinson (approved):
👍 can't go wrong with this, normally adding the `out/` folder also for MSVC uses this by default for presets I think.

Depending on maintainers though it may be preferred to configure a global ignore rule i.e. `git config --global core.excludesfile ~/.gitignore_global`

Issue #1094, opened 2022-03-20
Misc: To release 1.0.27


tormodvolden:
Postponed: (mostly Windows-only, retargeted for 1.0.28)
- #1297 (new API)
- #1181
- #749
- #1227
- #1067
- #890
- #1018 (libusb0.sys)
- #1127

Important changes merged:
- #1057 (major new feature)
- #1126 
- #1125
- #1139 (non-critical regression)
- #1200 (regression)
- #1026 (new API)
- #1231
- #965
- #1167 (new API)
- #1133 (new API)
- #1270
- #1266 (fix macOS version checks)

Other changes merged:
- #1080
- #1109
- #1116 (and assc 1111-1115)
- #1186 (fixes segfault with libusbK)
- #1183
- #1215
- #1117 (fixes e.g. Sane)
- #1188
- #1230
- #1257
- #1263
- #1284 (regression since 1.0.25)
- #1174
- #1202
- #1308
- #1331
- #1322
- #1432
- #1410
- #1412
- #1413 
- #1418
- #1407
- <s>#1208 (new API)</s>
- <s>#1217</s>

mcuee:
This is a very popular pull request for better Android support. Not so sure if we can just pull it in as long as it does not affect other platforms (especially Linux).
* https://github.com/libusb/libusb/pull/874

mcuee:
I am also not so sure how we deal with new platforms like this.
- https://github.com/libusb/libusb/pull/1057

tormodvolden:
I think #1026 must be settled before merging #874. Both #874 and #1057 need some work (and ideally more user feedback). They can be be merged in any release when they are ready (like any PR), but I wouldn't let 1.0.27 drag on waiting for these to finalize. IMO the only important thing (that I know about now) for 1.0.27 is #1026, and all other items are bonus points.

mcuee:
Is it possible to create a branch for #874 (once it is rebased to git head)? I think #1057 is still far away but #874 seems to be more popular.

tormodvolden:
[...]

PR #1208, opened 2022-10-17
Add support for enabling or disabling the WinUSB RAW_IO option on an endpoint

### LINKED ISSUE #490: WinUSB backend stutters on bulk IN requests and RAW_IO pipe policy

martinling:
This PR adds support for enabling or disabling the WinUSB RAW_IO pipe policy for a given endpoint, which is [documented here](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification).

This is necessary to increase performance. Without this option the WinUSB backend will only queue one inbound operation at a time, even if the libusb async API is used to submit multiple transfers.

For real-time sampling devices with high sample rates and small buffers, the use of RAW_IO combined with queued async transfers is essential to maintaining the necessary throughput and avoiding lost samples or buffer overruns.

Examples of devices affected include Cypress FX2 based logic analyzers accessed using [Sigrok](http://sigrok.org/), and the [HackRF](https://github.com/greatscottgadgets/hackrf/) software defined radio.

The new option must be set by calling `libusb_set_option` with the arguments:

```c
libusb_set_option(ctx, LIBUSB_OPTION_WINUSB_RAW_IO, dev_handle, endpoint_address, enable, max_transfer_size_ptr)
```
where the types of the variadic arguments are:
```c
libusb_device_handle *dev_handle;
unsigned int endpoint_address;
unsigned int enable;
unsigned int *max_transfer_size_ptr;
```

The `dev_handle` and `endpoint_address` parameters must identify a valid IN endpoint on an open device. If `enable` is nonzero, RAW_IO is enabled, otherwise it is disabled. Unless `max_transfer_size_ptr` is NULL, then on a successful call to enable `RAW_IO`, the pointer destination will be written with the `MAXIMUM_TRANSFER_SIZE` value for the endpoint.

Whilst RAW_IO is enabled for an endpoint, all transfers on that endpoint must meet the following two requirements:

* The buffer length must be a multiple of the maximum endpoint packet size.
* The length must be less than or equal to the `MAXIMUM_TRANSFER_SIZE` value.

This option should not be changed when any transfer is in progress on the specified endpoint.

This option only affects the WinUSB backend. On other backends it is ignored and returns `LIBUSB_ERROR_NOT_SUPPORTED`, without modifying the value pointed to by `max_transfer_size_ptr`.

This PR fixes #490, and replaces #1069. A great deal of credit is due to @PetteriAimonen for the previous PR, to @patstew for his alternative solution on which this PR is partly based, and to everyone else who participated in previous discussions.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2233,6 +2233,7 @@ int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
 		/* Handle all backend-specific options here */
 	case LIBUSB_OPTION_USE_USBDK:
 	case LIBUSB_OPTION_NO_DEVICE_DISCOVERY:
+	case LIBUSB_OPTION_WINUSB_RAW_IO:
 		if (usbi_backend.set_option)
 			return usbi_backend.set_option(ctx, option, ap);
 
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -146,7 +146,7 @@ typedef SSIZE_T ssize_t;
  * Internally, LIBUSB_API_VERSION is defined as follows:
  * (libusb major << 24) | (libusb minor << 16) | (16 bit incremental)
  */
-#define LIBUSB_API_VERSION 0x01000109
+#define LIBUSB_API_VERSION 0x0100010A
 
 /* The following is kept for compatibility, but will be deprecated in the future */
 #define LIBUSBX_API_VERSION LIBUSB_API_VERSION
@@ -2131,7 +2131,40 @@ enum libusb_option {
 
 #define LIBUSB_OPTION_WEAK_AUTHORITY LIBUSB_OPTION_NO_DEVICE_DISCOVERY
 
-	LIBUSB_OPTION_MAX = 3
+	/** Enable or disable WinUSB RAW_IO mode on an endpoint
+	 *
+	 * Requires four additional arguments of:
+	 *
+	 *   libusb_device_handle *dev_handle
+	 *   unsigned int endpoint_address
+	 *   unsigned int enable
+	 *   unsigned int *max_transfer_size_ptr
+	 *
+	 * The dev_handle and endpoint_address parameters must identify a valid
+	 * IN endpoint on an open device. If enable is nonzero, RAW_IO is
+	 * enabled, otherwise it is disabled. Unless max_transfer_size_ptr is
+	 * NULL, then on a successful call to enable RAW_IO, it will be written
+	 * with the MAXIMUM_TRANSFER_SIZE value for the endpoint.
+	 *
+	 * Whilst RAW_IO is enabled for an endpoint, all transfers on that endpoint
+	 * must meet the following two requirements:
+	 *
+	 * * The buffer length must be a multiple of the maximum endpoint packet size.
+	 *
+	 * * The length must be less than or equal to the MAXIMUM_TRANSFER_SIZE value.
+	 *
+	 * This option should not be changed when any transfer is in progress on the
+	 * specified endpoint.
+	 *
+	 * This option only affects the WinUSB backend. On other backends it is ignored
+	 * and returns LIBUSB_OPTION_NOT_SUPPORTED, without modifying the value pointed
+	 * to by max_transfer_size_ptr.
+	 *
+	 *  Since version 1.0.27, \ref LIBUSB_API_VERSION >= 0x0100010A
+	 */
+	LIBUSB_OPTION_WINUSB_RAW_IO = 3,
+
+	LIBUSB_OPTION_MAX = 4
 };
 
 int LIBUSB_CALLV libusb_set_option(libusb_context *ctx, enum libusb_option option, ...);
--- a/libusb/os/windows_common.c
+++ b/libusb/os/windows_common.c
@@ -607,8 +607,6 @@ static int windows_set_option(struct libusb_context *ctx, enum libusb_option opt
 {
 	struct windows_context_priv *priv = usbi_get_context_priv(ctx);
 
-	UNUSED(ap);
-
 	if (option == LIBUSB_OPTION_USE_USBDK) {
 		if (!usbdk_available) {
 			usbi_err(ctx, "UsbDk backend not available");
@@ -617,6 +615,10 @@ static int windows_set_option(struct libusb_context *ctx, enum libusb_option opt
 		usbi_dbg(ctx, "switching context %p to use UsbDk backend", ctx);
 		priv->backend = &usbdk_backend;
 		return LIBUSB_SUCCESS;
+	}
+
+	if (priv->backend->set_option) {
+		return priv->backend->set_option(ctx, option, ap);
 	}
 
 	return LIBUSB_ERROR_NOT_SUPPORTED;
--- a/libusb/os/windows_common.h
+++ b/libusb/os/windows_common.h
@@ -334,6 +334,7 @@ struct windows_backend {
 	int (*cancel_transfer)(struct usbi_transfer *itransfer);
 	void (*clear_transfer_priv)(struct usbi_transfer *itransfer);
 	enum libusb_transfer_status (*copy_transfer_data)(struct usbi_transfer *itransfer, DWORD length);
+	int (*set_option)(struct libusb_context *ctx, enum libusb_option option, va_list args);
 };
 
 struct windows_context_priv {
--- a/libusb/os/windows_usbdk.c
+++ b/libusb/os/windows_usbdk.c
@@ -721,4 +721,5 @@ const struct windows_backend usbdk_backend = {
 	NULL,	/* cancel_transfer */
 	usbdk_clear_transfer_priv,
 	usbdk_copy_transfer_data,
+	NULL /* usbdk_set_option */,
 };
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -42,6 +42,8 @@
 		continue;			\
 	}
 
+static int interface_by_endpoint(struct winusb_device_priv *priv,
+	struct winusb_device_handle_priv *handle_priv, uint8_t endpoint_address);
 // WinUSB-like API prototypes
 static bool winusbx_init(struct libusb_context *ctx);
 static void winusbx_exit(void);
@@ -2129,6 +2131,111 @@ static enum libusb_transfer_status winusb_copy_transfer_data(struct usbi_transfe
 	return priv->apib->copy_transfer_data(SUB_API_NOTSET, itransfer, length);
 }
 
+static int winusb_set_option(struct libusb_context *ctx, enum libusb_option option, va_list ap)
+{
+	struct libusb_device_handle *dev_handle;
+	unsigned int endpoint;
+	unsigned int enable;
+	unsigned int *max_transfer_size_ptr;
+	struct winusb_device_handle_priv *handle_priv;
+	struct winusb_device_priv *priv;
+	UCHAR policy;
+	int current_interface;
+	HANDLE winusb_handle;
+	int sub_api = SUB_API_NOTSET;
+	ULONG max_transfer_size = 0;
+
+	switch (option) {
+	case LIBUSB_OPTION_WINUSB_RAW_IO:
+		dev_handle = va_arg(ap, struct libusb_device_handle *);
+		endpoint = va_arg(ap, unsigned int);
+		enable = va_arg(ap, unsigned int);
+		max_transfer_size_ptr = va_arg(ap, unsigned int *);
+
+		policy = enable != 0;
+
+		if (dev_handle == NULL) {
+			usbi_err(ctx, "device handle passed for RAW_IO was NULL");
+			return LIBUSB_ERROR_INVALID_PARAM;
+		}
+
+		if (HANDLE_CTX(dev_handle) != ctx) {
+			usbi_err(ctx, "device handle passed for RAW_IO has wrong context");
+			return LIBUSB_ERROR_INVALID_PARAM;
+		}
+
+		if (endpoint & ~(LIBUSB_ENDPOINT_DIR_MASK | LIBUSB_ENDPOINT_ADDRESS_MASK)) {
+			usbi_err(ctx, "invalid endpoint %X passed for RAW_IO", endpoint);
+			return LIBUSB_ERROR_INVALID_PARAM;
+		}
+
+		if (!(endpoint & LIBUSB_ENDPOINT_DIR_MASK)) {
+			usbi_err(ctx, "endpoint %02X passed for RAW_IO is OUT, not IN", endpoint);
+			return LIBUSB_ERROR_INVALID_PARAM;
+		}
+
+		handle_priv = get_winusb_device_handle_priv(dev_handle);
+		priv = usbi_get_device_priv(dev_handle->dev);
+		current_interface = interface_by_endpoint(priv, handle_priv, (uint8_t) endpoint);
+
+		if (current_interface < 0) {
+			usbi_err(ctx, "unable to match endpoint to an open interface for RAW_IO");
+			return LIBUSB_ERROR_NOT_FOUND;
+		}
+
+		if (priv->usb_interface[current_interface].apib->id != USB_API_WINUSBX) {
+			usbi_err(ctx, "interface is not winusb when setting RAW_IO");
+			return LIBUSB_ERROR_NOT_SUPPORTED;
+		}
+
+		winusb_handle = handle_priv->interface_handle[current_interface].api_handle;
+
+		if (!HANDLE_VALID(winusb_handle)) {
+			usbi_err(HANDLE_CTX(dev_handle), "WinUSB handle not valid when setting RAW_IO");
+			return LIBUSB_ERROR_NOT_FOUND;
+		}
+
+		CHECK_WINUSBX_AVAILABLE(sub_api);
+
+		if (enable && max_transfer_size_ptr != NULL) {
+			ULONG size = sizeof(ULONG);
+			if (!WinUSBX[sub_api].GetPipePolicy(winusb_handle, (UCHAR) endpoint,
+				MAXIMUM_TRANSFER_SIZE, &size, &max_transfer_size)) {
+				usbi_err(ctx, "failed to get MAXIMUM_TRANSFER_SIZE for endpoint %02X", endpoint);
+				switch (GetLastError()) {
+				case ERROR_INVALID_HANDLE:
+					return LIBUSB_ERROR_INVALID_PARAM;
+				default:
+					return LIBUSB_ERROR_OTHER;
+				}
+			}
+		}
+
+		if (!WinUSBX[sub_api].SetPipePolicy(winusb_handle, (UCHAR) endpoint,
+			RAW_IO, sizeof(UCHAR), &policy)) {
+			usbi_err(ctx, "failed to change RAW_IO for endpoint %02X", endpoint);
+			switch (GetLastError()) {
+			case ERROR_INVALID_HANDLE:
+			case ERROR_INVALID_PARAMETER:
+				return LIBUSB_ERROR_INVALID_PARAM;
+			case ERROR_NOT_ENOUGH_MEMORY:
+				return LIBUSB_ERROR_NO_MEM;
+			default:
+				return LIBUSB_ERROR_OTHER;
+			}
+		}
+
+		usbi_dbg(ctx, "%s RAW_IO for endpoint %02X", enable ? "enabled" : "disabled", endpoint);
+
+		if (enable && max_transfer_size_ptr != NULL)
+			*max_transfer_size_ptr = max_transfer_size;
+
+		return LIBUSB_SUCCESS;
+	default:
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+	}
+}
+
 // NB: MSVC6 does not support named initializers.
 const struct windows_backend winusb_backend = {
 	winusb_init,
@@ -2151,6 +2258,7 @@ const struct windows_backend winusb_backend = {
 	winusb_cancel_transfer,
 	winusb_clear_transfer_priv,
 	winusb_copy_transfer_data,
+	winusb_set_option,
 };
 
 /*
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2306,6 +2306,7 @@ static bool winusbx_init(struct libusb_context *ctx)
 		WinUSB_Set(hWinUSB, ResetPipe, true);
 		WinUSB_Set(hWinUSB, SetCurrentAlternateSetting, true);
 		WinUSB_Set(hWinUSB, SetPipePolicy, true);
+		WinUSB_Set(hWinUSB, GetPipePolicy, true);
 		WinUSB_Set(hWinUSB, WritePipe, true);
 
 		// Check for isochronous transfers support (available starting with Windows 8.1)
--- a/libusb/os/windows_winusb.h
+++ b/libusb/os/windows_winusb.h
@@ -531,6 +531,13 @@ typedef BOOL (WINAPI *WinUsb_SetPipePolicy_t)(
 	ULONG ValueLength,
 	PVOID Value
 );
+typedef BOOL (WINAPI *WinUsb_GetPipePolicy_t)(
+	WINUSB_INTERFACE_HANDLE InterfaceHandle,
+	UCHAR PipeID,
+	ULONG PolicyType,
+	PULONG ValueLength,
+	PVOID Value
+);
 typedef BOOL (WINAPI *WinUsb_UnregisterIsochBuffer_t)(
 	WINUSB_ISOCH_BUFFER_HANDLE BufferHandle
 );
@@ -662,6 +669,7 @@ struct winusb_interface {
 	WinUsb_ResetPipe_t ResetPipe;
 	WinUsb_SetCurrentAlternateSetting_t SetCurrentAlternateSetting;
 	WinUsb_SetPipePolicy_t SetPipePolicy;
+	WinUsb_GetPipePolicy_t GetPipePolicy;
 	WinUsb_WritePipe_t WritePipe;
 	union {
 		struct {


PetteriAimonen:
Looks good in quick review. I'll try to find time to test it - I'll have to patch the Python wrapper to be able to set the option.

For anyone else testing, here is a `.dll` build:
https://github.com/devEmbedded/libusb/actions/runs/3274645213

mcuee:
@tormodvolden 
I think this can be merged as they have gone through quite some discussions. I admit that I am not capable of reviewing more ccomplex changes like this though.

mcuee:
@martinling 
Would you please help to provide your email address by sending an email to me? Thanks.
My email address is mentioned in the main libusb github README file (Xiaofan Chen).

mcuee:
@PetteriAimonen 

Just wonder how you patch the Python wrapper? Are you using pyusb or python-libusb1?

Ref:
* https://github.com/pyusb/pyusb/issues/433
* https://github.com/vpelletier/python-libusb1/issues/75

PetteriAimonen:
@mcuee I used `python-libusb1`, but never got around to patching it. By the point when this discussion reached a solution, we had already needed to deliver & I made some workarounds using proprietary libraries for Windows.

tormodvolden:
Merged to master. Thanks a lot for this contribution!

tormodvolden:
(from https://github.com/libusb/libusb/pull/1245#issuecomment-1435640547)
> > Would you even agree to converting WINUSB_RAW_IO into something like libusb_set_raw_io() already before next release? It could be a no-op on non-WinUSB backends or interfaces, and thus allow the same code to work on all platforms. Well the details we'll need to discuss elsewhere, with the people having worked on it.
> 
> Using a new function for this was one of the options discussed in [#1069 (comment)](https://github.com/libusb/libusb/pull/1069#issuecomment-1134465819). I don't recall anyone specifically objecting to it but at the time there didn't seem to be an appetite for adding new functions to the API, so I went along with the proposal to shoehorn it into `libusb_set_option` with a bunch of varargs.
> 
> However, if you'd rather have this be a separate function I think that would be a lot cleaner and I'd be happy to rewrite it.

Yes, I think it is becoming clearer that a separate function is better. There have been so many issues with libusb_set_option() and I think its use should be limited to a minimum. We have now un-deprecated libusb_set_debug() and we have libusb_init_context() that should be preferred for the two other options.

We should be careful with introducing new functions, but the same goes for new options, they add as much support burden and ABI breaks.

I appreciate the finesse getting RAW_IO into this option handling, but I guess everybody agrees it is not pretty. I also think it isn't an "option", it is more an operation that should be done at a given moment, and even returns output at that stage. So smells like a function, flies like a function, and it will probably be a lot easier to understand and be used properly by the users if it is a real function.

So yes, please, rewrite it as a function. The call for name proposals is open :) And what do to on non-Windows platforms?

mcuee:
> So yes, please, rewrite it as a function. The call for name proposals is open :) And what do to on non-Windows platforms?

@tormodvolden 
Similar to how Windows deal with the kernel driver detaching function?
https://github.com/libusb/libusb/blob/master/libusb/os/windows_common.c#L910-L912

As for function name, maybe `libusb_set_winusb_raw_io()`?

Do you want @martinling to create a new PR or continue using this PR?

martinling:
I suggest we have two functions, as follows.

The only reason for making retrieving `MAXIMUM_TRANSFER_SIZE` part of the enable call was to make it possible to shoehorn all this into `libusb_set_option`.

Now that we don't have that constraint, we can make that a separate operation. This also makes things easier for the application, because it can check the value and make sure it's suitably prepared for that *before* enabling RAW_IO.

My suggested API would be:

```c
/* Retrieve the WinUSB MAXIMUM_TRANSFER_SIZE policy for an endpoint on an open device.
 *
 * Returns the MAXIMUM_TRANSFER_SIZE value, or a negative error code.
 * If the backend is not WinUSB, returns LIBUSB_ERROR_NOT_SUPPORTED.
 */
int libusb_get_winusb_max_transfer_size(libusb_device_handle *dev_handle, unsigned int endpoint_address);

/* Set the WinUSB RAW_IO policy for an endpoint on an open device.
 *
 * Sets the MAXIMUM_TRANSFER_SIZE value, and returns `LIBUSB_SUCCESS` or a negative error code.
 * If the backend is not WinUSB, returns LIBUSB_ERROR_NOT_SUPPORTED.
 */
int libusb_set_winusb_raw_io(libusb_device_handle *dev_handle, unsigned int endpoint_address, int enable);
```
Let me know if you're happy with that.

To keep the discussion history understandable for others, let's leave this PR closed, and I'll submit a new one with the new version.

tormodvolden:
> To keep the discussion history understandable for others, let's leave this PR closed, and I'll submit a new one with the new version.

Good, do as you please. I hope everybody who followed this PR will follow the new one also. However I don't see a new one open yet, so I'll keep this one open until then.

tormodvolden:
> My suggested API would be:

Can you please show an example of portable code using the new functions?

I was really hoping we could use functions names not having "winusb" in them. We have functions that are only useful on some platforms, but we don't have driver names or platforms in any function names I believe. In the worst case I would be happy with something with "platform" in its name. I know these functions only makes sense on winusb today, but who knows what we have in a few years time.

There was also libusb_set_endpoint_option() suggested earlier.

And also Linux has (had?) max transfer sizes, but for control transfers and synchronous calls.

I have a lot more questions but I should check all the associated PRs and issues first, it has probably all been discussed.

martinling:
> Can you please show an example of portable code using the new functions?

Exactly how the API would be used will depend what the application's own requirements for transfer sizes are. Assuming a simple case, in which the application just needs to select an appropriate size that will be used for all future reads from the endpoint, the code might look like this.

```c
// Set up a bulk IN endpoint for fastest throughput, and choose a transfer
// size to be used when reading data from it.
//
// Returns a transfer size, or a negative error code.
//
int setup_fast_bulk_in_endpoint(libusb_device_handle *dev, int ep)
{
	int transfer_size = 0xA00000; // default to 10MB transfers.
	int max_transfer_size;
	int max_packet_size;
	int result;

	// Get WinUSB MAX_TRANSFER_SIZE policy for endpoint.
	result = libusb_get_winusb_max_transfer_size(dev, ep);

	if result == LIBUSB_ERROR_NOT_SUPPORTED {
		// Not using WinUSB backend. Use any transfer size.
		return transfer_size;
	} else if result < 0 {
		// Error occured.
		return result;
	} else {
		// Retrieved the value.
		max_transfer_size = result;
	}

	// Get maximum packet size for endpoint.
	result = libusb_get_max_packet_size(dev, ep);
	if result < 0 {
		// Error occured.
		return result;
	} else {
		// Retrieved the value.
		max_packet_size = result;
	}

	// Adjust the transfer size to comply with RAW_IO requirements:
	
	// The size must be a multiple of the endpoint max packet size.
	transfer_size -= transfer_size % max_packet_size;

	// The size must be less than or equal to the max transfer size.
	transfer_size = MIN(transfer_size, max_transfer_size);

	// Having ensured the requirements will be met, enable RAW_IO on the endpoint.
	result = libusb_set_winusb_raw_io(dev, ep, 1);
	if result < 0 {
		// Error occured.
		return result;
	}

	// Return transfer size to be used.
	return transfer_size;
}
```

martinling:
> I was really hoping we could use functions names not having "winusb" in them.

Whilst I appreciate the desire for that, the problem is that the meanings of these functions are completely specific to WinUSB. There is no direct equivalent to `RAW_IO` on any other platform. Even if some future platform introduced a feature with a similar name or effect, it would likely not place exactly the same set of requirements on the application.

Say we call the enable function `libusb_set_platform_raw_io`. Then a future WidgetOS platform adds its own "raw I/O" mode which is similar but not mutually compatible (perhaps it additionally requires the buffer be page aligned, for instance). Now you can't actually reuse the generic `libusb_set_platform_raw_io` function, because the meaning is different.

The same goes for the size function. You could call it `libusb_get_platform_max_transfer_size`, but that name is misleading, because the WinUSB MAXIMUM_TRANSFER_SIZE policy is not actually a limit on the size of transfers on the platform. It only applies to bulk and interrupt endpoints, for one, and even then, only matters if RAW_IO is enabled. If some other platform also had a maximum size, but which applied in different circumstances, then it wouldn't be safe to reuse the same function for both.

As alien as it is to the idea of a cross-platform API, I think naming these functions for what they are - WinUSB-specific details - is actually the best solution, and helps keep it clear to the user that what they're dealing with.

martinling:
> BTW, there is a bug with the current implementation which makes it not working with USB Composite Device

Thanks, I'll try to ensure that's fixed when I write the new version.

How do people feel about the API, aside from the function names? If we can agree on the function signatures at least, then I can go ahead and start implementing it and the names can be adjusted later if needed.

Counter-intuitively, I think it may actually result in cleaner cross-platform code if we add a third WinUSB-specific function:

```c
/* Check whether the backend in use is WinUSB. Returns 1 if so, otherwise 0.
 *
 * When libusb is using the WinUSB backend, an application must take some extra steps
 * if it needs to achieve the highest possible inbound throughput on bulk and interrupt
 * endpoints. See libusb_set_winusb_raw_io and libusb_get_winusb_max_transfer_size.
 */
int libusb_backend_is_libusb(void);
```

This would allow for cleaner usage, as it'd no longer be necessary to special-case `LIBUSB_ERROR_NOT_SUPPORTED`, so generic error handling code can be used throughout (which should be helpful for language bindings as well):

```c
// Set up a bulk IN endpoint for fastest throughput, and choose a transfer
// size to be used when reading data from it.
//
// Returns a transfer size, or a negative error code.
//
int setup_fast_bulk_in_endpoint(libusb_device_handle *dev, int ep)
{
	int transfer_size = 0xA00000; // default to 10MB transfers.

	if (libusb_backend_is_winusb()) {
		// Get MAX_TRANSFER_SIZE policy for endpoint.
		int max_transfer_size = check_errors(libusb_get_winusb_max_transfer_size(dev, ep));

		// Get maximum packet size for endpoint.
		int max_packet_size = check_errors(libusb_get_max_packet_size(dev, ep));

		// Adjust the transfer size to comply with RAW_IO requirements:
	
		// The size must be a multiple of the endpoint max packet size.
		transfer_size -= transfer_size % max_packet_size;

		// The size must be less than or equal to the max transfer size.
		transfer_size = MIN(transfer_size, max_transfer_size);

		// Having ensured the requirements will be met, enable RAW_IO on the endpoint.
		check_errors(libusb_set_winusb_raw_io(dev, ep, 1));

		// Return transfer size to be used.
		return transfer_size;
        } else {
		// Not using WinUSB backend. Use default transfer size.
		return transfer_size;
	}
}

// Generic application/binding specific error handling.
void check_errors(int result) {
	if (result < 0) {
		printf("Error: %s", libusb_strerror(result));
		exit(1);
	} else {
		return result;
	}
}
```

Or as a less-specific alternative to `libusb_backend_is_winusb`, we could add a more general-purpose function that queries libusb for what the backend currently in use is. I'm not sure what that should return, though. A string? An enum? A `struct libusb_backend *` ?

patstew:
If it's
```
libusb_get_winusb_max_transfer_size(dev, ep)
```
then what you said above about it not being a max transfer size for the platform doesn't really apply. You could have generic:
```
int libusb_get_max_transfer_size(dev, ep); // Maximum size of a transfer on this endpoint
int libusb_get_transfer_unit_size(dev, ep); // Transfer size must be a multiple of this
```
functions, which return 0x7FFFFFFF and 1 by default, or whatever's appropriate on that endpoint e.g. MAX_TRANSFER_SIZE and packet size on a WinUSB endpoint with RAW_IO enabled. They wouldn't return LIBUSB_ERROR_NOT_SUPPORTED or any other error, so you wouldn't need to handle the error case or have a platform specific switch.

martinling:
Yes, it could be done as @patstew suggests.

That approach feels like premature generalization to me, though. At present there are no similar oddities on other platforms. If new platforms appear in the future, they may have completely different oddities.

There is also a subtle difference, which is that with the API I proposed, the application can query what the size constraints will be _before_ enabling RAW_IO - and it may decide not to use RAW_IO if it sees that it can't meet the size constraints.

If we instead add generic functions that retrieve whatever the _current_ max transfer size and unit size are for a given endpoint, then the application has to enable RAW_IO first, then query what the new size constraints are, and then turn RAW_IO off again if it can't meet them. That's not necessarily a big problem, and it's one that exists with the current `libusb_set_option` based API from this PR as well, but there may be cases where it's undesirable.

martinling:
Thinking about a more general way of dealing with this:

- The fundamental reason that this RAW_IO feature requires API changes, is that it requires the application to commit to placing some constraints on its transfers in advance.

- Right now, there are two constraints involved: max transfer size, and transfer unit size. In future, there might be similar features available that require different sets of constraints, e.g. page-aligned buffers.

- Can we define an API in terms of **what constraints the application is committing to**, which can be extended to other constraints in the future, without breaking backwards compatibility?

For the current use case, it could look like the following:

```c

/* Check what constraints can be applied to an endpoint to increase performance. */
int libusb_get_available_constraints(libusb_device_handle *dev, int ep);

/* Promise that subsequent transfers on an endpoint will satisfy some constraints. */
int libusb_apply_constraints(libusb_device_handle *dev, int ep, int constraints);

/* Stop promising that subsequent transfers on an endpoint will satisfy constraints. */
int libusb_relax_constraints(libusb_device_handle *dev, int ep, int constraints);

/* Get the maximum size for an endpoint with the MAX_TRANSFER_SIZE constraint. */
int libusb_get_max_transfer_size_constraint(libusb_device_handle *dev, int ep);

/* Get the unit size for an endpoint with the TRANSFER_UNIT_SIZE constraint. */
int libusb_get_transfer_unit_size_constraint(libusb_device_handle *dev, int ep);

/* Bitmasks for all currently supported constraints. */
enum libusb_constraint {

    /* The application will ensure that transfers are less than or equal to the size returned
     * by libusb_get_max_size_constraint(). */
    #define LIBUSB_CONSTRAINT_MAX_TRANSFER_SIZE = (1 << 0)

    /* The application will ensure that transfers are a multiple of the size returned
     * by libusb_get_unit_size_constraint. */
    #define LIBUSB_CONSTRAINT_MAX_UNIT_SIZE = (1 << 1),

    /* Other constraints can be added in future as needed, e.g:
        - Transfer buffer must be page aligned.
        - Transfer buffer must have been allocated for DMA with libusb_dev_mem_alloc. */
}
```

Then the example code could look like this:

```c
// Set up a bulk IN endpoint for fastest throughput, and choose a transfer
// size to be used when reading data from it.
//
// Returns a transfer size, or a negative error code.
//
int setup_fast_bulk_in_endpoint(libusb_device_handle *dev, int ep)
{
	int transfer_size = 0xA00000; // default to 10MB transfers.

	int constraints = check_errors(libusb_get_available_constraints(dev, ep));

        if (constraints & LIBUSB_CONSTRAINT_TRANSFER_UNIT_SIZE) {
                int transfer_unit_size = check_errors(libusb_get_transfer_unit_size_constraint(dev, ep));
                transfer_size -= transfer_size % transfer_unit_size;
        }

        if (constraints & LIBUSB_CONSTRAINT_MAX_TRANSFER_SIZE) {
		int max_transfer_size = check_errors(libusb_get_max_transfer_size_constraint(dev, ep));
		transfer_size = MIN(transfer_size, max_transfer_size);
        }

        // Commit to the constraints we have guaranteed above.
        check_errors(libusb_apply_constraints(dev, ep,
            LIBUSB_CONSTRAINT_MAX_TRANSFER_SIZE |
            LIBUSB_CONSTRAINT_TRANSFER_UNIT_SIZE));

        return transfer_size;
}
```

Note that RAW_IO is not even mentioned in the API at all. The WinUSB RAW_IO policy would be enabled automatically as a side effect of `libusb_apply_constraints`, and may be disabled in future as a side effect of `libusb_relax_constraints`.

If any other platform has a performance optimization that requires the same constraints, or any subset of them, then that optimization can be enabled automatically by any future version of libusb, without the application having to be modified or even recompiled.

If further optimizations are added in the future which require additional constraints, then new constraint flags can be added for those extra constraints, without affecting existing programs. Applications can "opt-in" to the new optimizations by adding the necessary constraint flags.

mcuee:
> Yes, I think it is becoming clearer that a separate function is better. There have been so many issues with libusb_set_option() and I think its use should be limited to a minimum. We have now un-deprecated libusb_set_debug() and we have libusb_init_context() that should be preferred for the two other options.
> 
> We should be careful with introducing new functions, but the same goes for new options, they add as much support burden and ABI breaks.

To me one of the main problems with `libusb_set_option` is that it is a Variadic function. 

I am thinking we can still use the same idea of `libusb_set_option` but make the parameter number fixed. Say we can have two more similar function `libusb_set_option1` with single parameter and `libusb_set_option2` with two parameters. 

We can then deprecate `libusb_set_option` in libusb 1.1 version or 2.0 version.

mcuee:
@hjelmn 
Just wondering if you have any preferences of the function names here. Thanks.

mcuee:
The other option is to leave the change to another day and just fix the anomaly mentioned in #1260, if it takes time to align on the function names.
* #1260

martinling:
The decision can't be put off if this feature is going to be released with 1.0.27. Once the API is shipped in a release it will need to be kept as-is.

I'm happy to develop this further but there needs to be agreement on the API before I can go ahead and write it.

If there's no agreement on API then the merge of this current PR will need to be reverted before release.

PetteriAimonen:
For me, the currently implemented API with `libusb_set_option()` seems fine, and I don't really see the point of delaying this further.

tormodvolden:
There are some good ideas and comments coming up here, and I'd like to hear more, and rather not rush it. Yes, it would mean reverting the added option if we want to release 1.0.27 before this gets sorted out.

martinling:
@tormodvolden So far, the constraint API in [this comment](https://github.com/libusb/libusb/pull/1208#issuecomment-1469137442) is the version I'd be happiest with. It generalises the actual problem, and it's the only option that actually satisfies the desire not to have backend/platform specific details showing up in the cross-platform API (in fact there's no need to mention WinUSB or RAW_IO in the API at all). I'd be interested in your thoughts on that direction.

tormodvolden:
Yes, the constraint API is my favorite so far.

mcuee:
> @tormodvolden So far, the constraint API in [this comment](https://github.com/libusb/libusb/pull/1208#issuecomment-1469137442) is the version I'd be happiest with. It generalises the actual problem, and it's the only option that actually satisfies the desire not to have backend/platform specific details showing up in the cross-platform API (in fact there's no need to mention WinUSB or RAW_IO in the API at all). I'd be interested in your thoughts on that direction.

I will agree that the above is good and pretty generic. 

Somehow I do not quite like `constraints` in the function names. In the end it is an option for a transfer (related to an Endpoint). So it may be good to have `endpoint` or `transfer` in the function name.

patstew:
I think that in practice people who want to turn on RAW_IO know that's what they want to do, the constraints based API makes it unnecessarily hard to know if you've actually achieved that with your constraints calls.
Also, when I added the windows ZLP setting I got pushback because it allowed you to change the ZLP setting, and that could interfere with async transfers. RAW_IO is the same, you should probably call it once when you create the endpoint and before you issue any transfers. So I'm not sure it makes sense to allow people to 'relax' the constraints.

mcuee:
@patstew 

Just FYI, XP support was dropped in libusb-1.0.24.
https://github.com/libusb/libusb/wiki/Windows#supported-environments

mcuee:
@martinling 

Just looking at WinUSB pipe policy, we have many possibilities.
https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification

There were many discussions on WinUSB pipe policy in libusb-devel mailing list. 

RAW_IO has always been mentioned as the main request.

Ref 1: https://marc.info/?l=libusb-devel&m=131651480720220&w=2
Agreed. On Windows, from past user requests, I would very much like to 
have a backend_init call that can set the WinUSB pipe policy parameters 
according to the developer's choice.

Pete

Ref 2: https://marc.info/?l=libusb-devel&m=145654906508594&w=2

Using RAW_IO introduces certain complexities for libusb. In
particular, if the provided transfer buffer is not a multiple of the
endpoint max packet size, we have two options:

1) Round the buffer size down to a max packet size multiple. If the
buffer is smaller than the endpoint max packet size, libusb has to
allocate a buffer.
2) Round the buffer size up to the next max packet size multiple and
allocate a buffer of that size.

Now if we have to allocate and the transfer completes with more data
than the user's buffer will allow, we have to keep the excess data
until the next transfer.

Then there is the case where the user's buffer is greater than
MAXIMUM_TRANSFER_SIZE. Now the Windows backend is getting into the
business of breaking up the transfer into multiple smaller ones.

Regards,
Chris

mcuee:
FYI, libusbK has the API to set/get Pipe Policy.
https://libusbk.sourceforge.net/UsbK3/group__usbk.html#ga10f9e367e4241c431edc2cecd1bb827e

And set/get Power Policy.
https://libusbk.sourceforge.net/UsbK3/group__usbk.html#gadc938a613b1d64f7bba4ddc53591d317

patstew:
Ok then, the documentation would just be 

> With the RAW_IO transfer flag set the transfer size must be a multiple of 1kB and a maximum of 4MB with an ordinary bulk endpoint. (The maximum may be 256kB on ancient full speed controllers, and is different on isochronous endpoints. For more details see https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-bandwidth-allocation#maximum-transfer-size)

I think the issues raised in your ref 2 are (unsolvable) problems with libusb enabling RAW_IO under the covers for better performance. I think we're agreed that it needs to be opt-in for the user, then the user submits transfers of the right size or they get rejected.

As for other pipe policies, SHORT_PACKET_TERMINATE is already supported, AUTO_CLEAR_STALL, PIPE_TRANSFER_TIMEOUT, IGNORE_SHORT_PACKETS, ALLOW_PARTIAL_READS and AUTO_FLUSH sound like they would potentially cause problems or duplicate things that libusb does, which leaves RESET_PIPE_ON_RESUME that might be useful for someone, but I don't think it has ever been requested.

martinling:
> RAW_IO is the same, you should probably call it once when you create the endpoint and before you issue any transfers. So I'm not sure it makes sense to allow people to 'relax' the constraints.

Relaxing the constraints corresponds to turning RAW_IO off again on the endpoint. We want to support doing that, e.g. for the scenario I wrote about in #1069 as follows:

> Consider a device that implements a serial protocol on an endpoint. To drive it you need to make a lot of small, odd-size transfers containing commands and responses. But some commands may result in large blocks of data being sent back, so when reading these, it will make sense to enable RAW_IO and then queue up multiple transfers for speed. After the large data block is finished, you then need to return to sending small, odd-size transfers for the next commands. It's necessary to turn RAW_IO off again at that point, and safe to do so - provided that no transfers are in flight on that endpoint, which is the same requirement as when enabling the option.

martinling:
> Can we not hide these backend constraints in the libusb backend implementation? Maybe by submitting multiple requests and adding buffering depending on the transfer submitted by the user?

No, we can't. See [this comment](https://github.com/libusb/libusb/pull/1069#issuecomment-1134722559) on #1069 for my reasoning.

patstew:
> Relaxing the constraints corresponds to turning RAW_IO off again on the endpoint. We want to support doing that, e.g. for the scenario I wrote about in #1069 as follows:

And what if there are transfers in flight or the RAW_IO call returns an error? You need multiple constraints to enable RAW_IO with your system. Since the API operation doesn't correspond to the underlying operation that can fail, or ought to be rejected in the case of in flight transfers, there's no useful way to return errors to the user, and no way for the user to know if the constraints they've set have actually set RAW_IO successfully. If you start returning errors on constraint calls that happen to trigger the underlying RAW_IO call, then the user ends up in odd states with half their constraints set.

martinling:
@patstew 

I think you raise some valid concerns with constraint-based API approach. However, I think there are also some reasonable answers possible:

> What if there are transfers in flight?

We would have to make it a requirement of the constraint API that you can't apply or relax constraints on an endpoint whilst it has transfers in flight. That seems like a reasonable limitation to have in such an API. It doesn't make much sense to make guarantees about future transfers, when some transfers not subject to those guarantees are already queued.

> or the RAW_IO call returns an error?

In the case of an error while _enabling_ RAW_IO through the constraint API, I think we ought to just log a warning and continue without RAW_IO. Everything would still work correctly in that scenario. You would miss out on some performance, but the libusb API doesn't make any guarantees about performance. The constraint functions would just be a way for the application to make commitments about its future behaviour, which may allow libusb to opportunistically apply performance optimizations, on some backends, if and when it can. If libusb can't apply some optimization, it wouldn't make sense for that to count as a failure or return an error.

An error when _disabling_ RAW_IO is more tricky, because it means that it has to continue to maintain the transfer size constraints, even if it no longer wants to. I have no idea whether disabling RAW_IO can fail in practice, but the WinUSB API allows for it to happen, so we have to treat it as possible. So in that case we would have to signal that condition to the application.

I think the constraint API actually handles that quite cleanly, though: disabling RAW_IO would happen as an effect of a call to `libusb_relax_constraints`, so if that call returns an error as a result, the meaning of that makes sense: the application has failed to relax the constraints, and therefore has to either continue to follow the constraints, or try again to relax them. Of course, there may not be much the application can do in that situation except throw some error to the user and give up. But that's true regardless of what API we use.

> I think that in practice people who want to turn on RAW_IO know that's what they want to do, the constraints based API makes it unnecessarily hard to know if you've actually achieved that with your constraints calls.

I think this is the key point.

If we want an API in which you explicitly request RAW_IO, with an error if you fail to get it, then the constraint-based approach isn't appropriate, because as you say, it provides no way to tell if you succeeded.

But if we want an API that's completely generic, and that doesn't leak any backend-specific details into the libusb API, then it's not _possible_ to specifically request a backend-specific mode, nor to get an error if that mode isn't available - because such operations aren't allowed to exist.

So as I see it, we have some conflicting requirements. Either we accept having some WinUSB-specific details in the libusb API, or we accept that libusb can't tell you whether RAW_IO was enabled or not.

I've provided API suggestions for both options: [here](https://github.com/libusb/libusb/pull/1208#issuecomment-1466580112) plus [here](https://github.com/libusb/libusb/pull/1208#issuecomment-1468077686) for the WinUSB-visible approach, or the [constraint API](https://github.com/libusb/libusb/pull/1208#issuecomment-1469137442) for the WinUSB-invisible approach.

What I think we need a decision on is which approach the libusb maintainers want to accept. Either that, or someone needs to find some compromise between them.

martinling:
Any further thoughts?

I'd like to move ahead with implementing this in a form that's releasable.

mcuee:
@hjelmn and @tormodvolden 

Just wondering if you have any concerns of proceeding with @martinling's proposal? 

It seems to me that we do not have better proposals so it is probably good just to proceed with his idea which seems to be acceptable.

martinling:
This feature is currently in the "much wanted" list for 1.0.27 (#1094), and I'm still available to work on it, but I can't do so without a decision on which approach to take with the API.

PetteriAimonen:
So, there is opposition to the currently implemented libusb_option() approach. Myself I find that approach good enough and lowest risk of breaking things for anyone.

The [constraints approach](https://github.com/libusb/libusb/pull/1208#issuecomment-1469137442) is ok for me also, even though adding new functions will definitely cause some ABI problems for users and bindings. If you link against a new libusb and use the constraints functions, there will be "undefined symbol" errors on older libusb. But changes like that happen sometimes with any library.

mcuee:
Ping @hjelmn and @tormodvolden

Please help here. Thanks.

PetteriAimonen:
To elaborate on the ABI issues: when everything goes correctly, it is not a problem. But for example these situations are potentially more problematic with new constraints functions instead of existing `libusb_set_option()`:

1. Language bindings (e.g. some of Python modules) that use compiled modules. These would need to be matched to the libusb version, or use platform-specific methods such as `dlopen()` to detect if the functions are present.
2. Languages that can link to C functions directly at compile time, but cannot include C headers or use `#ifdef` (e.g. C#, Go). Using the new functions could make such code unbuildable on e.g. Ubuntu 22.04 LTS, which is supported to 2027.
3. Multiple versions of libusb installed on same machine, causing different version headers to be used for compilation vs. runtime.
4. On Windows, classic DLL conflicts if libusb-1.0.dll is installed system-wide.

The problem will be application-wide, so even if the user doesn't really have any device that requires RAW_IO, the whole application could be rendered unusable if its code has calls to new libusb functions. In comparison, having libusb version conflict with currently implemented `libusb_set_option()` would only cause `RAW_IO` not to work, and the application will be notified of that by the return status.

On positive side, the problems would only affect applications that do use the constraints calls. In practice this could lead to many bindings libraries to opt to only start including them by default once the major version is eventually incremented to 1.1.

martinling:
The API/ABI implications of adding new functions are well understood, and there are mechanisms for handling this. There are already many new functions which have been added since the 1.0.0 release. Each one is documented in the header and the API documents with the release and `LIBUSB_API_VERSION` at which it was added.

None of the issues you're talking about are specific to this feature, or even to libusb.

tormodvolden:
@hjelmn Have you had any more thoughts on this? I would be interested in the opinion of an active looong-time libusb developer.

More thoughts from me: The maintenance burden factor: A plain libusb_winusb_rawio function (although oh-so-ugly API wise) would be straight-forward to implement and to maintain. OTOH a beautiful abstraction can be a pain to maintain in the future, even if we have brilliant people at hand to implement it today. I don't want to bash libusb_set_option too much, it is a nice extensible interface for everything but I have the impression it has needed quite some fix-ups and continued work and I am not even sure it is all done yet. Granted that had much to do with the "default-context options" or "default context-options" which we are moving away from.

martinling:
On further reflection, I am much less keen on the constraint API idea if there is no other use case for it. So far nobody here has come up with one, and I haven't either.

Whilst potentially new use cases might appear in the future, there are not a lot of new operating systems being written these days, and USB is old news. It does not seem likely that libusb is going to gain new quirky USB APIs to deal with.

I'm therefore inclined to withdraw that suggestion: it was an interesting exercise in generalising the problem, but really the only reason for doing so was to avoid the ugliness of having explicitly WinUSB-specific functions in the libusb API.

I think perhaps a better way to address that issue is to just give the functions more generic names. This is what has already been done for another OS-specific detail in the case of `libusb_dev_mem_alloc`, which is actually completely specific to the usbfs backend, just not named as such.

So I have a new API proposal, with three functions as follows:

```c
/* Check whether the backend in use supports raw I/O. Returns 1 if so, otherwise 0.
 *
 * When raw I/O is available, an application must take some extra steps if it needs to
 * achieve the highest possible inbound throughput on bulk and interrupt endpoints.
 * See libusb_set_raw_io.
 */
int libusb_backend_supports_raw_io(void);

/* Retrieve the maximum transfer size supported for raw I/O for an inbound bulk or
*  interrupt endpoint on an open device.
 *
 * Returns a maximum transfer size, or a negative error code.
 * If the backend does not support raw I/O, returns LIBUSB_ERROR_NOT_SUPPORTED.
 */
int libusb_get_max_raw_io_transfer_size(libusb_device_handle *dev_handle, unsigned int endpoint_address);

/* Enable or disable raw I/O for an inbound bulk or interrupt endpoint on an open device.
 *
 * When raw I/O is in use on an endpoint, all transfers on the endpoint must have a size that is
 * a multiple of the maximum packet size returned by `libusb_get_max_packet_size`, and less
 * than or equal to the maximum transfer size returned by `libusb_get_max_raw_io_transfer_size`.
 *
 * Returns LIBUSB_SUCCESS or a negative error code.
 * If the backend does not support raw I/O, returns LIBUSB_ERROR_NOT_SUPPORTED.
 */
int libusb_set_raw_io(libusb_device_handle *dev_handle, unsigned int endpoint_address, int enable);
```

This is functionally exactly the same API as I  suggested [here](https://github.com/libusb/libusb/pull/1208#issuecomment-1466580112) and [here](https://github.com/libusb/libusb/pull/1208#issuecomment-1468077686), except that neither the function names nor the documentation explicitly mention WinUSB. Instead, they are named and documented in terms of the raw I/O feature itself.

In practice this may be a distinction without a difference, but it does at least keep the API focused on the relevant matter (do we have to jump through these raw I/O hoops on this system), rather than the internal detail of which backend libusb is using.

And it does at least generalise _slightly_: if there's ever a new backend that shares this quirk (most likely in some future version of Windows), then it can just reuse the same API, rather than this feature being tied explicitly to a current named backend.

Usage example updated to the new proposal:
```c
// Set up a bulk IN endpoint for fastest throughput, and choose a transfer
// size to be used when reading data from it.
//
// Returns a transfer size, or a negative error code.
//
int setup_fast_bulk_in_endpoint(libusb_device_handle *dev, int ep)
{
	int transfer_size = 0xA00000; // default to 10MB transfers.

	if (libusb_backend_supports_raw_io()) {
		// Get maximum raw I/O transfer size for endpoint.
		int max_transfer_size = check_errors(libusb_get_max_raw_io_transfer_size(dev, ep));

		// Get maximum packet size for endpoint.
		int max_packet_size = check_errors(libusb_get_max_packet_size(dev, ep));

		// Adjust the transfer size to comply with raw I/O requirements:
	
		// The size must be a multiple of the endpoint max packet size.
		transfer_size -= transfer_size % max_packet_size;

		// The size must be less than or equal to the max transfer size.
		transfer_size = MIN(transfer_size, max_transfer_size);

		// Having ensured the requirements will be met, enable raw I/O on the endpoint.
		check_errors(libusb_set_raw_io(dev, ep, 1));

		// Return transfer size to be used.
		return transfer_size;
        } else {
		// Backend does not support raw I/O.
		return transfer_size;
	}
}

// Generic application/binding specific error handling.
void check_errors(int result) {
	if (result < 0) {
		printf("Error: %s", libusb_strerror(result));
		exit(1);
	} else {
		return result;
	}
}
```

Would this be a reasonable compromise?

martinling:
I have fleshed out the new proposal in a draft PR #1297. This is only compile-tested so far, but allows seeing the full effect of the proposed change on the codebase. The implementation is adapted from the option-based version.

martinling:
Closing in favour of #1512.

PR #1217, opened 2022-11-15
windows: Fix off-by-one error in HID backend for interrupt and bulk transfers without reportId.

kleinerm:
When the windows hid backend is used for interrupt or bulk transfers on USB-HID compliant devices which don't support HID reportId's, the current code properly adds a zero prefix byte for reportId 0x00 before an OUT transfer, and properly strips the leading zero prefix byte from buffers received by IN transfers. Length of transmitted data is increased by +1 accordingly to account for the extra prefix byte.

However, on transfer completion the length of sent or received data is not adjusted back again by one. This leads to misreporting of effective transfer length for both OUT and IN transfers, reporting one byte too much. It also causes a memcpy for IN transfers which copies one byte of data too much into the client provided target buffer transfer_priv->hid_dest, appending an extra zero byte at the end and thereby writing one byte past the size of the client target buffer. This could lead to memory corruption or a access violation if the client application is unlucky.

This commit fixes the problem by detecting this type of HID transfers and adjusting the effective length of the transfer down one byte.

Fix successfully tested with a Griffin PowerMate USB-HID device that doesn't support reportId's, performing interrupt transfers from/to the device.

Signed-off-by: Mario Kleiner <mario.kleiner.de@gmail.com>

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -4228,12 +4228,15 @@ static enum libusb_transfer_status hid_copy_transfer_data(int sub_api, struct us
 				}
 
 				if (transfer_priv->hid_buffer[0] == 0) {
+					length--;
 					memcpy(transfer_priv->hid_dest, transfer_priv->hid_buffer + 1, length);
 				} else {
 					memcpy(transfer_priv->hid_dest, transfer_priv->hid_buffer, length);
 				}
 			}
 			transfer_priv->hid_dest = NULL;
+		} else if ((length > 0) && (transfer_priv->hid_buffer[0] == 0)) {
+			length--;
 		}
 		// For write, we just need to free the hid buffer
 		safe_free(transfer_priv->hid_buffer);


mcuee:
@kleinerm 

Have you also tested with HIDAPI? We have issues in HIDAPI as well previously. 

We actually recommend HIDAPI over libusb for HID device. There are quite a few limitations of the libusb HID backend under Windows.

kleinerm:
> @kleinerm
> 
> Have you also tested with HIDAPI? We have similar issues in HIDAPI as well previously.

Only lightly. When reading the libusb HID backend code I realized I can also use HID report via HIDAPI, but my own HIDAPI backend is not actually returning transfer length info. But if i remember correctly HIDAPI's code, I think it handles that correctly - or makes it the applications task to deal with it.

> 
> We actually recommend HIDAPI over libusb for HID device. There are quite a few limitations of the libusb HID backend under Windows.

I know. The Griffin PowerMate which I used for testing is a USB HID device, usable with HIDAPI. I only used the PowerMate as a convenient test device I had laying around for basic testing of my apps new libusb-1 based implementation of USB interrupt/control transfers. Worked beautifully on Linux and macOS, showed this inconsistency and off-by-one bug on MS-Windows, which turned out to be a problem of libusb HID backend. I also verified with a manually assigned WinUSB driver and that worked fine.

Btw. I did find another oddity though when trying a libusbh bulk/interrupt out transfer (or equivalent HID SetReport via HIDAPI) when trying to send data to the PowerMate. Regardless of output buffer / out report size, only 2-Bytes (One zero reportId prefix byte and the first actual payload  byte) were written to the device. Indeed the Win32 api function WriteFile() in both libusb and HIDAPI reported no error, but only a 2 Byte write, regardless of requested length of the write. I'm not sure why, or if this is only limited to some oddity of my PowerMate, but these out transfers worked perfectly fine on Linux and macOS. The USB endoint descriptor of the device reported a wMaxPacket size of 1 Byte, but I didn't have a tool around to decode HID report descriptors, so I don't know if this is some bug, or some quirk the PowerMate's report descriptors. I didn't investigate any further, because it doesn't really matter for my use cases at all.

tormodvolden:
Thanks for fixing this!

mcuee:
As mentioned in #1285, this PR causes regression and I will suggest we revert this PR.

mcuee:
@kleinerm 

I will suggest that you test your device with hidapitester/hidapi first. The PR is causing regression as mentioned in #1285.

kleinerm:
Ok, not a problem. I do have workarounds for this in my app.

tormodvolden:
If we revert this, what would be a correct fix for the buffer overflow reported here?

mcuee:
> If we revert this, what would be a correct fix for the buffer overflow reported here?

Basically I believe there is no buffer overflow isues. That  is why I have asked the OP to test with HIDAPI.

From the discussion here, it is more about a documentation here. The application has to use buffer length = HID Report Length + 1 to devices without Report ID (even though it is not strictly for Input Report, but it is necessary for Output Report and Feature Report). So the original code is correct. I have done quite extensive test under Linux and Windows (as well as some tests under macOS) to come to the conclusion.
https://github.com/libusb/hidapi/issues/589

This PR is clearly causing regressions so it has to be reverted, as it is wrong for Output Report and Feature Report.

mcuee:
> Ok, not a problem. I do have workarounds for this in my app.

Thanks. What do you mean by workarounds in your application? 

Please post the output of hidapitester and your libusb code so that we can see that if there is a real problem or not with libusb-1.0.26 release. 

As mentioned above, application has to use buffer size = HID Report Length + 1 to devices without Report ID (even though it is not strictly for Input Report, but it is necessary for Output Report and Feature Report), across different OS.

mcuee:
@tormodvolden 

Basically there are different school of thoughts for libusb (and hidapi), if we want to be consistent among all platforms in terms of number of bytes written/read, for devices without Numbered Report ID (using Fake Report ID 0), as per the following discussion.
* https://github.com/libusb/hidapi/issues/589

1) Only report the bytes transferred on the USB Bus -- then report ID 0 needs to be removed.
Using Windows ReadFile and WriteFile -- report ID 0 will not be in the buffer, no need to minus 1.
Using Windows IOCTL -- report ID 0 will be in the buffer, need to minus 1 to report the correct number of size

2) Only report the bytes transferred using the buffer size -- current implementation in libusb 1.0.26 release and HIDAPI. It is also used for HIDAPI Linux and macOS
Using Windows ReadFile and WriteFile -- report ID 0 will not be in the buffer (same for Linux)
Using Windows IOCTL -- report ID 0 will be in the buffer (same for Linux)

kleinerm:
> > Ok, not a problem. I do have workarounds for this in my app.
> 
> Thanks. What do you mean by workarounds in your application?
> 
> Please post the output of hidapitester and your libusb code so that we can see that if there is a real problem or not with libusb-1.0.26 release.
> 
> As mentioned above, application has to use buffer size = HID Report Length + 1 to devices without Report ID (even though it is not strictly needed for Input Report, but it is necessary for Output Report and Feature Report), across different OS.

My current workaround is in this commit:

https://github.com/kleinerm/Psychtoolbox-3/commit/396ae58654a6ff690f03bac7eeab6812ca221cb6

In short: For bulk- and interrupt IN transfers, it allocates a receive buffer of +1 byte longer, and does not report/return such an extra byte of received data if any, essentially stripping away that one extra postfix byte that would happen for reportId 0.

It doesn't do any special handling for control transfers or any OUT transfers. This worked for me. I do have a remark in my testing notes though, that control OUT transfers did not work for my one test device on MS-Windows, even though no error was reported by libusb, whereas the same transfers worked on Linux (and macOS iirc).

That said, my software uses the HIDAPI library for all USB-HID transfers, which are the majority of uses. libusb-1 is only used directly for non-HID devices, and afaik only for about three niche use cases, some control transfers on non-Windows systems, and some bulk transfer. The workaround was needed to make bulk transfers work on Windows for that one use case, and that was not tested by myself but by a user of my software with his devices.

So the best I can say is that my workaround works for this n=1 case of bulk transfers on Windows, and I don't have any good ability to test further without access to the needed hardware. Sorry that I can't do further testing atm.

So in the end, this a bug/omission in the documentation if i understand correctly, and some update to libusb docs will tell us how to deal with possible extra bytes for reportId = 0 cases?

mcuee:
@kleinerm 

Firstly there is no bulk transfer in HID. The following code has a bad name, it should be `hid_submit_bulk_transfer`.
```
static int hid_submit_bulk_transfer(int sub_api, struct usbi_transfer *itransfer);
```

> So in the end, this a bug/omission in the documentation if i understand correctly, and some update to libusb docs will tell us how to deal with possible extra bytes for reportId = 0 cases?

For hidapi, the issue is classified as documentation.

For libusb, it depends on the direction to go. Option 1 is more in line with libusb project's USB part but it requires changes to the codes. Option 2 is easier as it does not require code change. If Option 2 is decided to be the way to go, then I will update the Wiki to mention it as a limitation of the HID backend.

1) Only report the bytes transferred on the USB Bus -- then report ID 0 needs to be removed.
Using Windows ReadFile and WriteFile -- report ID 0 will not be in the buffer, no need to minus 1.
Using Windows IOCTL -- report ID 0 will be in the buffer, need to minus 1 to report the correct number of size

2) Only report the bytes transferred using the buffer size -- current implementation in libusb 1.0.26 release and HIDAPI. It is also used for HIDAPI Linux and macOS
Using Windows ReadFile and WriteFile -- report ID 0 will not be in the buffer (same for Linux)
Using Windows IOCTL -- report ID 0 will be in the buffer (same for Linux)

And in the end, why do you need to use libusb for the HID device since you are already using HIDAPI? What is the limitation of HIDAPI are you facing? Take note libusb project recommends HIDAPI as the library to use for HID device and not libusb.

mcuee:
BTW, the number of bytes to report can be confusing. Even the Linux hidraw documentation has been wrong from the very beginning.
* https://github.com/libusb/hidapi/issues/589#issuecomment-1602290106

kleinerm:
> @kleinerm
> 
> Firstly there is no bulk transfer in HID. The following code has a bad name, it should be `hid_submit_bulk_transfer`.
> 
> ```
> static int hid_submit_bulk_transfer(int sub_api, struct usbi_transfer *itransfer);
> ```
> 
> > So in the end, this a bug/omission in the documentation if i understand correctly, and some update to libusb docs will tell us how to deal with possible extra bytes for reportId = 0 cases?
> 
> For hidapi project, the issue is classified as documentation issue.
> 
> For libusb, it depends on the direction to go. Option 1 is more in line with libusb project's USB part but it requires changes to the codes. Option 2 is easier as it does not require code change. If Option 2 is decided to be the way to go, then I will update the Wiki to mention it as a limitation of the HID backend.
> 
>     1. Only report the bytes transferred on the USB Bus -- then report ID 0 needs to be removed.
>        Using Windows ReadFile and WriteFile -- report ID 0 will not be in the buffer, no need to minus 1.
>        Using Windows IOCTL -- report ID 0 will be in the buffer, need to minus 1 to report the correct number of size
> 
>     2. Only report the bytes transferred using the buffer size -- current implementation in libusb 1.0.26 release and HIDAPI. It is also used for HIDAPI Linux and macOS
>        Using Windows ReadFile and WriteFile -- report ID 0 will not be in the buffer (same for Linux)
>        Using Windows IOCTL -- report ID 0 will be in the buffer (same for Linux)
> 
> 
> And in the end, why do you need to use libusb for the HID device since you are already using HIDAPI? What is the limitation of HIDAPI are you facing? Take note libusb project recommends HIDAPI as the library to use for HID device and not libusb.

Oh, I don't to use it for HID transfers. For HID transfers to HID devices I use HIDAPI. I only need libusb for non-HID transfers (control, bulk, interrupt) to non-HID devices. The issue I encountered happened only on Windows when testing libusb transfers to a device that happened to be a HID device, but I'd normally not use libusb for controlling that device, it was just that that was the only device I had around for testing libusb at all.

The background is that I added support for interrupt and bulk transfers to my software by use of libusb, for the needs of a user (see here: https://psychtoolbox.discourse.group/t/request-help-in-implementing-bulk-transfer-in-psychhid/4657). But only that user has the relevant hardware. I don't have suitable hw for testing, so I used a HID device instead, to make sure my software behaves the same on Linux, macOS and Windows, and there I found the discrepancy on Windows.

Maybe the whole problem I thought I found only occurred because I was testing bulk or interrupt transfers on a HID device instead of a non-HID device? libusb on Windows has many backends, and in my case with my device the hid backend was chosen for handling my requests. So maybe this is a complete non-issue for non-HID devices, essentially a ghost hunt?

In any case, I can't remember or reconstruct the exact details of my tests from over six months ago, so the best is to simply disregard this whole issue here. It does work properly for the end user on their actual non-HID USB device.

Thanks.

mcuee:
> Oh, I don't to use it for HID transfers. For HID transfers to HID devices I use HIDAPI. I only need libusb for non-HID transfers (control, bulk, interrupt) to non-HID devices. The issue I encountered happened only on Windows when testing libusb transfers to a device that happened to be a HID device, but I'd normally not use libusb for controlling that device, it was just that that was the only device I had around for testing libusb at all.
> 
> The background is that I added support for interrupt and bulk transfers to my software by use of libusb, for the needs of a user (see here: https://psychtoolbox.discourse.group/t/request-help-in-implementing-bulk-transfer-in-psychhid/4657). But only that user has the relevant hardware. I don't have suitable hw for testing, so I used a HID device instead, to make sure my software behaves the same on Linux, macOS and Windows, and there I found the discrepancy on Windows.
> 
> Maybe the whole problem I thought I found only occurred because I was testing bulk or interrupt transfers on a HID device instead of a non-HID device? libusb on Windows has many backends, and in my case with my device the hid backend was chosen for handling my requests. So maybe this is a complete non-issue for non-HID devices, essentially a ghost hunt?
> 
> In any case, I can't remember or reconstruct the exact details of my tests from over six months ago, so the best is to simply disregard this whole issue here. It does work properly for the end user on their actual non-HID USB device.
> 

I see. Then the original reported is really a non-issue.

Windows HID backend -- only for USB HID device using Windows native HID driver, which does not support non-HID USB device. 

Windows WinUSB backend without libusbK.dll -- only support USB device with WinUSB driver, it does not support HID device. 

Windows WinUSB backend with libusbK.dll -- support USB device with WinUSB driver, libusbK driver and libusb0.sys driver, it does not support HID device. 

Windows usbdk backend -- only for USB device with usbdk driver

PR #1230, opened 2023-01-17
Adjust NetBSD's udi_bus/udi_addr to libusb's 1-based notion

efuss:
NetBSD's `udi_bus`/`udi_addr` values in `struct usb_device_info` start from 0. OTOH, libusb's `bus_number`/`device_address` in `struct libusb_device` use 0 to represent a missing value. Adjust between the two worlds by adding an offset of 1.

While there, add a comment that, because NetBSD's `#define`s for USB speeds happen to match the corresponding libusb `enum` values, it's OK to assign `struct usb_device_info`'s `udi_speed` field to a `struct libusb_devices`'s `speed` field.

### DIFF
--- a/libusb/os/netbsd_usb.c
+++ b/libusb/os/netbsd_usb.c
@@ -146,8 +146,8 @@ netbsd_get_device_list(struct libusb_context * ctx,
 			if (dev == NULL)
 				return (LIBUSB_ERROR_NO_MEM);
 
-			dev->bus_number = di.udi_bus;
-			dev->device_address = di.udi_addr;
+			dev->bus_number = 1 + di.udi_bus;
+			dev->device_address = 1 + di.udi_addr;
 			dev->speed = di.udi_speed; /* NetBSD #define's happen to match libusb enum */
 
 			dpriv = usbi_get_device_priv(dev);


tormodvolden:
Are the udi_bus/udi_addr numbers displayed in OS messages and device trees (if applicable)? Would this change be confusing for users while debugging?

efuss:
> Are the udi_bus/udi_addr numbers displayed in OS messages
The corresponding autoconfig (boot) message looks like
	ugen0 at uhub4 port 1
(which seems to correspond to udi_bus=0 and udi_addr=2)

> and device trees (if applicable)?
I don't have a FTD-aware NetBSD port.

> Would this change be confusing for users while debugging?
May depend on how people debug.
But the libusb API states that a value of 0 returned by libusb_get_bus_number() or libusb_get_device_address() means "not available" so there's software (nut, in my case) that (as it seems, correctly) refuses to match on zero values.

tormodvolden:
Thanks for the 

Meanwhile I pushed the new comment, seems useful in any case.

> But the libusb API states that a value of 0 returned by libusb_get_bus_number() or libusb_get_device_address() means "not available" so there's software (nut, in my case) that (as it seems, correctly) refuses to match on zero values.

That sounds like a good

tormodvolden:
> and device trees (if applicable)?
> I don't have a FTD-aware NetBSD port.

I meant such trees in a wider sense. On Linux you would have for instance /dev/bus/usb/xxx/yyy for bus xxx and device yyy.

PR #1234, opened 2023-01-24
Adding github action for Visual Studio

mcuee:
This PR adds github action for Visual Studio.

Together with the existing MSYS2 github action, this wil address #1045.
* #1045

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -1,0 +1,20 @@
+name: Visual Studio build
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  build:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        platform: [Win32, x64, ARM, ARM64]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: microsoft/setup-msbuild@v1.1
+    - name: MSBuild ${{ matrix.platform }}
+      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=${{ matrix.platform }}


tormodvolden:
This is probably fine, but do we need both this and appveyor? I am already uncomfortable that a tiny comment typo push causes such a waste of resources :) Appveyor is creating build artifacts archives, that would be great to improve to generate full binary packages. Is this possible with GitHub actions?

mcuee:
> Appveyor is creating build artifacts archives, that would be great to improve to generate full binary packages. Is this possible with GitHub actions?

Yes github actions can keep the artifcates.

Examples:
https://github.com/avrdudes/avrdude/blob/main/.github/workflows/build.yml (for all buids, binary and cmake build files)
https://github.com/libusb/hidapi/blob/master/.github/workflows/builds.yml (only for Windows msbuild)

mcuee:
> Maybe we could configure appveyor to build only on tags, and let this github action (it does indeed look a lot more simple and elegant) do the per-commit build.

I think this is a good idea.

We can also improve github action build scripts later to keep the artifacts.

mcuee:
> Maybe we could configure appveyor to build only on tags, and let this github action (it does indeed look a lot more simple and elegant) do the per-commit build.

Ah, Appveyor has Cygwin build now -- so we may need to add cygwin github action first before going with this idea.

.github/workflows/msvc.yml
```diff
@@ -0,0 +1,18 @@
...
+    - uses: actions/checkout@v3
+    - uses: microsoft/setup-msbuild@v1.1
+    - name: MSBuild x86
+      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=Win32
+    - name: MSBuild x64
+      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=x64
+    - name: MSBuild ARM
+      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=ARM
+    - name: MSBuild ARM64
+      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=ARM64
```
Youw, line 18: this one is a perfect place to use a Matrix:
```suggestion
    strategy:
      matrix:
        platform: [Win32, x64, ARM, ARM64]
    - name: MSBuild ${{ matrix.platform }}
      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=${{ matrix.platform }}
```
maybe, even, multiplied by the configuration (later?)

mcuee:
@Youw 

Your suggestion is good. Just wondering if you can help to suggest how to add the feature to keep the binary for MSVC build. Thanks.

mcuee:
@Youw 
Hmm, strange, it says there is a syntax error which I can not figure out.

Youw:
> syntax error

Fixed. Coding inside of the `suggestion` box is tricky.

Youw:
> how to add the feature to keep the binary for MSVC build

Lets start with the general theory a bit: we want to store some binaries (probably right here on Github) which are produced by the build.
To store a binary, we need a storage, and when in comes to storage, we should ask what kind of storage do we want:
- lifetime: temporary/permanent;
- scope: generic for a repo? build? bound for some other entity?

NOTE: Github doesn't provide a generic repo/organisation-level storage (but there used to be - obsolete).

The only permanent storage there is on Github - an artifacts attached to a _release_. I don't think that is what we need right now.

The only available alternative - build artifacts bound to a specific build: https://github.com/marketplace/actions/upload-a-build-artifact
Example used by HIDAPI: https://github.com/libusb/hidapi/blob/b516e970480a8d05b1c392d106ca110a9d06d668/.github/workflows/builds.yml#L313
Available here: https://github.com/libusb/hidapi/actions/runs/4891807115

But that storage is always temporary (see [Retention Period](https://github.com/marketplace/actions/upload-a-build-artifact#retention-period)).

.github/workflows/msvc.yml
```diff
@@ -0,0 +1,14 @@
+name: Visual Studio build
+on: [push, pull_request]
```
Youw, line 2: NOTE: this way there will be two instances of the build started when a PR created: one for PR and the another one for a push to branch.
See [here](https://github.com/orgs/community/discussions/26276). Right now it is not triggered only because this rule is not in master yet. That discussion provides a good solution to that problem (the one marked with "answer").

mcuee:
I think for the purpose of this PR, we just need to keep the binary files for a specific build. I understand it is a temporary storage.

.github/workflows/msvc.yml
mcuee: This is good. Please help to follow the answer and push the fix. Thanks,

Youw:
```suggestion
on:
  push:
    branches:
    - master
  pull_request:
    branches:
    - master
```

Youw: when other branches are needed - those lists should be extended

mcuee:
@tormodvolden 

I think we can merge this PR first. Later we can discuss whether we want to keep the binaries or not.

tormodvolden:
And if it cannot store artifacts, does it have any advantage over the appveyor builds?

Youw:
> if it cannot store artifacts

Oh it can (see [here](https://github.com/libusb/libusb/pull/1234#issuecomment-1539817416)), it is just a matter of getting it done.

> any advantage over the appveyor builds?

At least one thing - all in one place. The build, the logs, etc. - all at github and no additional integration is needed.

tormodvolden:
> all in one place

Could seem like a pragmatic advantage. On a general note though, if we all stop supporting the alternatives we may well be finding ourselves without alternatives one day.

Youw:
> a pragmatic advantage

Just a convenience I'd say.

---
Regarding the alternatives note. This doesn't seem like a replacement for Appveyor builds - only additional alternative.

BTW: @mcuee what is the original reason to have this? As a potential way to store automatic artifacts, or is there other reason?

mcuee:
> BTW: @mcuee what is the original reason to have this? As a potential way to store automatic artifacts, or is there other reason?

The idea is to have github action for Windows. To me it is faster than Appveyor.

Then the next step is to keep the artifacts, that is not the original intention of the PR, but rather the next step.

mcuee:
> Can you please squash all the fixup commits?

@Youw 
Can you help here? I am not good at git and I am afraid that I may screw up the PR.

tormodvolden:
To make sure you are not screwing up anything while rebasing, just check out a branch copy to work on:
(note what the original branch is called, e.g. my_master)
git checkout -b new_b
Now rebase it on top of master:
git rebase -i origin/master
In the editor you can now replace "pick" with "s" (for squash) on the second and further commits. And clean up the commit message.
Afterwards you can compare this fixed-up branch with the orginal:
git diff my_master
Or, since many other files were changed since then:
git diff my_master -- .github/
If everything is fine, you can force-push the new branch directly, or maybe easier without having to know the remote name etc, update your original branch and repush it:
git checkout my_master
git reset --hard new_b
git push -f

mcuee:
@tormodvolden 

Thanks for the great help. I have tried your steps and it seems to work. Hope I am doing it correctly.

The first step -- I am not so sure what to do so I actually created the new_b branch from the github Web interface using the github_action as the base.

```
git clone https://github.com/mcuee/libusb.git -b new_b
cd .\libusb\
git rebase -i origin/master (replace "pick" with "s" (for squash) on the second and further commits. And clean up the commit message.)
git diff github_action
 git checkout github_action
git reset --hard new_b
git push -f
```

mcuee:
@tormodvolden 

Just wondering if you are okay with this PR now. Thanks.

tormodvolden:
What is the situation about the artifacts, and the duplication of AppVeyor?

mcuee:
> What is the situation about the artifacts, and the duplication of AppVeyor?

Hmm, I think we can keep AppVeyor and the artifacts generated by AppVeyor.

github action is faster and we can have future PRs to improve this and then totally replace AppVeyor (eg: support more versions of MSVC, keeping the artifacts). But I do not know how to do that myself.

.github/workflows/msvc.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        platform: [Win32, x64, ARM, ARM64]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: microsoft/setup-msbuild@v1.1
+    - name: MSBuild ${{ matrix.platform }}
+      run: msbuild msvc\libusb.sln /p:Configuration=Release /p:Platform=${{ matrix.platform }}
```
neheb, line 20: should use msbuild-architecture instead.

.github/workflows/msvc.yml
```diff
@@ -0,0 +1,20 @@
...
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  build:
+    runs-on: windows-latest
+    strategy:
+      matrix:
```
neheb, line 14: AppVeyor tests various versions of MSVC. You should use vs-version.

mcuee:
I think I will close this PR first. 

@neheb 
You seem to be familiar with this topic. PR is welcome.

PR #1245, opened 2023-02-10
libusb_set_options: adjust semantics of libusb_set_option for log level

hjelmn:
This commit updates the actual semantics of:

libusb_set_option(_, LIBUSB_OPTION_LOG_LEVEL, _)

to match the documentation. The documentation states that if LIBUSB_DEBUG is set then the level set there always overrides the value to set with libusb_set_option. Added a set of unit tests to cover some of libusb_set_option to prevent regressions.


hjelmn:
@tormodvolden @mcuee This fixes an inconsistency with the documentation that In found. Plan to push this shortly if there are no objections.

hjelmn:
Ok, that should do it. One interesting note about this PR. Our default options are sticky and can not be unset by the user API. I added an exported function that is not in `libusb.h` to clear them from tests. The method is `usbi_clear_default_option()`.

tormodvolden:
Is this the first function we export that is not in the user API?

hjelmn:
@tormodvolden Correct. I don't know if there is a good way to not export it yet have it available to our tests.

hjelmn:
An alternative is to define a user API to clear options.

mcuee:
> An alternative is to define a user API to clear options.

It seems to me this is a better option to go.

tormodvolden:
What are the use cases for clearing options?

hjelmn:
The only use case I could come up was for unit testing since default options for one test could change the behavior of another. Ideally options like `LIBUSB_OPTION_NO_DEVICE_DISCOVERY` should be unsettable but guaranteeing the behavior to users might be difficult.

tormodvolden:
> The only use case I could come up was for unit testing since default options for one test could change the behavior of another. Ideally options like `LIBUSB_OPTION_NO_DEVICE_DISCOVERY` should be unsettable but guaranteeing the behavior to users might be difficult.

Yes, unit testing isn't a "use case" the way I meant it :)

BTW, currently the API libusb_set_options() is called by libusb_init_context(). I think we have to refactor this and forbid (or at least warn strongly about) LIBUSB_OPTION_NO_DEVICE_DISCOVERY and LIBUSB_OPTION_USE_USBDK in libusb_set_options().

tormodvolden:
\<rant\>In my ideal world we wouldn't have libusb_set_option() at all, it is too complex with its argument handling, the default <context options> which are not <default context> options etc. Only libusb_set_debug would be needed for dynamic adjustment, and e.g. WINUSB_RAW_IO could be a new function.\</rant\>

hjelmn:
@tormodvolden Agreed. `libusb_init_context()` is the much better option overall since  options should be a per-context thing with no changeable defaults. Ideally we should eliminate it in 1.1.0 and leave `libusb_set_debug()`.

tormodvolden:
Would you even agree to converting WINUSB_RAW_IO into something like libusb_set_raw_io() already before next release? It could be a no-op on non-WinUSB backends or interfaces, and thus allow the same code to work on all platforms. Well the details we'll need to discuss elsewhere, with the people having worked on it.

tests/set_option.c
```diff
@@ -0,0 +1,194 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+#include <inttypes.h>
+#include "libusbi.h"
+#include "libusb_testlib.h"
+
+#if defined(_WIN32) || defined(__CYGWIN__)
+/* Windows does not offer unsetenv. To simplify the test code just declare
+ * an empty implementation. */
+static void unsetenv(const char *env) {
```
tormodvolden, line 33: This fails on Cygwin: ../../tests/set_option.c:33:13: error: conflicting types for ‘unsetenv’
 static void unsetenv(const char *env) {
             ^~~~~~~~
In file included from /usr/include/stdlib.h:26:0,
                 from ../../tests/set_option.c:24:
/usr/include/cygwin/stdlib.h:27:5: note: previous declaration of ‘unsetenv’ was here
 int unsetenv (const char *);
     ^~~~~~~~

tormodvolden: And gives a warning both on MinGW and Cygwin:
../../tests/set_option.c:33:34: warning: unused parameter ‘env’ [-Wunused-parameter]
 static void unsetenv(const char *env) {
                                  ^~~

mcuee: Cygwin has `unsetenv`. So the code needs to be changed.

mcuee:
```
MINGW64 /c/work/libusb/libusb_pr1245
$ ./autogen.sh
...

$ make
...
...
make[2]: Entering directory '/c/work/libusb/libusb_pr1245/tests'
  CC       stress.o
  CC       testlib.o
  CCLD     stress.exe
  CC       set_option.o
set_option.c: In function 'unsetenv':
set_option.c:33:34: warning: unused parameter 'env' [-Wunused-parameter]
   33 | static void unsetenv(const char *env) {
      |                      ~~~~~~~~~~~~^~~
  CCLD     set_option.exe
make[2]: Leaving directory '/c/work/libusb/libusb_pr1245/tests'
make[2]: Entering directory '/c/work/libusb/libusb_pr1245'
make[2]: Leaving directory '/c/work/libusb/libusb_pr1245'
make[1]: Leaving directory '/c/work/libusb/libusb_pr1245'

$ ./tests/set_option.exe
Starting test run: test_set_log_level_basic...
Success (0)
Starting test run: test_set_log_level_default...
Success (0)
Starting test run: test_set_log_level_env...
Skip (3)
Starting test run: test_no_discovery...
Skip (3)
---
Ran 4 tests
Passed 2 tests
Failed 0 tests
Error in 0 tests
Skipped 2 tests
```

tests/set_option.c
hjelmn: Gotta love it. Let me work around this.

tormodvolden:
Would be good if this could be split up in several commits, one for the bug fix and another for tests and possibly new exported functions.

hjelmn:
Split it into 3 commits and changed it to `usbi_clear_default_options()` since I just want it to clear everything.

hjelmn:
Ok, looks like appveyor is happy with the build now. Windows is always the odd one out that requires extra fiddling.

tormodvolden:
Thanks for splitting it up, it is much easier to review now. You may wanna leave out version_nano in pull requests.

tests/set_option.c
```diff
@@ -0,0 +1,200 @@
...
+#include <stdio.h>
+#include <inttypes.h>
+#include "libusbi.h"
+#include "libusb_testlib.h"
+
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <winbase.h>
+/* Windows does not offer unsetenv. To simplify the test code just declare
+ * an empty implementation. */
+static void unsetenv(const char *env) {
```
tormodvolden, line 34: FTR, the "real" setenv returns int.

tests/set_option.c
```diff
@@ -0,0 +1,200 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+#include <inttypes.h>
+#include "libusbi.h"
+#include "libusb_testlib.h"
+
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <winbase.h>
+/* Windows does not offer unsetenv. To simplify the test code just declare
+ * an empty implementation. */
```
tormodvolden, line 33: Is it still "empty"? It looks like implemented, although I don't know that Windows function in detail.

tormodvolden:
An alternative rule could be that libusb_set_option() and libusb_set_debug() would be allowed to set higher debug levels than LIBUSB_DEBUG. It could be useful in some cases. OTOH it is very clear-cut that LIBUSB_DEBUG overrules everything.

tests/set_option.c
mcuee: So it is better to change to return int as well. Right?

tests/set_option.c
mcuee: Indeed not empty any more.
https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setenvironmentvariable
```
SetEnvironmentVariable function (winbase.h)
Sets the contents of the specified environment variable for the current process.
```

mcuee:
Latest test results under Windows, using MSYS2 mingw64.
```
$ ./tests/set_option.exe
Starting test run: test_set_log_level_basic...
Success (0)
Starting test run: test_set_log_level_default...
Success (0)
Starting test run: test_set_log_level_env...
Expected test_ctx->debug (0) == 4 (4) at set_option.c:147
Failure (1)
Starting test run: test_no_discovery...
Skip (3)
---
Ran 4 tests
Passed 2 tests
Failed 1 tests
Error in 0 tests
Skipped 1 tests

```

martinling:
> Would you even agree to converting WINUSB_RAW_IO into something like libusb_set_raw_io() already before next release? It could be a no-op on non-WinUSB backends or interfaces, and thus allow the same code to work on all platforms. Well the details we'll need to discuss elsewhere, with the people having worked on it.

Using a new function for this was one of the options discussed in https://github.com/libusb/libusb/pull/1069#issuecomment-1134465819. I don't recall anyone specifically objecting to it but at the time there didn't seem to be an appetite for adding new functions to the API, so I went along with the proposal to shoehorn it into `libusb_set_option` with a bunch of varargs.

However, if you'd rather have this be a separate function I think that would be a lot cleaner and I'd be happy to rewrite it.

mcuee:
@hjelmn 
Any updates on this PR to fix the test failure for Windows? I think the problem is only with the `set_option.c` and specific to Windows.

@tormodvolden 
Just wondering if you can help fix `set_option.c` test code for Windows?

tormodvolden:
The counterpart to getenv() is _putenv() on Windows. Can you try that? It seems getenv/_putenv live in a parallel universe to GetEnvironmentVariable/SetEnvironmentVariable: "getenv operates only on the data structures accessible to the run-time library and not on the environment "segment" created for the process by the operating system. "

mcuee:
> The counterpart to getenv() is _putenv() on Windows. Can you try that? It seems getenv/_putenv live in a parallel universe to GetEnvironmentVariable/SetEnvironmentVariable: "getenv operates only on the data structures accessible to the run-time library and not on the environment "segment" created for the process by the operating system. "

I am not so sure how to do that but I find an implementation from PostgreSQL. Looks like not that straightfoward.
https://doxygen.postgresql.org/win32__port_8h_source.html

```
 /* in port/win32env.c */
 extern int  pgwin32_putenv(const char *);
 extern int  pgwin32_setenv(const char *name, const char *value, int overwrite);
 extern int  pgwin32_unsetenv(const char *name);
  
 #define putenv(x) pgwin32_putenv(x)
 #define setenv(x,y,z) pgwin32_setenv(x,y,z)
 #define unsetenv(x) pgwin32_unsetenv(x)
```

https://doxygen.postgresql.org/win32env_8c_source.html
```
 /*-------------------------------------------------------------------------
  *
  * win32env.c
  *    putenv(), setenv(), and unsetenv() for win32.
  *
  * These functions update both the process environment and caches in
  * (potentially multiple) C run-time library (CRT) versions.
  *
  * Portions Copyright (c) 1996-2023, PostgreSQL Global Development Group
  * Portions Copyright (c) 1994, Regents of the University of California
  *
  *
  * IDENTIFICATION
  *    src/port/win32env.c
  *
  *-------------------------------------------------------------------------
  */
  
 #include "c.h"
  
  
 /*
  * Note that unlike POSIX putenv(), this doesn't use the passed-in string
  * as permanent storage.
  */
 int
 pgwin32_putenv(const char *envval)
 {
     char       *envcpy;
     char       *cp;
     typedef int (_cdecl * PUTENVPROC) (const char *);
     static const char *const modulenames[] = {
         "msvcrt",               /* Visual Studio 6.0 / MinGW */
         "msvcrtd",
         "msvcr70",              /* Visual Studio 2002 */
         "msvcr70d",
         "msvcr71",              /* Visual Studio 2003 */
         "msvcr71d",
         "msvcr80",              /* Visual Studio 2005 */
         "msvcr80d",
         "msvcr90",              /* Visual Studio 2008 */
         "msvcr90d",
         "msvcr100",             /* Visual Studio 2010 */
         "msvcr100d",
         "msvcr110",             /* Visual Studio 2012 */
         "msvcr110d",
         "msvcr120",             /* Visual Studio 2013 */
         "msvcr120d",
         "ucrtbase",             /* Visual Studio 2015 and later */
         "ucrtbased",
         NULL
     };
     int         i;
  
     /*
      * Update process environment, making this change visible to child
      * processes and to CRTs initializing in the future.  Do this before the
      * _putenv() loop, for the benefit of any CRT that initializes during this
      * pgwin32_putenv() execution, after the loop checks that CRT.
      *
      * Need a copy of the string so we can modify it.
      */
     envcpy = strdup(envval);
     if (!envcpy)
         return -1;
     cp = strchr(envcpy, '=');
     if (cp == NULL)
     {
         free(envcpy);
         return -1;
     }
     *cp = '\0';
     cp++;
     if (*cp)
     {
         /*
          * Only call SetEnvironmentVariable() when we are adding a variable,
          * not when removing it. Calling it on both crashes on at least
          * certain versions of MinGW.
          */
         if (!SetEnvironmentVariable(envcpy, cp))
         {
             free(envcpy);
             return -1;
         }
     }
     free(envcpy);
  
     /*
      * Each CRT has its own _putenv() symbol and copy of the environment.
      * Update the environment in each CRT module currently loaded, so every
      * third-party library sees this change regardless of the CRT it links
      * against.  Addresses within these modules may become invalid the moment
      * we call FreeLibrary(), so don't cache them.
      */
     for (i = 0; modulenames[i]; i++)
     {
         HMODULE     hmodule = NULL;
         BOOL        res = GetModuleHandleEx(0, modulenames[i], &hmodule);
  
         if (res != 0 && hmodule != NULL)
         {
             PUTENVPROC  putenvFunc;
  
             putenvFunc = (PUTENVPROC) (pg_funcptr_t) GetProcAddress(hmodule, "_putenv");
             if (putenvFunc)
                 putenvFunc(envval);
             FreeLibrary(hmodule);
         }
     }
  
     /*
      * Finally, update our "own" cache.  This is redundant with the loop
      * above, except when PostgreSQL itself links to a CRT not listed above.
      * Ideally, the loop does visit all possible CRTs, making this redundant.
      */
     return _putenv(envval);
 }
  
 int
 pgwin32_setenv(const char *name, const char *value, int overwrite)
 {
     int         res;
     char       *envstr;
  
     /* Error conditions, per POSIX */
     if (name == NULL || name[0] == '\0' || strchr(name, '=') != NULL ||
         value == NULL)
     {
         errno = EINVAL;
         return -1;
     }
  
     /* No work if variable exists and we're not to replace it */
     if (overwrite == 0 && getenv(name) != NULL)
         return 0;
  
     envstr = (char *) malloc(strlen(name) + strlen(value) + 2);
     if (!envstr)                /* not much we can do if no memory */
         return -1;
  
     sprintf(envstr, "%s=%s", name, value);
  
     res = pgwin32_putenv(envstr);
     free(envstr);
     return res;
 }
  
 int
 pgwin32_unsetenv(const char *name)
 {
     int         res;
     char       *envbuf;
  
     envbuf = (char *) malloc(strlen(name) + 2);
     if (!envbuf)
         return -1;
  
     sprintf(envbuf, "%s=", name);
     res = pgwin32_putenv(envbuf);
     free(envbuf);
     return res;
 }
```

tormodvolden:
Yes, this seems to be complicated on Windows: https://learn.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=msvc-170#example-pass-environment-variables-across-dll-boundary

> But it seems to me `GetEnvironmentVariable/SetEnvironmentVariable` can still be used with some modifications for the purpose of `set_option.c`, probably no need to use the more complex PostgreSQL implementation.

Good, I hope so.

mcuee:
@tormodvolden and @hjelmn 

After some trial and error, the following seems to work. Please check. Thanks.

```
PS C:\work\libusb\libusb_pr1245\tests> git diff
diff --git a/tests/set_option.c b/tests/set_option.c
index 876b68a2..d09f6f76 100644
--- a/tests/set_option.c
+++ b/tests/set_option.c
@@ -28,17 +28,38 @@
 #include "libusb_testlib.h"

 #if defined(_WIN32) && !defined(__CYGWIN__)
+/* Windows does not offer unsetenv and setenv. The following are the replacement functions. */
 #include <winbase.h>
-/* Windows does not offer unsetenv. To simplify the test code just declare
- * an empty implementation. */
-static void unsetenv(const char *env) {
-  SetEnvironmentVariable(env, NULL);
+int unsetenv(const char *env);
+int setenv(const char* name, const char* value, int overwrite);
+
+int unsetenv(const char* name) {
+    int err = _putenv_s(name, "");
+    if (err != 0) {
+        return -1; // Failed to unset variable
+    }
+    return 0; // Success
 }

-static void setenv(const char *env, const char *value, int overwrite) {
-  (void)overwrite;
-  SetEnvironmentVariable(env, value);
+int setenv(const char* name, const char* value, int overwrite) {
+    // Check if the variable already exists
+    char* old_value = getenv(name);
+    if (old_value != NULL) {
+        // Variable already exists, check if we should overwrite it
+        if (!overwrite) {
+            return 0; // Variable exists and overwrite flag is not set
+        }
+    }
+
+    // Set the variable
+    int err = _putenv_s(name, value);
+    if (err != 0) {
+        return -1; // Failed to set variable
+    }
+
+    return 0; // Success
 }
+
 #endif

 #define LIBUSB_TEST_CLEAN_EXIT(code) \

MINGW64 /c/work/libusb/libusb_pr1245
$ ./tests/set_option.exe
Starting test run: test_set_log_level_basic...
Success (0)
Starting test run: test_set_log_level_default...
Success (0)
Starting test run: test_set_log_level_env...
Success (0)
Starting test run: test_no_discovery...
Skip (3)
---
Ran 4 tests
Passed 3 tests
Failed 0 tests
Error in 0 tests
Skipped 1 tests

```

mcuee:
Reference from Microsoft.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-170
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-170

`getenv_s` is supposed to be the more secure version of `getenv` but then the usage is not the same and more complicated. I do not know how to use that.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-s-wgetenv-s?view=msvc-170

mcuee:
Reference from MinGW.org.
https://osdn.net/projects/mingw/scm/git/mingw-org-wsl/blobs/5.5-trunk/mingwrt/mingwex/setenv.c

tormodvolden:
WIN32 putenv requires a "var=value" string, so _putenv_s seems more straightforward. It could be that putenv is older and has higher availability, but _putenv_s is available here on old MinGW 7.0.0 so I am happy with it.

I added an implementation of this to the PR, so that Appveyor will check on old MSVC.

tormodvolden:
I won't hold back the first commit, which fixes a real issue, so I'll merge that one the next time I am in merging mode.

When it comes to the tests, I wonder if we can do without an exported usbi_clear_default_options() if we rearrange the tests a bit and live with marginally lower test coverage. What if we simply run test_set_log_level_default() last?

Actually the tests all succeed if I omit the usbi_clear_default_options() call...

mcuee:
> WIN32 putenv requires a "var=value" string, so _putenv_s seems more straightforward. It could be that putenv is older and has higher availability, but _putenv_s is available here on old MinGW 7.0.0 so I am happy with it.
> 
> I added an implementation of this to the PR, so that Appveyor will check on old MSVC. (EDIT: looks good)

It works fine under MSYS2 MinGW
```
$ ./tests/set_option.exe
Starting test run: test_set_log_level_basic...
Success (0)
Starting test run: test_set_log_level_default...
Success (0)
Starting test run: test_set_log_level_env...
Success (0)
Starting test run: test_no_discovery...
Skip (3)
---
Ran 4 tests
Passed 3 tests
Failed 0 tests
Error in 0 tests
Skipped 1 tests
```

mcuee:
> I added an implementation of this to the PR, so that Appveyor will check on old MSVC. (EDIT: looks good)

@tormodvolden 

Hmm, I do not see the added Visual Studion project file, I am not so sure how you can check the old MSVC version.

tormodvolden:
Good point :) Of course, it didn't even try to compile it. A project file would need to be added.

mcuee:
Now it is okay for MSVC build, tested with VS2022 under Windows 11.

```
PS C:\work\libusb\libusb_pr1245\msvc> msbuild -m -v:m -p:PlatformToolset=v143,Platform=x64,Configuration=Release .\libusb.sln
...
PS C:\work\libusb\libusb_pr1245> .\build\v143\x64\Release\set_option.exe
Starting test run: test_set_log_level_basic...
Success (0)
Starting test run: test_set_log_level_default...
Success (0)
Starting test run: test_set_log_level_env...
Success (0)
Starting test run: test_no_discovery...
Skip (3)
---
Ran 4 tests
Passed 3 tests
Failed 0 tests
Error in 0 tests
Skipped 1 tests

```

tormodvolden:
I merged it without usbi_clear_default_option(). If we see this makes false positives or hides bugs, we should come back to that discussion.

PR #1246, opened 2023-02-11
Build stress_mt on Windows with pthreads

tormodvolden:
There is a bit of autoconf crafting to make sure we only add pthreads for stress_mt, and not for library building. Would be good if some autoconf aficinados can review this.

Maybe it would have been easier to add native Windows thread support to stress_mt :-/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,6 +179,15 @@ if test "x$platform" = xposix; then
 	fi
 elif test "x$platform" = xwindows; then
 	AC_DEFINE([PLATFORM_WINDOWS], [1], [Define to 1 if compiling for a Windows platform.])
+	dnl windows platforms may have pthread support, use for portable tests only
+	saved_LIBS="${LIBS}"
+	AC_SEARCH_LIBS([pthread_create], [pthread],
+		[win_has_pthread=yes], [], [])
+	WIN_PTHREAD_LIBS="${LIBS}"
+	LIBS="${saved_LIBS}"
+	if test "x$win_has_pthread" = xyes; then
+		AC_DEFINE([HAVE_WIN_PTHREAD], [1], [Windows platform has pthread support])
+	fi
 else
 	AC_MSG_ERROR([Unknown platform])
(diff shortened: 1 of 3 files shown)


tormodvolden:
When it comes to variables and flags naming, it should be noted that Windows might have pthread available without linking -lpthread. On our Appveyor CI build for MinGW:
```
checking for library containing pthread_create... none required
checking for pthread_create... yes
```
When I cross-build with MinGW (7.0.0) on Ubuntu 20.04:
```
checking for library containing pthread_create... -lpthread
checking for pthread_create... no
```

mcuee:
My test under Windows 11 MSYS2 mingw64 is good, The output is the same as the github action MSYS2 mingw64 build.
```
checking for library containing pthread_create... none required
checking for struct timespec... yes
checking for sys/time.h... yes
checking for pthread_create... yes
checking if linker supports --add-stdcall-alias... yes
```

tormodvolden:
I tried cross-building with newer MinGW (10.0.0) but it was the same as before.

mcuee:
So this PR seems to be good to go.

mcuee (approved):

tormodvolden:
> Maybe it would have been easier to add native Windows thread support to stress_mt :-/

Done in PR #1252.

mcuee:
#1252 is better as it uses native Windows threading and works with MSVC along with MinGW.

So we can probable merge #1252 and drop this one.

Youw:
Agree.
Trying to have/use `ptread`s on Windows is like trying to drive a car on alcohol - sometime it works, but it is not a solution in today's realities.

tormodvolden:
I would think pthreads works fine on Windows, but with much poorer performance, well, like everything Posix on Windows. For sure, the other PR obsoletes this one. It could have some value to stress-test the Posix threading as well, maybe other libusb bugs will be exposed, but it is not worth the extra complexity. Nice autoconf exercise though.

PR #1247, opened 2023-02-20
threads_posix: Use proper function for thread ID on OpenBSD

brad0:


tormodvolden:
Thanks, looks good. BTW, we should probably use gettid() on Linux as well.

tormodvolden:
I just added the Linux change here for simplicity.

mcuee:
@tormodvolden 

Interestingly nginx code still metions the following.
https://github.com/nginx/unit/blob/master/src/nxt_thread_id.h#L13-L16
```
/*
 * Linux thread id is a pid of thread created by clone(2),
 * glibc does not provide a wrapper for gettid().
 */
```

As per the following page, that has been added only in glibc 2.30.
https://man7.org/linux/man-pages/man2/gettid.2.html

```
The gettid() system call first appeared on Linux in kernel
       2.4.11.  Library support was added in glibc 2.30.  (Earlier glibc
       versions did not provide a wrapper for this system call,
       necessitating the use of [syscall(2)](https://man7.org/linux/man-pages/man2/syscall.2.html).)
```

Then I checked the date of glibc 2.30 release, it was released on 2019-08-01.
https://sourceware.org/glibc/wiki/Glibc%20Timeline

In the end I think it is better not to proceed with the change on Linux side as older Linux distro will not work.

Youw:
This make me wounder why would the original change be needed for OpenBSD?
The author doesn't explain reasoning.

mcuee:
> This make me wounder why would the original change be needed for OpenBSD? The author doesn't explain reasoning.

I think the original PR for OpenBSD is correct as per the OpenBSD manual page and it is available since OpenBSD 3.9 (a very old version released on 1 May 2006)
https://man.openbsd.org/getthrid.2

```
#include <unistd.h>

pid_t getthrid(void);
```

tormodvolden:
@mcuee You are absolutely right, glibc 2.30 is relatively recent. Thanks for catching this. We'll await that Linux change for some years.

brad0:
> This make me wounder why would the original change be needed for OpenBSD? The author doesn't explain reasoning.

1. We use the function like this everywhere else.
2. -current no longer allows syscalls like this.

Youw (approved):
> -current no longer allows syscalls like this.

This is hard to argue.
For historical reasons: [block system calls that are not made via the C library (libc) system-call wrappers](https://lwn.net/Articles/806776/).

Issue #1249, opened 2023-02-23
Handling broken descriptors


robert-billing:
I have run into a device which has a broken implementation of USB. In the device descriptor bNumConfigurations is returned by the device as 2, but only one configuration descriptor exists. This causes "Entity not found" to be reported, and the device is inaccessible.

This patch makes missing configuration descriptors report as the first descriptor, and bypasses the problem. I now have the device working.
[conf.txt](https://github.com/libusb/libusb/files/10814269/conf.txt)

mcuee:
Please create a Pull Request for this issue. Thanks.

tormodvolden:
What device is this?

mcuee:
@robert-billing 

Please help to answer the question from @tormodvolden. Thanks. It will be good to know the problematic device to see if the patch is acceptable or not.

mcuee:
Reference:
* #1250

mcuee:
I tend to think we can close this issue as the device is really broken.


@sonatique
What is your opinion?

PR #1250, opened 2023-02-23
Fix broken config

robert-billing:
Some devices have a broken descriptor. The device config specifies more configurations than exist. This patch supplies the missing config descriptors by copying config [0]. This prevents an "Entity not found" error.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -820,6 +820,17 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
 		priv->config_descriptor[i] = cd_data;
 		cd_buf_actual = NULL;
 	}
+
+	/* Some USB devices have broken descriptors in which the
+	   number of descriptors does not match
+	   bNumConfigurations. This can result in "Entity not found"
+	   errors. If this happens simply replicate the first
+	   configuration. */
+	for ( i = 0; i < num_configurations; i++ )
+	  {
+	    if ( priv -> config_descriptor [ i ] == NULL )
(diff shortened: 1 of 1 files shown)


libusb/os/windows_winusb.c
```diff
@@ -820,6 +820,17 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
...
+
+	/* Some USB devices have broken descriptors in which the
+	   number of descriptors does not match
+	   bNumConfigurations. This can result in "Entity not found"
+	   errors. If this happens simply replicate the first
+	   configuration. */
+	for ( i = 0; i < num_configurations; i++ )
+	  {
+	    if ( priv -> config_descriptor [ i ] == NULL )
+	      priv -> config_descriptor [ i ] = priv -> config_descriptor [ 0 ] ;
```
Youw, line 832: we probably want at least log an `dbg` entry that something is going on with the descriptor(s)

sonatique:
Hello @robert-billing , thanks. 2 questions:

1) Why is this fix Windows-specific when the problem seem to be caused by issues on the device side? Seems like a general fix / workaround would have some value. Do you think that there is a way to move this code to a more general place?

2) I think that this device related issue might be better handled by the client code, because in the way you propose, clients will have a dupplicated entry without being notified that this is not exactly what the device provided, hence preventing any feedback to user or high-level action. What are the reasons for silently workarounding this inside libusb?

Thanks in advance for your reply.

mcuee:
I will close this PR but keep #1249 open.

#1249 is a generic issue worth looking at if we know about this device. 

* #1249

PR #1257, opened 2023-03-02
windows: fix handle usbd_status translate

HotIce0:
``` device descriptor
Device Descriptor:
bcdUSB:             0x0200
bDeviceClass:         0x00
bDeviceSubClass:      0x00
bDeviceProtocol:      0x00
bMaxPacketSize0:      0x40 (64)
idVendor:           0x058F (Alcor Micro, Inc.)
idProduct:          0x6387
bcdDevice:          0x0100
iManufacturer:        0x01
0x0409: "Generic"
iProduct:             0x02
0x0409: "Mass Storage"
iSerialNumber:        0x03
0x0409: "5B546574"
bNumConfigurations:   0x01
```
when i use libusb to proxy usb transfer of usb disk device. (libusb + spice + qemu)
it work fine in linux. but not in windows with usbdk backend.

i use wrieshark to capture usb packet in guest os ( win10).
1. in linux, data flow as below
```
1_bulk_out: cbw scsi opcode=0x23
2_bulk_in: data (transfered length=0. usbd_status_success)
3_bulk_in: data (transfered length=0, usbd_status_stall_pid)
> and then clear_stall, device work fine.
```

the usbd_status_stall_pid status is created by libusb/os/linux_usbfs.c: handle_bulk_completion:
```c
case -EPIPE:
usbi_dbg(TRANSFER_CTX(transfer), "detected endpoint stall");
if (tpriv->reap_status == LIBUSB_TRANSFER_COMPLETED)
	tpriv->reap_status = LIBUSB_TRANSFER_STALL;
goto cancel_remaining;
```
2. in windows, i got usbd_status_stall_pid(0xC0000004) in 3_bulk in. libusb translate it to LIBUSB_IO_ERROR.
i try to recover this device with libusb_reset_device. but device was no response after reset.


libusb/os/windows_common.h
```diff
@@ -131,8 +131,9 @@ typedef LONG USBD_STATUS;
 
 #define USBD_SUCCESS(Status)		((USBD_STATUS)(Status) >= 0)
 
+#define USBD_STATUS_STALL_PID		((USBD_STATUS)0xC0000004L)
 #define USBD_STATUS_ENDPOINT_HALTED	((USBD_STATUS)0xC0000030L)
-#define USBD_STATUS_TIMEOUT		((USBD_STATUS)0xC0006000L)
+#define USBD_STATUS_TIMEOUT			((USBD_STATUS)0xC0006000L)
```
mcuee, line 136: This change is not necessary.

libusb/os/windows_common.c
```diff
@@ -279,6 +279,7 @@ enum libusb_transfer_status usbd_status_to_libusb_transfer_status(USBD_STATUS st
 	case USBD_STATUS_CANCELED:
 		return LIBUSB_TRANSFER_CANCELLED;
 	case USBD_STATUS_ENDPOINT_HALTED:
+	case USBD_STATUS_STALL_PID:
```
mcuee, line 282: Just wondering what is the reason to add this one? Any potential side-effect?

mcuee:
Does the above PR really sort out your issue?

HotIce0:
> @HotIce0 Does the above PR really sort out your issue?
> 
> Have you asked the Spice/Qemu side too?

yes, the issue is solved.
i have not asked the spice/qemu side.

according to MSDN: https://learn.microsoft.com/en-us/previous-versions/windows/hardware/usb/ff539136(v=vs.85)?redirectedfrom=MSDN
```
USBD_STATUS_STALL_PID  0xC0000004 The device returned a stall packet identifier (defined for backward compatibility with the USB 1.0)
```

ref: Universal Serial Bus Specification 2.0 (usb_20.pdf)   5.8.5 Bulk Transfer Data  Sequences

If a halt condition is detected on a bulk pipe
due to transmission errors or **a STALL handshake being returned from the endpoint**, all pending IRPs are
retired. Removal of the halt condition is achieved **via software intervention through a separate control pipe.
This recovery will reset the data toggle bit to DATA0 for the endpoint on both the host and the device**.

problem: (stall token packet return )STALL handshake being returned from the endpoint
solution:  (clear halt to sync data toggle)via software intervention through a separate control pipe.
This recovery will reset the data toggle bit to DATA0 for the endpoint on both the host and the device

libusb/os/windows_common.h
HotIce0: thks, this change is already restored

HotIce0:
Look at it another way.
I was used USB Logic Analyzer to capture packet between device and hcd. The device is indeed a STALL that is returned in bulk IN transfer. The return of the device does not differ between the two platforms. I think the return value of libusb should not be different.

1. linux:
image
the usb flash disk device send STALL token back. the urb->status is EPIPE.
libusb translate EPIPE to LIBUSB_TRANSFER_STALL (libusb/libusb/os/linux_usbfs.c: handle_bulk_completion)

3. windows
image
the same device, STALL token back, the urb status is USBD_STATUS_STALL_PID.
but libusb translate USBD_STATUS_STALL_PID to  LIBUSB_TRANSFER_ERROR.

HotIce0:
@mcuee How can i resolved "appveyor build failed" , Is it an build environment problem?
```
Build succeeded.
    0 Warning(s)
    0 Error(s)
Time Elapsed 00:00:41.71
...
Uploading artifacts...
Error uploading artifact to the storage: Unable to connect to the remote server
```

mcuee:
@HotIce0 

No worried about the appveyor issue.

libusb/os/windows_common.c
HotIce0: https://github.com/libusb/libusb/pull/1257#issuecomment-1451687864

no side-effect i think, this changes follows the usb specification.

mcuee (approved):

HotIce0:
> @HotIce0
> 
> No worried about the appveyor issue.

:smile: OK, thks!

Youw (approved):
Great analysys.
LGTM

libusb/os/windows_common.h
tormodvolden: You can also amend the first commit and repush it with git push -f. But no worries, we can do this ourselves when we merge it.

The change would have made sense with 4-chars tab width. But we default to 8-chars tab width.

HotIce0: I see, I did not realize that libusb was currently using 8-chars tab width. VSCode defualt to show it with 4-chars. 

Can you tell me how to determine the tab width for a particular project. thks!

tormodvolden:
Looks good, and obviously a win to have the same result on both platforms. I am just wondering why this was not there before. Maybe as the Windows docs indicate, this will only be seen with USB 1.0 devices? Or only with the usbdk backend and not WinUSB?

So is USBD_STATUS_STALL_PID returned when the device returns a stall, while USBD_STATUS_ENDPOINT_HALTED is returned early by Windows if a new transfer is attempted without having cleared the halt condition in between?

HotIce0:
>

HotIce0:
> Looks good, and obviously a win to have the same result on both platforms. I am just wondering why this was not there before. Maybe as the Windows docs indicate, this will only be seen with USB 1.0 devices? Or only with the usbdk backend and not WinUSB?
> 
> So is USBD_STATUS_STALL_PID returned when the device returns a stall, while USBD_STATUS_ENDPOINT_HALTED is returned early by Windows if a new transfer is attempted without having cleared the halt condition in between?

HotIce0:
> So is USBD_STATUS_STALL_PID returned when the device returns a stall, while USBD_STATUS_ENDPOINT_HALTED is returned early by Windows if a new transfer is attempted without having cleared the halt condition in between?

I think your point is right, 

```
ref: msdn
USBD_STATUS_ENDPOINT_HALTED: A transfer was submitted to an endpoint that is stalled.
USBD_STATUS_STALL_PID: The device returned a stall packet identifier (defined for backward compatibility with the USB 1.0)
```

HotIce0:
> Looks good, and obviously a win to have the same result on both platforms. I am just wondering why this was not there before. Maybe as the Windows docs indicate, this will only be seen with USB 1.0 devices? Or only with the usbdk backend and not WinUSB?
> 
> So is USBD_STATUS_STALL_PID returned when the device returns a stall, while USBD_STATUS_ENDPOINT_HALTED is returned early by Windows if a new transfer is attempted without having cleared the halt condition in between?

Unfortunately my stm32 board cannot find now... 😞  maybe in the company, I will try to find it tomorrow.

but i did find something useful in MSDN. 
ref: https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/how-to-recover-from-usb-pipe-errors
```
Transfer failures can result from a device error, such as USBD_STATUS_STALL_PID or USBD_STATUS_BABBLE_DETECTED. They can also result due to an error reported by the host controller, such as USBD_STATUS_XACT_ERROR.
```

HotIce0:
> So is USBD_STATUS_STALL_PID returned when the device returns a stall, while USBD_STATUS_ENDPOINT_HALTED is returned early by Windows if a new transfer is attempted without having cleared the halt condition in between?

It's very strange. I can't construct a case where USBD_STATUS_ENDPOINT_HALTED  is returned with this code.
```
1. bulk out: send CBW(scsi command opcode=0x23), URB_STATUS=USBD_STATUS_SUCCESS.
2. bulk in: recv DATA, URB_STATUS=USBD_STATUS_STALL_PID.
3. bulk in: recv CSW, URB_STATUS=USBD_STATUS_TIMEOUT. （expectation is USBD_STATUS_ENDPOINT_STALL）
```

```c
FileName: xusb.c (with set_option LIBUSB_OPTION_USE_USBDK)
...
// Mass Storage device to test bulk transfers (non destructive test)
static int test_mass_storage(libusb_device_handle *handle, uint8_t endpoint_in, uint8_t endpoint_out)
{
	int r, size;
	uint8_t lun;
	uint32_t expected_tag;
	double device_size;
	uint8_t cdb[16];	// SCSI Command Descriptor Block
	uint8_t buffer[READ_FORMAT_CAPACITIES_LENGTH];

	printf("Reading Max LUN:\n");
	r = libusb_control_transfer(handle, LIBUSB_ENDPOINT_IN|LIBUSB_REQUEST_TYPE_CLASS|LIBUSB_RECIPIENT_INTERFACE,
		BOMS_GET_MAX_LUN, 0, 0, &lun, 1, 1000);
	// Some devices send a STALL instead of the actual value.
	// In such cases we should set lun to 0.
	if (r == 0) {
		lun = 0;
	} else if (r < 0) {
		perr("   Failed: %s", libusb_strerror((enum libusb_error)r));
	}
	printf("   Max LUN = %d\n", lun);

	// Send READ_FORMAT_CAPACITIES 0x23
	printf("Sending ReadFormatCapacityes:\n");
	memset(buffer, 0, sizeof(buffer));
	memset(cdb, 0, sizeof(cdb));
	cdb[0] = 0x23;
	cdb[8] = 0xfc;

	// CBW
	send_mass_storage_command(handle, endpoint_out, lun, cdb, LIBUSB_ENDPOINT_IN, READ_FORMAT_CAPACITIES_LENGTH, &expected_tag);
	// DATA
	r = libusb_bulk_transfer(handle, endpoint_in, (unsigned char*)&buffer, READ_FORMAT_CAPACITIES_LENGTH, &size, 5000);
	if (r < 0) {
		printf("data in: bulk transfer failed, r=%d %s, expect=%d LIBUSB_ERROR_PIPE\n",
			r, libusb_strerror(r), LIBUSB_ERROR_PIPE);
	}
	printf("   received %d bytes\n", size);
	// CSW
	struct command_status_wrapper csw;
	r = libusb_bulk_transfer(handle, endpoint_in, (unsigned char*)&buffer, sizeof(struct command_status_wrapper), &size, 5000);
	if (r < 0) {
		printf("data in: bulk transfer failed, r=%d %s, expect=%d LIBUSB_ERROR_PIPE\n",
			r, libusb_strerror(r), LIBUSB_ERROR_PIPE);
	}

	return 0;
}
```

I executed "xusb.exe -k 058f:6387" and output as follows.
```
C:\Users\Administrator\SynologyDrive\projects\debug\endpoint_halt_confirm>xusb.exe -k 058f:6387
Using libusb v1.0.26.11724

Opening device 058F:6387...

Reading device descriptor:
            length: 18
      device class: 0
               S/N: 3
           VID:PID: 058F:6387
         bcdDevice: 0100
   iMan:iProd:iSer: 1:2:3
          nb confs: 1
....
Sending ReadFormatCapacityes:
   sent 10 CDB bytes
libusb: info [usbd_status_to_libusb_transfer_status] zengguang: USBD_STATUS 0xc0000004 translated to LIBUSB_TRANSFER_STALL
data in: bulk transfer failed, r=-9 Pipe error, expect=-9 LIBUSB_ERROR_PIPE
   received 0 bytes
data in: bulk transfer failed, r=-7 Operation timed out, expect=-9 LIBUSB_ERROR_PIPE

Releasing interface 0...
Closing device...
```

tormodvolden:
> It's very strange. I can't construct a case where USBD_STATUS_ENDPOINT_HALTED is returned with this code.

Thanks for investigating this. Maybe it is dependent on the driver, or the documentation is not so precise. Anyway, I think your code is correct in any case.

tormodvolden:

PR #1266, opened 2023-04-04
Adjust IOUSBLib interface versions used based on the running macOS (or Mac OS X) version

hjelmn:
This is an RFC (needs some cleanup) on a potential way to handle using newer IOUSBLib interfaces even if the minimum macOS version of the binary is older. This is done by dynamically choosing the appropriate interface based on the environment at compile time and run time.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -167,6 +167,73 @@ static int get_interface_interface_version(void) {
   return get_interface_interface()->version;
 }
 
+static const struct darwin_iokit_interface *get_device_interface(void) {
+  struct darwin_iokit_interface interfaces[] = {
+#if defined(kIOUSBDeviceInterfaceID650)
+    {
+      .min_os_version = 100900,
+      .version = 650,
+      .interface_id = kIOUSBDeviceInterfaceID650,
+    },
+#endif
+#if defined(kIOUSBDeviceInterfaceID500)
+    {
+      .min_os_version = 100703,
+      .version = 500,
+      .interface_id = kIOUSBDeviceInterfaceID500,
+    },
+#endif
+#if defined(kIOUSBDeviceInterfaceID320)
+    {
+      .min_os_version = 100504,
+      .version = 320,
+      .interface_id = kIOUSBDeviceInterfaceID320,
+    },
+#endif
+#if defined(kIOUSBDeviceInterfaceID300)
+    {
+      .min_os_version = 100500,
+      .version = 300,
+      .interface_id = kIOUSBDeviceInterfaceID300,
+    },
+#endif
+#if defined(kIOUSBDeviceInterfaceID245)
+    {
+      .min_os_version = 100407,
+      .version = 245,
+      .interface_id = kIOUSBDeviceInterfaceID245,
+    },
+#endif
+    {
+      .min_os_version = 100000,
+      .version = 197,
+      .interface_id = kIOUSBDeviceInterfaceID197,
+    },
+    {
+      .version = 0,
+    },
+  };
+  static struct darwin_iokit_interface cached_interface = {.version = 0};
+  if (0 == cached_interface.version) {
+    uint32_t os_version = get_running_version();
+    for (int i = 0 ; interfaces[i].version > 0 ; ++i) {
+      if (os_version >= interfaces[i].min_os_version && cached_interface.min_os_version < interfaces[i].min_os_version) {
+        cached_interface = interfaces[i];
+      }
+    }
+  }
+
+  return &cached_interface;
+}
+
+static CFUUIDRef get_device_interface_id(void) {
+  return get_device_interface()->interface_id;
+}
+
+static int get_device_interface_version(void) {
+  return get_device_interface()->version;
+}
+
 struct darwin_pipe_properties {
   uint8_t number;
   uint8_t direction;
@@ -343,7 +410,7 @@ static void darwin_deref_cached_device(struct darwin_cached_device *cached_dev) 
     list_del(&cached_dev->list);
 
     if (cached_dev->device) {
-      (*(cached_dev->device))->Release(cached_dev->device);
+      (*cached_dev->device)->Release(cached_dev->device);
       cached_dev->device = NULL;
     }
     IOObjectRelease (cached_dev->service);
@@ -463,12 +530,12 @@ static bool get_ioregistry_value_data (io_service_t service, CFStringRef propert
   return success;
 }
 
-static usb_device_t **darwin_device_from_service (struct libusb_context *ctx, io_service_t service)
+static int darwin_device_from_service (struct libusb_context *ctx, io_service_t service, usb_device_t* device)
 {
   io_cf_plugin_ref_t *plugInInterface = NULL;
-  usb_device_t **device;
   IOReturn kresult;
   SInt32 score;
+  
   const int max_retries = 5;
 
   /* The IOCreatePlugInInterfaceForService function might consistently return
@@ -490,15 +557,15 @@ static usb_device_t **darwin_device_from_service (struct libusb_context *ctx, io
 
   if (kIOReturnSuccess != kresult || !plugInInterface) {
     usbi_dbg (ctx, "could not set up plugin for service: %s", darwin_error_str (kresult));
-    return NULL;
+    return darwin_to_libusb(kresult);
   }
 
-  (void)(*plugInInterface)->QueryInterface(plugInInterface, CFUUIDGetUUIDBytes(DeviceInterfaceID),
-                                           (LPVOID)&device);
+  (void)(*plugInInterface)->QueryInterface(plugInInterface, CFUUIDGetUUIDBytes(get_device_interface_id()),
+                                           (LPVOID)device);
   /* Use release instead of IODestroyPlugInInterface to avoid stopping IOServices associated with this device */
   (*plugInInterface)->Release (plugInInterface);
 
-  return device;
+  return LIBUSB_SUCCESS;
 }
 
 static void darwin_devices_attached (void *ptr, io_iterator_t add_devices) {
@@ -568,7 +635,7 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) {
 
           /* the device object is no longer usable so go ahead and release it */
           if (old_device->device) {
-            (*(old_device->device))->Release(old_device->device);
+            (*old_device->device)->Release(old_device->device);
             old_device->device = NULL;
           }
 
@@ -848,12 +915,12 @@ static int get_configuration_index (struct libusb_device *dev, UInt8 config_valu
   IOReturn kresult;
 
   /* is there a simpler way to determine the index? */
-  kresult = (*(priv->device))->GetNumberOfConfigurations (priv->device, &numConfig);
+  kresult = (*priv->device)->GetNumberOfConfigurations (priv->device, &numConfig);
   if (kresult != kIOReturnSuccess)
     return darwin_to_libusb (kresult);
 
   for (i = 0 ; i < numConfig ; i++) {
-    (*(priv->device))->GetConfigurationDescriptorPtr (priv->device, i, &desc);
+    (*priv->device)->GetConfigurationDescriptorPtr (priv->device, i, &desc);
 
     if (desc->bConfigurationValue == config_value)
       return i;
@@ -905,7 +972,7 @@ static int darwin_get_config_descriptor(struct libusb_device *dev, uint8_t confi
 
 /* check whether the os has configured the device */
 static enum libusb_error darwin_check_configuration (struct libusb_context *ctx, struct darwin_cached_device *dev) {
-  usb_device_t **darwin_device = dev->device;
+  usb_device_t darwin_device = dev->device;
 
   IOUSBConfigurationDescriptorPtr configDesc;
   IOUSBFindInterfaceRequest request;
@@ -941,7 +1008,7 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
   request.bInterfaceProtocol = kIOUSBFindInterfaceDontCare;
   request.bAlternateSetting  = kIOUSBFindInterfaceDontCare;
 
-  kresult = (*(darwin_device))->CreateInterfaceIterator(darwin_device, &request, &interface_iterator);
+  kresult = (*darwin_device)->CreateInterfaceIterator(darwin_device, &request, &interface_iterator);
   if (kresult != kIOReturnSuccess)
     return darwin_to_libusb (kresult);
 
@@ -970,7 +1037,7 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
   return LIBUSB_SUCCESS;
 }
 
-static IOReturn darwin_request_descriptor (usb_device_t **device, UInt8 desc, UInt8 desc_index, void *buffer, size_t buffer_size) {
+static IOReturn darwin_request_descriptor (usb_device_t device, UInt8 desc, UInt8 desc_index, void *buffer, size_t buffer_size) {
   IOUSBDevRequestTO req;
 
   assert(buffer_size <= UINT16_MAX);
@@ -991,7 +1058,7 @@ static IOReturn darwin_request_descriptor (usb_device_t **device, UInt8 desc, UI
 }
 
 static enum libusb_error darwin_cache_device_descriptor (struct libusb_context *ctx, struct darwin_cached_device *dev) {
-  usb_device_t **device = dev->device;
+  usb_device_t device = dev->device;
   int retries = 1;
   long delay = 30000; // microseconds
   int unsuspended = 0, try_unsuspend = 1, try_reconfigure = 1;
@@ -1042,15 +1109,17 @@ static enum libusb_error darwin_cache_device_descriptor (struct libusb_context *
 
     if (kIOReturnSuccess != ret && is_open && try_unsuspend) {
       /* device may be suspended. unsuspend it and try again */
-#if DeviceVersion >= 320
-      UInt32 info = 0;
+#if MAX_DEVICE_VERSION >= 320
+      if (get_device_interface_version() >= 320) {
+        UInt32 info = 0;
 
-      /* IOUSBFamily 320+ provides a way to detect device suspension but earlier versions do not */
-      (void)(*device)->GetUSBDeviceInformation (device, &info);
+        /* IOUSBFamily 320+ provides a way to detect device suspension but earlier versions do not */
+        (void)(*IODEVICE_V(device, 320))->GetUSBDeviceInformation (device, &info);
 
-      /* note that the device was suspended */
-      if (info & (1U << kUSBInformationDeviceIsSuspendedBit) || 0 == info)
-        try_unsuspend = 1;
+        /* note that the device was suspended */
+        if (info & (1U << kUSBInformationDeviceIsSuspendedBit) || 0 == info)
+          try_unsuspend = 1;
+      }
 #endif
 
       if (try_unsuspend) {
@@ -1163,7 +1232,7 @@ static enum libusb_error darwin_get_cached_device(struct libusb_context *ctx, io
   UInt64 sessionID = 0, parent_sessionID = 0;
   UInt32 locationID = 0;
   enum libusb_error ret = LIBUSB_SUCCESS;
-  usb_device_t **device;
+  usb_device_t device;
   UInt8 port = 0;
 
   /* assuming sessionID != 0 normally (never seen it be 0) */
@@ -1206,9 +1275,8 @@ static enum libusb_error darwin_get_cached_device(struct libusb_context *ctx, io
 
     usbi_dbg(ctx, "caching new device with sessionID 0x%" PRIx64, sessionID);
 
-    device = darwin_device_from_service (ctx, service);
-    if (!device) {
-      ret = LIBUSB_ERROR_NO_DEVICE;
+    ret = darwin_device_from_service (ctx, service, &device);
+    if (LIBUSB_SUCCESS != ret) {
       break;
     }
 
@@ -1323,7 +1391,7 @@ static enum libusb_error process_new_device (struct libusb_context *ctx, struct 
       dev->parent_dev = usbi_get_device_by_session_id (ctx, (unsigned long) cached_device->parent_session);
     }
 
-    (*(priv->dev->device))->GetDeviceSpeed (priv->dev->device, &devSpeed);
+    (*priv->dev->device)->GetDeviceSpeed (priv->dev->device, &devSpeed);
 
     switch (devSpeed) {
     case kUSBDeviceSpeedLow: dev->speed = LIBUSB_SPEED_LOW; break;
@@ -1388,11 +1456,12 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
 static int darwin_open (struct libusb_device_handle *dev_handle) {
   struct darwin_device_handle_priv *priv = usbi_get_device_handle_priv(dev_handle);
   struct darwin_cached_device *dpriv = DARWIN_CACHED_DEVICE(dev_handle->dev);
+  usb_device_t darwin_device = dpriv->device;
   IOReturn kresult;
 
   if (0 == dpriv->open_count) {
     /* try to open the device */
-    kresult = (*(dpriv->device))->USBDeviceOpenSeize (dpriv->device);
+    kresult = (*darwin_device)->USBDeviceOpenSeize (darwin_device);
     if (kresult != kIOReturnSuccess) {
       usbi_warn (HANDLE_CTX (dev_handle), "USBDeviceOpen: %s", darwin_error_str(kresult));
 
@@ -1407,12 +1476,13 @@ static int darwin_open (struct libusb_device_handle *dev_handle) {
     }
 
     /* create async event source */
-    kresult = (*(dpriv->device))->CreateDeviceAsyncEventSource (dpriv->device, &priv->cfSource);
+    kresult = (*darwin_device)->CreateDeviceAsyncEventSource (darwin_device,
+                                                                                &priv->cfSource);
     if (kresult != kIOReturnSuccess) {
       usbi_err (HANDLE_CTX (dev_handle), "CreateDeviceAsyncEventSource: %s", darwin_error_str(kresult));
 
       if (priv->is_open) {
-        (*(dpriv->device))->USBDeviceClose (dpriv->device);
+        (*darwin_device)->USBDeviceClose (darwin_device);
       }
 
       priv->is_open = false;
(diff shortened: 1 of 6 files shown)


mcuee:
@seanm 

Just wondering if you can review this when you have some time. Thanks.

hjelmn:
@seanm Have you had a chance to test this. If not I will try to set up an environment to test. I think I still have an Intel Mac in a box somewhere..... Best chance I have at compiling across a large range of macOS versions.

libusb/version_nano.h
```diff
@@ -1 +1 @@
-#define LIBUSB_NANO 11792
```
tormodvolden: The nano jumps by +2 here.

libusb/os/darwin_usb.c
```diff
@@ -481,15 +549,15 @@ static usb_device_t **darwin_device_from_service (struct libusb_context *ctx, io
 
   if (kIOReturnSuccess != kresult || !plugInInterface) {
     usbi_dbg (ctx, "could not set up plugin for service: %s", darwin_error_str (kresult));
-    return NULL;
+    return darwin_to_libusb(kresult);
   }
```
tormodvolden, line 553: This looks like an independent fix so should also be mentioned in the commit message.

tormodvolden:
Nice stuff. Looks like /the/ solution to this problem.

libusb/os/darwin_usb.c
```diff
@@ -91,6 +91,179 @@ static enum libusb_error process_new_device (struct libusb_context *ctx, struct
...
+  uint16_t max_packet_size;
+  uint8_t interval;
+};
+typedef struct darwin_pipe_properties darwin_pipe_properties_t;
+
+static IOReturn darwin_get_pipe_properties(struct darwin_interface *cInterface, uint8_t pipe, darwin_pipe_properties_t *out) {
+  IOReturn kresult;
+
+#if (MAX_INTERFACE_VERSION >= 550)
+  if (get_interface_interface_version() >= 550) {
```
Youw, line 249: nit: this `if` (and a cast below) is only required if `MIN_INTERFACE_VERSION < 550`
but I'm pretty sure it would make the code unreadable if one would try to take that into account

libusb/os/darwin_usb.h
```diff
@@ -29,125 +29,132 @@
...
+#endif
+#if defined (kIOUSBDeviceInterfaceID245)
+  /* New in OS 10.4.7. */
+  IOUSBDeviceInterface245 **device245;
+#endif
+  /* New in OS 10.2.3 but can't test deployment target to that granularity, so round up. Oldest
+   * version supported. */
+  IOUSBDeviceInterface197 **device197;
+};
+typedef union darwin_device usb_device_t;
```
Youw, line 124: because the value of the `usb_device_t` is always casted to the "currently required" type, and most of the time only a "default" type is used, I believe it would be much more convenient to typedef a `usb_device_t` to a MIN_DEVICE_VERSION, so in most places it can be used as is w/o any cast, and use a macro or an explicit cast only on those few exceptions there are in the code

same regarding the interface interface

Youw (commented):
overall I'm very agree with the approach/implementation

libusb/os/darwin_usb.h
hjelmn: That is a good point. Will make that change.

libusb/version_nano.h
hjelmn: Opps. Forgot to do one more `git amend --no-edit` to fix that. Need to figure out why the scripts do this.

libusb/os/darwin_usb.h
tormodvolden: > That is a good point. Will make that change.

@hjelmn gentle reminder :)

mcuee:
@hjelmn 

Any updates on this PR? Thanks.

libusb/os/darwin_usb.c
```diff
@@ -165,6 +165,74 @@ static int get_interface_interface_version(void) {
...
+    },
+#endif
+#if defined(kIOUSBDeviceInterfaceID300)
+    {
+      .min_os_version = 100500,
+      .version = 300,
+      .interface_id = kIOUSBDeviceInterfaceID300,
+    },
+#endif
+#if defined(kIOUSBInterfaceInterfaceID245)
```
seanm, line 198: All the others use `DeviceInterface` but this one is `InterfaceInterface`, is that accidental?

libusb/os/darwin_usb.c
```diff
@@ -167,6 +167,60 @@ static enum libusb_error darwin_to_libusb (IOReturn result) {
...
 }
 
+uint32_t get_running_version(void) {
+  static int32_t running_version = 0;
+  if (running_version == 0) {
+    char os_version_string[64];
+    size_t os_version_string_len = sizeof(os_version_string) - 1;
+
+    /* newer versions of macOS provide a sysctl for the OS version */
+    int ret = sysctlbyname("kern.osproductversion", os_version_string, &os_version_string_len, NULL, 0);
```
seanm, line 350: I can't find `kern.osproductversion` in any man page or header, which gives me pause.  Are we sure it's public API?

I'm not a big fan of falling back to using the file system either, since that would probably run afoul of App Sandbox, and require apps that don't otherwise use the file system to whitelist that path.

Seems preferable to use `kern.osrelease` which works since forever. It would require shifting from Darwin versions to macOS versions, but the wikipedia macOS articles have tables mapping the two.

libusb/os/darwin_usb.c
```diff
@@ -167,6 +167,60 @@ static enum libusb_error darwin_to_libusb (IOReturn result) {
   }
 }
 
+uint32_t get_running_version(void) {
+  static int32_t running_version = 0;
```
seanm, line 344: Using a `static` here makes this not very thread safe...

I tried running the sysctlbyname + sscanf in a loop 1 million times, and it takes only about 1.3 µs on my Mac Studio.

Maybe the caching is not needed, especially if we remove the file system usage.  Is `get_running_version` called in any hot code paths?

libusb/os/darwin_usb.h
```diff
@@ -160,15 +160,16 @@
...
 /* Testing availability */
-#ifndef __has_builtin
-  #define __has_builtin(x) 0  // Compatibility with non-clang compilers.
-#endif
-#if __has_builtin(__builtin_available)
-  #define HAS_CAPTURE_DEVICE() __builtin_available(macOS 10.10, *)
-#else
-  #define HAS_CAPTURE_DEVICE() 0
-#endif
+#define HAS_CAPTURE_DEVICE() (get_running_version() >= 101000)
```
seanm, line 179: Since the implementation of `get_running_version` uses a `static`, it's not very thread-safe, and so this macro has changed from thread-safe to not.

Looks like your other commits remove 2 of the 3 uses of this macro, perhaps this macro should just die.

libusb/os/darwin_usb.c
hjelmn: Good catch. That was indeed accidental.

libusb/os/darwin_usb.c
hjelmn: It helps if we have to fall back to a file even though it is not in a hot path. The main path it may be called in is the re-enumerate path which doesn't have to be fast but reading a file every time seemed like a bad idea since the answer never changes. As for the thread safety, it doesn't really matter since this is called under the first time lock so is always set by the time we are thread hot. I can change it to atomic_int32_t to make it truly safe.

hjelmn: Though.... I can't remember whether we require C11 for this backend. I remember wanting to but can't remember if I ever went through with it.

libusb/os/darwin_usb.h
hjelmn: Good point about the macro. Will remove it since it doesn't help with readability.

libusb/os/darwin_usb.c
hjelmn: It has to do with an edge case for `kern.osrelease` that was kind of nasty. It was something like there was a new version of IOKit but the darwin version did not change. The systems where this occurs (10.4.7 and 10.5.4) don't really matter much so it is probably ok to treat them as 10.4.0 and 10.5.0.

Apple has not documented `kern.osproductversion` but it is well understood and widely used. It is set at boot time by launchd. See:

https://github.com/apple/darwin-xnu/commit/5bbb823c13f3ab1ab58878f96b35433a29882676#diff-1daf70413d68262120e719485aac47dd7da773d2333383c9fe5fd34dc93a5ffcR1694

Apple has serious documentation issues and there is no longer a way to communicate these back to them. radar is no longer accessible outside of Apple.

I have no problem with making it fall back to `kern.osrelease` instead of a file for the older releases and just remove the point release granularity. I might have to retest given this code was tested in VMs going back to 10.4.x.

libusb/os/darwin_usb.c
hjelmn: Ok, I reorganized this code to use sysctl only which removes any need to cache. I also remembered what I was doing. The first time init was calling with a lock to fill in the cached values so there was never a threading issue anyway.

libusb/os/darwin_usb.h
hjelmn: Done.

tormodvolden: Great! Since I am queuing up some merges now, please let me merge this afterwards and I'll take care of the nanoversioning.

tormodvolden:
Just to confirm @hjelmn, is this finished now?

libusb/os/darwin_usb.c
```diff
@@ -69,6 +69,11 @@ static usbi_mutex_t darwin_cached_devices_mutex = PTHREAD_MUTEX_INITIALIZER;
 static struct list_head darwin_cached_devices;
 static const char *darwin_device_class = "IOUSBDevice";
 
+DEFAULT_VISIBILITY uint32_t libusb_testonly_fake_running_version;
+DEFAULT_VISIBILITY int libusb_testonly_using_running_interface_version;
+DEFAULT_VISIBILITY int libusb_testonly_using_running_device_version;
+DEFAULT_VISIBILITY bool libusb_testonly_clear_running_version_cache;
```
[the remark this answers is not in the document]
Youw: oh and BTW:
```suggestion
DEFAULT_VISIBILITY uint32_t libusb_testonly_fake_running_version = 0;
DEFAULT_VISIBILITY int libusb_testonly_using_running_interface_version = 0;
DEFAULT_VISIBILITY int libusb_testonly_using_running_device_version = 0;
DEFAULT_VISIBILITY bool libusb_testonly_clear_running_version_cache = 0;
```
?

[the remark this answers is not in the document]
hjelmn: C defines all variables with static storage duration as initialized to 0 (C89 § 3.5.7). So, no need to set them as they are already set.

As for making them not available in a non-debug. That was the original intent. Must have lost part of the change. Fixing.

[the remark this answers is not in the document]
hjelmn: Even better. I can force using the static version of the library. That should be cleaner.

hjelmn:
Ok, that should do it. I think I addressed all the comments and fixed it so CI runs the new test (we probably should not hard code test names in CI but will come back to that later).

libusb/os/darwin_usb.c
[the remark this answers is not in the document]
Youw: > C defines all variables with static storage duration as initialized to 0 (C89 § 3.5.7). So, no need to set them as they are already set.

Just my oppinion: if you have to remind this to people - better make it explicit. That fact that some behavior is well defined and documented - doesn't mean it is obious or readable.

Youw (approved):
LGTM

hjelmn:
gIOKitLibServerVersion is not documented and should not be relied on. A Google search for it only returns a single result (the file it is used in). `kern.osproductversion` and `kern.osrelease` are well known and should be used here. You are correct that iOS will get a different result. I will tweak it to not use `kern.osproductversion` in that case.

Youw:
> gIOKitLibServerVersion is not documented and should not be relied on.

Now that you've mentioned it: there is `kCFCoreFoundationVersionNumber` from `CFBase.h` and that one is well [documented](https://developer.apple.com/documentation/corefoundation/base_utilities/framework_version_numbers?language=objc) for the purpose we need:

> Compare these values to the value of the [kCFCoreFoundationVersionNumber](https://developer.apple.com/documentation/corefoundation/kcfcorefoundationversionnumber?language=objc) variable to identify on which version of the Core Foundation framework your code is executing.

Core Foundation framework versioning fully matched macOS versioning.

HIDAPI is successfully [using](https://github.com/libusb/hidapi/blob/7011fa98af2dde00c298105735e470de800288c7/mac/hid.c#L468) it.

hjelmn:
`kCFCoreFoundationVersionNumber` is a compile time constant. We need what the running version is so we can select the right version.

I could also just walk through the versions from newest to oldest and record whichever one works first. Kind of ugly but it would probably be less fragile.

Youw:
> `kCFCoreFoundationVersionNumber` is a compile time constant

It is not. It looks like one, but it is defined as an externa variable (CoreFoundation framework/library). Look at `CFBase.h`.

image

hjelmn:
Leave it to Apple to use the k prefix (universally meaning constant) for something non-constant.

hjelmn:
Well, that is annoying. Apple stopped adding constants for the CoreFoundation version with 10.12.0. So, not sure I want to use that either. I have an idea. Will work through it now.

hjelmn:
Lol. Our Wiki page says we do not support iOS. The code as is technically fits with that. I can just protect the osreleaseversion check with `#if TARGET_OSX` to ensure we don't break anything for those who use it anyway.

seanm:
I don't think adding Objective-C to libusb just for this is worth it.  osproductversion seems unlikely to go away.

hjelmn:
Also, `NSProcessInfo.operatingSystemVersion` appears to be a wrapper around the sysctl given that the was added at the same time.

hjelmn:
The latest version I pushed a short while ago disables the `osproductversion` check for iOS targets. That should be sufficient to not break iOS until I can determine a good workaround.

Another thing I am working on is just trying each version until it works. I need more time to work on it and get it right before moving to it. Let's proceed with a review of this version and get it in ASAP so 1.0.27 can get out.

Youw (approved):
No more comments from me.

libusb/os/darwin_usb.c
```diff
@@ -170,6 +170,55 @@ static enum libusb_error darwin_to_libusb (IOReturn result) {
...
+  size_t os_version_string_len = sizeof(os_version_string) - 1;
+
+#if !defined(TARGET_OS_OSX) || TARGET_OS_OSX == 1
+  /* newer versions of macOS provide a sysctl for the OS version but this is not useful for iOS without
+   * code detecting this is iOS and a mapping from iOS -> macOS version. it is still useful to have since
+   * it provides the exact macOS version instead of the approximate version (as below). */
+  int ret = sysctlbyname("kern.osproductversion", os_version_string, &os_version_string_len, NULL, 0);
+  if (ret == 0) {
+    int major = 10, minor = 0, patch = 0;
+    sscanf(os_version_string, "%i.%i.%i", &major, &minor, &patch);
```
seanm, line 374: I've been playing with fuzzing lately, so can't help but see things like: what if `os_version_string` is garbage?  Might be good to check the return value of `sscanf` and make sure it's 3. Also, defaulting `major` to 0 seems preferable.

libusb/os/darwin_usb.c
```diff
@@ -170,6 +170,55 @@ static enum libusb_error darwin_to_libusb (IOReturn result) {
   }
 }
 
+uint32_t get_running_version(void) {
+  char os_version_string[64];
+  size_t os_version_string_len = sizeof(os_version_string) - 1;
```
seanm, line 365: These variable declarations best inside the `#if` below I believe...

libusb/os/darwin_usb.h
```diff
@@ -160,15 +160,13 @@
...
-/* Testing availability */
-#ifndef __has_builtin
-  #define __has_builtin(x) 0  // Compatibility with non-clang compilers.
-#endif
-#if __has_builtin(__builtin_available)
-  #define HAS_CAPTURE_DEVICE() __builtin_available(macOS 10.10, *)
-#else
-  #define HAS_CAPTURE_DEVICE() 0
-#endif
+/* returns the current macos version in a format similar to the
```
seanm, line 117: nit: macos -> macOS

libusb/os/darwin_usb.h
```diff
@@ -71,54 +71,37 @@ typedef IOUSBInterfaceInterface220 **usb_interface_t;
...
+#define	MAX_DEVICE_VERSION 197
+#endif
 
-#error "IOUSBFamily is too old. Please upgrade your SDK and/or deployment target"
-
+#if (MAC_OS_X_VERSION_MIN_REQUIRED >= 100900)
+typedef IOUSBDeviceInterface650 **usb_device_t;
+#elif (MAC_OS_X_VERSION_MIN_REQUIRED >= 100700)
+typedef IOUSBDeviceInterface500 **usb_device_t;
+#elif (MAC_OS_X_VERSION_MIN_REQUIRED >= 100503)
```
seanm, line 92: The old comment says "New in OS 10.5.4". I checked the macOS SDK and it says the same. So this should be `>= 100504` I think.

seanm: But wait, also, like the old comment said, back then `MAC_OS_X_VERSION_MIN_REQUIRED` did not have such granularity...

libusb/os/darwin_usb.c
hjelmn: Based on the minimum version being 10.0. It is find if it only scans two number given that there are plenty of examples of that (14.1 for example). Might be worth reporting a parting error if the number is smaller than 2 given Apple has never had a version without a minor release.

libusb/os/darwin_usb.c
hjelmn: Fixed.

libusb/os/darwin_usb.c
```diff
@@ -170,6 +170,55 @@ static enum libusb_error darwin_to_libusb (IOReturn result) {
...
+  /* the minimum released kernel version is 1.3.1 (OS X 10.0) */
+  int darwin_major = 1, darwin_minor = 3;
+  (void) sscanf(os_release_string, "%i.%i", &darwin_major, &darwin_minor);
+
+  int major = 10, minor = 0, patch = 0;
+
+  if (1 == darwin_major && darwin_minor < 4) {
+    /* 10.0.x */
+  } else if (darwin_major < 6) {
+    /* assume 10.1 for anything in this rang */
```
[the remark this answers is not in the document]
hjelmn: Fixed.

libusb/os/darwin_usb.h
hjelmn: Fixed.

libusb/os/darwin_usb.c
```diff
@@ -168,6 +168,74 @@ static int get_interface_interface_version(void) {
   return get_interface_interface()->version;
 }
 
+static const struct darwin_iokit_interface *get_device_interface(void) {
+  struct darwin_iokit_interface interfaces[] = {
```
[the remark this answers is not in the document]
hjelmn: Fixed.

libusb/os/darwin_usb.h
hjelmn: Ok, I remember now. Before 10.10 the version was only xxyy major + minor without a patch level. Let me fix this.

hjelmn: See https://github.com/JamieBerghmans/Cylinder-iOS13/blob/master/deps/iPhoneOS13.2.sdk/usr/include/AvailabilityMacros.h for reference.

hjelmn: Thanks for catching that. This code went through a bunch of revisions and lost that min required can't catch patch releases in 10.9 and earlier.

hjelmn: Actually, this is not as useful as it once was. Will just define it to the minimum version. We can bump that up if needed at some point.

hjelmn:
Weird that the macOS build check failed. Running fine on my mac. Looking into what is going on.

libusb/os/darwin_usb.c
seanm: I'm not too concerned about it ... but yeah, an `if >= 2` would be more correct.

libusb/os/darwin_usb.c
```diff
@@ -94,6 +94,112 @@ static enum libusb_error process_new_device (struct libusb_context *ctx, struct
...
+      .version = 0,
+    },
+  };
+  static struct darwin_iokit_interface cached_interface = {.version = 0};
+  if (0 == cached_interface.version) {
+    uint32_t os_version = get_running_version();
+    for (int i = 0 ; interfaces[i].version > 0 ; ++i) {
+      if (os_version >= interfaces[i].min_os_version) {
+        cached_interface = interfaces[i];
+        break;
```
seanm, line 165: On further thought, maybe better to avoid this need to sort, remove the `break`, and just iterate over all of them and find the max.

`struct darwin_iokit_interface interfaces[]` can also be const.

libusb/os/darwin_usb.c
```diff
@@ -2335,19 +2407,19 @@ static int darwin_abort_transfers (struct usbi_transfer *itransfer) {
...
 
   /* abort transactions */
-#if InterfaceVersion >= 550
-  if (LIBUSB_TRANSFER_TYPE_BULK_STREAM == transfer->type)
-    kresult = (*(cInterface->interface))->AbortStreamsPipe (cInterface->interface, pipeRef, itransfer->stream_id);
-  else
+#if MAX_INTERFACE_VERSION >= 550
+  if (LIBUSB_TRANSFER_TYPE_BULK_STREAM == transfer->type && get_interface_interface_version() >= 550) {
+    kresult = (*IOINTERFACE_V(cInterface, 550))->AbortStreamsPipe (IOINTERFACE(cInterface), pipeRef, itransfer->stream_id);
+  } else
```
seanm, line 2516: would be nice for the `else` body have a `{}` (despite the `#endif`)

libusb/os/darwin_usb.h
```diff
@@ -29,74 +29,34 @@
 #include <IOKit/IOCFBundle.h>
 #include <IOKit/usb/IOUSBLib.h>
 #include <IOKit/IOCFPlugIn.h>
+#include <CoreServices/CoreServices.h>
```
seanm, line 32: Why is CoreServices needed now?

libusb/os/darwin_usb.c
hjelmn: Agreed. Less error prone that way.

libusb/os/darwin_usb.h
hjelmn: Probably was needed for some prior iteration of the code. Doesn't seem to be needed now.

tormodvolden:
> Weird that the macOS build check failed. Running fine on my mac. Looking into what is going on.

This is the stress_mt failing on CI, #1358.

.private/ci-build.sh
```diff
@@ -82,10 +82,12 @@ make -j4 -k
 if [ "${test}" = "yes" ]; then
 	# Load custom shim for WebUSB tests that simulates Web environment.
 	export NODE_OPTIONS="--require ${scriptdir}/../tests/webusb-test-shim/"
-	for test_name in init_context set_option stress stress_mt; do
+	for test_name in init_context macos set_option stress stress_mt; do
+	    if test -x ./tests/${test_name} ; then
```
RReverser, line 86: Why is this change necessary? It doesn't seem great that it might quietly skip some tests.

hjelmn: Ok, opening a PR to add `make check`.

hjelmn: See #1361

RReverser: This is much better, thanks.

hjelmn: Merged into the PR. Done.

libusb/os/darwin_usb.c
hjelmn: Done. Added a couple of error checks and added some logging.

hjelmn:
Found the typo that caused the test failure. Should be good now.

tormodvolden:
Would be nice if you also could format your commit messages to 72 chars width (see HACKING).

hjelmn:
> Would be nice if you also could format your commit messages to 72 chars width (see HACKING).

Ugh, 72. As if 80 was not bad enough. Will reformat it despite how much I despise the 80 character limit :). It is so 1920s and is archaic in the modern world.

seanm:
> Ugh, 72. As if 80 was not bad enough. Will reformat it despite how much I despise the 80 character limit :). It is so 1920s and is archaic in the modern world.

Agreed. :)  It also breaks automatic line wrapping, especially if you use very large fonts.

hjelmn:
Commit descriptions now limited to 72 characters.

hjelmn:
Rebase mistake signing all of these. Fixing.

tests/Makefile.am
```diff
@@ -1,11 +1,12 @@
 AM_CPPFLAGS = -I$(top_srcdir)/libusb
 LDADD = ../libusb/libusb-1.0.la
-LIBS =
+LDFLAGS = -static
```
tormodvolden, line 3: Maybe worth mentioning in the commit message since it affects all platforms.

tormodvolden:
It is obvious for most people due to the IOUSB* mention, but the two "Add..." commits can safely have "darwin:" prepended.

tormodvolden:
> Ugh, 72. As if 80 was not bad enough. Will reformat it despite how much I despise the 80 character limit :). It is so 1920s and is archaic in the modern world.

Now, now. It is more like 1970! Actually exactly 80 like you did in the last master commits is almost worse than no line breaks at all, since git log indents it slightly. But 72 works for everybody. (Even on macOS the terminal windows are 80 columns by default...)

hjelmn:
> > Ugh, 72. As if 80 was not bad enough. Will reformat it despite how much I despise the 80 character limit :). It is so 1920s and is archaic in the modern world.
> 
> Now, now. It is more like 1970! Actually exactly 80 like you did in the last master commits is almost worse than no line breaks at all, since git log indents it slightly. But 72 works for everybody. (Even on macOS the terminal windows are 80 columns by default...)

Dates back to 1928 and IBM punchcards which were 80 in width! Stuck since then. Somewhat due to a rather contentious fight over readability and whether more lines is more readable vs wider lines. I have mostly honored the 80 character limit but tend to use 120 where I can. That is the minimum with of any terminal and emacs windows I have open. I don't enforce anything but 120 in the darwin backend.

Anyway, the 72 was news to me but never gave any thought to the indenting from git log. Especially since the commit line will go way over 80 characters when marking the various branch heads.

hjelmn:
> It is obvious for most people due to the IOUSB* mention, but the two "Add..." commits can safely have "darwin:" prepended.

Considering I started that practice in libusb I can't believe I forgot to tag them. Will do so.

tests/Makefile.am
hjelmn: True. Using static makes it so tests can access things with hidden visibility. Will make that clear.

hjelmn:
Let's roll an RC with this and put it out to test. If there are any issues should be easy enough to fix in a subsequent RC.

tormodvolden:
No, let's wait till I have merged all that should be in 1.0.27. Otherwise we'll end up with several RCs and people have limited time for testing so we shouldn't cry wolf.

You're free to test on your macs though, I expect no changes to core or macos code by now.

tormodvolden:
These changes seem to have caused a regression, with heap-use-after-free crashes #1386.

PR #1269, opened 2023-04-07
windows: get_dev_port_number try using VID / PID

sonatique:
Since SPDRP_ADDRESS may return the actual port number, but not always, try another way of getting port number by matching VID/PID obtained via IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX with values encoded in dev_id string, by brute-forcing initialization of temporary device objects.

If a match is found, Serial Number is verified as well. If all is succeed then port_nr is assigned to the value found, otherwise revert to port_nr and return value from SPDRP_ADDRESS.

EDIT: updated description for the new work:

SPDRP_ADDRESS is the port number only for the Microsoft hub driver. Under a FabulaTech Virtual USB Host Controller it holds an unrelated instance number, which init_device() passes on as a ConnectionIndex; the hub rejects it and the
device is dropped from the device list entirely.

When the registry properties yield nothing usable, ask the parent hub what is connected to each of its ports and match the VID/PID from the device instance ID, using the serial number to choose between identical devices. Where nothing can tell them apart, no guess is made and the SPDRP_ADDRESS value stands.

Everything goes through the hub handle, so no device is opened. That is deliberate: the caller holds priv->interface_lock while auto_claim() takes autoclaim_lock before interface_lock, so opening the device here would invert that order. The requests are bounded by hub_device_io_control() and the first timeout abandons the scan.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -98,6 +98,8 @@ static enum libusb_transfer_status composite_copy_transfer_data(int sub_api, str
 static int composite_endpoint_supports_raw_io(int sub_api, struct libusb_device_handle* dev_handle, uint8_t endpoint);
 static int composite_endpoint_set_raw_io(int sub_api, struct libusb_device_handle* dev_handle, uint8_t endpoint, int enable);
 static int composite_get_max_raw_io_transfer_size(int sub_api, struct libusb_device_handle* dev_handle, uint8_t endpoint);
+// Other prototypes
+static bool find_dev_port_number_by_descriptor(struct libusb_device *dev, struct libusb_device *parent_dev, DWORD hint_port, DWORD *port_nr);
 
 static usbi_mutex_t autoclaim_lock;
 
@@ -1562,13 +1564,14 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 	return LIBUSB_SUCCESS;
 }
 
-static bool get_dev_port_number(HDEVINFO dev_info, SP_DEVINFO_DATA *dev_info_data, DWORD *port_nr)
+static bool get_dev_port_number(struct libusb_device *dev, HDEVINFO dev_info, SP_DEVINFO_DATA *dev_info_data, struct libusb_device *parent_dev, DWORD *port_nr)
 {
 	char buffer[MAX_KEY_LENGTH];
 	DWORD size;
 	const char *start = NULL;
 	char *end = NULL;
 	long long port;
+	bool address_is_port;
 
 	// First try SPDRP_LOCATION_INFORMATION, which returns a REG_SZ. The string *may* have a format
 	// similar to "Port_#0002.Hub_#000D", in which case we can extract the port number. However, we
@@ -1612,11 +1615,20 @@ static bool get_dev_port_number(HDEVINFO dev_info, SP_DEVINFO_DATA *dev_info_dat
 		}
 	}
 
-	// Lastly, try SPDRP_ADDRESS, which returns a REG_DWORD. The address *may* be the port number,
-	// which is true for the Microsoft driver but may not be true for other drivers. However, we
-	// have no other options here but to accept what it returns.
-	return pSetupDiGetDeviceRegistryPropertyA(dev_info, dev_info_data, SPDRP_ADDRESS,
+	// Next try SPDRP_ADDRESS, which returns a REG_DWORD. The address *may* be the port number,
+	// which is true for the Microsoft driver but may not be true for other drivers.
+	address_is_port = pSetupDiGetDeviceRegistryPropertyA(dev_info, dev_info_data, SPDRP_ADDRESS,
 			NULL, (PBYTE)port_nr, sizeof(*port_nr), &size) && (size == sizeof(*port_nr));
+
+	// Lastly, ask the parent hub which of its ports carries this device. Some hub drivers
+	// (notably virtual ones) store an unrelated instance number in SPDRP_ADDRESS, so prefer
+	// a positive descriptor match over it when we can get one. SPDRP_ADDRESS is unreliable
+	// rather than useless, so hand it over as the first port to check.
+	if (parent_dev != NULL && find_dev_port_number_by_descriptor(dev, parent_dev,
+			address_is_port ? *port_nr : 0, port_nr))
+		return true;
+
+	return address_is_port;
 }
 
 static int enumerate_hcd_root_hub(struct libusb_context *ctx, const char *dev_id,
@@ -2328,7 +2340,9 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 				case USB_API_HUB:
 					parent_dev = get_ancestor(ctx, dev_info_data.DevInst, NULL);
 					if (parent_dev == NULL) {
-						if (!get_dev_port_number(*dev_info, &dev_info_data, &hub_port_nr) || hub_port_nr == 0) {
+						// parent_dev is NULL here, so this is a registry-only lookup:
+						// with no parent hub to interrogate there is nothing to match against.
+						if (!get_dev_port_number(dev, *dev_info, &dev_info_data, NULL, &hub_port_nr) || hub_port_nr == 0) {
 							if (bus_number == UINT8_MAX) {
 								usbi_warn(ctx, "program assertion failed - found more than %u buses, skipping the rest", UINT8_MAX);
 								break;
@@ -2379,7 +2393,7 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 					libusb_unref_device(parent_dev);
 					r = LIBUSB_SUCCESS;
 				} else {
-					if (!get_dev_port_number(*dev_info, &dev_info_data, &port_nr))
+					if (!get_dev_port_number(dev, *dev_info, &dev_info_data, parent_dev, &port_nr))
 						usbi_warn(ctx, "could not retrieve port number for device '%s': %s", dev_id, windows_error_str(0));
 					r = init_device(dev, parent_dev, (uint8_t)port_nr, dev_info_data.DevInst);
 					usbi_mutex_unlock(&priv->interface_lock);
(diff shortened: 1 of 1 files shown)


sonatique:
> @sonatique
> 
> Please help to share some background info and show one example that `SPDRP_ADDRESS is incorrect`. Thanks.

As stated in a comment in the current code: "try SPDRP_ADDRESS, which returns a REG_DWORD. The address *may* be the port number", or it may not.

When experimenting with "FabulaTech Virtual USB Host Controller" (USB-Ethernet bridge) under Hyper-V virtual machine, I found out that indeed, SPDRP_ADDRESS access returns true (so this key is found in registry) but we get an obviously wrong port number. This is because SPDRP_ADDRESS registry key does not seem to be standardised. In my case the number found here was simply an instance number not related to port number at all.

Under regular Windows, I experienced that either we get an port number from SPDRP_LOCATION_INFORMATION or SPDRP_LOCATION_PATHS, or trying SPDRP_ADDRESS return false (such key don't exist). 

Therefore I devised another alternate way to obtain a valid port_number. In my experience, either this code is not called at all (nor is the call using SDRP_ADDRESS) because earlier access to registry lead to a valid port_number, or the SDRP_ADDRESS is not valid and it makes sense to call my new, somehow heavier, implementation.

mcuee:
@sonatique 

I see. I think your explanation is good. I do not have such device to test though.

mcuee:
@tormodvolden 

I think this PR is good to go. Please review as well. Thanks.

mcuee:
@tormodvolden and other reviewers, please take a look at this PR when you have some time. Thanks.

seanm:
I don't know Windows, but I looked this over and don't see anything obviously wrong.

mcuee:
I have set this for milestone 1.0.31.

But the conflicts need to be fixed and we need to wait for the review of the other maintainers.

sonatique:
@mcuee @seanm : I decided to rework on this PR. Since it was 3 years old, I first add to adapt it to several changes that came in between. Some of them forced me to reconsider the implementation from scratch, and I found issues along the way.

This present form is much better, and I tested it as much as I could: a) for regression in the normal case b) for the case the PR address, i.e. the Fabula Virtual USB.

@mcuee : you asked back for a concrete example where SPDRP_ADDRESS is wrong. Here it is, from a FabulaTech Virtual USB Host Controller under Hyper-V. 

I used `listdevs`. The symptom turns out to be worse than a wrong port number: the device disappears from the device list entirely.

master:

```
warning [init_device] could not get node connection information for device
        'USB\VID_9999&PID_0200\XYZ-33100': [87] The parameter is incorrect.
warning [winusb_get_device_list] failed to initialize device 'USB\VID_9999&PID_0200\XYZ-33100'
```

`listdevs` reports only the root hub. SPDRP_ADDRESS hands back an unrelated instance number, init_device() passes it as a ConnectionIndex, the hub rejects it as an invalid parameter, and the device is dropped: invisible to applications.

this PR:

```
debug [find_dev_port_number_by_descriptor] matched device 'USB\VID_9999&PID_0200\XYZ-33100'
      to port 1 of hub \\?\ROOT#SYSTEM#0001#{...} by serial number
debug [init_device] (bus: 1, addr: 0, depth: 1, port: 1): 'USB\VID_9999&PID_0200\XYZ-33100'
```

`listdevs: 9999:0200 (bus 1, device 0) path: 1`

I found no regression on normal Windows: on a 21-device tree (3 root hubs, nested external hubs) the device list and testlibusb strings are byte-identical to master, and the new code never runs at all: the registry lookups always win, confirmed with debug logging.

The fallback itself was exercised by forcing every device through it: ports identical to the registry in every case. Also confirmed: serial-number disambiguation of identical twins on one hub, refusal to guess when identical serial-less devices cannot be told apart, and the request-timeout path (bounded by hub_device_io_control() from #1882, scan abandoned, port still correct).

The implementation has been reworked since the last review so it is worth a fresh look. 

The main change: it no longer opens the device: the parent hub is queried instead, which avoids a lock-order inversion with auto_claim() and works whatever driver is bound.

mcuee:
Let me carry out some simple tests to see if there are any regression or not over the weekend.

mcuee:
No regressions here with my Dell Desktop.

I tested stress_mt and testcore as well.


```
 libusb_pr1269  $Env:LIBUSB_DEBUG=3
 libusb_pr1269  .\build\v145\x64\Release-Hotplug-MT\listdevs.exe
libusb: info [winusbx_init] WinUSB DLL available (with isoch support)
libusb: info [windows_init] UsbDk backend is not available
1a2c:7fff (bus 3, device 11) path: 2.3.2.2
047f:c056 (bus 1, device 4) path: 6
413c:b06e (bus 3, device 7) path: 2.5
413c:b06f (bus 3, device 10) path: 2.3.5
1bcf:2bb4 (bus 1, device 1) path: 3
7768:642c (bus 3, device 15) path: 2.3.2.1
0bda:402e (bus 3, device 9) path: 2.3.4
24ae:4057 (bus 3, device 12) path: 2.3.2.3
0424:2740 (bus 3, device 13) path: 2.3.2.5
8087:0033 (bus 1, device 3) path: 10
0b95:772b (bus 1, device 6) path: 7.1
0bda:8153 (bus 3, device 5) path: 4.4
16c0:05dc (bus 1, device 10) path: 7.4
0bda:0413 (bus 3, device 2) path: 4.3
0bda:0487 (bus 3, device 1) path: 4
05e3:0608 (bus 1, device 5) path: 7
8086:15f0 (bus 3, device 0)
0bda:5487 (bus 3, device 3) path: 2
8086:7ec0 (bus 2, device 0)
0424:2744 (bus 3, device 8) path: 2.3.2
8086:7e7d (bus 1, device 0)
0bda:5413 (bus 3, device 6) path: 2.3
0424:5744 (bus 3, device 4) path: 4.3.2

 libusb_pr1269  .\build\v145\x64\Release-Hotplug-MT\testlibusb.exe
libusb: info [winusbx_init] WinUSB DLL available (with isoch support)
libusb: info [windows_init] UsbDk backend is not available
Dev (bus 3, device 11): 1A2C - 7FFF speed: 12M
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
  Manufacturer:
  Product:                   2.4G Wireless Device
Dev (bus 1, device 4): 047F - C056 speed: 12M
  Manufacturer:              Plantronics
  Product:                   Plantronics Blackwire 3220 Series
Dev (bus 3, device 7): 413C - B06E speed: 480M
Dev (bus 3, device 10): 413C - B06F speed: 480M
Dev (bus 1, device 1): 1BCF - 2BB4 speed: 480M
  Manufacturer:              CN03N9728LU00548DCG8A00
  Product:                   Integrated_Webcam_FHD
Dev (bus 3, device 15): 7768 - 642C speed: 12M
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
  Manufacturer:              Dongle
  Product:                   Inphic IN9 SLE
Dev (bus 3, device 9): 0BDA - 402E speed: 480M
Dev (bus 3, device 12): 24AE - 4057 speed: 1.5M
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
  Manufacturer:              SEMICO
  Product:                   USB Keyboard
Dev (bus 3, device 13): 0424 - 2740 speed: 480M
  Manufacturer:              Microchip Tech
  Product:                   Hub Controller
Dev (bus 1, device 3): 8087 - 0033 speed: 12M
Dev (bus 1, device 6): 0B95 - 772B speed: 480M
Dev (bus 3, device 5): 0BDA - 8153 speed: 5G
Dev (bus 1, device 10): 16C0 - 05DC speed: 1.5M
  Manufacturer:              www.fischl.de
  Product:                   USBasp
Dev (bus 3, device 2): 0BDA - 0413 speed: 5G
Dev (bus 3, device 1): 0BDA - 0487 speed: 10G
Dev (bus 1, device 5): 05E3 - 0608 speed: 480M
Dev (bus 3, device 0): 8086 - 15F0 speed: 10G
Dev (bus 3, device 3): 0BDA - 5487 speed: 480M
Dev (bus 2, device 0): 8086 - 7EC0 speed: 5G
Dev (bus 3, device 8): 0424 - 2744 speed: 480M
Dev (bus 1, device 0): 8086 - 7E7D speed: 5G
Dev (bus 3, device 6): 0BDA - 5413 speed: 480M
Dev (bus 3, device 4): 0424 - 5744 speed: 5G
```

mcuee (approved):

mcuee:
Output from current git.

```
 libusb  .\build\v145\x64\Release-Hotplug-MT\listdevs.exe
libusb: info [winusbx_init] WinUSB DLL available (with isoch support)
libusb: info [windows_init] UsbDk backend is not available
1a2c:7fff (bus 3, device 11) path: 2.3.2.2
047f:c056 (bus 1, device 4) path: 6
413c:b06e (bus 3, device 7) path: 2.5
413c:b06f (bus 3, device 10) path: 2.3.5
1bcf:2bb4 (bus 1, device 1) path: 3
7768:642c (bus 3, device 15) path: 2.3.2.1
0bda:402e (bus 3, device 9) path: 2.3.4
24ae:4057 (bus 3, device 12) path: 2.3.2.3
0424:2740 (bus 3, device 13) path: 2.3.2.5
8087:0033 (bus 1, device 3) path: 10
0b95:772b (bus 1, device 6) path: 7.1
0bda:8153 (bus 3, device 5) path: 4.4
16c0:05dc (bus 1, device 10) path: 7.4
0bda:0413 (bus 3, device 2) path: 4.3
0bda:0487 (bus 3, device 1) path: 4
05e3:0608 (bus 1, device 5) path: 7
8086:15f0 (bus 3, device 0)
0bda:5487 (bus 3, device 3) path: 2
8086:7ec0 (bus 2, device 0)
0424:2744 (bus 3, device 8) path: 2.3.2
8086:7e7d (bus 1, device 0)
0bda:5413 (bus 3, device 6) path: 2.3
0424:5744 (bus 3, device 4) path: 4.3.2
 libusb  .\build\v145\x64\Release-Hotplug-MT\testlibusb.exe
libusb: info [winusbx_init] WinUSB DLL available (with isoch support)
libusb: info [windows_init] UsbDk backend is not available
Dev (bus 3, device 11): 1A2C - 7FFF speed: 12M
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
  Manufacturer:
  Product:                   2.4G Wireless Device
Dev (bus 1, device 4): 047F - C056 speed: 12M
  Manufacturer:              Plantronics
  Product:                   Plantronics Blackwire 3220 Series
Dev (bus 3, device 7): 413C - B06E speed: 480M
Dev (bus 3, device 10): 413C - B06F speed: 480M
Dev (bus 1, device 1): 1BCF - 2BB4 speed: 480M
  Manufacturer:              CN03N9728LU00548DCG8A00
  Product:                   Integrated_Webcam_FHD
Dev (bus 3, device 15): 7768 - 642C speed: 12M
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
  Manufacturer:              Dongle
  Product:                   Inphic IN9 SLE
Dev (bus 3, device 9): 0BDA - 402E speed: 480M
Dev (bus 3, device 12): 24AE - 4057 speed: 1.5M
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
libusb: warning [hid_open] could not open HID device in R/W mode (keyboard or mouse?) - trying without
  Manufacturer:              SEMICO
  Product:                   USB Keyboard
Dev (bus 3, device 13): 0424 - 2740 speed: 480M
  Manufacturer:              Microchip Tech
  Product:                   Hub Controller
Dev (bus 1, device 3): 8087 - 0033 speed: 12M
Dev (bus 1, device 6): 0B95 - 772B speed: 480M
Dev (bus 3, device 5): 0BDA - 8153 speed: 5G
Dev (bus 1, device 10): 16C0 - 05DC speed: 1.5M
  Manufacturer:              www.fischl.de
  Product:                   USBasp
Dev (bus 3, device 2): 0BDA - 0413 speed: 5G
Dev (bus 3, device 1): 0BDA - 0487 speed: 10G
Dev (bus 1, device 5): 05E3 - 0608 speed: 480M
Dev (bus 3, device 0): 8086 - 15F0 speed: 10G
Dev (bus 3, device 3): 0BDA - 5487 speed: 480M
Dev (bus 2, device 0): 8086 - 7EC0 speed: 5G
Dev (bus 3, device 8): 0424 - 2744 speed: 480M
Dev (bus 1, device 0): 8086 - 7E7D speed: 5G
Dev (bus 3, device 6): 0BDA - 5413 speed: 480M
Dev (bus 3, device 4): 0424 - 5744 speed: 5G
```

sonatique:
@Youw @seanm : I wish to merge this PR soon, would you like more time to review it? Thanks!

libusb/os/windows_winusb.c
```diff
@@ -2587,6 +2601,371 @@ static int usbi_utf16le_to_utf8(uint8_t const *src, int src_length, char *dst, i
...
+		return false;
+
+	langid = (uint16_t)sd.desc.bString[0] | ((uint16_t)sd.desc.bString[1] << 8);
+
+	memset(&sd, 0, sizeof(sd));
+	sd.req.ConnectionIndex = connection_index;
+	sd.req.SetupPacket.bmRequest = LIBUSB_ENDPOINT_IN;
+	sd.req.SetupPacket.bRequest = LIBUSB_REQUEST_GET_DESCRIPTOR;
+	sd.req.SetupPacket.wValue = (LIBUSB_DT_STRING << 8) | string_idx;
+	sd.req.SetupPacket.wIndex = langid;
```
Youw, line 2655: [P2] Use LANGID 0x0409 for this comparison

This string is compared with the serial suffix that Windows placed in the device instance ID. Microsoft states that Windows always requests USB serial numbers with LANGID 0x0409. This code instead uses the first LANGID from descriptor zero. If another language is first, the strings can differ. The scan can then retain the untrusted `SPDRP_ADDRESS` for devices that the serial number could distinguish. Request `string_idx` with `wIndex = 0x0409`; the descriptor-zero request and `langid` variable then become unnecessary.

[Microsoft USB FAQ](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-faq--introductory-level#what-langid-is-used-to-extract-a-devices-serial-number)

libusb/os/windows_winusb.c
```diff
@@ -2587,6 +2601,371 @@ static int usbi_utf16le_to_utf8(uint8_t const *src, int src_length, char *dst, i
...
+	if (vid == 0 || vid > UINT16_MAX || pid > UINT16_MAX)
+		return false;
+
+	// The instance ID ends with the serial number when the device reports one,
+	// e.g. "USB\VID_0403&PID_6001\A6008isP". When it does not, Windows synthesises
+	// an ID containing '&', which can never match a real serial number.
+	dev_id_serial = strrchr(priv->dev_id, '\\');
+	if (dev_id_serial != NULL) {
+		dev_id_serial++;
+		if (*dev_id_serial == '\0' || strchr(dev_id_serial, '&') != NULL)
```
Youw, line 2800: [P2] Keep valid serial numbers that contain `&`

`&` is byte 0x26. Microsoft rejects byte 0x2C, bytes below 0x20, and bytes above 0x7F in USB serial numbers, so `&` is valid. A serial such as `ABC&123` is discarded here. With identical VID/PID devices, this removes the discriminator and can leave the invalid `SPDRP_ADDRESS` in use. A synthesized ID for a serial-less device cannot match a descriptor because `iSerialNumber` is zero, so reject only an empty suffix here.

[Microsoft USB FAQ](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-faq--introductory-level#what-characters-or-bytes-are-valid-in-a-usb-serial-number)

```suggestion
		if (*dev_id_serial == '\0')
```

libusb/os/windows_winusb.c
```diff
@@ -2587,6 +2601,371 @@ static int usbi_utf16le_to_utf8(uint8_t const *src, int src_length, char *dst, i
...
+		bool timed_out = false;
+
+		if (!get_port_connection_info(ctx, hub_handle, (ULONG)port, &conn_info, &timed_out)) {
+			if (timed_out) {
+				usbi_warn(ctx, "port %u of hub %s timed out, abandoning the scan for '%s'",
+					port, parent_priv->path, priv->dev_id);
+				scan_incomplete = true;
+				break;
+			}
+			continue;
```
Youw, line 2882: [P2] Mark a failed port query as an incomplete scan

A non-timeout IOCTL failure skips this port but leaves `scan_incomplete` false. A single VID/PID match on a later port can then pass the final uniqueness check. The skipped port can hold the current device while the match is an identical device, so `init_device()` receives the other device's port. Mark the scan incomplete before continuing. A serial match can still return before the incomplete-scan check.

```suggestion
			scan_incomplete = true;
			continue;
```

libusb/os/windows_winusb.c
```diff
@@ -2587,6 +2601,371 @@ static int usbi_utf16le_to_utf8(uint8_t const *src, int src_length, char *dst, i
...
+	// that was never reached could hold the same VID/PID.
+	if (scan_incomplete)
+		return false;
+
+	// A single VID/PID match is unambiguous only if every port could be identified. 
+	// A port in a failure state may hold this device while the lone match is an
+	// identical twin elsewhere on the hub, and taking it would hand back the twin's port.
+	// Where the instance ID carries a serial number the two could have been told apart,
+	// so not having matched it is a failure here rather than a reason to fall back on the count.
+	if (vid_pid_matches == 1 && !(dev_id_serial != NULL && unidentified_port_seen)) {
```
Youw, line 2947: [P2] Reject VID/PID-only matches when any port is unidentified

When the target has no serial number, this expression ignores `unidentified_port_seen`. If the failed-status port holds the target and another port holds a serial-less twin, the function returns the twin's port. That is the guess that the PR description says to avoid. Require `!unidentified_port_seen` for every VID/PID-only result. Also mark `DeviceEnumerating` and `DeviceReset` as unidentified; those ports can hold the target during hotplug or reset, and the incorrect mapping becomes initialized before a later scan can correct it.

```suggestion
	if (vid_pid_matches == 1 && !unidentified_port_seen) {
```

Youw (changes requested):
Codex found four blocking correctness cases in the new port fallback:

- The serial request does not use the LANGID that Windows uses for the device instance ID.
- A valid serial number that contains `&` is treated as a synthetic ID.
- A failed port query does not make the scan incomplete.
- A VID/PID-only result can be accepted while another port is unidentified.

`git diff --check` also reports trailing whitespace on `windows_winusb.c:2942`.

The Windows CI matrix is green. The Linux failure is an unrelated Emscripten link failure.

PR #1270, opened 2023-04-17
windows: no longer wait for device to get an active configuration

sonatique:
This commit essentially revert commit eb164d75 (from 2018). This commit made the assumption that when a device is not configured (i.e. it returns CurrentConfigurationValue == 0 during init_device, then this is only a temporary condition caused by Windows already listing the device but not having completed its setup yet. The commit added a delay of up to 1 second (20 x 50ms) for re-obtaining information and checking until CurrentConfigurationValue would eventually change to a value >0. This commit also forced libusb internal structure to contain 1 as the active configuration even if the wait failed to retreive an active configuration, creating a discrepancy in between the value seen from libusb and actual value managed by Windows.

There is problem with this commit because devices could be in the unconfigured state for permanent reasons. For instance when a device is disabled by Windows (either manually using Device Manager or automatically because of a resource conflict). In this case the device is listed with CurrentConfigurationValue == 0. Therefore init_device will add a 1 second delay for each disabled device, at every device enumeration.

A device which is not configured has no active interface nor available endpoint. Therefore it cannot be used (except for standard request on enpoint 0) and therefore shall not be returned by libusb as an active device. It is the responsability of the OS to bring the device to a configured state. When libusb returns active configuration == 1 to a user, the behavior of actually using this device is not defined.

The potential gain of being able to wait for Windows actually completing device setup is not in par with the delay caused by the waiting which has high-level impact (for instance in GUI listing devices) for standard scenarios.

The initial problem addressed by commit eb164d75 is simply a race condition occuring (rarely) when libusb enumeration happens exactly at the instant when Windows is not yet done loading the driver etc. for an enabled device. Calling init_device slightly earlier would have return no result for this device, and slightly later would have return a fully active device. In the first case a new enumeration would be necessary anyway. Given hotplug is not supported under Windows it is a good practive to re-enumerate devices to detect newly plugged ones anyway, so the race condition could be easily solved by simply enumerating regularly.

To summarize: this commit revert to handling the described race condition by ignoring any not configured device, instead of waiting for them to maybe get an active configuration at a later time. The benefit is to avoid introducing arbitrary large delay in common scenarios where some devices are disabled at the OS levl.

Moreover this commit revert assigning a wrong value to priv->active_config and returning such device from get_device_list, preventing later missused of not configured devices.

Note: disabled devices might not necessarily be listed under USB devices in Windows Device Manager, but can be other system peripherals, one example being Bluetooth chips with HCI USB interfaces.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1074,8 +1074,7 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 	DWORD size;
 	uint8_t bus_number, depth;
 	int r;
-	int ginfotimeout;
-
+	
 	priv = usbi_get_device_priv(dev);
 
 	// If the device is already initialized, we can stop here
@@ -1135,61 +1134,45 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 
 		conn_info.ConnectionIndex = (ULONG)port_number;
 		// coverity[tainted_data_argument]
-		ginfotimeout = 20;
-		do {
-			if (!DeviceIoControl(hub_handle, IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, &conn_info, sizeof(conn_info),
-				&conn_info, sizeof(conn_info), &size, NULL)) {
-				usbi_warn(ctx, "could not get node connection information for device '%s': %s",
-					priv->dev_id, windows_error_str(0));
-				CloseHandle(hub_handle);
-				return LIBUSB_ERROR_NO_DEVICE;
-			}
 
-			if (conn_info.ConnectionStatus == NoDeviceConnected) {
-				usbi_err(ctx, "device '%s' is no longer connected!", priv->dev_id);
-				CloseHandle(hub_handle);
-				return LIBUSB_ERROR_NO_DEVICE;
-			}
+		if (!DeviceIoControl(hub_handle, IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, &conn_info, sizeof(conn_info),
+			&conn_info, sizeof(conn_info), &size, NULL)) {
+			usbi_warn(ctx, "could not get node connection information for device '%s': %s",
+				priv->dev_id, windows_error_str(0));
+			CloseHandle(hub_handle);
+			return LIBUSB_ERROR_NO_DEVICE;
+		}
 
-			if ((conn_info.DeviceDescriptor.bLength != LIBUSB_DT_DEVICE_SIZE)
-				 || (conn_info.DeviceDescriptor.bDescriptorType != LIBUSB_DT_DEVICE)) {
-				SleepEx(50, TRUE);
-				continue;
-			}
-
-			static_assert(sizeof(dev->device_descriptor) == sizeof(conn_info.DeviceDescriptor),
-				      "mismatch between libusb and OS device descriptor sizes");
-			memcpy(&dev->device_descriptor, &conn_info.DeviceDescriptor, LIBUSB_DT_DEVICE_SIZE);
-			usbi_localize_device_descriptor(&dev->device_descriptor);
-
-			priv->active_config = conn_info.CurrentConfigurationValue;
-			if (priv->active_config == 0) {
-				usbi_dbg(ctx, "0x%x:0x%x found %u configurations (not configured)",
-					dev->device_descriptor.idVendor,
-					dev->device_descriptor.idProduct,
-					dev->device_descriptor.bNumConfigurations);
-				SleepEx(50, TRUE);
-			}
-		} while (priv->active_config == 0 && --ginfotimeout >= 0);
+		if (conn_info.ConnectionStatus == NoDeviceConnected) {
+			usbi_err(ctx, "device '%s' is no longer connected!", priv->dev_id);
+			CloseHandle(hub_handle);
+			return LIBUSB_ERROR_NO_DEVICE;
+		}
 
 		if ((conn_info.DeviceDescriptor.bLength != LIBUSB_DT_DEVICE_SIZE)
-			 || (conn_info.DeviceDescriptor.bDescriptorType != LIBUSB_DT_DEVICE)) {
+			|| (conn_info.DeviceDescriptor.bDescriptorType != LIBUSB_DT_DEVICE)) {
 			usbi_err(ctx, "device '%s' has invalid descriptor!", priv->dev_id);
 			CloseHandle(hub_handle);
 			return LIBUSB_ERROR_OTHER;
 		}
 
-		if (priv->active_config == 0) {
-			usbi_info(ctx, "0x%x:0x%x found %u configurations but device isn't configured, "
-				"forcing current configuration to 1",
-				dev->device_descriptor.idVendor,
-				dev->device_descriptor.idProduct,
-				dev->device_descriptor.bNumConfigurations);
-			priv->active_config = 1;
-		} else {
-			usbi_dbg(ctx, "found %u configurations (current config: %u)", dev->device_descriptor.bNumConfigurations, priv->active_config);
+		static_assert(sizeof(dev->device_descriptor) == sizeof(conn_info.DeviceDescriptor),
+			      "mismatch between libusb and OS device descriptor sizes");
+		memcpy(&dev->device_descriptor, &conn_info.DeviceDescriptor, LIBUSB_DT_DEVICE_SIZE);
+		usbi_localize_device_descriptor(&dev->device_descriptor);
+
+		if (conn_info.CurrentConfigurationValue == 0) {
+			usbi_dbg(ctx, "found %u configurations for device '%s' but device is not configured (i.e. current config: 0), ignoring it",
+				dev->device_descriptor.bNumConfigurations,
+				priv->dev_id);
+			CloseHandle(hub_handle);
+			return LIBUSB_ERROR_OTHER;
 		}
 
+		priv->active_config = conn_info.CurrentConfigurationValue;
+		usbi_dbg(ctx, "found %u configurations (current config: %u) for device '%s'", 
+			dev->device_descriptor.bNumConfigurations, priv->active_config,	priv->dev_id);
+		
 		// Cache as many config descriptors as we can
 		cache_config_descriptors(dev, hub_handle);
 


mcuee:
@sonatique 

I think your explanation makes sense and the original patch does not seem to be correct. But let's hear the opinion of @tormodvolden and @hjelmn as well.

nxpfrankli (commented):
Windows backend is quite complex. I am not sure if it works. Can you update your pull request and let me test again locally.

libusb/os/windows_winusb.c
```diff
@@ -1135,59 +1134,42 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
...
 		if (priv->active_config == 0) {
-			usbi_info(ctx, "0x%x:0x%x found %u configurations but device isn't configured, "
-				"forcing current configuration to 1",
-				dev->device_descriptor.idVendor,
-				dev->device_descriptor.idProduct,
-				dev->device_descriptor.bNumConfigurations);
-			priv->active_config = 1;
+			usbi_info(ctx, "found %u configurations for device '%s' but device isn't configured (current config: 0)",
+				dev->device_descriptor.bNumConfigurations,
+				priv->dev_id);
```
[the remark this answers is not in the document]
sonatique: > Windows backend is quite complex. I am not sure if it works. Can you update your pull request and let me test again locally.

Done. Also testing on my side under various conditions, of course.

@nxpfrankli : Please note that before this PR, devices that were still not configured after the 20 attempts would not be discarded but would rather be returned by get_device_list with a forced active_config == 1 (instead of 0) opening the door to undefined behavior.

sonatique:
> > get devices list should return only configured devices
@Youw : I do completly agree, my PR is now doing this. 
With current master, a device that cannot get a proper configuration after 20 attempts, because it is disabled and not just temporary initializing, will be returned with a forced active configuration of 1, but the device is not usable.

Here is the output of testlibusb on a machine where device "1.3" 0CF3 - E300 is disabled.
With master:
```
Dev (bus 1, device 0): 8086 - A12F speed: 5G
Dev (bus 1, device 3): 0CF3 - E300 speed: 12M
Dev (bus 1, device 1): 413C - 3012 speed: 1.5M
  Manufacturer:              Dell
  Product:                   Dell USB Optical Mouse
Dev (bus 1, device 3): 1770 - FF00 speed: 12M
  Manufacturer:              MSI EPF USB
  Product:                   MSI EPF USB
Dev (bus 1, device 2): 04F2 - B536 speed: 480M

```
With this PR:
```
Dev (bus 1, device 0): 8086 - A12F speed: 5G
Dev (bus 1, device 1): 413C - 3012 speed: 1.5M
  Manufacturer:              Dell
  Product:                   Dell USB Optical Mouse
Dev (bus 1, device 3): 1770 - FF00 speed: 12M
  Manufacturer:              MSI EPF USB
  Product:                   MSI EPF USB
Dev (bus 1, device 2): 04F2 - B536 speed: 480M
```

As you can see the line "`Dev (bus 1, device 3): 0CF3 - E300 speed: 12M`" is now missing, indicating that the disabled device is no longer returned.

mcuee:
It seems to me this PR is good to go.

Test results with 04d8:FA2E benchmark device disabled, under Windows 11. 

This PR:
```
PS C:\work\libusb\libusb_pr1270> .\build\v143\x64\Release\testlibusb.exe
Dev (bus 2, device 1): 05E3 - 0612 speed: 5G
Dev (bus 1, device 3): 8087 - 0026 speed: 12M
Dev (bus 1, device 1): 04F2 - B6DD speed: 480M
Dev (bus 1, device 2): 1C7A - 0575 speed: 480M
Dev (bus 1, device 0): 8086 - A0ED speed: 5G
Dev (bus 1, device 4): 05E3 - 0610 speed: 480M
Dev (bus 1, device 7): 046D - C52B speed: 12M
  Manufacturer:              Logitech
  Product:                   USB Receiver
Dev (bus 2, device 0): 8086 - 9A13 speed: 5G

[ 0.025814] [0000167c] libusb: debug [winusb_get_device_list] allocating new device for session [85]
[ 0.025906] [0000167c] libusb: debug [init_device] found 1 configurations for device 'USB\VID_04D8&PID_FA2E\LUSBW1' but device is not configured (i.e. current config: 0), ignoring it
[ 0.025963] [0000167c] libusb: warning [winusb_get_device_list] failed to initialize device 'USB\VID_04D8&PID_FA2E\LUSBW1'
```

Mater branch
```
PS C:\work\libusb\libusb> .\build\v143\x64\Release\testlibusb.exe
Dev (bus 2, device 1): 05E3 - 0612 speed: 5G
Dev (bus 1, device 3): 8087 - 0026 speed: 12M
Dev (bus 1, device 1): 04F2 - B6DD speed: 480M
Dev (bus 1, device 2): 1C7A - 0575 speed: 480M
Dev (bus 1, device 0): 8086 - A0ED speed: 5G
Dev (bus 1, device 4): 05E3 - 0610 speed: 480M
Dev (bus 1, device 9): 04D8 - FA2E speed: 12M
Dev (bus 1, device 7): 046D - C52B speed: 12M
  Manufacturer:              Logitech
  Product:                   USB Receiver
Dev (bus 2, device 0): 8086 - 9A13 speed: 5G

[ 0.022795] [00004dfc] libusb: debug [winusb_get_device_list] allocating new device for session [85]
[ 0.022888] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.080312] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.140344] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.200314] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.260441] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.320445] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.380766] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.431368] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.491353] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.551785] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.611838] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.671798] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.722476] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.782493] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.842406] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.892847] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 0.952669] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 1.012668] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 1.063625] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 1.114419] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 1.174423] [00004dfc] libusb: debug [init_device] 0x4d8:0xfa2e found 1 configurations (not configured)
[ 1.234256] [00004dfc] libusb: info [init_device] 0x4d8:0xfa2e found 1 configurations but device isn't configured, forcing current configuration to 1
[ 1.234556] [00004dfc] libusb: debug [cache_config_descriptors] cached config descriptor 0 (bConfigurationValue=1, 32 bytes)
[ 1.234883] [00004dfc] libusb: debug [init_device] (bus: 1, addr: 9, depth: 2, port: 3): 'USB\VID_04D8&PID_FA2E\LUSBW1'
```

mcuee:
@nxpfrankli 

Please give this PR a test to see if it sorted out your original issue. Thanks.

nxpfrankli:
I just hit once unconfiguration situation.  So far our program have not report error.  Let me contact original reporter check if some condition missed.  It needs some time to confirm.

nxpfrankli:
I can not exactly reproduce the original problem.  But I hit several time unconfiguration condition,  and track my uuu code flow, it should be okay.  so far so good.

sonatique:
@mcuee : do you think we need more information about this PR? Do you expect me to do or provide something or we're good to go? 
Thanks!

mcuee:
> @mcuee : do you think we need more information about this PR? Do you expect me to do or provide something or we're good to go? Thanks!

I think this PR is good to go. But I will let @tormodvolden to have a final look and merge this PR.

tormodvolden:
I definitely agree that libusb should not wait and repoll, but just provide a snapshot of what is available, as fast as possible. Whether unconfigured devices should be returned (marked as unconfigured) or not, is another question. On other platforms there is the possibility of driver-less devices where libusb can set the configuration instead of an OS driver doing it, but I don't know if that is a possibility on Windows - here libusb can only access a device through a driver, and the driver will (always?) have configured the device. A device can be deconfigured with a Set Configuration requests, and as long as libusb keeps a handle on it it can configure it again later. But if it doesn't show up on enumeration, it will be lost for new libusb contexts. However if I understand (our wiki) correctly, WinUSB and libusbK cannot deconfigure a device anyway.

mcuee:
>  However if I understand (our wiki) correctly, WinUSB and libusbK cannot deconfigure a device anyway.

Yes that is correct. libusbK is based on KMDF and it is using the framework's USB I/O Targets. So it has the same limitation as WinUSB. libusb0.sys is based on the older WDM framework so it does not have the limitation.

Reference:
https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/configuring-usb-devices#limitations-for-selecting-a-configuration

<I>
A client driver for a composite device that manages interfaces or interface collections through the [USB Generic Parent Driver](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-common-class-generic-parent-driver) (Usbccgp.sys) cannot change the device's configuration value. However, the client driver can configure Usbccgp.sys to select a configuration other than the first (default) configuration. For more information, see [Configuring Usbccgp.sys to Select a Non-Default USB Configuration](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/selecting-the-configuration-for-a-multiple-interface--composite--usb-d).

A KMDF-based client driver that is using the framework's [USB I/O Targets](https://learn.microsoft.com/en-us/windows-hardware/drivers/wdf/usb-i-o-targets) can select only the first configuration.

[WinUSB](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb) supports only the first configuration.

A class driver frequently lacks support for multiple configurations. If your device implements a class that is defined by a USB class specification, see the [USB technology](https://www.usb.org/defined-class-codes) website for information about device classes and class specifications. Microsoft provides class drivers for the supported USB device classes. For more information, see [Drivers for the Supported USB Device Classes](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/supported-usb-classes).
</I>

libusb/os/windows_winusb.c
```diff
@@ -1135,61 +1134,45 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
...
+			CloseHandle(hub_handle);
+			return LIBUSB_ERROR_NO_DEVICE;
+		}
 
 		if ((conn_info.DeviceDescriptor.bLength != LIBUSB_DT_DEVICE_SIZE)
-			 || (conn_info.DeviceDescriptor.bDescriptorType != LIBUSB_DT_DEVICE)) {
+			|| (conn_info.DeviceDescriptor.bDescriptorType != LIBUSB_DT_DEVICE)) {
 			usbi_err(ctx, "device '%s' has invalid descriptor!", priv->dev_id);
 			CloseHandle(hub_handle);
 			return LIBUSB_ERROR_OTHER;
```
tormodvolden, line 1156: The original code would sleep and retry here. Is this another case of a device not ready yet? In that case, shouldn't we just ignore it and usbi_dbg warn about it instead of writing an error message?

tormodvolden: But it is ambiguous because the original code also checked this after the loop, and at that point an error message was maybe appropriate. I just base my comments on the original code, I don't know in which cases this failure could be transient.

tormodvolden: @Youw please help me understand this, you added these checks in commit 851ff40. Why the wait and retry on failed descriptor length/type?

libusb/os/windows_winusb.c
```diff
@@ -1135,61 +1134,45 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
...
+			      "mismatch between libusb and OS device descriptor sizes");
+		memcpy(&dev->device_descriptor, &conn_info.DeviceDescriptor, LIBUSB_DT_DEVICE_SIZE);
+		usbi_localize_device_descriptor(&dev->device_descriptor);
+
+		if (conn_info.CurrentConfigurationValue == 0) {
+			usbi_dbg(ctx, "found %u configurations for device '%s' but device is not configured (i.e. current config: 0), ignoring it",
+				dev->device_descriptor.bNumConfigurations,
+				priv->dev_id);
+			CloseHandle(hub_handle);
+			return LIBUSB_ERROR_NO_DEVICE;
```
Youw, line 1169: this case looks like `LIBUSB_ERROR_OTHER` to me

Youw (commented):
Other then my last comment, this change looks fine to me.

Unfortunately I no longer have the environment to reproduce the issue I attempted to fix with #806, thus I'm unable to test this PR agains that.
But I don't believe that is really required.

libusb/os/windows_winusb.c
tormodvolden: OK. After looking more into those older issues, I think the previous sleep and retry was not useful for the broken device descriptor - those will just stay broken. So the error message suggested here is correct.

tormodvolden:
Just want to add that the previous code did not allow users to deal with a (non-transient) unconfigured device either (since it would eventually wrongly report it as configured - and basically give up on it as per the comments), so there is no requirement for us to provide this possibility here either.

libusb/os/windows_winusb.c
tormodvolden: Back in winusb_get_device_list() the exact failure return code is not looked at and gets rewritten, so I guess it doesn't matter much. It could otherwise have been nice to differentiate between a permanent and real problem with the device (broken descriptor) and a transient OS/timing issue (not configured) failure, but again, we cannot tell if the not-configured case is transient or if it is a device that will stay unconfigured. All in all I agree LIBUSB_ERROR_OTHER is best, since there is actually a device, but we choose to ignore it because it has a problem. Whereas with LIBUSB_ERROR_NO_DEVICE we communicate that the device has disappeared.

sonatique:
Hello, I am just wondering what is preventing this PR for being merged? May I do something about it?

mcuee:
@tormodvolden 

I think this PR is good to go. Please take a look to see this can be merged. Thanks.

libusb/os/windows_winusb.c
tormodvolden: Is there agreement on which error code to return?

sonatique: @tormodvolden @Youw : done (and rebased)

Youw (approved):
LGTM
+1

libusb/os/windows_winusb.c
```diff
@@ -1135,61 +1134,45 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
...
+			usbi_dbg(ctx, "found %u configurations for device '%s' but device is not configured (i.e. current config: 0), ignoring it",
+				dev->device_descriptor.bNumConfigurations,
+				priv->dev_id);
+			CloseHandle(hub_handle);
+			return LIBUSB_ERROR_OTHER;
 		}
 
+		priv->active_config = conn_info.CurrentConfigurationValue;
+		usbi_dbg(ctx, "found %u configurations (current config: %u) for device '%s'", 
+			dev->device_descriptor.bNumConfigurations, priv->active_config,	priv->dev_id);
```
tormodvolden, line 1174: You have simplified the debug output to not print idVendor and idProduct, so do you think these are not useful, or even possibly misleading? I just noticed because I am bit frustrated by how little I get out of the current debug output from the Windows enumeration, and I was thinking of adding these fields more consistently, then remembered you were changing this part of the code and now I see that you are removing them. We can discuss further elsewhere how we can improve the debug output...

sonatique: For this particular case: I think removal of VID/PID was not intentional: I reused the second usbi_dbg string (that was on line 1190) instead of the one for the error case. I will change this to re-add VID/PID on the debug line.
I think it is always good to have VID/PID (even if we can match it using devid with another line of debug output) because it greatly help when debugging.
Thanks.

tormodvolden: Yes, consistency is good, and I also see that the VID/PID always appears in following output so it is not needed here.

Absolutely, enhancing debug output should be another merge request (or at least separate commits). I prefer that commits with functional changes like this one are kept as small and simple as possible to make reviewing easier, and other non-critical enhancements that are not directly prerequisites are done separately.

tormodvolden:
Merged, thanks for your work on this.

PR #1279, opened 2023-05-11
Fix advertisement of detach kernel driver capability in sunos backend

braewoods:
The code in the backend already supports this feature but fails to disclose it properly so application code may not be aware this is even available at times. This adds the corresponding capability flag so detach kernel driver support is properly advertised.


mcuee:
Just curios what kind of Solaris based system support this feature. Thanks.

Issue #1280, opened 2023-05-16
[WinUSB] Incorrect handling of composite devices with more than 9 interfaces


afeinman-snap:
https://github.com/libusb/libusb/blob/07441f54244991af55df158b61fd69ca95b39662/libusb/os/windows_winusb.c#L1390-L1391
The logic above fails to account for the fact that the MI_XX is a hex number. So on a device with 10 interfaces it is MI_0A. As a result the interface is being ignored.

mcuee:
Just curious what is this device which has more than 9 interfaces. It will be great to post the USB descriptors for such device. Thanks.

mcuee:
The issue comes from pyocd.
* https://github.com/pyocd/pyOCD/issues/1556

afeinman-snap:
> The issue comes from pyocd.
> 
> * [Unable to force cmsis-dap v1 when flashing pyocd/pyOCD#1556](https://github.com/pyocd/pyOCD/issues/1556)

The issue was discovered via pyocd by me and initially blamed on pyocd.

afeinman-snap:
[usbdesc.txt](https://github.com/libusb/libusb/files/11493079/usbdesc.txt)
USB descriptor details. This is a custom CMSIS-DAP -based debug interface extended to provide 4 virtual UARTs. 2 interfaces per UART + 1 CMSIS-DAP v1 + 1 CMSIS-DAP v2. 
Regardless, the logic in the quoted code is wrong. For the moment I have replaced it with 
```
if (mi_str != NULL) {
   interface_number = strtoul(mi_str + 3, NULL, 16);
```

mcuee:
@afeinman-snap 

Thanks. Please help to create a PR to address the issue.

afeinman-snap:
Done: https://github.com/libusb/libusb/pull/1281

PR #1281, opened 2023-05-17
Fix enumeration of complex composite devices on Windows

afeinman-snap:
On Windows when a composite device has multiple interfaces, they are addressed as VID_XXXX&PID_YYYY&MI_WZ. Windows support layer in libusb currently uses naive conversion (W - '0') * 10 + (Z-'0'). However the WZ is actually hex-encoded. This results in a failure enumerating interfaces above 9 (e.g. MI_0A).

This patch addresses the above inconsistency using strtoul to convert the index. It fixes [this issue](https://github.com/libusb/libusb/issues/1280)

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -29,7 +29,6 @@
 #include <setupapi.h>
 #include <ctype.h>
 #include <stdio.h>
-#include <errno.h>
 
 #include "libusbi.h"
 #include "windows_winusb.h"
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1381,6 +1381,7 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
 	int interface_number;
 	const char *mi_str;
 	int iadi, iadintfi;
+	char* endptr;
 	struct libusb_interface_association_descriptor_array *iad_array;
 	const struct libusb_interface_association_descriptor *iad;
 
@@ -1388,14 +1389,15 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
 	// devices will have only MI_00 & MI_03 for instance), we retrieve the actual
 	// interface number from the path's MI value
 	mi_str = strstr(device_id, "MI_");
-        // Initialize to default
-        interface_number = 0;
-		errno = 0;
-		char* endptr = NULL;
+
+	endptr = NULL;
+	interface_number = 0;
+	
 	if (mi_str != NULL) {
 		interface_number = strtoul(&mi_str[3], &endptr, 16);
 	}
-    if (mi_str == NULL || errno != 0 || endptr - &mi_str[3] != 2) {
+
+	if (mi_str == NULL || endptr - &mi_str[3] != 2) {
 		usbi_warn(ctx, "failure to read interface number for %s, using default value", device_id);
 		interface_number = 0;
 	}
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1390,7 +1390,8 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
 	mi_str = strstr(device_id, "MI_");
 
 	endptr = NULL;
-	interface_number = 0;
+	// This initialization, while redundant, is needed to make MSVC happy
+	interface_number = -1;
 	
 	if (mi_str != NULL) {
 		interface_number = strtoul(&mi_str[3], &endptr, 16);
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1390,11 +1390,14 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
 	mi_str = strstr(device_id, "MI_");
         // Initialize to default
         interface_number = 0;
+		errno = 0;
+		char* endptr = NULL;
 	if (mi_str != NULL) {
-		interface_number = strtoul(&mi_str[3], NULL, 16);
+		interface_number = strtoul(&mi_str[3], &endptr, 16);
 	}
-        if (mi_str == NULL || errno != 0) {
+    if (mi_str == NULL || errno != 0 || endptr - &mi_str[3] != 2) {
 		usbi_warn(ctx, "failure to read interface number for %s, using default value", device_id);
+		interface_number = 0;
 	}
 
 	if (interface_number >= USB_MAXINTERFACES) {
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -29,6 +29,7 @@
 #include <setupapi.h>
 #include <ctype.h>
 #include <stdio.h>
+#include <errno.h>
 
 #include "libusbi.h"
 #include "windows_winusb.h"
@@ -1387,11 +1388,13 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
 	// devices will have only MI_00 & MI_03 for instance), we retrieve the actual
 	// interface number from the path's MI value
 	mi_str = strstr(device_id, "MI_");
-	if ((mi_str != NULL) && isdigit((unsigned char)mi_str[3]) && isdigit((unsigned char)mi_str[4])) {
-		interface_number = ((mi_str[3] - '0') * 10) + (mi_str[4] - '0');
-	} else {
+        // Initialize to default
+        interface_number = 0;
+	if (mi_str != NULL) {
+		interface_number = strtoul(&mi_str[3], NULL, 16);
+	}
+        if (mi_str == NULL || errno != 0) {
 		usbi_warn(ctx, "failure to read interface number for %s, using default value", device_id);
-		interface_number = 0;
 	}
 
 	if (interface_number >= USB_MAXINTERFACES) {


libusb/os/windows_winusb.c
```diff
@@ -1387,11 +1388,13 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
...
 	mi_str = strstr(device_id, "MI_");
-	if ((mi_str != NULL) && isdigit((unsigned char)mi_str[3]) && isdigit((unsigned char)mi_str[4])) {
-		interface_number = ((mi_str[3] - '0') * 10) + (mi_str[4] - '0');
-	} else {
+        // Initialize to default
+        interface_number = 0;
+	if (mi_str != NULL) {
+		interface_number = strtoul(&mi_str[3], NULL, 16);
+	}
+        if (mi_str == NULL || errno != 0) {
```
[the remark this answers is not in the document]
Youw: > provide a pointer instead of NULL and check that the value is mit_str + 5.

I always recommend this way.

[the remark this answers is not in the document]
afeinman-snap: Addressed

mcuee:
Interestingly I do not see the limit of number of interfaces an USB device can have in USB 2.0 spec Section 9.6.5. It uses "Number" without saying the limit.
https://www.usb.org/document-library/usb-20-specification
Capture

But then I can see Windows uses UCHAR. So we should be fine here.
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/usbspec/ns-usbspec-_usb_interface_descriptor
```
typedef struct _USB_INTERFACE_DESCRIPTOR {
  UCHAR bLength;
  UCHAR bDescriptorType;
  UCHAR bInterfaceNumber;
  UCHAR bAlternateSetting;
  UCHAR bNumEndpoints;
  UCHAR bInterfaceClass;
  UCHAR bInterfaceSubClass;
  UCHAR bInterfaceProtocol;
  UCHAR iInterface;
} USB_INTERFACE_DESCRIPTOR, *PUSB_INTERFACE_DESCRIPTOR;
```

mcuee:
What will happen if the devices have 256 interfaces? Doe the PR work in that case?

It is actually trivial to add an USB interfaces (just the Control Endpoint).

afeinman-snap:
> What will happen if the devices have 256 interfaces? Doe the PR work in that case?
> 
> It is actually trivial to add an USB interfaces (just the Control Endpoint).

The PR will not work, because of the check for interface number being 2 hex digits. However this is a ridiculous corner case and I expect Windows to crap out first. Since their interface descriptor is UCHAR and this is Windows-only branch of code, we are ok

mcuee:
> > What will happen if the devices have 256 interfaces? Doe the PR work in that case?
> > It is actually trivial to add an USB interfaces (just the Control Endpoint).
> 
> The PR will not work, because of the check for interface number being 2 hex digits. However this is a ridiculous corner case and I expect Windows to crap out first. Since their interface descriptor is UCHAR and this is Windows-only branch of code, we are ok

What about 255 interfaces?

afeinman-snap:
> > > What will happen if the devices have 256 interfaces? Doe the PR work in that case?
> > > It is actually trivial to add an USB interfaces (just the Control Endpoint).
> > 
> > 
> > The PR will not work, because of the check for interface number being 2 hex digits. However this is a ridiculous corner case and I expect Windows to crap out first. Since their interface descriptor is UCHAR and this is Windows-only branch of code, we are ok
> 
> What about 255 interfaces?

What about it? The parsing code will succeed. The rest of libusb - I have no idea

mcuee:
> > What about 255 interfaces?
> 
> What about it? The parsing code will succeed. The rest of libusb - I have no idea

Thanks for the confirmation. I am good with this PR now.

libusb/os/windows_winusb.c
```diff
@@ -1387,9 +1388,14 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
 	// devices will have only MI_00 & MI_03 for instance), we retrieve the actual
 	// interface number from the path's MI value
 	mi_str = strstr(device_id, "MI_");
-	if ((mi_str != NULL) && isdigit((unsigned char)mi_str[3]) && isdigit((unsigned char)mi_str[4])) {
-		interface_number = ((mi_str[3] - '0') * 10) + (mi_str[4] - '0');
-	} else {
+        // Initialize to default
+        interface_number = 0;
```
tormodvolden, line 1392: Don't think you need to initialize this one, because there should always be a MI_xx field, right?

libusb/os/windows_winusb.c
```diff
@@ -1387,9 +1388,14 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
...
+		errno = 0;
+		char* endptr = NULL;
```
tormodvolden, line 1394: Looks like whitespace error.

libusb/os/windows_winusb.c
```diff
@@ -1387,9 +1388,14 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
...
+	if (mi_str != NULL) {
+		interface_number = strtoul(&mi_str[3], &endptr, 16);
+	}
+    if (mi_str == NULL || errno != 0 || endptr - &mi_str[3] != 2) {
```
tormodvolden, line 1398: Since in our case we always want to see 2 characters converted and there is no possibility of overflow value, if you check the end pointer, there is no need to additionally check errno.

tormodvolden:
> What will happen if the devices have 256 interfaces? Doe the PR work in that case?

Note that the  bNumInterfaces field in the configuration descriptor is a single byte, so 255 is the hard limit for USB.

mcuee:
> > What will happen if the devices have 256 interfaces? Doe the PR work in that case?
> 
> Note that the bNumInterfaces field in the configuration descriptor is a single byte, so 255 is the hard limit for USB.

It is a hard limit for Windows implementation (I have not checked Linux and macOS), but not USB 2.0 Specification.

mcuee:
But of course this is just a theoretical question. I do not think there are real practical use of devices with 256 interfaces or more...

mcuee:
Quick glance for Linux implementation.
https://github.com/torvalds/linux/blob/4d6d4c7f541d7027beed4fb86eb2c451bd8d6fff/drivers/usb/core/usb.c#L285-L299

```
/**
 * usb_find_alt_setting() - Given a configuration, find the alternate setting
 * for the given interface.
 * @config: the configuration to search (not necessarily the current config).
 * @iface_num: interface number to search in
 * @alt_num: alternate interface setting number to search for.
 *
 * Search the configuration's interface cache for the given alt setting.
 *
 * Return: The alternate setting, if found. %NULL otherwise.
 */
struct usb_host_interface *usb_find_alt_setting(
		struct usb_host_config *config,
		unsigned int iface_num,
		unsigned int alt_num)
```

mcuee:
> > It is a hard limit for Windows implementation (I have not checked Linux and macOS), but not USB 2.0 Specification.
> 
> No, it is not about host software. It is about the bNumInterfaces of the configuration descriptor of any USB device. A device cannot declare more than 255 interfaces because this is a single byte in the descriptor.

I know this is a theoretical issue and not affecting this PD. But please refer to USB 2.0 spec Section 9.6.5. It uses "Number" without saying the limit. It never mentions that is is a single byte.
https://www.usb.org/document-library/usb-20-specification
Capture

tormodvolden:
It says "Size" = 1 for that field. Offset is 4, next field has offset 5.

mcuee:
> It says "Size" = 1 for that field. Offset is 4, next field has offset 5.
I have not found that the unit of Size is BYTE yet. Maybe I did not look deep enough.

afeinman-snap:
Also, there is this:
https://github.com/libusb/libusb/blob/07441f54244991af55df158b61fd69ca95b39662/libusb/libusbi.h#L138

libusb/os/windows_winusb.c
afeinman-snap: Unfortunately, MSVC appears to be not smart enough to figure out that the following will initialize the variable one way or another:
```C
	if (mi_str != NULL) {
		interface_number = strtoul(&mi_str[3], &endptr, 16);
	}

        if (mi_str == NULL || endptr - &mi_str[3] != 2) {
		usbi_warn(ctx, "failure to read interface number for %s, using default value", device_id);
		interface_number = 0;
	}
```
It throws _uninitialized variable_ warning treated as error

libusb/os/windows_winusb.c
afeinman-snap: Fixed

libusb/os/windows_winusb.c
afeinman-snap: Removed

libusb/os/windows_winusb.c
```diff
@@ -29,6 +29,7 @@
 #include <setupapi.h>
 #include <ctype.h>
 #include <stdio.h>
+#include <errno.h>
```
Youw, line 32: I don't believe it is needed
```suggestion
```

Youw (approved):
LGTM

libusb/os/windows_winusb.c
tormodvolden: Okay :) Same for endptr I guess.

libusb/os/windows_winusb.c
tormodvolden: Yes, please don't add it unless needed.

libusb/os/windows_winusb.c
```diff
@@ -1380,16 +1381,23 @@ static int set_composite_interface(struct libusb_context *ctx, struct libusb_dev
...
 	// Because MI_## are not necessarily in sequential order (some composite
 	// devices will have only MI_00 & MI_03 for instance), we retrieve the actual
 	// interface number from the path's MI value
 	mi_str = strstr(device_id, "MI_");
-	if ((mi_str != NULL) && isdigit((unsigned char)mi_str[3]) && isdigit((unsigned char)mi_str[4])) {
-		interface_number = ((mi_str[3] - '0') * 10) + (mi_str[4] - '0');
-	} else {
+
+	endptr = NULL;
+	interface_number = 0;
```
tormodvolden, line 1394: Since this is only for the compiler and not for setting a default value, I think it would be best to initialize it to e.g. -1 here, just to make a point to the human reader.

mcuee:
> Also, there is this:
> 
> https://github.com/libusb/libusb/blob/07441f54244991af55df158b61fd69ca95b39662/libusb/libusbi.h#L138

Nice finding.

I wonder why we limited the number of interfaces to 32 though.
```
#define USB_MAXENDPOINTS	32 // as per Standard
#define USB_MAXINTERFACES	32 // articifical limit --> can be up to 255 as per standard
#define USB_MAXCONFIG		8  // another artificial limit --> but take note WinUSB does not support changing the default configuration
```

Anyway this code was from the very beginning and nobody complained, so it is not a real issue.
https://github.com/libusb/libusb/blob/77cea822788e024c848c7f554915f771a2dc1e0f/libusb/libusbi.h

libusb/os/windows_winusb.c
afeinman-snap: What point would that be?

tormodvolden: That this is an initialization value that won't be carried over, and is there only to make the compiler happy. Otherwise someone may smell redundancy and be tempted to remove it. I would have added a comment too.

afeinman-snap: Fair enough. Although removing it will instantly result in a failed build, so it's not like it is a hidden bug. Changed/commented

mcuee:
> > I wonder why we limited the number of interfaces to 32 though.
> 
> There are some arrays that are sized by this value, and some code paths are iterating through all values. So unless there actually are devices with more interfaces than that, bumping the number would unnecessarily slow down things and use more memory.

As the codes have been there from the very beginning and there are no complaints, I agree not to change.

libusb/os/windows_winusb.c
tormodvolden: Thanks! Yes, the CI would catch it. But gcc wouldn't complain, so it might not be seen immediately by everyone.

PR #1282, opened 2023-05-17
Adds missing files

top-master:
Added files that are missing from GitHub (but exist on SourceForge).

If these are intentionally missing, is there any way to generate them???


mcuee:
Hmm, those files are generated when you run bootstrap or autogen.

top-master:
Even the "compile" and "missing" named files?
(Maybe at least those are NOT auto generated, right?)

tormodvolden:
> Added files that are missing from GitHub (but exist on SourceForge).

What do you exactly mean by exist on SourceForge? Such files should indeed exist in the release tarballs, but not in the git repository. As mcuee said, they are autogenerated.

PR #1284, opened 2023-05-18
core: Avoid crash due to premature va_end

### LINKED ISSUE #1274: usbi_backend.set_option(...) is called with invalid va_list
Hello,

I have compiled the libusb with MSVC2019 and found that in the function _libusb_set_option(...)_ the access to _va_list ap_ is terminated too early. This leads to the fact that in the cases LIBUSB_OPTION_USE_USBDK, LIBUSB_OPTION_NO_DEVICE_DISCOVERY and LIBUSB_OPTION_WINUSB_RAW_IO when calling _usbi_backend.set_option(..., ap)_ the already deleted ap-structure is passed. This leads to a crash if later in the backend (e.g. winusb_set_option(..., ap)) a nullptr is accessed via va_arg(ap, ...).

tormodvolden:
After va_end(ap) the variable ap is undefined and cannot be accessed.

Fixes #1274

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2312,58 +2312,65 @@ int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
 		log_cb = (libusb_log_cb) va_arg(ap, libusb_log_cb);
 	}
 
-	if (LIBUSB_SUCCESS != r) {
-		return r;
-	}
-
-	if (option >= LIBUSB_OPTION_MAX) {
-		return LIBUSB_ERROR_INVALID_PARAM;
-	}
-
-	if (NULL == ctx) {
-		usbi_mutex_static_lock(&default_context_lock);
-		default_context_options[option].is_set = 1;
-		if (LIBUSB_OPTION_LOG_LEVEL == option) {
-			default_context_options[option].arg.ival = arg;
-		} else if (LIBUSB_OPTION_LOG_CB) {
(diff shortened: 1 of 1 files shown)


libusb/core.c
```diff
@@ -2362,6 +2361,7 @@ int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
 	default:
 		return LIBUSB_ERROR_INVALID_PARAM;
 	}
+	va_end(ap);
```
Youw, line 2364: there is multiple `return` statements before `va_end` and after `va_start`, and I believe the C standard states, that if the function that invoked `va_start` doesn't invoke `va_end` before return - the behavior is undefined.

Youw: effectively it makes this fix a lot more complicated

tormodvolden: Thanks for the review. That is a good point. The man page doesn't warn about it, but I can imagine possible memory leaks.

tormodvolden: I have attempted a do loop to make single point of return, the diff is best viewed with git diff -w (without whitespace since I had to indent everything). The alternative would be a lot of "else if" and it would be harder to get it right and difficult to read.

Youw: that should work

Youw (approved):

Issue #1287, opened 2023-05-31
libusb_free_device_list(ptr, 1) does not unref all devices when the device list has a null entry 


diablodale:
`libusb_free_device_list(ptr, 1)` does not handle the scenario of having `null` entries in the list. It will errantly stop its unref walk of the list thinking it has reached the terminating `null`.

### Setup

* all os and platforms
* libusb at `master`

### Repro

Found in code review as it is hard to repro.

`libusb_get_device_list()` returns both a `count` and a `libusb_device**` which points to the allocated list of devices `libusb_device*`.
The `count` is the allocated `listSize - 1`. The list is null terminated as per libusb docs.
In some scenarios, it is possible for the allocated list to have `null` entries.

`libusb_free_device_list(list, 1)` accepts the pointer to the allocated list of devices. No count parameter.
This api's code does not handle the scenario of having `null` entries in the list. It will errantly stop its walk of the list thinking it has reached the terminating `null`.

https://github.com/libusb/libusb/blob/cc498ded18fb2c6e4506c546d0351c4ae91ef2cc/libusb/core.c#L895-L901

### Workaround

Don't use `libusb_free_device_list(ptr, 1)`. Instead, write custom code to walk the known `count` size of the list, unref devices that are not null, and then free the list without the auto-unref. Pseudocode...
```c++
for (size_type i = 0; i < countDevices; ++i) {
    if (deviceList[i] != nullptr)
        libusb_unref_device(deviceList[i]);
}
libusb_free_device_list(deviceList, 0);
```

Youw:
I do not believe it is possible to fix this w/o introducing something like `libusb_free_device_list_n(libusb_device ** list, size_t device_count, int unref_devices)` and it is up to the core team to decide is this is needed at all.

The real question is: how do you come up with `NULL` devices in between your list? I believe the second parameter in the `libusb_free_device_list` is designed to include cases like this - when a user code decides to modify the list itself.

mcuee:
> The api libusb_get_device_list() makes no guarantee that every pointer it returns will be non-null. 

I believe it will not contain a null entry by definition. In a system libusb is expected to work, there will be USB devices (eg: at least the root hub).

As far as I know, a lot of libusb APIs expect you to call them in pair.

[libusb_get_device_list()](https://libusb.sourceforge.io/api-1.0/group__libusb__dev.html#gac0fe4b65914c5ed036e6cbec61cb0b97)
[...]

PR #1289, opened 2023-06-09
clarify doc that libusb_device* in lists are non-NULL

diablodale:
Doc update to clarify that the lists get/free by libusb are null-terminated and will have no NULL pointers within their list.

- fixes libusb/libusb#1287

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -806,9 +806,9 @@ struct libusb_device *usbi_get_device_by_session_id(struct libusb_context *ctx,
  * not to unreference a device you are about to open until after you have
  * opened it.
  *
- * This return value of this function indicates the number of devices in
- * the resultant list. The list is actually one element larger, as it is
- * NULL-terminated.
+ * The return value of this function indicates the number of libusb_device
+ * pointers in the contiguous non-NULL resultant list. The list is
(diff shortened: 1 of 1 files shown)


mcuee:
@tormodvolden 

I believe this PR is correct for the platforms where libusb is expected to work (libusb_init will not fail). Please check as well. Thanks.

tormodvolden:
The original text says "The list [...] is NULL-terminated.". A null-terminated list doesn't have NULL elements, per definition. I think the suggested text would just add confusion.

diablodale:
When libusb functions perfectly always and never writes a null into any place in the list except at the "end", and the contiguous list of pointers is not interntionally/accidentially manipulated by a user, then it is true that a contiguous non-null list terminated by a null would work correctly.

Unfortuantunately, libusb isn't perfect, users manipulate lists, etc. And so using this API will unsuccessfully and prematurely release only a subset of the originally ref'd devices.

Personally, I check for null and don't use the "1" option so that I correctly and completely free all devices. This was purely for the community so that the documentation was updated to be clear and declare its dependency on a non-null contiguous list.
As this PR was closed, the doc won't be as clear. But this PR and the related issue will serve as a bridge for those that do searches. Cheers.

PR #1295, opened 2023-07-06
BSD style and minor debug fixes

tormodvolden:
Not of greatest importance, but I put it out here and it would be nice if someone familiar with *BSD codestyle comments on it. @efuss @kucharskim @botovq anyone?


botovq:
On Thu, Jul 06, 2023 at 03:36:02AM -0700, Tormod Volden wrote:
> Not of greatest importance, but I put it out here and it would be nice if someone familiar with *BSD codestyle comments on it. @efuss @kucharskim @botovq anyone?

In OpenBSD we tend not to use parentheses after return in new code. An
example in https://man.openbsd.org/style.9 was changed to have the
parentheses removed since that's the consensus among devs.

I would definitely also remove the extra parentheses around the ioctl
calls. The change to hexadecimal in format strings looks sensible as
well.

If it helps you by all means go for it.

efuss:
I haven't looked into that in great detail, just some quick comments:

>   * netbsd/openbsd: Drop parentheses around return values
Looks reasonable.

>   * netbsd/openbsd: Remove extraneous parentheses around ioctl call
The old code looks totally weird.

>   * netbsd: Prefix debug hex values with 0x
You don't prefix them, you print %x instead of %d. That my well make sense, just the description of the change looks incomplete.

>   * netbsd: Debug print all errors with code
I can't tell whether that one makes sense. It will call `strerror()` for every call to `_errno_to_libusb()` (not only for `LIBUSB_ERROR_OTHER`) and may flood the debug output. I haven't looked into what other implementations do.

If you apply that change, I would move the `return LIBUSB_ERROR_OTHER` into a `default` clause of the `switch` statement.

kucharskim:
All commits not explicitly mentioned in my comment make sense to me. I didn't compile just read the pull request here in GitHub.

Per @efuss maybe commit related to `"type 0x%x request 0x%x value 0x%x index %d length %d timeout %d"` could be reworded, but change itself of that commit make sense to me.

The `usbi_dbg(NULL, "erro...` change I am not sure what implications it has to move it earlier in the file / to the top of the `_errno_to_libusb()` function. Adding `%d` for `err` I like, moving earlier - I cannot judge.

tormodvolden:
Thanks for the quick and good feedback, guys, you are the best!

> You don't prefix them, you print %x instead of %d. That my well make sense, just the description of the change looks incomplete.

Indeed, must have brain-leaked from the referenced commit. I have fixed the description.

> I can't tell whether that one makes sense. It will call `strerror()` for every call to `_errno_to_libusb()` (not only for `LIBUSB_ERROR_OTHER`) and may flood the debug output. I haven't looked into what other implementations do.

The openbsd version does this, that's why I suggested doing this for netbsd as well. But it should probably wait until someone can test it for us. And if applied, add the "default" clause (thanks Edwin) - that could need changing for openbsd then as well. I have split out these parts.

tormodvolden:
After all this (and a temporary s/obsd/netbsd/g) the *bsd_usb.c diff is rather overseeable. What is left in _errno_to_libusb is:
```
@@ -493,6 +455,7 @@ _errno_to_libusb(int err)
 	case ENOMEM:
 		return LIBUSB_ERROR_NO_MEM;
+	case EWOULDBLOCK:
 	case ETIMEDOUT:
 		return LIBUSB_ERROR_TIMEOUT;
 	default:
```
Would it be correct to add EWOULDBLOCK for NetBSD as well?

efuss:
> Would it be correct to add EWOULDBLOCK for NetBSD as well?
I don't know, I didn't write that code.

tormodvolden:
I merged the four first and will rebase the PR to keep the last two pending.

seanm (approved):

PR #1298, opened 2023-07-16
Add 20Gbps speed support for Linux and macOS backend.

naruto-abc:
As USB 3.2 devices come into market, a new speed needs to be supported.
There are three commits here in order to do the minimal things in each one.
One adds a new speed called LIBUSB_SPEED_SUPER_PLUS_BY2.
One adds support for macOS backend and the third one adds support for Linux backend.

### DIFF
--- a/android/examples/unrooted_android.c
+++ b/android/examples/unrooted_android.c
@@ -205,12 +205,13 @@ static void print_device(libusb_device *dev, libusb_device_handle *handle)
     uint8_t i;
 
     switch (libusb_get_device_speed(dev)) {
-        case LIBUSB_SPEED_LOW:		speed = "1.5M"; break;
-        case LIBUSB_SPEED_FULL:		speed = "12M"; break;
-        case LIBUSB_SPEED_HIGH:		speed = "480M"; break;
-        case LIBUSB_SPEED_SUPER:	speed = "5G"; break;
-        case LIBUSB_SPEED_SUPER_PLUS:	speed = "10G"; break;
-        default:			speed = "Unknown";
+        case LIBUSB_SPEED_LOW:			speed = "1.5M"; break;
+        case LIBUSB_SPEED_FULL:			speed = "12M"; break;
+        case LIBUSB_SPEED_HIGH:			speed = "480M"; break;
+        case LIBUSB_SPEED_SUPER:		speed = "5G"; break;
+        case LIBUSB_SPEED_SUPER_PLUS:		speed = "10G"; break;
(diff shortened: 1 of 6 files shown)


mcuee:
Just wondering what is the main difference between this one and the previous PR #680.
* https://github.com/libusb/libusb/pull/680

naruto-abc:
Hi mcuee,

Just checked PR #680, which involve more technical details like rx and tx lanes. USB3.2  Gen2x2 is actually 20Gbps. For me, I care more about the speed itself instead of lanes and try to get it directly from OS-depedent backends, which can make things a lot easier. And luckily, as in my changes, macOS and Linux both have supported this.

tormodvolden:
> Latest Linux kernel has supported value 20000 via sysfs attribute file 'speed'.

I wonder, is that a kernel change which has made the more complicated stuff in PR 680 obsolete?

Since which  Linux kernel version is this value available?

> But this speed is still not supported based on kernel implementation of ioctl USBDEVFS_GET_SPEED, which is the counterpart of ioctl IOCTL_USBFS_GET_SPEED in libusb. So nothing is changed in usbfs_get_speed().

So do you mean, this is actually not working yet?

naruto-abc:
Hi Tormod,

```
is that a kernel change which has made the more complicated stuff in PR 680 obsolete?
``` 
Emm...To be honest ,I don't know the answer to this question. My point is that, if we just want to get speed value instead of more technical details, PR #680 seems a little bit too heavy.  Why not let OSes handle the details for us?

```
Since which Linux kernel version is this value available?
```
As I can know, starting from v5.13, 20000 is supported, https://elixir.bootlin.com/linux/v5.13/source/drivers/usb/core/sysfs.c#L169

```
So do you mean, this is actually not working yet?
```
Yes, the kernel side implementation can only return at most USB_SPEED_SUPER_PLUS(10Gbps) via ioctl USBDEVFS_GET_SPEED, based on what I understand now.

naruto-abc:
I'm also think about letting ioctl USBDEVFS_GET_SPEED support 20Gbps first, then we can fix it accordingly in libusb.

mcuee:
Related discussion in linux-usb mailing list here.
https://lore.kernel.org/linux-usb/da536c80-7398-dae0-a22c-16e521be697a@suse.com/T/#t

fabiensanglard:
It seems this PR can only go in once the Linux kernel patch is accepted. Is Dingyan Li still working on it?

naruto-abc:
Unfortunately, after rounds of discussion, it seems to me that there is no perfect solution to add the corresponding part for Linux kernel. Therefore I decide to drop it. I think we can close this PR too.

PR #1309, opened 2023-09-18
Wrap ioctl with EINTR loop to handle spurious interrupt

addisoncrump:
Somewhat related to #1308: found another issue where ioctl was infrequently setting errno to EINTR, which caused some local tests to be flaky. Not sure the exact conditions of this error, but this is the patch I applied locally.

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -167,6 +167,20 @@ struct linux_transfer_priv {
 	int iso_packet_offset;
 };
 
+#define MAX_IOCTL_TRIES 16
+
+/* try ioctl up to MAX_IOCTL_TRIES times, retrying if r < 0 and errno is set to EINTR */
+static inline int ioctl_retry(int fd, unsigned long int request, void *extra) {
+	int r;
+	for (unsigned int i = 0; i < MAX_IOCTL_TRIES; i++) {
+		r = ioctl(fd, request, extra);
+		if (r >= 0 || errno != EINTR) {
+			break;
+		}
+	}
+	return r;
+}
+
 static int dev_has_config0(struct libusb_device *dev)
 {
 	struct linux_device_priv *priv = usbi_get_device_priv(dev);
@@ -852,7 +866,7 @@ static int usbfs_get_active_config(struct libusb_device *dev, int fd)
 		.data = &active_config
 	};
 
-	r = ioctl(fd, IOCTL_USBFS_CONTROL, &ctrl);
+	r = ioctl_retry(fd, IOCTL_USBFS_CONTROL, &ctrl);
 	if (r < 0) {
 		if (errno == ENODEV)
 			return LIBUSB_ERROR_NO_DEVICE;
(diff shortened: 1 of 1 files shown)


Youw:
`EINTR` may be caused by a number of things, e.g. a debugger (gdb) attached

but I believe Ctrl+C generates `EINTR` as well, so having an infinite loop like that may fail the application to shutdown by a Ctrl+C from terminal in fase if the defice hang on that call

at minimum I'd suggest to add a limit of retries

ultimately I'd suggest to find our what exactly is causing this, and if my assumption is correct - the debugger is to blame, in which case I'd make a number of retries only if debugger is attached (there're a number of ways to detect if the debugger is present)

addisoncrump:
This occurred outside of a debugger, and my tests here are extensive.

The only part of my source code which I think may be relevant is the presence of timers and signal handlers for those timers, which may intermittently interrupt the syscall. This feels like a fairly common use case.

Without explicitly handling CTRL-C, the program will simply terminate, it will not enter an infinite loop. Even in the presence of a SIGINT handler, the EINTR would only happen once per SIGINT, so I don't think this is a realistic scenario. I don't believe there are any cases where EINTR can be issued infinitely from ioctl without explicitly trying to do so :sweat_smile:

Youw:
I'd still limit the number of retries to something reasonable (10?)

addisoncrump:
Need to add the tries limiter and apply this to the other ioctl calls in linux_usbfs.c.

Youw:
> other ioctl calls

Make a wrapper funciton (a macro?) to avoid heavy copy-pasting

addisoncrump:
Sadly, neither a macro nor a function addresses the use cases seen elsewhere in the file. Can't evaluate a `do {} while (0)` macro as an expression, and there's no va_args friendly version of ioctl!

tormodvolden:
In your case, what is the ioctl doing?

addisoncrump:
> In your case, what is the ioctl doing?

I encountered this issue while testing with umockdev in the presence of timers. However, I note also from the [manual pages](https://linux.die.net/man/3/ioctl):

> Under the following general conditions, ioctl() shall fail if:
> ...
> EINTR
>    A signal was caught during the ioctl() operation.

I expect this to occur in real scenarios as well. Perhaps also in different architectures, though I do not have the means to test this.

addisoncrump:
Any updates on this?

tormodvolden:
We're in a release freeze, I will come back to this afterwards.

tormodvolden:
Would using SA_RESTART for your signal handlers be another solution? https://stackoverflow.com/a/59795677

mcuee:
@addisoncrump 

Just wondering if you want to further persue this PR. Thanks.

addisoncrump:
To be honest, I will likely not pursue this further. I found this with some fuzzing ~2.5 years ago and I don't think I have the time in the near future to work on this.

I'll happily release any copyright/authorship for this and my other PR if anyone else wants to follow up.

Issue #1307, opened 2023-09-12
libusb_open() returns LIBUSB_ERROR_NOT_FOUND


pazourek:
I am using a full-speed device with own firmware, 0x2846:1780. The device is a composite device with 4 simplified CDC-ACM and one DFU interface. I implemented WCID (Extended compat ID only) that automatically installs WinUSB driver for all device interfaces. That works for Win10 as well for Win11. I can see WinUSB of version 10.0.22621.608 is installed. 

I am able to read a device descriptor calling the libusb_get_device_descriptor(), but every attempt to open the device (libusb_open()) fails with an error LIBUSB_ERROR_NOT_FOUND (-5). I have also tried to disable WCID and install custom drivers that left the CDC-ACM interface to use default driver (usbser.sys), the DFU interface is used with libusb0 driver. Neither this configuration worked as expected. Version of Libusb is 1.0.26.11724, both x86 and amd64 binaries behave the same way.

Other devices using the WinUSB seems to work well.

[230912_WCID_beagle.txt](https://github.com/libusb/libusb/files/12586484/230912_WCID_beagle.txt)
[230912_WCID_open.txt](https://github.com/libusb/libusb/files/12586485/230912_WCID_open.txt)
[230912_WCID_treeview.txt](https://github.com/libusb/libusb/files/12586486/230912_WCID_treeview.txt)

pazourek:
I also tried to modify registry records (_Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Enum\USB\VID_2846&PID_1780\01\Service_). I replaced the usbccgp driver by the WinUSB, so the WinUSB became be used on the top level parent node. This fixed the issue, now I can open the device. But it seem that WCID is not able to do such assignation. So using separate interfaces for automatic driver installation remains the only way.

mcuee:
I tend to think this is a duplicate of https://github.com/libusb/libusb/issues/1238.
* https://github.com/libusb/libusb/issues/1238

pazourek:
Digging around this issue I found out, the libusb is able to open any WinUsb interface of the composite device only if the device driver provides GUID. This also works with the automatic driver installation (WCID) if the device implements "Extended Properties OS Feature Descriptor" with one GUID. If the device reports more GUIDs, opening the device fails with the error LIBUSB_ERROR_NOT_FOUND (-5). 

[...]

PR #1322, opened 2023-10-11
windows: Fix getting GUID from REG_MULTI_SZ string

pazourek:
A composite device with four WinUSB interfaces, the device implements "Extended Properties OS Feature Descriptor" with multiple GUIDs. An attempt to open the device fails with an error LIBUSB_ERROR_NOT_FOUND (-5).

This commit allows getting REG_MULTI_SZ string with up to 5 GUIDs, only the first GUID is taken;

It solves #1307.


tormodvolden:
What does "only the first GUID is taken" mean? Should we ideally add all of the GUIDs to the guid list?

libusb/os/windows_winusb.c
```diff
@@ -1689,8 +1689,8 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 				// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks
 				// - "string may not have been stored with the proper terminating null characters"
 				// - "Note that REG_MULTI_SZ strings could have two terminating null characters"
-				if ((reg_type == REG_SZ && size >= sizeof(guid_string) - sizeof(char))
-				    || (reg_type == REG_MULTI_SZ && size >= sizeof(guid_string) - 2 * sizeof(char))) {
+				if ((reg_type == REG_SZ && size <= sizeof(guid_string) && guid_string[size-1] == 0)
+					|| (reg_type == REG_MULTI_SZ && size <= sizeof(guid_string) && guid_string[size-1] == 0 && guid_string[size-2] == 0)) {
```
tormodvolden, line 1693: We usually have spaces around arithmetic operators "size - 1". And when checking for null characters the character '\0' constant is used for readability.

pazourek:
> What does "only the first GUID is taken" mean? Should we ideally add all of the GUIDs to the guid list?

I am not a Win32 specialist,  but it seems that the GUID points to a whole device. When the winusb_get_device_list() searching for composite devices, it passes the GUID of the device to get_interface_details() function that enumerates in a loop all device interfaces.
If the device uses more GUIDs, passing the second (or n-th) GUID results in enumeration the same device interfaces.

The "only the first GUID is taken" mean, that the winusb_get_device_list() uses only the first GUID for listing all interfaces of the composite device.

libusb/os/windows_winusb.c
pazourek: resolved as you suggested

tormodvolden:
This is great! Thanks for walking that extra mile and refactoring it out of that nasty enumeration function, which now becomes more digestible. And the proper dynamic allocation. I have some review comments coming up later.

libusb/os/windows_winusb.c
```diff
@@ -1480,6 +1480,110 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+		} else {
+			usbi_warn(ctx, "unexpected error from pRegQueryValueExA for '%s'", dev_id);
+			err = LIBUSB_ERROR_ACCESS;
+			goto exit;
+		}
+	}
+
+	// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks
+	// - "string may not have been stored with the proper terminating null characters"
+	// - "Note that REG_MULTI_SZ strings could have two terminating null characters"
```
tormodvolden, line 1548: The comment suggests that it can happen that there is not two (or not even one?) terminating null character...

libusb/os/windows_winusb.c
```diff
@@ -1480,6 +1480,110 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+	if (reg_type == REG_SZ && size > 1 && guid_string[size - 1] == 0) {
+		guid = guid_string;
+	} else if (reg_type == REG_MULTI_SZ && size > 2 && guid_string[size - 1] == 0 && guid_string[size - 2] == 0) {
+		/* get the n-th guid indexed by guid_number since the DeviceInterfaceGUIDs may
```
tormodvolden, line 1552: ... and this doesn't allow for only one terminating null character any longer, right?

libusb/os/windows_winusb.c
```diff
@@ -1480,6 +1480,110 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+		   contain more GUIDs */
+		guid = guid_string;
+		while(guid_number && *guid) {
+			len = strlen(guid);
```
tormodvolden, line 1556: _If_ we would allow only one (or zero) terminating NUL (as of above comment), we'd need to check for overrun, something like: guid - guidstring < size

libusb/os/windows_winusb.c
```diff
@@ -1662,68 +1764,38 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
 						}
 						guid_list = new_guid_list;
 						guid_size += GUID_SIZE_STEP;
 					}
-					if_guid = malloc(sizeof(*if_guid));
-					if (if_guid == NULL) {
-						usbi_err(ctx, "failed to alloc if_guid");
-						LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
+
+					// Check if we've already seen this GUID
```
tormodvolden, line 1785: This check could ideally be done before potentially extending guid_list.

libusb/os/windows_winusb.c
```diff
@@ -1662,68 +1764,38 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
 					break;
-				}
-				// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks
-				// - "string may not have been stored with the proper terminating null characters"
-				// - "Note that REG_MULTI_SZ strings could have two terminating null characters"
-				if ((reg_type == REG_SZ && size >= sizeof(guid_string) - sizeof(char))
-				    || (reg_type == REG_MULTI_SZ && size >= sizeof(guid_string) - 2 * sizeof(char))) {
+				} else if(r == LIBUSB_ERROR_NO_MEM) {
+					LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
+				} else if(r == LIBUSB_SUCCESS) {
```
tormodvolden, line 1773: Suggestion (especially if below gets reworked), to avoid more indentation, check for != SUCCESS first.

libusb/os/windows_winusb.c
```diff
@@ -1480,6 +1480,110 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
 	return LIBUSB_SUCCESS;
 }
 
+static int get_guid(struct libusb_context *ctx, char *dev_id, HDEVINFO *dev_info, SP_DEVINFO_DATA *dev_info_data,
```
tormodvolden, line 1483: For future use, maybe return number of guids discovered on success?

tormodvolden: Do I understand correctly that dev_id is only used for the warning messages? I guess that is still worth it though.

libusb/os/windows_winusb.c
```diff
@@ -1480,6 +1480,110 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+	DWORD size, reg_type;
+	HKEY key;
+	char *guid_string, *new_guid_string;
+	char *guid;
+	LONG s;
+	size_t len;
+	int i;
+	int err = LIBUSB_SUCCESS;
+
+	if (!pSetupDiGetDeviceRegistryPropertyA(*dev_info, dev_info_data, SPDRP_DRIVER,
```
tormodvolden, line 1495: Potentially we could leave this check in the caller if is kind of independent? Or is it more "efficient" to have it here? Maybe at least duplicate here the comment in the caller about missing driver check.

libusb/os/windows_winusb.c
```diff
@@ -1480,6 +1480,110 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+		NULL, NULL, 0, NULL) && (GetLastError() != ERROR_INSUFFICIENT_BUFFER)) {
+		usbi_info(ctx, "The following device has no driver: '%s'", dev_id);
+		usbi_info(ctx, "libusb will not be able to access it");
+	}
+	// ...and to add the additional device interface GUIDs
```
tormodvolden, line 1500: On the same note, maybe leave this comment in the caller too.

libusb/os/windows_winusb.c
```diff
@@ -1662,68 +1764,38 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
-				}
-				// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks
-				// - "string may not have been stored with the proper terminating null characters"
-				// - "Note that REG_MULTI_SZ strings could have two terminating null characters"
-				if ((reg_type == REG_SZ && size >= sizeof(guid_string) - sizeof(char))
-				    || (reg_type == REG_MULTI_SZ && size >= sizeof(guid_string) - 2 * sizeof(char))) {
+				} else if(r == LIBUSB_ERROR_NO_MEM) {
+					LOOP_BREAK(LIBUSB_ERROR_NO_MEM);
+				} else if(r == LIBUSB_SUCCESS) {
 					if (nb_guids == guid_size) {
```
tormodvolden, line 1774: Also here a comment would be nice, like "// extend guid_list (in chunks) if needed". Not that it is difficult to see this from reading the code, but such a comment allows not reading the code :)

tormodvolden:
I added some review comments, tentatively in order of importance.

libusb/os/windows_winusb.c
pazourek: > For future use, maybe return number of guids discovered on success?
There is a parameter 'guid_number', which should ensure taking the n-th GUID. So for going through all GUIDs it is possible to increasing the guid_number as long as the get_guid() returns SUCCESS. From my angle, returning number of GUIDs is a little strange. However if you wish, I have no problem with it.

> Do I understand correctly that dev_id is only used for the warning messages? I guess that is still worth it though.
Yes, the dev_id is used only for the warning messages. It makes me sense to keep it here.

libusb/os/windows_winusb.c
pazourek: Yes, you're right, we might allow GUIDs also without terminating NULs with checking the overrun. First I'll try to make any firmware with such an error to see how it behaves like.

libusb/os/windows_winusb.c
pazourek: The new GUID is compared with the others in the line 1787 and added into the guid_list on the line 1792 after the check. So the check already is before extending the guid_list as you suggest. Or did I miss something?

tormodvolden: Yes, by "extending" I actually meant the resizing/reallocation done on line 1775. If the GUID has been seen and added before in a previous invocation, this reallocation would be unnecessary, right?

libusb/os/windows_winusb.c
tormodvolden: You're probably right, there is no potential use of returning the number of GUIDs. Iterating over this function like you suggest, would work well.

libusb/os/windows_winusb.c
pazourek: You're right, it makes sense to leave this check in the caller because this function reads only from registry key that might be independent on the driver state.

libusb/os/windows_winusb.c
pazourek: According to documentation the "DeviceInterfaceGUID" format is "{xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx}\0".
However at least the win10 accepts also the GUID string without the last \0 character. The GUID string is 1char shorter then.

The situation with "DeviceInterfaceGUIDs" is similar. The individual GUIDs are separated by \0 wchar by default. The list ends with two '\0' wchars, for e.g: "{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}\0\0"
But the win10 accepts the GUID list without the last two \0 chars. The GUID list is also accepted when there is missing the \0 wchars as the GUID separator. So the following GUIDs should be accepted as well:
"{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}\0"
"{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}"
"{xxx.....xx}{xxx.....xx}{xxx.....xx}"

pazourek: In addition "DeviceInterfaceGUID" and "DeviceInterfaceGUIDs" with paddings should also be considered as valid, for e.g.:
"{xxx.....xx}\0\0\0"
"{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}\0\0\0\0"

pazourek:
The get_guid() has been reworked to be able to parse all types of GUID as mentioned  in comments above. All other review comments have been included.

My tests showed, that although the GUIDs were correctly parsed, if the GUID string is non-standard, libusb fails when opening it. Tests result:

- Reporting standard DeviceInterfaceGUID - working
- Reporting non-standard DeviceInterfaceGUID with padding - W10 reports: "This device cannot start. (Code 10) An invalid parameter was passed to a service or function."
- Reporting standard DeviceInterfaceGUIDs - working
- Reporting non-standard DeviceInterfaceGUIDs with padding - working
- Reporting non-standard DeviceInterfaceGUIDs with missing the last terminating char - W10 reports: "This device is working properly." but the libusb cannot open such a device. The problem seems to be in get_interface_details() which cannot find out the dev_interface_path.
- Reporting non-standard DeviceInterfaceGUIDs with missing two terminating chars  - libusb cannot open such a device, same issue as above
- Reporting non-standard DeviceInterfaceGUIDs with missing all separators and two terminating chars - libusb cannot open such a device, same issue as above

libusb/os/windows_winusb.c
```diff
@@ -1450,6 +1450,140 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+	// "{xxx.....xx}\0\0\0", "{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}\0",
+	// "{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}", "{xxx.....xx}{xxx.....xx}{xxx.....xx}",
+	// "{xxx.....xx}\0{xxx.....xx}\0{xxx.....xx}\0\0\0\0"
+	if ((reg_type == REG_SZ ) || (reg_type == REG_MULTI_SZ)) {
+		/* get the n-th guid indexed by guid_number since the DeviceInterfaceGUIDs may
+		   contain more GUIDs */
+		guid = guid_string;
+		// add two terminating chars for not overrunning the allocated memory during iterating the guid
+		guid[size] = '\0';
+		guid[size + 1] = '\0';
```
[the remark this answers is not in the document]
tormodvolden: On a second look, I think I lost track of "size" here. It has the length of the whole string buffer, so no problem then.

libusb/os/windows_winusb.c
```diff
@@ -1450,6 +1450,140 @@ static int set_hid_interface(struct libusb_context *ctx, struct libusb_device *d
...
+			// skip the terminating char if available
+			if (*guid == '\0') {
+				guid++;
+			}
+
+			guid_number--;
+		}
+		// add terminating char to the string
+		tmp_str = strchr(guid, '}');
+		if ((tmp_str == NULL) || (guid_number > 0)) {
```
tormodvolden, line 1553: Isn't guid_number always zero at this point, after the while loop? So also the below warning will always say "index 0" ?

tormodvolden:
Just added some minor whitespace fixups that I will squash when I merge this. If you have WIP stuff that conflicts you can also ignore it and force-push.

libusb/os/windows_winusb.c
pazourek: Yes, you're right, the guid_number is always zero at this point. And thus the warning message prints a wrong guid number.

tormodvolden: Does my last commit sort this out correctly? I cannot test this.

pazourek: I performed a few tests with all supported variants. All my tests passed, the guid index is correctly printed now.

tormodvolden: Thanks for testing (again)!

tormodvolden:
Merged. Thanks a lot for your work on this.

tormodvolden:
I see these warnings when I test 1.0.27-rc1:
```
libusb: warning [get_guid] no DeviceInterfaceGUID registered for 'USB\ROOT_HUB30\4&DB2FC44&0&0'
libusb: warning [get_guid] no DeviceInterfaceGUID registered for 'USB\VID_045E&PID_091A\5&2C705BFE&0&1'
libusb: warning [get_guid] no DeviceInterfaceGUID registered for 'USB\VID_045E&PID_07CD\6&3A757EEC&0&4'
libusb: warning [get_guid] no DeviceInterfaceGUID registered for 'USB\VID_0483&PID_5730\TEST1'

```
Maybe they should be demoted to "info" or "debug"?

Droid-MAX:
I have a vaguely bad feeling about this, could be the next potential supply chain poisoning similar to xz-utils?
And I found that programs built with the libusb-1.0.26 source code at some point in time always output some Libusb-related warning information, which had not been the case before. A warning message appears even after the upgrade to 1.0.27.

https://github.com/Droid-MAX/SNANDer/issues/15

mcuee:
@Droid-MAX

Don't spam the github issues. Thanks.

I have re-opened the relevant issue.
* https://github.com/libusb/libusb/issues/1394

If you feel that is not good enough, please open a new issue.

Issue #1177, opened 2022-08-20
Multiple WinUSB devices in single composite device


Novakov:
Let's assume USB device reporting itself as composite device with two interfaces (0 and 1) of class vendor. Using Microsoft OS 1.0 descriptors they are assigned WinUSB driver (using WINUSB CompatibilityId descriptor with two sections) and unique `DeviceInterfaceGUID` (using two separate Extended Compatibilty descriptor, one for each interface)

Some screenshots:
Device manager:
![image](https://user-images.githubusercontent.com/420393/185757154-9ffef3ea-83a4-442e-8f52-2c6bad7117de.png)

Registry, interface 0 with GUID:
![image](https://user-images.githubusercontent.com/420393/185757246-f28f6762-3f16-4c24-a9ee-2a2fc4526c5b.png)

Registry, interface 1 with GUID:
![image](https://user-images.githubusercontent.com/420393/185757265-50073192-54ed-4971-aed6-ac349f127e3b.png)

Now, there is well known limitation of WinUSB:  "WinUSB is not an option if: Your device is accessed by multiple applications". However in the context of composite device interface 0 and interface 1 are **separate devices**.
Given that by using SetupAPI and WinUSB native API I'm able to lookup specific interface (single child device from composite device) that gives me paths:
* `\\?\usb#vid_4d4e&pid_4455&mi_00#7&57ec2a6&2&0000#{644991b2-1f10-45f7-a445-9a50ee4ce73d}` for interface 0
* `\\?\usb#vid_4d4e&pid_4455&mi_01#7&57ec2a6&2&0001#{13b2b252-adbf-45f8-9512-07342e35f305}` for interface 1
As intended each path points to specific interface. Opening them (`CreateFile`) and wraping it in WinUSB (`WinUSB_Initialize`) in two different applications (one application for interface 0 and one for interface 1) gives me two separate handles for two separate devices. I can use these handles to issue control transfers and they are correctly interpreted by USB device as transfers to two different interfaces.

**Observation:** From WinUSB point of view it is possible to access **different** child devices of composite device from multiple application. It is not possible to access **the same** child device of composite device from multiple application.

Sorry for long introduction and probably stating facts that might be obvious but I wanted to describe context and what is possible before going to libusb specifics.

[...]

Issue #16, opened 2014-06-02
Unable to claim interface multiple times


ssoloski:
When performing multiple open/close cycles and claiming an interface, libusbK fails after 32 times with Access Denied error.

This error was found during unit testing of my code using libFTDI; further testing using only libusb 1.0.19-rc2 code reproduces the error. This happens under Win8.1 x64; USB drivers are installed using libusbK-inf-wizard. If I install only the libusbK driver, the problems occurs. I've installed the driver on both the composite device (an FT2232H chip) and on Interface A; the error occurs either way.

If I install the WinUSB driver, the problem occurs... but if you then remove the libusb0 and libusbK dll files (installed by the inf-wizard) and only use the WinUSB driver the error does not occur. 

I have a sample program to demonstrate the issue; not sure how to attach it here though. Let me know if you need additional info...

Steve

dickens:
Hi,

Please try the attached patch and see if that resolves the issue you are
observing.

Regards,
Chris

dickens:
Hi,

Please try the patch below and see if that resolves the issue you are observing.

Regards,
Chris


```
From 4bb0fffcefc507f301afdfd97a40f17ecd839a25 Mon Sep 17 00:00:00 2001
From: Chris Dickens <christopher.a.dickens@gmail.com>
Date: Sat, 3 Jan 2015 01:07:07 -0800
Subject: [PATCH] Windows: Fix auto-claim functionality for WinUSB API

* Correctly auto-release interface 0 when releasing associated interface(s)
* Correctly handle when interface 0 is claimed after one or more associated
  interfaces so as to not release interface 0 prematurely
* Closes #16
---
 libusb/os/windows_usb.c | 170 +++++++++++++++++++++++++++++++++---------------
 libusb/os/windows_usb.h |   1 +
 2 files changed, 119 insertions(+), 52 deletions(-)

diff --git a/libusb/os/windows_usb.c b/libusb/os/windows_usb.c
index a36ab45..fd8597e 100644
--- a/libusb/os/windows_usb.c
+++ b/libusb/os/windows_usb.c
@@ -2835,6 +2835,45 @@ static int winusbx_configure_endpoints(int sub_api, struct libusb_device_handle
 	return LIBUSB_SUCCESS;
 }
 
+static int winusbx_autoclaim_iface0(int sub_api, struct libusb_device_handle *dev_handle)
+{
+	struct windows_device_handle_priv *handle_priv = _device_handle_priv(dev_handle);
[...]

PR #1323, opened 2023-10-12
Windows: Add support for claim_interface, release_interface for libusbK

sonatique:
Problem addressed by this PR: 
under Windows (as opposed to Linux, MacOS) multiple applications can all successfully call claim_interface on the same device with the same interface number. All calls will succeed, letting all application concurrently access non-default endpoints.

This issue prevent Windows application to use claim_interface for simple built-in access management and force implementer to add some Windows dedicated mechanism in order to mimic the default behavior os other platforms.

libusbK support claim_interface / release_interface at libusbK level, but these functions where not used by libusb-1.0. I guess that the reason behind it is that libusb-1.0 was designed to only used the more restraint WinUSB API which doesn't expose claim_interface / release_interface equivalent function, given WinUSB.dll / WinUSB.sys only permit one open (making multiple concurrent claim_interface irrelevant).

This PR add proper support for claim_interface / release_interface for libusbK-based sub API.

EDIT:
The behavior when using WinUSB.dll / WinUSB.sys remains unchanged.
With libusbK.dll and a driver supporting it, claim_interface now succeeds only once at a time for a given interface, but only if the application opts in by setting the new LIBUSB_OPTION_WINDOWS_EXCLUSIVE_CLAIM option. **The default is the
existing behavior, so applications relying on claiming the same interface from several processes are unaffected.**

This also fixes a pre-existing bug that the change would otherwise have made worse: an interface claimed automatically to service a control transfer was recorded as claimed by the application, so an application claim on that interface silently did nothing, and the completing transfer released the interface underneath it.


mcuee:
@tormodvolden 

I think this is a good change to have. 

Previously I was assuming that this is a libusbK issue.
* https://github.com/mcuee/libusbk/issues/37

sonatique:
@mcuee : 
> Previously I was assuming that this is a libusbK issue.
From what I can see from current code (I didn't check how it was in May 2022): libusbK.dll has a "ClaimInterface" API function
It  is explicitly known to libusb-1.0 even before my PR, see windows_winusb.h line 574), but libusb_claim_interface was not calling it (see my main comment).
What I can see in libusbK code, in addition to the comment
/*
Claim the interface. You can use the claim/release interface functions
cooperatively in libusbK application to ensure only one application is
usng the interface at a time.
*/

is that ClaimInterface implementation calls LIBUSBK_IOCTL_CLAIM_INTERFACE against the driver, so assuming that the driver correctly implement it, the call should work.

I tested with libsub0.sys (not with libusbK.sys yet).

sonatique:
Please note: after further testing, I think that the "auto_claim" feature may conflict with this change.

The expectation is that application B shall be able to do control transfer even after application A claimed and hold an interface. But because "auto_claim" wants to claim first interface before executing control transfer, and since it cannot do so in application B, the whole auto_claim doesn't work any longer.

I need to think about how to correct this. My first thought is that some assumptions made in the code around what could be claim or not is only relevant for WinUSB.dll not libusbK,dll. Especially "It is a requirement for multiple interface devices on Windows that, you must first claim the first interface before you claim the others".

Given no interface has never been properly claimed so far by libsub-1.0, we may probably assume that this requirement is not exactly true as it is stated. There should be some subtleties.

I will investigate.

sonatique:
Yes, this is what I think as well. I will update my PR with this in sight . Le 12 oct. 2023 à 17:19, mcuee ***@***.***> a écrit :﻿
I think autoclaim is only applicable to WinUSB and should not be applicable to libusbK.
Even with WinUSB, we may need to revisit this topic.

mcuee:
@pbatard 

Sorry to disturb this again, just wondering if you still remember the background of auto-claim. Thanks.

mcuee:
@sonatique 

I am not so sure if the following issue is related.
* https://github.com/libusb/libusb/issues/16

pbatard:
@mcue, yeah, as far as I can recall, the issue that autoclaim tried to address was that, on Linux/Mac platforms, you can claim any interface, but when using WinUSB on Windows, you need to claim interface 0 before you can claim any other interface. So of course, this created issues with people developing a libusb app on Linux/Mac and then wondering why it failed on Windows...

sonatique:
Hello @mcuee @pbatard ,
I just pushed a second commit, my first attempt at solving the "auto_claim vs claim" issue I mentionne earlier.

I made a few attempts, but in the end this one seemed to be the most promising in the sens that changes are quite contained close to where it is relevant.

If we decide to go this route, I guess a few more renaming would be necessary (for instance "auto_claim" to "auto_initialize" because, as far as I could tell, there was a lof confusion caused by the fact "claim" was used where the actual implementation was only "initializing" interface, not really "claiming" them.

All my test so far seems to indicate that this solution work for all my use cases (which are of some non negligible complexity) but of course I am looking forward for your comments.

sonatique:
I think that I noticed a possible race condition that is pre-existing to this PR, but I don't think any normal usage would produce it and therefore I am wondering whether it is necessary to try to prevent it (I think it is not):

1) a control transfer is issued before claiming an interface, therefore an interface is auto_claimed
2) before the control transfer ends, so before the auto_claimed interface is auto_released, another thread claim this same interface explicitly.
3) when the control transfer complete, the interface is auto_released which leaves the system in an unexpected state from the user point of view: the interface that he just claimed is now released and subsequent calls on it will fail.

What do you think?

sonatique:
Another thought: I had time to think about it more and my conclusion so far is that the best move would be to remove "auto_claim" altogether, in favor of simple doing the init during Open (and "DeInit" during Close). This way control transfer would be immediately available for an open device (which is the goal of auto_claim) and claim would be a real manual claim. 
This way the behavior would be the same as what we have for non-Windows OS, and the code would be much simpler, less error-prone. For WinUSB.dll it would still behave the same, and for libusbK.dll we would have the extra feature of a working claim for getting exclusive access.

Any though before I start working on this?

mcuee:
> Another thought: I had time to think about it more and my conclusion so far is that the best move would be to remove "auto_claim" altogether, in favor of simple doing the init during Open (and "DeInit" during Close). This way control transfer would be immediately available for an open device (which is the goal of auto_claim) and claim would be a real manual claim. This way the behavior would be the same as what we have for non-Windows OS, and the code would be much simpler, less error-prone. For WinUSB.dll it would still behave the same, and for libusbK.dll we would have the extra feature of a working claim for getting exclusive access.
> 
> Any though before I start working on this?

I think this is a good idea.

Let's hear from @tormodvolden as well.

diabolo38:
I'mot much using winusb back end but libusbk , i have plenty composite device where libusb control non 0 itf while itf 0 are system one (cdc, ether, mass storage  ...) 

also i heavily used and need the ability to open/claim  multiple time the same interface in many application with libusbk (done with proper mutex handling in app/lib ) breaking this behavior would be a serious/major regression.

They are behavior difference in the various os and that's just one user shall be aware and take care off . Can't user just used winusb backend to get single exclusive open/use ?

mcuee:
> I'mot much using winusb back end but libusbk , i have plenty composite device where libusb control non 0 itf while itf 0 are system one (cdc, ether, mass storage ...)
> 
> also i heavily used and need the ability to open/claim multiple time the same interface in many application with libusbk (done with proper mutex handling in app/lib ) breaking this behavior would be a serious/major regression.
> 
> They are behavior difference in the various os and that's just one user shall be aware and take care off . Can't user just used winusb backend to get single exclusive open/use ?

One possibility is to have a libusb_set_option for libusbK to allow or disallow claiming the interface multiple times. The default can be the existing behavior (allow).

mcuee:
Discussion in Issue #1177 and PR #1181 may also be applicable here.
* https://github.com/libusb/libusb/pull/1181
* https://github.com/libusb/libusb/issues/1177#issuecomment-1401892549

mcuee:
Rebase to master using the tip from @tormodvolden.
No issues with `make check` even with the Dell Docking Station.

```
C:\work\libusb> git clone https://github.com/libusb/libusb.git .\libusb_pr1323
C:\work\libusb> cd .\libusb_pr1323\
C:\work\libusb> git fetch origin pull/1323/head
C:\work\libusb> git checkout -b pr1323_rebase FETCH_HEAD
C:\work\libusb> git rebase origin/master

 CLANG64 /c/work/libusb/libusb_pr1323
$ make check
Making check in libusb
make[1]: Entering directory '/c/work/libusb/libusb_pr1323/libusb'
make[1]: Nothing to be done for 'check'.
make[1]: Leaving directory '/c/work/libusb/libusb_pr1323/libusb'
Making check in examples
make[1]: Entering directory '/c/work/libusb/libusb_pr1323/examples'
make[1]: Nothing to be done for 'check'.
make[1]: Leaving directory '/c/work/libusb/libusb_pr1323/examples'
Making check in tests
make[1]: Entering directory '/c/work/libusb/libusb_pr1323/tests'
make  check-TESTS
make[2]: Entering directory '/c/work/libusb/libusb_pr1323/tests'
make[3]: Entering directory '/c/work/libusb/libusb_pr1323/tests'
PASS: stress.exe
PASS: stress_mt.exe
PASS: set_option.exe
PASS: init_context.exe
Testsuite summary for libusb-1.0 1.0.27-rc1
# TOTAL: 4
# PASS:  4
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
make[3]: Leaving directory '/c/work/libusb/libusb_pr1323/tests'
make[2]: Leaving directory '/c/work/libusb/libusb_pr1323/tests'
make[1]: Leaving directory '/c/work/libusb/libusb_pr1323/tests'
make[1]: Entering directory '/c/work/libusb/libusb_pr1323'
make[1]: Leaving directory '/c/work/libusb/libusb_pr1323'
```

mcuee:
I have set this for milestone 1.0.31.

But we need to wait for the review of the other maintainers.

sonatique:
@mcuee @diabolo38  @pbatard : I rebased and fix an issue I noticed lately in a new commit.

I still believe this PR would be a nice addition to libusb, but I understand the change in behavior could be problematic.

In order to solve this we could (apart from documenting in in the next release changelog obviously):
1) make it configurable at runtime
2) enable/disable it at compilation time

In both cases we'll need to decide what is the new default behavior.

sonatique:
@mcuee @diabolo38 @pbatard: actually I decided to restructure the branch and force-pushed. 

Exclusive claiming is opt-in via a new LIBUSB_OPTION_WINDOWS_EXCLUSIVE_CLAIM, defaulting to the current behavior.
@diabolo38, your case of several processes claiming the same interface keeps working untouched unless an application explicitly asks for exclusion.

Per driver:
- libusbK.sys and libusb0.sys honor the option (both go through libusbK.dll)
- WinUSB is unaffected: it already permits a single open of the device
- HID returns LIBUSB_ERROR_NOT_SUPPORTED, since Windows shares HID devices by
  design. Control transfers are never affected.

The first commit is independent of the option and fixes a bug that exists in master today: auto-claiming an interface for a control transfer records it in dev_handle->claimed_interfaces, so a later libusb_claim_interface() on that interface returns success without reaching the backend, and the completing transfer then releases it underneath the application. 
That is the race I described in this thread on 2023-10-19.

I have tested libusbK extensively on hardware since the original submission.

I hope that this new way of doing will address all of your concerns. As far as I can tell, all obstacles that blocked the adoption of the first attempt of this PR are now lifted such that we could promptly converge to an agreement and merge the PR.

Thank you very much in advance for your reviews.

mcuee:
Let me carry out some simple tests over the weekend.

BTW, the Linux emscripten build failed -- which has nothing to do with this PR.

mcuee:
Hmm, sorry I forgot about this PR. I will see what I can do with this PR.

@sonatique 

You may want to resolve the conflicts and rebase with the latest git.

sonatique:
@mcuee : rebased on current master

Issue #1001, opened 2021-09-24
Unable to open Android phone on Windows.


AlynxZhou:
I am implementing [HID over AOAv2 for scrcpy](https://github.com/Genymobile/scrcpy/pull/2632), while libusb works quite well on Linux, it failed to open my Android phone on Windows.

I've tried `xusb.exe`, and it says `Failed to open device` (or something like that, I am under Linux so no screenshot).

I tried to use Zadig to install driver, however there is no device without driver, and I can see multiply devices for my phone if check `List all devices`, one is MTP, and another is `Samsung_ADB_Interface` (or something like it), I tried to force install WinUSB driver for them, but install failed.

Could you please tell me what's wrong with it? Or what should I do?

mcuee:
Read here on how to use Zadig. You may want to use WinUSB driver to replace the composite parent driver.
https://github.com/pbatard/libwdi/wiki/Zadig

mcuee:
BTW, if you have issues with Zadig, please create an issue in Zadig. It is a separate project.

AlynxZhou:
> Read here on how to use Zadig. You may want to use WinUSB driver to replace the composite parent driver.
> https://github.com/pbatard/libwdi/wiki/Zadig

How two decide which one is the composite parent driver? I've tried every one related and failed...

mcuee:
> > Read here on how to use Zadig. You may want to use WinUSB driver to replace the composite parent driver.
> > https://github.com/pbatard/libwdi/wiki/Zadig
> 
> How two decide which one is the composite parent driver? I've tried every one related and failed...

The one with the USB VID/PID of your phone and without saying Interface 0 or interface 1.

If you have issues using Zadig, you can use libusbk-inf-wizard which some users like more.
https://github.com/mcuee/libusbk/releases (install libusbK-3.1.0.0-setup.exe)

AlynxZhou:
> > > Read here on how to use Zadig. You may want to use WinUSB driver to replace the composite parent driver.
> > > https://github.com/pbatard/libwdi/wiki/Zadig
> > 
> > 
> > How two decide which one is the composite parent driver? I've tried every one related and failed...
> 
> You need to set the options so that the parent device will be listed.
> The one with the USB VID/PID of your phone and without saying Interface 0 or interface 1.
> 
> If you have issues using Zadig, you can use libusbk-inf-wizard which some users like more.
[...]

PR #1331, opened 2023-10-23
windows: Recognize Samsung composite device driver

yume-chan:
Relates to #1001 and https://github.com/Genymobile/scrcpy/issues/3654

This is a simple fix that treats SAMSUNG Mobile USB Composite Device driver (download: https://developer.samsung.com/android-usb-driver) as a composite device driver. This allows Samsung Android devices to be enumrated and opened by libusb.

I had previously filed a bug for Chrome's WebUSB implementation at https://crbug.com/1127206 and this PR is derived from their fix at https://chromium-review.googlesource.com/c/chromium/src/+/2436786.

Besides the name, Samsung's custom driver doesn't seem to behave differently from the Windows built-in driver (`usbccgp`). I have verified that Scrcpy OTG mode works on Samsung devices with their custom driver after this fix.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2311,7 +2311,10 @@ const struct windows_backend winusb_backend = {
  * USB API backends
  */
 
-static const char * const composite_driver_names[] = {"USBCCGP"};
+static const char * const composite_driver_names[] = {
+  "USBCCGP", // (Windows built-in) USB Composite Device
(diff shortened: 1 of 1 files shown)


tormodvolden:
Thanks. LGTM, this seems to make sense.

tormodvolden:
"Thanks" to GitHub your commit only has a "github.noreply" address, even when I fetch your branch. Do you have a proper e-mail address we can use as commit author?

tormodvolden:
Thanks for your contribution!

PR #1057, opened 2022-02-04
WebAssembly + WebUSB support

RReverser:
(This is a follow-up from https://github.com/libusb/libusb/issues/501#issuecomment-1017750127.)

I implemented WebAssembly (via Emscripten) + WebUSB support in libusb as a separate backend specifically so that it would be easier to upstream the changes. There are a few changes outside of the backend's folder, but only where they were unavoidable.

In the process of making this port I've made various improvements to Emscripten itself to make porting of this and other similar libraries easier in the future, especially in area of async integration, embedding JavaScript snippets and more.

Afterwards, I published an article on how this port was made at https://web.dev/porting-libusb-to-webusb/ with a follow-up in https://web.dev/porting-gphoto2-to-the-web/, which show how the libusb fork was used almost "out of the box" with libgphoto2 to build a web UI and control DSLR cameras from the web.

Given the complexity and variety of operations of gPhoto2, I'm hoping that such testing was extensive enough to eliminate most of the critical issues by now, but I wouldn't be surprised if there are more. As suggested by @mcuee in https://github.com/libusb/libusb/issues/501#issuecomment-1018403790, I'm opening this PR to get more eyes on it and have more people test it for now, with the hope that it can be merged upstream afterwards.

I have instructions for port usage in https://web.dev/porting-libusb-to-webusb/#using-the-port, but the TL;DR is:
 - point application to use the local checkout of the libusb fork
 - configure the libusb fork with `emconfigure`/`emmake` wrappers from [Emscripten toolchain](https://github.com/emscripten-core/emsdk) and `--host=wasm32`
 - build & link the application with `--bind -s ASYNCIFY -s ALLOW_MEMORY_GROWTH -pthread`
 - if you're building without optimizations, sometimes you'll get weird errors from WebAssembly due to insufficient Asyncify stack space; to increase it, add `-s ASYNCIFY_STACK_SIZE=10485760` (for example this would set it 10MB which should be more than enough for most apps, and it's only for debug builds anyway).

Let me know if there are any questions or concerns.

Fixes #501.

PR #1339, opened 2023-11-15
Wasm+WebUSB backend fixes and improvements

RReverser:
I've been recently working on a large project that uses libusb and WebUSB and had to make lots of improvements to the backend in the process.

Here's a list of primary fixes & changes:

 - Added long-awaited support for multithreading. Since WebUSB still doesn't support accessing same device from multiple threads, this works by proxying I/O operations to the main thread as discussed on the original PR #1057. For applications that run on a single (main) thread nothing changes and they will continue to access WebUSB via Asyncify like before, while other threads will use blocking mechanism until an asynchronous response is available from the main thread.
 - Rewrote notification mechanism to use atomic waiting via sync Wasm instructions or `Atomics.waitAsync` depending on the thread. This results in simpler and faster notifications than the previous `dispatchEvent`-based approach (which was used because `Atomics.waitAsync` wasn't yet available), as well as allows to send notifications across threads for multithreading scenario described above.
 - Fixed notification access to only wait/notify on the event we're interested instead of using a global lock.
 - Rewrote descriptor reading to query device for raw device & configuration descriptors instead of re-serializing them from JavaScript object representation. This incurs slight extra cost for the initial device opening, but fixes number of issues with information that is not yet exposed via WebUSB and allows to read supplementary descriptors such as string and BOS which were previously not supported.
 - Fixed listing of the available alternates where previously it would only return the first one.
 - Fixed device closing & re-opening which could previously spuriously error out with a "device busy" error.
 - Added mandatory Emscripten-specific linking flags to the generated pkgconfig so that they're automatically included in the final linking command.
 - Added device speed inference. This is not yet exposed via WebUSB, but we can at least make a best effort guess based on USB version and packet size, like some other backends already do.
 - Simplified & fixed device session ID generation which is now guaranteed to be truly unique, whereas previously it could clash with other devices of the same `vid:pid` type.
 - Prepare code to support building for the Web without mandatory multithreading (which is costly on the Web) in the future. For now those `#ifdef`s are defunct as libusb is always compiled with `-pthread`, but some upcoming changes on the Emscripten side will allow to leverage those codepaths for smaller Wasm binaries.
 - Require explicit `--host=wasm32-unknown-emscripten` instead of the previously accepted `--host=wasm32` as we might want to add non-Emscripten WebAssembly backends in the future.
 - Various smaller fixes and improvements.

Note that this requires Emscripten 3.1.48 or newer to support some of the features used here, namely `co_await` support for JavaScript values (without it code would be both more complex and slower) and some proxying APIs. It shouldn't be a big deal in practice as most users retrieve Emscripten via the official emsdk installer or Docker images.

I've tested those changes to work on two projects (web-gphoto2 which I used for the initial backend, and a closed-source project I've been working on) and it worked well in both scenarios, but I'd appreciate feedback from more users who'd like to try it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -83,10 +83,16 @@ case $host in
 	backend=haiku
 	platform=posix
 	;;
-wasm32-**)
+wasm*-emscripten)
 	AC_MSG_RESULT([Emscripten])
 	backend=emscripten
 	platform=posix
+	;;
+wasm*-unknown-none)
+	AC_MSG_ERROR([
+--host=$host_alias is not accepted as it might become ambiguous in the future.
+Please use an explicit --host=$host_cpu-emscripten instead.
+	])
 	;;
 *-linux* | *-uclinux*)
 	dnl on Android Linux, some functions are in different places
@@ -228,7 +234,9 @@ windows)
 	;;
 emscripten)
 	# Note: LT_LDFLAGS is not enough here because we need link flags for executable.
-	AM_LDFLAGS="${AM_LDFLAGS} --bind -s ASYNCIFY -s ASSERTIONS -s ALLOW_MEMORY_GROWTH"
+	EM_LDFLAGS="--bind -s ASYNCIFY"
+	AM_LDFLAGS="${AM_LDFLAGS} ${EM_LDFLAGS} -s ASSERTIONS -s ALLOW_MEMORY_GROWTH"
+	LIBS="${LIBS} ${EM_LDFLAGS}"
 	;;
 *)
 	dnl no special handling required
--- a/libusb/os/emscripten_webusb.cpp
+++ b/libusb/os/emscripten_webusb.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2021 Google LLC
+ * Copyright © 2023 Ingvar Stepanyan <me@rreverser.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -19,23 +20,47 @@
  *		Ingvar Stepanyan <me@rreverser.com>
  */
 
+#include <emscripten/version.h>
+
+static_assert((__EMSCRIPTEN_major__ * 100 * 100 + __EMSCRIPTEN_minor__ * 100 +
+			   __EMSCRIPTEN_tiny__) >= 30148,
+			  "Emscripten 3.1.48 or newer is required.");
+
+#include <assert.h>
 #include <emscripten.h>
 #include <emscripten/val.h>
+
+#include <type_traits>
+#include <utility>
 
 #include "libusbi.h"
 
 using namespace emscripten;
 
+#ifdef _REENTRANT
+#include <emscripten/proxying.h>
+#include <emscripten/threading.h>
+#include <pthread.h>
+
+static ProxyingQueue queue;
+#endif
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wmissing-prototypes"
 #pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wshadow"
+
 namespace {
+
 // clang-format off
-	EM_JS(EM_VAL, em_promise_catch_impl, (EM_VAL handle), {
-		let promise = Emval.toValue(handle);
-		promise = promise.then(
-			value => ({error : 0, value}),
-			error => {
+EM_JS(EM_VAL, usbi_em_promise_catch, (EM_VAL handle), {
+	let promise = Emval.toValue(handle);
+	promise = promise.then(
+		value => ({error : 0, value}),
+		error => {
+			console.error(error);
+			let errorCode = -99; // LIBUSB_ERROR_OTHER
+			if (error instanceof DOMException) {
 				const ERROR_CODES = {
 					// LIBUSB_ERROR_IO
 					NetworkError : -1,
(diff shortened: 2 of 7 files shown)


libusb/os/events_posix.c
```diff
@@ -37,26 +37,21 @@
...
-		});
-	} finally {
-		removeEventListener('em-libusb', onEvent);
-		clearTimeout(timeoutId);
+static int em_libusb_wait(int *ptr, int expected_value, int timeout)
+{
+	if (emscripten_is_main_runtime_thread()) {
+		return em_libusb_wait_async(ptr, expected_value, timeout);
+	} else {
+		return emscripten_atomic_wait_u32(ptr, expected_value, timeout) == ATOMICS_WAIT_OK;
```
goncalomb, line 52: `emscripten_atomic_wait_u32` uses nanosecond units, `timeout` is milliseconds here.

libusb/os/emscripten_webusb.cpp
```diff
@@ -68,559 +88,697 @@ namespace {
...
-  }
-  return LIBUSB_SUCCESS;
+void em_clear_transfer_priv(usbi_transfer* itransfer) {
+	WebUsbTransferPtr(itransfer).free();
 }
 
-void em_clear_transfer_priv(usbi_transfer *itransfer) {
-  WebUsbTransferPtr(itransfer).take();
+int em_cancel_transfer(usbi_transfer* itransfer) {
+	return LIBUSB_SUCCESS;
```
goncalomb, line 704: Shouldn't we call `usbi_signal_transfer_completion` here to signal transfer completion, even if we can't do it on WebUSB side?

goncalomb (commented):
As you know, I have a patch for your original implementation that adds rudimentary multithreading support and some extra fixes for my sane-wasm project (goncalomb/sane-wasm#1, [libusb.patch](https://github.com/goncalomb/sane-wasm/blob/00a40eea30755f9560cda6633fdf32a29592f55a/deps/libusb.patch)).

I could not resist waiting to test this PR on sane-wasm. I can tell you is mostly working but I'm having some issues with canceled transfers and some deadlocks, that I need to investigate further (I had similar issues when I was working on my implementation).

I'll leave a couple of comments for now. In a few days I'll be able to do some more testing.

libusb/os/events_posix.c
RReverser: Ohh good catch, must have missed it when switching from JS-based API. It's pretty unfortunate that JS `Atomics.wait` and Wasm `atomics.wait` take different units 🤦🏼‍♂️

RReverser:
```suggestion
		return emscripten_atomic_wait_u32(ptr, expected_value, 1'000'000LL * timeout) == ATOMICS_WAIT_OK;
```

RReverser:
> I can tell you is mostly working but I'm having some issues with canceled transfers and some deadlocks, that I need to investigate further (I had similar issues when I was working on my implementation)

Hm interesting, I didn't see issues like that but admittedly neither of the projects I worked on rely on transfer cancellation. I'd love to see what you find!

libusb/os/emscripten_webusb.cpp
RReverser: Also - you probably meant `usbi_handle_transfer_cancellation`, right? Either way, that's always done from transfer callback even in other backends.

RReverser: Ah I think I see what you mean - take a look at `promiseThen` in `em_submit_transfer`. It will invoke `usbi_signal_transfer_completion` when transfer is completed, regardless of whether it was marked as cancelled internally or not.

goncalomb: I should have explained my reasoning better. I understand that `usbi_signal_transfer_completion` is called when the promise resolves, but it might never resolve, by that I mean the device might never respond (no timeout). It will eventually fail, when the device is closed.

When a cancel is requested (i.e. `libusb_cancel_transfer`, backend method `cancel_transfer`), we just say "yes, sure, we are canceling it", but we don't actually do anything, because WebUSB does not have that function.

I guess what I'm trying to say here is that, we are not "pretending" all the way, we say "we are canceling" but not "it's canceled" (even if it's all fake, because we can't really cancel).

The specific case that I have is a call to [libusb_interrupt_transfer](https://libusb.sourceforge.io/api-1.0/group__libusb__syncio.html#ga0f171a72904a552fc43e6e6564d108a3) with a short timeout (300 ms). This is part of a specific SANE backend that I'm testing with, where it starts a transfer to see if the device has some "events", or it quickly times out.

I suspect this happens on other transfers when the timeout is involved (WebUSB not timing out) and the device not responding. It goes something like this:

1. call to `libusb_interrupt_transfer` with a short timeout
2. the device never responds
3. libusb times out and [calls `libusb_cancel_transfer`](https://github.com/libusb/libusb/blob/67ac8a0803034bc6022c6ce53cd813848cd4fe77/libusb/io.c#L2057)
4. we say "yes, we are cancelling"
5. the device continues to never respond (promise does not resolve)
6. libusb continues waiting to completion (cancellation), `libusb_interrupt_transfer` continues blocked. 

Now, that's why I was suggesting calling `usbi_signal_transfer_completion` on `em_cancel_transfer` to fake the cancellation all the way.

That was [one of the fixes on my original patch](https://github.com/goncalomb/sane-wasm/blob/00a40eea30755f9560cda6633fdf32a29592f55a/deps/libusb.patch#L79), I added a call to `em_signal_transfer_completion_impl` (that [itself was just a wrap of `usbi_signal_transfer_completion`](https://github.com/libusb/libusb/blob/fcf0c710ef5911ae37fbbf1b39d48a89f6f14e8a/libusb/os/emscripten_webusb.cpp#L133)). You can also see that I wrapped the transfer promise with cancellable promise, to avoid it resolving later (i.e. on device close), at that point the transfer is finished (from the viewpoint of libusb), and avoids use-after-free, etc. My implementation is not very pretty, I'm sure there are other ways of doing it.

Well I hope that I explained this right. My understanding of libusb internals is very limited. I'm inferring everything just from reading the code, following the execution path and skimming the documentation.

I'll continue testing on my end in a couple of days.

goncalomb:
> [...] about deadlocks - I remember now I did see that before in one of these projects. Assuming your cause is the same, it can happen if you lock a mutex on the main UI thread waiting for a result [...]

Yes, it's related to that, it's not your implementation. In my case SANE is calling `pthread_join` from the main thread, to wait on the worker. That causes a deadlock because the main event loop never runs, and the worker is "waiting on USB promises".

At the time I think I tried `PROXY_TO_PTHREAD`, but in my case I don't actually have a "main", sane-wasm is a lib, we (as the user) call exported methods "from the main thread". I don't think `PROXY_TO_PTHREAD` helps in this case.

I actually [have a weird fix for this](https://github.com/goncalomb/sane-wasm/blob/00a40eea30755f9560cda6633fdf32a29592f55a/deps/libusb.patch#L187), while the worker was waiting for USB events, I enqueue `emscripten_sleep(0)` on the main thread every 50ms using `setInterval`, if the main thread starts blocking that unwinds the stack and lets the event loop run. I guess I'll need to continue using this hack.

libusb/os/emscripten_webusb.cpp
RReverser: > we just say "yes, sure, we are canceling it", but we don't actually do anything, because WebUSB does not have that function.

It's because libusb is marking it as cancelled internally, and later we just check that flag in the transfer callback.

> I suspect this happens on other transfers when the timeout is involved (WebUSB not timing out) and the device not responding. It goes something like this:

Hm step 5 doesn't sound right here... The promise from transfer APIs should always either resolve with some data, resolve with no data or reject with error - that is, either way our callback with PromiseResult should always be called. That matches what I observed with other devices when added this implementation too.

So cancellation will act as if it was slightly delayed - it will only trigger when device responded or when the transfer failed - but since apps should generally expect that cancellation is asynchronous anyway, the idea is that it's not really a problem.

I'm curious about the scenario where it fails. IIUC, you're saying that with some scanners the transfer promise might hang indefinitely and be neither resolved or rejected?

RReverser:
> At the time I think I tried `PROXY_TO_PTHREAD`, but in my case I don't actually have a "main", sane-wasm is a lib, we (as the user) call exported methods "from the main thread". I don't think `PROXY_TO_PTHREAD` helps in this case.
> 
> I actually [have a weird fix for this](https://github.com/goncalomb/sane-wasm/blob/00a40eea30755f9560cda6633fdf32a29592f55a/deps/libusb.patch#L187), while the worker is waiting for USB events, I enqueue `emscripten_sleep(0)` on the main thread every 50ms using `setInterval`, if the main thread starts blocking that unwinds the stack and lets the event loop run. I guess I'll need to continue using this hack.

Ah yeah, for library the fix would be more involved as it requires manual proxying.

Conceptually there are two ways to go about this problem:

1. As mentioned, we could allow moving libusb calls to a separate thread - which sounds like what you did - in which case what happens is roughly:
 - Main thread asks the worker to run some WebUSB task.
 - Worker runs it asynchronously (since it's all Promise-based).
 - Main thread is synchronously blocked until an asynchronous task is finished.
2. Applications/libraries should move their own work to a worker, so that main thread is free to run asynchronous tasks. E.g.:
 - Your library asks the worker to run some SANE task.
 - When SANE task reaches a WebUSB call, it proxies that to the main thread.
 - Main thread runs this asynchronous WebUSB task w/o blocking.
 - The helper worker is blocked while waiting for a result.

They're mirroring each other, the only difference is which thread is running an asynchronous non-blocking task and which one is blocked.

(1) is more intuitive to implement, and sounds like what you're doing, but it's not ideal because it means blocking the main thread for potentially long periods of time, especially for slow devices - which is discouraged on the web because it's also the thread responsible for UI and can lead to poor UX through non-responsiveness.

Instead - I know it's a bit of work - but I'd suggest to rewrite bindings to do (2) where you Embind C++ API proxies all the `sane_` calls to a worker and either invokes a callback or returns a Promise on the main thread when that's done. This approach will be compatible with proxying-to-main implemented in this PR, and will provide the best UX.

libusb/os/emscripten_webusb.cpp
RReverser: > I'm curious about the scenario where it fails. IIUC, you're saying that with some scanners the transfer promise might hang indefinitely and be neither resolved or rejected?

If this is the case, your proposed solution sounds reasonable and I'd ask you to make a follow-up PR to integrate it (I know we talked about it last time over email when you mentioned your changes).

It makes sense to cancel early, it's just that it might be better to keep PR separate so code attribution purposes and because it wasn't something I had to change in this PR to make mentioned apps work yet.

goncalomb: > [...] libusb is marking it as cancelled internally [...]

Yes, I understand that.

> [...] you're saying that with some scanners the transfer promise might hang indefinitely and be neither resolved or rejected?

Yes, that was what I was trying get at, with my over-complicated explanation. Isn't that kind of expected because WebUSB does not have timeouts, and we are not really triggering the a cancellation on WebUSB's side?

Here's a quick example of what I'm seeing (the errors are just debug messages form SANE and libusb):

image

1. call to a specific SANE function;
2. starts a USB transfer, timeout 300ms;
3. asyncify kicks in and we get a pending promise;
4. eventually the transfer times out, libusb calls `libusb_cancel_transfer`, we fake it, BUT the transfer promise never resolves, it will stay in this state "forever"
5. i eventually close the device "from the outside", by calling WebUSB directly
6. transfer promise fails
7. transfer completes (canceled)
8. i get the response from the original SANE call

> [...] your proposed solution sounds reasonable and I'd ask you to make a follow-up PR to integrate it [...]

Sure, I'll have to work on it either way, and I have this test case to test it with.

RReverser: > Isn't that kind of expected because WebUSB does not have timeouts

It depends on the device. When I implemented backend initially, I was testing with DSLRs and there if cancellation was implemented as no-op, device itself eventually just returned transfer result, we ignored it, and cancellation still visibly works as expected just with a tiny delay.

If the device itself might never neither respond nor reject the transfer, that's of course a wholly different situation and then a fix like the one you implemented is necessary. Problem is, I haven't encountered such devices yet so wouldn't even have anything to test this behaviour on.

goncalomb:
> but I'd suggest to rewrite bindings to do (2)

Yes, I'll probably do that eventually. I was kind of thinking that at beginning, moving all the blocking calls to a helper worker. I don't really want to mess with SANE's code, just proxy the bindings. I'll have to explore the best way to do it. 

Thanks for all the great insight.

RReverser:
> I don't really want to mess with SANE's code, just proxy the bindings.

Yeah that sounds about right. Just wrapping all such bindings into `queue.proxyCallback` should be enough for a start, and then on  the JS side you can turn them into promises or whatever is most convenient for the high-level API.

libusb/os/emscripten_webusb.cpp
RReverser: Btw, as for your mention of WebUSB not supporting timeouts - we actually support them in the translation layer, but looking at your screenshot, that wouldn't help because the 2nd call to `usbi_wait_for_events` asks for a timeout of 1 minute. So even if it eventually expires, that's obviously too long for any real-time feedback.

RReverser:
Found couple more "interesting" 😅 synchronisation issues by somewhat augmenting stress_mt. I split out test changes in https://github.com/libusb/libusb/pull/1347 and pushed couple more fixes in this PR.

libusb/os/emscripten_webusb.cpp
RReverser: @goncalomb Have you had more chance to play with this / any more issues you noticed?

goncalomb: Yes, I've done some extra testing, overall it works very well. No new issues.

goncalomb:
I have a working version of sane-wasm that is more stable now using this PR, when compared with my old implementation. Especially now that I also started moving my bindings to an helper thread, as we discussed. I kind of followed your footsteps and went down the C++20 coroutines rabbit hole for that. :sweat_smile:

https://github.com/goncalomb/sane-wasm/compare/libusb-threads

Note that I'm not really testing libusb alone, I'm integrating it with sane-wasm and testing that. I'll leave standalone tests for the libusb pros, I see that you are doing some commits in that regard.

I have to say that I'm still using a patch for that cancelled transfer problem, where I signal the cancellation and short-circuit the transfer promise on WebUSB's side.

https://github.com/goncalomb/sane-wasm/blob/b44878f4ea3b96d0339d579f138521f411c06514/deps/libusb.patch

I'll work on cleaning that up a bit (probably starting with storing the promise in libusb's `priv` instead of a global), and creating a follow-up PR, as per your suggestion.

No extra outstanding issues on my part.

libusb/os/emscripten_webusb.cpp
RReverser: Thank you!

RReverser:
@goncalomb I suspect that patch can be simplified somewhat by using Promise.race, which already guarantees that the callback is invoked only once.

Basically it should do something like (JS pseudocode) 

```js
let cancel;

transfer_promise = Promise.race([
  transfer_promise, 
  new Promise((_, reject) => cancel = reject) 
]);

//...somewhere later

cancel(someCancellationTokenThatYouCanCompareWithToKnowItsACancellationAndNotAnActualError);
```

goncalomb:
> I suspect that patch can be simplified somewhat by using Promise.race, which already guarantees that the callback is invoked only once.

That is very nice construction using `Promise.race` that I never thought of, or seen anywhere. I'll definitely use that in the future.

RReverser:
If it helps, I can also split this up back into slightly smaller commits. Probably not the backend itself, as it's almost a rewrite, but at least to separate build script improvements from backend changes.

tormodvolden:
Yes, I think it makes sense to include this, instead of shipping an outdated version with so much rewritten. We just have to trust your and Goncalomb's testing, and your opinion on the state of this. There were many changes here lately but it seems to have come more quiet. Do you consider this in good condition to release now?

configure.ac
```diff
@@ -83,11 +83,18 @@ case $host in
...
+*-emscripten)
 	AC_MSG_RESULT([Emscripten])
 	backend=emscripten
 	platform=posix
 	;;
+wasm32*)
+	AC_MSG_ERROR([
+--host=wasm32 used to be recommended and assumed Emscripten backend.
+This is no longer accepted as we might add more WebAssembly backends in the future.
+Please use an explicit --host=wasm32-unknown-emscripten instead.
```
tormodvolden, line 95: I guess we don't need this in a first release.

libusb/os/events_posix.c
```diff
@@ -257,22 +256,14 @@ int usbi_wait_for_events(struct libusb_context *ctx,
...
-#else
-	num_ready = poll(fds, nfds, timeout_ms);
+	// Emscripten's poll doesn't actually block, so we need to use an out-of-band
+	// waiting signal.
+	em_libusb_wait(&ctx->event.has_event, 0, timeout_ms);
+	// Emscripten ignores timeout_ms, but set it to 0 for future-proofing in case
+	// they ever implement real poll.
+	timeout_ms = 0;
 #endif
+	num_ready = poll(fds, nfds, timeout_ms);
```
tormodvolden, line 266: What is this doing, outside the emscripten section?

tormodvolden: nvm

tormodvolden:
> Note that this requires Emscripten 3.1.48+ to build

I'd prefer saying "3.1.48 or later", because I was first thinking git snapshot when I saw this, although without good reason.

RReverser:
> There were many changes here lately but it seems to have come more quiet. Do you consider this in good condition to release now?

Yeah there were some silly mistakes such as I was accidentally returning `LIBUSB_SUCCESS` from `get_config_descriptor_by_value` implementation instead of the configuration length but none of our projects used that function so it flew under the radar.

For this kind of non-Emscripten-specific stuff another pair of eyes (and cross-platform unit tests in the future) is always welcome, but, as I said, I'm happy to leave it as follow-up fixes if anything else comes up.

Other than that, it does seem to work well on the projects we tested on so I'd say it's in a good condition, yes. For the release it could be nice to include upcoming @goncalomb's patch too but you probably don't need to block on it as we all use libusb from git anyway for Wasm targets.

configure.ac
RReverser: Problem is that there are several projects that I know of that already use this backend from libusb git + potentially more private projects, so I'd rather keep it around for a couple months if you don't mind so that everyone can migrate with graceful error message.

RReverser: > Anyway, I would still prefer leaving out the past tense talk and just say "wasm32 is not accepted, please use...".

Sure, sounds good for future-proofing, I'll update it.

RReverser:
> I'd prefer saying "3.1.48 or later", because I was first thinking git snapshot when I saw this, although without good reason.

FWIW `static_assert` says "Emscripten 3.1.48 or newer is required.". I only said `3.1.48+` in the PR description, but can update here too.

configure.ac
RReverser: Done & also realised that I can suggest the `wasm32-emscripten` shorthand instead of `wasm32-unknown-emscripten` instead, as they're equivalent.

whitequark:
This would be really nice to have! I'm probably going to look into porting openocd on top of libusb once this PR lands.

RReverser:
> This would be really nice to have! I'm probably going to look into porting openocd on top of libusb once this PR lands.

Oh hi, Catherine!

If you want, you can try to start porting by adding my branch as a submodule for now - that's how we did it in another project - and switching over to mainstream libusb later. 

> Just wondering what you mean here. OpenOCD is using libusb library as base lib to access many supported JTAG/SWD debuggers.

I think they want to use this PR to port OpenOCD to the web.

whitequark:
> If you want, you can try to start porting by adding my branch as a submodule for now - that's how we did it in another project - and switching over to mainstream libusb later.

My final goal is to have this packaged as an extension available from something like https://vscode.dev that also includes YoWASP Yosys and nextpnr as well as Python for Amaranth, all compiled to JS. I think I might want to check the state of all the other moving parts involved first...

tormodvolden:
Is "webusb:" a reasonable platform tag that I can use in the commit summary?

RReverser:
> Is "webusb:" a reasonable platform tag that I can use in the commit summary?

Yeah either `webusb:` or `emscripten:` should do. I'd only be wary of using `wasm:` as it's less future-proof in case we add more Wasm backends.

tormodvolden:
Thanks for all your work on this!

RReverser:
Thank you!

RReverser:
Btw, do I need to do something separately to update [the website](https://libusb.info/) to mention this backend?

I added "and WebAssembly via WebUSB" to the repo README back in the original PR, but I guess the website sources live somewhere separately?

image

whitequark:
@RReverser Have you tried using a batch (as opposed to interactive) application with this patch? It seems that `callMain` returns the first time Wasm blocks on an async call, and it further seems that there's no way to find out when the application *actually* quits (aside from patching it).

whitequark:
Nevermind, I figured it out. You have to do something like this:

```js
    const argv = ['openFPGALoader', ...args, null];
    const argvPtr = inst.stackAlloc(argv.length * 4);
    argv.forEach((arg, off) =>
        inst.HEAPU32[(argvPtr>>2)+off] = arg === null ? 0 : inst.stringToUTF8OnStack(arg));

    const exitCode = await inst.ccall('main', 'number', ['number', 'number'],
        [argv.length - 1, argvPtr], { async: true });
    inst._fflush(0);
```

PR #1340, opened 2023-11-17
Fix additional bad descriptor read issues

addisoncrump:
Extends #1308.

This fixes a few additional potential memory issues associated with bad descriptor reads. I will annotate the commit below.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -163,6 +163,18 @@ static int parse_endpoint(struct libusb_context *ctx,
 	return parsed;
 }
 
+static void clear_interface_descriptor(struct libusb_interface_descriptor *ifp) {
+	free((void *)ifp->extra);
+	if (ifp->endpoint) {
+		uint8_t j;
+
+		for (j = 0; j < ifp->bNumEndpoints; j++)
+			clear_endpoint((struct libusb_endpoint_descriptor *)
+								   ifp->endpoint + j);
+	}
+	free((void *)ifp->endpoint);
+}
+
 static void clear_interface(struct libusb_interface *usb_interface)
 {
 	int i;
@@ -173,15 +185,7 @@ static void clear_interface(struct libusb_interface *usb_interface)
 				(struct libusb_interface_descriptor *)
 				usb_interface->altsetting + i;
 
-			free((void *)ifp->extra);
-			if (ifp->endpoint) {
-				uint8_t j;
-
-				for (j = 0; j < ifp->bNumEndpoints; j++)
-					clear_endpoint((struct libusb_endpoint_descriptor *)
-						       ifp->endpoint + j);
-			}
-			free((void *)ifp->endpoint);
+			clear_interface_descriptor(ifp);
 		}
 	}
 	free((void *)usb_interface->altsetting);
@@ -212,7 +216,19 @@ static int parse_interface(libusb_context *ctx,
 		usb_interface->altsetting = altsetting;
 
 		ifp = altsetting + usb_interface->num_altsetting;
+
+		usb_interface->num_altsetting++;
+
+		ifp->endpoint = NULL;
+		ifp->extra = NULL;
+		ifp->extra_length = 0;
+
 		parse_descriptor(buffer, "bbbbbbbbb", ifp);
+
+		// stash in case we have to clear it
+		uint8_t stashedNumEndpoints = ifp->bNumEndpoints;
+		ifp->bNumEndpoints = 0;
+
 		if (ifp->bDescriptorType != LIBUSB_DT_INTERFACE) {
 			usbi_err(ctx, "unexpected descriptor 0x%x (expected 0x%x)",
 				 ifp->bDescriptorType, LIBUSB_DT_INTERFACE);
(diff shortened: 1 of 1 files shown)


libusb/descriptor.c
```diff
@@ -212,7 +216,19 @@ static int parse_interface(libusb_context *ctx,
 		usb_interface->altsetting = altsetting;
 
 		ifp = altsetting + usb_interface->num_altsetting;
+
+		usb_interface->num_altsetting++;
```
addisoncrump, line 220: Moving this above ensures that the altsetting is freed in the case of an error.

libusb/descriptor.c
```diff
@@ -212,7 +216,19 @@ static int parse_interface(libusb_context *ctx,
...
+
+		ifp->endpoint = NULL;
+		ifp->extra = NULL;
+		ifp->extra_length = 0;
+
 		parse_descriptor(buffer, "bbbbbbbbb", ifp);
+
+		// stash in case we have to clear it
+		uint8_t stashedNumEndpoints = ifp->bNumEndpoints;
+		ifp->bNumEndpoints = 0;
```
addisoncrump, line 230: Stashing this ensures that we clear only the endpoints that are actually instantiated.

libusb/descriptor.c
```diff
@@ -291,23 +302,22 @@ static int parse_interface(libusb_context *ctx,
...
+			endpoint = calloc(stashedNumEndpoints, sizeof(*endpoint));
 			if (!endpoint) {
 				r = LIBUSB_ERROR_NO_MEM;
 				goto err;
 			}
 
 			ifp->endpoint = endpoint;
-			for (i = 0; i < ifp->bNumEndpoints; i++) {
-				r = parse_endpoint(ctx, endpoint + i, buffer, size);
+			for (; ifp->bNumEndpoints < stashedNumEndpoints; ifp->bNumEndpoints++) {
```
addisoncrump, line 315: Making this explicit with the loop ensures that the endpoints are cleared in the case of an error.

libusb/descriptor.c
```diff
@@ -384,7 +393,10 @@ static int parse_configuration(struct libusb_context *ctx,
 	buffer += config->bLength;
 	size -= config->bLength;
 
-	for (i = 0; i < config->bNumInterfaces; i++) {
+	uint8_t stashedNumInterfaces = config->bNumInterfaces;
+	config->bNumInterfaces = 0;
+
+	for (; config->bNumInterfaces < stashedNumInterfaces; config->bNumInterfaces++) {
```
addisoncrump, line 399: Same as previous: ensures that the interfaces are cleared in the case of an error.

libusb/descriptor.c
```diff
@@ -1218,7 +1229,7 @@ static int parse_iad_array(struct libusb_context *ctx,
 
 	// First pass: Iterate through desc list, count number of IADs
 	iad_array->length = 0;
-	while (consumed < size) {
+	while (consumed + 1 < size) {
```
addisoncrump, line 1232: Prevents off-by-one heap overflow in the case that the amount consumed less than size, but by less than the descriptor we're about to read.

libusb/descriptor.c
```diff
@@ -1242,7 +1253,7 @@ static int parse_iad_array(struct libusb_context *ctx,
 		// Second pass: Iterate through desc list, fill IAD structures
 		consumed = 0;
 		i = 0;
-		while (consumed < size) {
+		while (consumed + 9 < size) {
```
addisoncrump, line 1256: Prevents nine byte eight heap overflow in the case that the amount consumed less than size, but by less than the descriptor we're about to read.

addisoncrump:
Fair point, it hadn't been commented in my previous PR so I just assumed it wasn't necessary here. Will update and add a comment to my previous PR section as well.

tormodvolden:
Well, in the previous PR https://github.com/libusb/libusb/pull/1308/files the changed code is quite self-explanatory. This is not always the case here IMO. If there are e.g. subtle ordering requirements it is good to spell that out, since someone else might come later and change the code without thinking about that.

Am I the only one who think we could find better names than stashed_xxx multiple times? I understand it is for saving or backing up a value, but maybe e.g. the reason for doing so could be reflected in the name? Anyway it is just a matter of taste, and sometimes `i` is the best loop counter name.

addisoncrump:
This macOS test seems to be flaky... It did not fail in the previous CI run, and all that was changed in the last commit was some formatting items.

tormodvolden:
Yes, nothing to worry about. It is some kind of race condition that kicks in sometimes, and maybe limited to virtual machines.

libusb/descriptor.c
tormodvolden: Line 228 says: stash /in case/ we have to clear it. Then line 231 goes on to clear it. This is confusing.

> bNumEndpoints will represent the number of *initialised* endpoints

Can't we call it e.g. numInitialisedEndpoints then, instead of "stashedNumEndpoints"?

libusb/descriptor.c
tormodvolden: All comments should be /* */.

seanm:
@tormodvolden Looks like some of this will overlap/conflict with https://github.com/libusb/libusb/pull/1460.  I propose getting [1460](https://github.com/libusb/libusb/pull/1460) merged first, then rebase this one.

tormodvolden:
Yes, this one needs some fixing up now that parse_descriptor() is gone.

seanm:
The buffer overruns this PR fixes are hopefully already fixed by https://github.com/libusb/libusb/pull/1460

addisoncrump:
Sadly the code I wrote to test these cases is now wildly out of date. I will need to update it sometime after next week.

mcuee:
@addisoncrump 

Just wondering if you want to further persue this PR. Thanks.

PR #1342, opened 2023-11-19
Compiling libusb for wasm32/Emscripten now outputs a static library

andrewsomahony:
A static library allows libusb to be linked with other wasm projects to use USB within the web browser.  SIDE_MODULE and PIC were added as options to facilitate this, and preventing callMain from executing is no longer required.


RReverser:
> A static library allows libusb to be linked with other wasm projects to use USB within the web browser. SIDE_MODULE and PIC were added as options to facilitate this

Hm this doesn't sound right. It was already outputting a static library before this change.

SIDE_MODULE is doing the opposite - it's generating a "shared" library using Emscripten's custom dynamic linking implementation, but it's generally not recommended as it's much more costly and error-prone than static linking in WebAssembly world.

> and preventing callMain from executing is no longer required.

This was needed for tests & examples, not for final binaries, and still needed regardless of static vs dynamic linking.

andrewsomahony:
> > A static library allows libusb to be linked with other wasm projects to use USB within the web browser. SIDE_MODULE and PIC were added as options to facilitate this
> 
> Hm this doesn't sound right. It was already outputting a static library before this change and that's how it's used in all existing applications.
> 
> SIDE_MODULE is doing the opposite - it's generating a "shared" library using Emscripten's custom dynamic linking implementation ("side modules" which are a combination of JS+Wasm), but it's generally not recommended as it's much more costly and error-prone than static linking in the WebAssembly world.
> 
> Besides, to make this dynamic linking work, you'd need the applications to be built as MAIN_MODULE as only they can load those special "side modules", and I don't see that in the PR.
> 
> > and preventing callMain from executing is no longer required.
> 
> This was needed for tests & examples, not for library itself, and still needed regardless of static vs dynamic linking.

Hi Ingvar,

You are absolutely correct about SIDE_MODULE, and PIC was also not needed for static linking, so those were my mistake.  However, callMain is not included in the exports in the static library, so while the test "exe" (the HTML/JS output) requires it, an independent static library does not.  I'm not as experienced with automake as I am with CMake, so is there a way to filter out whether we are compiling the tests or the library, and I can implement something using it?

PS: Nice article about porting GPhoto, it helped me a lot!

-Andrew

RReverser:
> callMain is not included in the exports in the static library, so while the test "exe" (the HTML/JS output) requires it, an independent static library does not

callMain is Emscripren's own thing, it's just a helper so that users could pass custom command line flags via DevTools. As those are linking flags, they shouldn't affect the static library, as it's not linked, but merely archives all object files together. So it already should apply only to tests & examples as those are the only things linked inside libusb's build, and never the end user. 

Did you run into a case where those flags leak into your application and cause problems for you as an end user? If so, that's definitely a bug.

> Nice article about porting GPhoto, it helped me a lot!

Thanks! Glad you liked it.

BTW, if you're building something with this backend, you might be interested in trying out improvements in https://github.com/libusb/libusb/pull/1339. It could use more testing and feedback!

andrewsomahony:
> > callMain is not included in the exports in the static library, so while the test "exe" (the HTML/JS output) requires it, an independent static library does not
> 
> callMain is Emscripren's own thing, it's just a helper so that users could pass custom command line flags via DevTools. As those are linking flags, they shouldn't affect the static library, as it's not linked, but merely archives all object files together. So it already should apply only to tests & examples as those are the only things linked inside libusb's build, and should never affect the end user.
> 
> Did you run into a case where those flags leak into your application and cause problems for you as an end user? If so, that's definitely a bug.
> 
> > Nice article about porting GPhoto, it helped me a lot!
> 
> Thanks! Glad you liked it.
> 
> BTW, if you're building something with this backend, you might be interested in trying out improvements in #1339. It could use more testing and feedback!

Hi Ingvar,

You are correct; INVOKE_RUN is used as a sort of "preprocessor" macro in Emscripten's JS glue code.

My confusion stemmed from seeing these flags: they aren't needed for a .a file, as you said, it is just a bunch of .o files: this includes anything that would resolve an unresolved symbol, as the .a will be linked with an "exe" and the symbols will be resolved.  I merely cloned what I had in my CMake test setup for my libusb work, as my automake knowledge is a bit sparse: I figured that emcc's command line params had changed since you wrote that article.

I will have a look at your branch, thanks!  There is no need for this one.

-Andrew

PR #1347, opened 2023-11-25
Test open/transfer/close in stress_mt

RReverser:
Previously stress_mt was only testing that you can enumerate devices in parallel.

With this change it will also verify that you can open a device, do some transfer (a device descriptor control transfer was chosen as it should succeed with any connected device), and close a device from multiple threads in parallel as well.

This is particularly useful since libusb allows to `libusb_open` same device multiple times under different handles nowadays, but given an incorrect backend implementation it can accidentally e.g. close the actual underlying device on the first `libusb_close`, leading to transfer errors on other threads.

Adding those tests helped me catch couple more bugs related to the above in the WebUSB backend, and I figured it will be useful for other backends too.


RReverser:
@mcuee This one should be good to go. Let's merge it so that I can rebase https://github.com/libusb/libusb/pull/1349 on top of it and verify that those new tests work on all platforms too.

mcuee:
@tormodvolden 

Please help to review as well. Thanks.

tormodvolden:
Great, more test coverage is much welcome. That this helped you find bugs shows that this is useful for real.

Hmm, I kind of agree with the MSVC compiler that e.g. "if((ti->err = libusb_get_device_descriptor(dev, &desc)))" easily can be read by a human as a misspelled comparison, even with the extra parentheses, or the reader at least get distracted wondering if it is one. I would prefer adding != 0 if that pleases the compiler, or split it in two statements.

tormodvolden:
Just testing if that explicit !=0 helps any, we can drop or squash this later.

tormodvolden:
It is problematic to have this enabled by default because it requires that the user running stress_mt has write access to /all/ USB devices.

RReverser:
> because it requires that the user running stress_mt has write access to /all/ USB devices.

Hm I'd generally expect that control transfers shouldn't be problematic but I can see how on some backends (USBFS?) it would require more permissions than just listing devices.

I don't mind making it another option / level on top of `enumerate`.

tormodvolden:
Yes, I think another option would be nice. Even better if there's a way to specify a device (or several) to test against, but surely it would be more complex.

RReverser:
> Let's merge it so that I can rebase #1349 on top of it and verify that those new tests work on all platforms too.

@tormodvolden FWIW I was mistaken, it won't make a difference since both are operating on empty list of devices for now anyway, so open/transfer/close won't do anything on CI.

#1349 can be reviewed independently.

RReverser:
@tormodvolden I decided to try a slightly different approach: instead of either erroring out when device is readonly (which is not very useful error if some devices are genuinely inaccessible by the user) or having a mode that skips those tests altogether, I'm now showing a warning once per device only on `LIBUSB_ERROR_ACCESS` and skipping those subtests for such devices.

It's slightly more complicated than skipping all such tests altogether, but I believe it provides a better balance between testing as much as the environment allows vs hinting to the user what they could do for deeper testing.

Let me know what you think.

RReverser:
Ah old MSVC versions don't have stdatomic... Sigh.

tormodvolden:
> showing a warning once per device only on `LIBUSB_ERROR_ACCESS`

I think this is a great way of doing it. Granted, it won't test for somehow failing access that should have been working, but that we can consider safely out of scope for this test.

RReverser:
Ah there are Windows-only `InterlockedExchange*` functions that I can fallback to.

tormodvolden:
See for example usbi_atomic_store in libusb/libusbi.

RReverser:
Ah thanks. Looks like the same idea with falling back to Windows-only functions. I'll do that then.

tormodvolden:
Maybe it will work well enough If you only set a flag (int or long?) without atomics? (That's what we do on MSC anyway). You could then check the flags after all threads are finished and only then print the warning.

RReverser:
> You could then check the flags after all threads are finished and only then print the warning.

Oh, that's a great idea too. Although then I'd need to save device descriptors in some global table, or just not print VID:PID at all... I'll try atomic fallback first, but if it turns too difficult, I'll switch to this alternative approach.

hjelmn:
MSVC is the bane of all open source programs. It was literally the LAST compiler to support C99 and Microsoft took their sweet time with C11 as well. Intel was nearly as bad but MSVC was just &*#&$#.

tormodvolden:
> it queues all builds, even if new commit is added to the branch

Good to know, thanks. Usually it is not a problem for us since there are not so many PRs being pushed and repushed all the time. It would probably help Appveyor if they made "rolling builds" a global default though. We could probably enable it anyway, the only and very minor disadvantage I can see is that you won't see so easily afterwards which of the repushes actually broke the build. Our Appveyor is administered by @LudovicRousseau and @dickens
Actually it looks like they both have jobs running, which seems like a waste of Appveyor resources?
- https://ci.appveyor.com/project/dickens/libusb/history
- https://ci.appveyor.com/project/LudovicRousseau/libusb/history

RReverser:
> I would have printed the VID:PIDs with %x not %X.

Sure, I don't have a strong preference. Although it does mean adding yet another push to that long AppVeyor queue...

tormodvolden:
Looks good, and I have tested it.

To add insult to the injury, I added another commit. What do you think about it? It makes it more helpful when testing manually.

tormodvolden:
Oh no, they are smarter than me. If all threads would fail, last_devcount would be uninitialized.

RReverser:
Yeah this is the primary reason I dislike the style of declaring all variables at top (although I know it was the only accepted option in some old C standards).

It's way too easy to leave uninitialized variables or access them before they're initialized compared to when they are colocated with initializer assignment itself.

Checking the device count sounds like a good idea though.

tests/stress_mt.c
```diff
@@ -138,21 +221,26 @@ static int test_multi_init(int enumerate)
...
+			if (t > 0 && tinfo[t].devcount != last_devcount) {
+				devcount_mismatch++;
+				printf("Device count mismatch: Thread %d discovered %ld devices instead of %ld\n",
 					tinfo[t].number,
-					tinfo[t].iteration);
-			} else {
-				printf("Thread %d discovered %ld devices\n",
-				       tinfo[t].number,
-				       (long int) tinfo[t].devcount);
+					(long int) tinfo[t].devcount,
```
RReverser, line 228: Why are those casts necessary btw? Wouldn't it be easier to change the format string?

tormodvolden: IIRC some compilers (on non-Posix platforms) don't have the %z support for ssize_t.

tormodvolden: On Debian (bookworm) with gcc version 12 this is no longer a problem. I use the latter for compile testing, but the former is my daily driver so I'd like to keep it supported.

RReverser: Ah I thought device count fields are just an int, didn't notice ssize_t. Going over int limits is unrealistic anyway.

RReverser:
> Maybe a future improvement would be to synchronize the threads after enumeration and then let them all loose on the devices at the same time, for maximum stress levels?

Yeah sounds like a good idea.

RReverser:
> Oh, the stress test fails on macOS in #1357. @hjelmn - any ideas?

Heh, that's interesting, especially considering that CI doesn't have any emulated devices yet so it can't be the new open/transfer/close operations that are failing.

tormodvolden:
No, only whitespace (and version_nano).
```
$ git fetch https://github.com/RReverser/libusb.git stress-mt-open-transfer-close
From https://github.com/RReverser/libusb
 * branch            stress-mt-open-transfer-close -> FETCH_HEAD
$ git diff -w FETCH_HEAD master~3
diff --git a/libusb/version_nano.h b/libusb/version_nano.h
...
```

RReverser:
Odd. I guess must be a timing issue then and I happened to be lucky on this PR.

tormodvolden:
I think it is a race in the macOS backend that sometimes materialize, sometimes not.

RReverser:
> I think it is a race in the macOS backend that sometimes materialize, sometimes not.

So, the stress test is doing its job in triggering those as expected 😅

tormodvolden:
If you look at the macOS build log it looks random: https://github.com/libusb/libusb/actions/workflows/macos.yml

tormodvolden:
More details (and I'd suggest continued discussion) of the macOS failure in #1358.

tormodvolden:
> Heh, that's interesting, especially considering that CI doesn't have any emulated devices yet so it can't be the new open/transfer/close operations that are failing.

It is trying to open something so maybe there are real devices exposed on the CI host, that sometimes are busy.

PR #1350, opened 2023-11-26
Match any device in hotplugtest by default

RReverser:
I spent some time trying to figure out why, when I run this example, it's not showing any events, until I looked into the code and saw it has some hardcoded vid:pid.

Most people probably don't have the specific device hardcoded here, and it seems better to default to showing events for any device unless overridden.

While at it, also updated device detached message to match the attached one to show vid:pid.

### DIFF
--- a/examples/hotplugtest.c
+++ b/examples/hotplugtest.c
@@ -60,10 +60,18 @@ static int LIBUSB_CALL hotplug_callback(libusb_context *ctx, libusb_device *dev,
 
 static int LIBUSB_CALL hotplug_callback_detach(libusb_context *ctx, libusb_device *dev, libusb_hotplug_event event, void *user_data)
 {
+	struct libusb_device_descriptor desc;
+	int rc;
+
 	(void)ctx;
 	(void)dev;
 	(void)event;
 	(void)user_data;
+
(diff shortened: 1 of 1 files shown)


tormodvolden:
Thanks, I also think this makes sense.

While at it, I added some error string printing, which I noticed was missing when I tested this.

mcuee:
* https://github.com/libusb/libusb/pull/1406#issuecomment-1874646894

It would be good to apply the patch from @sonatique in the above discussion.

sonatique:
> It would be good to apply the patch from @sonatique in the above discussion.

@mcuee would you like me to make a sperate PR with just this patch?

mcuee:
> > It would be good to apply the patch from @sonatique in the above discussion.
> 
> @mcuee would you like me to make a sperate PR with just this patch?

Yes, I think that is a good idea. Thanks.

sonatique:
@mcuee : https://github.com/libusb/libusb/pull/1425

mcuee:
@RReverser and @tormodvolden 

Somehow I think this PR is flawed for Linux.

Please refer to the test results here.
https://github.com/libusb/libusb/pull/1425#issuecomment-1879482085

PR #1352, opened 2023-11-28
core: do not use m4 to define PRINTF_FORMAT

hjelmn:
The PRINTF_FORMAT macro is intended to allow changing how libusb gets warnings about coding warnings for internal logging methods. At this time there is only a single definition and it is always used. It is better to just define it in libusbi.h and not bother with the m4.


examples/ezusb.h
```diff
@@ -100,7 +100,7 @@ extern int ezusb_load_eeprom(libusb_device_handle *device,
 /* Verbosity level (default 1). Can be increased or decreased with options v/q  */
 extern int verbose;
 
-extern void logerror(const char *format, ...) PRINTF_FORMAT(1, 2);
+extern void logerror(const char *format, ...);
```
neheb, line 103: Why remove it here?

tests/libusb_testlib.h
```diff
@@ -20,7 +20,8 @@
 #ifndef LIBUSB_TESTLIB_H
 #define LIBUSB_TESTLIB_H
 
-#include <config.h>
+#include "config.h"
+#include "libusbi.h"
```
tormodvolden, line 24: I guess this is the disadvantage of not using m4 (and config.h), now "user" programs need to include this internal header file.

examples/ezusb.h
hjelmn: I don't think it is actually useful to have it in this particular case. I am not 100% convinced this file should even be including config.h really. Let me run through the code one more time and think this through a bit more. I am not a fan of overloading config.h with things that we shouldn't define using m4. This macro is a good example since the different versions can be determined with a simple #if check.

hjelmn:
I am leaning toward a new header: libusb_compat.h that will hold the two compatibility macros we currently define in config.h. These are `PRINTF_FORMAT` and `DEFAULT_VISIBILITY` neither of which would exist except that they differ on windows.

libusb/libusbi.h
```diff
@@ -310,6 +310,8 @@ int usbi_vsnprintf(char *dst, size_t size, const char *format, va_list args);
 #define LIBUSB_PRINTF_WIN32
 #endif /* defined(_MSC_VER) && (_MSC_VER < 1900) */
 
+#define PRINTF_FORMAT(a,b) __attribute__((__format__ (__printf__, a, b)))
```
Youw, line 313: shouldn't it be under `ifder`/compiler-specific?
I believe this is `__GNUC__`-specific - MSVC doesn't support it

Youw: and I believe we need an `#else` fallback as well

seanm:
+1 for not using m4.

How about just #defining PRINTF_FORMAT in libusb.h, just like LIBUSB_DEPRECATED_FOR and ZERO_SIZED_ARRAY.

tormodvolden:
This is still WIP or under discussion, right? Feel free to un-DRAFT it when it seems ready,

examples/ezusb.h
seanm: Removing this takes away the compiler's ability to warn about messing up the format string, and format string screwups can be a source of security issues.

I agree we don't need m4, autotools, cmake or anything else, we just need to create a WeWishCHadThis.h with a bunch of #if / #else to create a `PRINTF_FORMAT` macro.

PR #1353, opened 2023-11-29
Add Windows MSVC build to Github Actions

### LINKED ISSUE #1045: github actions for Window

RReverser:
Adds Windows build to Github actions.

Closes #1045.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,32 @@
+name: Windows build and Package
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os:
+          - 'windows-2019'
+          - 'windows-2022'
+        configuration:
+          - 'Debug'
+          - 'Release'
+        architecture:
+          - 'x86'
+          - 'x64'
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup MSBuild
+        uses: microsoft/setup-msbuild@v1.3
+        with:
+          msbuild-architecture: ${{ matrix.architecture }}
+
+      - name: Build solution
+        run: |
+          msbuild msvc/libusb.sln /m -property:Configuration=${{ matrix.configuration }}


.github/workflows/windows.yml
```diff
@@ -0,0 +1,52 @@
+name: Build and Package
```
Youw, line 1:
```suggestion
name: MSVC Build and Package
```

RReverser:
Ah right, Github runs actions defines on master branch not from PRs, which is sensible from the security standpoint but means we won't know if this works until it's merged. Only someone with push access will be able to tweak those configs properly.

Although maybe I can try to enable GH actions on my fork separately...

.github/workflows/windows.yml
```diff
@@ -0,0 +1,52 @@
...
+          - '2015'
+          - '2017'
+          - '2019'
+          - '2022'
+        configuration:
+          - 'Debug'
+          - 'Release'
+        architecture:
+          - 'x86'
+          - 'x64'
```
Youw, line 25: I believe Github only allows 15 concurent jobs defined by `matrix`
here we have 20
that is either has to be split into different jobs or drop some of the configuration combinations  (e.g. run Debug only for x64)

RReverser: > I believe Github only allows 15 concurent jobs defined by `matrix`

Huh, TIL, thanks.

> drop some of the configuration combinations (e.g. run Debug only for x64)

Yeah to me it seems sensible to drop Release for most modes and add only one we want to package, but for now I was trying to migrate what Appveyor did 1:1 (minus the Appveyor part).

Youw: unless libusb Org has different limitations...
all 20 seem to spawn just fine, resolving for now

RReverser:
> which is sensible from the security standpoint but means we won't know if this works until it's merged. Only someone with push access will be able to tweak those configs properly.
> 
> Although maybe I can try to enable GH actions on my fork separately...

Nvm, it was a bug in my config. It does try to run them on the PR now.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,49 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup MSBuild
+        uses: microsoft/setup-msbuild@v1.1
```
Youw, line 29: maybe:
```suggestion
        uses: microsoft/setup-msbuild@v1.3
```
?

RReverser: Stop being quicker than me 😅

But yeah, looks like their README needs an update.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,49 @@
+name: Windows build and Package
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    strategy:
+      matrix:
```
Youw, line 10: add `fail-fast: false`

RReverser:
It looks like MSBuild might be not supported on Github Actions unless you use self-hosted runner, in which case this was a pointless exercise. Reading through their issues for now... https://github.com/microsoft/setup-msbuild/issues/20

.github/workflows/windows.yml
RReverser: Hm do you think it's just that specific versions can't be found? I guess worth a try.

Youw:
> MSBuild might be not supported on Github Actions

HIDAPI [successfully](https://github.com/libusb/hidapi/blob/7011fa98af2dde00c298105735e470de800288c7/.github/workflows/builds.yml#L329) using it. I believe the limitation is that not all versions of Visual Studio are pre-installed on self-hosted runners.
We might need to limit the runs only to those few available by self-hosted runners.

.github/workflows/windows.yml
RReverser: No, that wasn't it. It actually can't find MSBuild. See linked issues.

tormodvolden:
Please see also #1234 

I am not particularly interested in migrating to GitHub, especially if it doesn't provide artifacts like AppVeyor does.

RReverser:
> especially if it doesn't provide artifacts like AppVeyor does.

It does, see the action code.

RReverser:
> Please see also #1234

Ah didn't see that previous attempt. Is there TL;DR on why it was closed?

mcuee:
> > Please see also #1234
> 
> Ah didn't see that previous attempt. Is there TL;DR on why it was closed?

It was closed as I do not know much about this thing and I can not address the issues raised.

RReverser:
Ohh I know what the issue is. Fixing...

RReverser:
Ok so the problem was that setup-msbuild takes VS version as in their internal version, not product name (12 instead of 2013 and so on).

This fixes finding MSBuild, but, unfortunately, it looks like only VS 2022 is installed in their runner, so this won't allow testing all configs.

We could install other VS version manually, but that would be pointless as it would be too slow.

I think we'll need to keep Appveyor for the matrix of all versions, but at least can use GH action for basic Windows testing with latest VS.

RReverser:
@Youw Basically what you said

> I believe the limitation is that not all versions of Visual Studio are pre-installed on self-hosted runners.

except it's not even a couple of versions, but only the latest.

mcuee:
> Ah right, Github runs actions defines on master branch not from PRs, which is sensible from the security standpoint but means we won't know if this works until it's merged. Only someone with push access will be able to tweak those configs properly.

It is actually running for PRs as well.

For example:
https://github.com/libusb/libusb/actions/runs/7033676554

RReverser:
> It is actually running for PRs as well.

Yeah I figured some comments above. It's running on this PR now.

RReverser:
> Yes that is the way to go.

Ok since we're keeping Appveyor anyway, I've removed version matrix and artifacts from this config then, so it provides only testing across x86/x64 and debug/release.

RReverser:
Although now there's a lot less value in this Github Action... Not sure if it's still worth it if we're still running AppVeyor anyway.

Unless AppVeyor is moved to only run on commits on master or something?

Youw:
https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md
MSVC 2019 is availabel if [windows-2019](https://github.com/actions/runner-images#available-images) is used - that's something.

But I guess that's it :)

Long-term we could use GithubActions to test latest available MSVC and AppVeyour to test on what's not avilable on GithubActions.
Of course it has its cons though (supportfing different CI scripts).

Also we need to thing if we even need a CI for old MSVC for current latest releases/builds.
If is not like we have a CI that checks build on gcc-4 or so...

hjelmn:
I think we are ok without gcc 4 given how old that is now. Not sure how many version of MSVC are actually useful to test. Is it something that is free or do users have to pay to upgrade? If they have to pay I can see testing multiple versions but if they are free there is not much of a reason to support older versions.

But I don't pay any attention to the Windows testing so whatever works is fine by me.

Youw:
It depends. One you can use MSVC build toolchain for free.
Visual Studio IDE has a free license which is not free for corpodate usage.
There is also a cost of upgrade - even if you have a fine code-base, upgrading to newer MSVC would most likely mean fixing some compilation errors/new warnings, etc., and sometime (when the code-base is huge) it is unaffirdable to upgrade at the moment.
On my current place of work I'm actively preparing code-base and builds infra to update from MSVC2019 to MSVC2022 and that takes a good amount of time.
On the other hand - if one cannot afford to upgrade their build infra, what is the chance one need to upgrade libusb specifically to the very latest version.

I'd say testing for MSVC 2019 and MSVC 2022 should be fine for all newest releases. Of we happen to break something e.g. for MSVC 2015 and people really need it - they can always contribute a fix, even if we don't have a CI for it.

mcuee:
> I'd say testing for MSVC 2019 and MSVC 2022 should be fine for all newest releases. If we happen to break something e.g. for MSVC 2015 and people really need it - they can always contribute a fix, even if we don't have a CI for it.

Yes that is a good plan.

On the other hand, we can always go with VS2022 first with this PR and later add VS2019 build.

As of now, I think the idea is to keep AppVeyor.

.github/workflows/windows.yml
Youw: It is not about MSBuild, it is about not canceling all matrix build if only one of the is failng.
Let others continue/finish independently.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os:
+          - 'windows-2019'
+          - 'windows-2022'
```
Youw, line 11: that looks about right

Youw:
The CI won't start unti the conflict is resolved.

> Eventually I'd like to see AppVeyor going away (for sure after 1.0.27 release, and most likely after 1.0.28 release).

Then lets not touch the AppVeyor build config and only introduce new GithubAction in addtion to what we already have.

RReverser:
I know, thanks :)

RReverser:
In retrospect, I probably should have marked this PR as draft to make it clear it's not quite ready for review but rather just me playing with CI config, as is common with CIs 😅

mcuee:
> > Eventually I'd like to see AppVeyor going away (for sure after 1.0.27 release, and most likely **after** 1.0.28 release).

Sorry I mean to say "and most likely **before** 1.0.28 release"

> Then lets not touch the AppVeyor build config and only introduce new GithubAction in addtion to what we already have.

Yes, that is better.

Youw:
> What if we at least remove VS2019 & VS2022 debug (where we don't care about artifacts) from AppVeyor?

Sound fine to me.

RReverser:
Reverted AppVeyor change altogether for now to keep things simple.

> > What if we at least remove VS2019 & VS2022 debug (where we don't care about artifacts) from AppVeyor?
> 
> Sound fine to me.

Might do this next.

Before spending more time on this, would be good to hear from @tormodvolden on whether this is worth pursuing further at all.

mcuee:
@tormodvolden 

I think this PR is good to go now. It will not replace AppVeryor but it can give the PR author a faster feedback on the Windows MSVC build.

minicarpet:
What about this ? can it be merged as well ? so we will have all worklow into github.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,32 @@
...
+        configuration:
+          - 'Debug'
```
sonatique, line 13: There are more configuration that those 2, because we have the -MT ones as well.
But I am ok to only use these 2 ones here.

.github/workflows/windows.yml
sonatique, line 11: Are these github names for Windows 10 and Windows 11 ? Are these the most recent ones?

sonatique (approved):

.github/workflows/windows.yml
minicarpet: The last currently version is windows-2022. we can use windows-latest as well.

.github/workflows/windows.yml
minicarpet: We can still update later (to add hotplug too 😃)

.github/workflows/windows.yml
minicarpet: Sorry, there is windows-2025 but not set to windows-latest (https://github.com/actions/runner-images). windows-2019 seems deprecated ?

sonatique: I have to say I don't really know. These seems to be Windows *server* version anyway, not common Windows 10 and 11

mcuee: I think we have to drop windows-2019.
* https://github.com/actions/runner-images/issues/12045

Now only the following two are supported.
Windows Server 2022
Windows Server 2025

.github/workflows/windows.yml
mcuee: I think we can keep only windows-2022 now that Windows 2019 github runner has been deprecated and is no longer supported.
https://github.com/actions/runner-images/issues/12045

.github/workflows/windows.yml
mcuee: I think we can keep only windows-2022 now that Windows 2019 github runner has been deprecated and is no longer supported.

sonatique: So: shall I remove line 10 'windows-2019" and merge? What about 'windows-2025'?
Shall we make it windows-2022 and windows-2025?

mcuee: Yes, I think this is the way to go.

sonatique: @mcuee : I am not the original author. I don't know whether it's fine for me to do the change here directly.
@minicarpet :I can handle the last change and finalize the PR,  ok for you?

RReverser: @sonatique Feel free to push changes directly to the PR, yeah, no problem from my side.

seanm: Better to have something than nothing; let's get a first draft merged ASAP.

minicarpet: Ok for me !

sonatique:
Well so the idea of merging this with windows-2022 and windows-2025 was not that good actually:
https://github.com/libusb/libusb/actions/runs/17219592028
I was under the impression that this has been tested, at least for 2022.

OK, so I'll try to figure out some fixes for it

minicarpet:
Well the réal question is: do libusb should support 32 bits anymore ?

sonatique:
> Well the réal question is: do libusb should support 32 bits anymore ?

That is a good question indeed. The issue with Windows action is related to ARM 32 only. Win 32 (x86) is still fine.

I fixed things here: https://github.com/libusb/libusb/pull/1681

PR #1362, opened 2023-12-05
ci: dump test suite output on test failure

hjelmn:
The test output is hidden by default with make check so we need to explicitly dump the test suite log on failure.

### DIFF
--- a/.private/ci-build.sh
+++ b/.private/ci-build.sh
@@ -83,6 +83,9 @@ if [ "${test}" = "yes" ]; then
 	# Load custom shim for WebUSB tests that simulates Web environment.
 	export NODE_OPTIONS="--require ${scriptdir}/../tests/webusb-test-shim/"
 	make check
+	if test "$?" != "0" ; then
+	    cat tests/test-suite.log
+	fi
 fi
 
 if [ "${install}" = "yes" ]; then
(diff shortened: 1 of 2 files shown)


hjelmn:
Quick test to make sure this does what it is supposed to do in the success case.

tormodvolden:
This didn't seem to work, telling from your latest repushes on #1266.

```
Testsuite summary for libusb-1.0 1.0.26
# TOTAL: 5
# PASS:  4
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
See tests/test-suite.log
make[3]: *** [test-suite.log] Error 1
Please report to libusb-devel@lists.sourceforge.net
make[2]: *** [check-TESTS] Error 2
make[1]: *** [check-am] Error 2
make: *** [check-recursive] Error 1
Error: Process completed with exit code 2.
```

hjelmn:
Yeah, strange. I checked it by hand and it worked. Will be looking into it shortly. I want to do a bit more and maybe set LIBUSB_DEBUG and run again as the default CI behavior. WDYT?

hjelmn:
set -e is responsible. i need to disable it before running the tests. See #1363

PR #1363, opened 2023-12-05
ci: disable set -e when running make check

hjelmn:
In order to dump the test log need to temporarily disable set -e to ensure the script runs the next line. This commit adds set +e before running the tests and set -e after.

### DIFF
--- a/.private/ci-build.sh
+++ b/.private/ci-build.sh
@@ -82,9 +82,9 @@ make -j4 -k
 if [ "${test}" = "yes" ]; then
 	# Load custom shim for WebUSB tests that simulates Web environment.
 	export NODE_OPTIONS="--require ${scriptdir}/../tests/webusb-test-shim/"
-	make check
-	if test "$?" != "0" ; then
+	if ! make check ; then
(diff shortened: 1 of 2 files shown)


tormodvolden:
To avoid set -e toggling you can also do:
```
if ! make check; then
    cat tests/test-suite.log
    exit 1
fi
```

hjelmn:
> To avoid set -e toggling you can also do:
> 
> ```
> if ! make check; then
>     cat tests/test-suite.log
>     exit 1
> fi
> ```

Indeed. Done.

tormodvolden:
Now if only the CI test could fail so it can see this is working :) I've repushed to see.

hjelmn:
Let me add a change that intentionally breaks a test to verify.

hjelmn:
Verified locally. No need to push the commit. Just made init_context always fail and ran:

`.private/ci-build.sh --build-dir ${PWD}/build`

tormodvolden:
And when you merged it to master, the stress_mt failed, so we got verified that your PR was correct :)

PR #1368, opened 2023-12-08
windows: when usb debug is on in android, libusb cannot open the android device

dayo7116:
If the usb debug is on in android device, libusb fails to open the android device on windows.

This PR fixes the problem. Please Review

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2688,7 +2688,7 @@ static int winusbx_open(int sub_api, struct libusb_device_handle *dev_handle)
 {
 	struct winusb_device_priv *priv = usbi_get_device_priv(dev_handle->dev);
 	struct winusb_device_handle_priv *handle_priv = get_winusb_device_handle_priv(dev_handle);
-	HANDLE file_handle;
+	HANDLE file_handle = INVALID_HANDLE_VALUE;
 	int i;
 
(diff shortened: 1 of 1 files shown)


tormodvolden:
Please provide a description of what this change does in general, not only the problem that you are seeing. Also explain what this has to do with "usb debug is on in android device".

libusb/os/windows_winusb.c
```diff
@@ -2708,9 +2708,11 @@ static int winusbx_open(int sub_api, struct libusb_device_handle *dev_handle)
 				default:
 					return LIBUSB_ERROR_IO;
 				}
+			} else if (INVALID_HANDLE_VALUE == file_handle){
```
tormodvolden, line 2711: It is a bit confusing that you use Yoda syntax here, while using normal syntax in line 2701.

Youw:
I believe the former logic was:
- try to open all interfaces;
- if any of them fails to open - fail the entire `winusbx_open`;

After the update:
- try to open all interfaces;
- if at least one (first) interface successfully opened - ignore all subsequent interface open failures;

The new implementation is a bit confusing to me. I'd replace `file_handle` with a bool flag instead.

Since the new logic no longer guarantees that `handle_priv->interface_handle[i].dev_handle` has only valid values, something else may break in a different place. Don't know the code base but it should be considered and checked.

PR #1374, opened 2023-12-11
tests/set_option: avoid use-after-free in case of test failure

aurel32:
In case num_devices equals 0, LIBUSB_EXPECT() calls LIBUSB_TEST_CLEAN_EXIT(), which calls libusb_exit() on test_ctx which has just been freed a few lines above by a call to libusb_exit(). This might cause a SEGFAULT or SIGBUS depending on the system.

Fixes that by assigning NULL to test_ctx just after calling libusb_exit() like it is done elsewhere in the file.

### DIFF
--- a/tests/set_option.c
+++ b/tests/set_option.c
@@ -174,6 +174,7 @@ static libusb_testlib_result test_no_discovery(void)
   ssize_t num_devices = libusb_get_device_list(test_ctx, &device_list);
   libusb_free_device_list(device_list, /*unref_devices=*/1);
   libusb_exit(test_ctx);
+  test_ctx = NULL;
 
   LIBUSB_EXPECT(>, num_devices, 0);
 


RReverser:
Hm I wonder why this wasn't picked by ASAN which is enabled on CI and CI does have `num_devices == 0`.

tormodvolden:
> Hm I wonder why this wasn't picked by ASAN which is enabled on CI and CI does have `num_devices == 0`.

Because that code is not included if CI_WITHOUT_DEVICES?

RReverser:
> Because that code is not included if CI_WITHOUT_DEVICES?

Oh right, I got confused and thought it was stress_mt that had this extra condition. Makes sense.

tormodvolden:
Another reason for making the CI_WITHOUT_DEVICES a run-time conditional :)

RReverser:
Maybe that test just should never error out when there are no devices? Would be simpler for all use cases.

tormodvolden:
> Maybe that test just should never error out when there are no devices? Would be simpler for all use cases.

Yes, I agree. We simply cannot assume there are USB devices on the system. We can warn about it so that a user running the test manually can see that no USB devices are detected and judge themselves if that is a correct outcome.

PR #1360, opened 2023-12-05
Run tests for AppVeyor too

RReverser:
Follow-up to #1349.

PR #1378, opened 2023-12-13
windows: Base HID device descriptor on cached values

tormodvolden:
Instead of filling in the blanks with hard-coded made-up values that are sometimes correct, use the cached descriptor values retrieved during enumeration, which should provide more correct values.

References #1360

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -3498,24 +3498,26 @@ static int _hid_wcslen(WCHAR *str)
 	return i;
 }
 
-static int _hid_get_device_descriptor(struct hid_device_priv *hid_priv, void *data, size_t *size)
+static int _hid_get_device_descriptor(struct libusb_device *dev, struct hid_device_priv *hid_priv, void *data, size_t *size)
 {
 	struct libusb_device_descriptor d;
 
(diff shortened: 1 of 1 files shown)


Youw (approved):
looks reasonable

Hamad-G (approved):

mcuee:
> looks reasonable

I agree. This PR should be good to go.

tormodvolden:
For little bit of background here, the code with these hard-coded values haven't been touched since the HID backend was (re?)introduced in 2012. It is pretty much the same in the oldest libusb-win32 git entry I could find from 2007. Back then it was maybe not practical to get the "true" device descriptor from the hub drivers, I don't know exactly. But no one has complained all this time, so we are not in a rush to change it, and definitely not now while we are in release freeze for 1.0.27. We cannot know what downstream code is dependent on the "made-up" values.

It is also interesting that some HID devices seem to display e.g. a non-standard bcdUSB value. I don't know if it is the device itself or a Windows driver that comes up with these values.

mcuee:
We recommend users to use HIDAPI anyway for USB HID devices. So indeed no rush to change the codes before 1.0.27 release.

PR #1167, opened 2022-07-21
core: Add libusb_get_max_iso_packet_size_for_alt_setting

aib:
As discussed in #36, [old issue #77](https://sourceforge.net/p/libusb/mailman/libusb-devel/thread/056.8f7cf80ce90539817d58effd879129d7@libusb.org/) and [the mailing list](https://sourceforge.net/p/libusb/mailman/libusb-devel/thread/4D52A463.5080108@redhat.com/).

(I don't have much more to add, but I don't want to sound like a bot, so please feel free to review and thank you!)

PR #1388, opened 2023-12-20
Fix function name in ChangeLog

larskanis:
It was renamed to libusb_get_max_alt_packet_size() after discussion in https://github.com/libusb/libusb/pull/1167 .


tormodvolden:
Thanks, good catch!

larskanis:
Force-pushed with signed commit, as this seems to be a requirement.

mcuee:
> Force-pushed with signed commit, as this seems to be a requirement.

You can ignore that warning. You are not required to do that.

seanm (approved):

seanm:
@mcuee @tormodvolden who is able to merge things?

mcuee:
> @mcuee @tormodvolden who is able to merge things?

@tormodvolden and @hjelmn are the two active maintainers who will commit to the git tree now.

tormodvolden:
I merge things from time to time, but it can be a little while in between because I don't always have time. I prefer to give some extra time for darwin and core changes so that @hjelmn have the chance to look at it. If we don't hear from him over longer time I commit it anyway if I am comfortable with the changes.

In this case of a Changelog change, I already have staged some other changes and combine them towards the release instead of lots of small commits during the release freeze. This one is milestoned so I won't forget it.

seanm:
> In this case of a Changelog change, I already have staged some other changes and combine them towards the release instead of lots of small commits during the release freeze.

Why not merge it to master now? I'm not very familiar with github, is it harder than pressing a 'merge' button?

By having things in master sooner, we spare people from re-discovering problems. For example, when I was working on https://github.com/libusb/libusb/pull/1408 and adding comments about new APIs, I read the changelog and searched for instances of new API in the code to see how they were commented, and I didn't find this `libusb_get_max_iso_packet_size_for_alt_setting`.

seanm:
>Sorry that you wasted time looking for the misspelled function.

It was only a few minutes, so no big deal.

Thanks for merging!

PR #1397, opened 2023-12-21
Fix setting the global log callback per `libusb_set_option()`

larskanis:
There are two problems with setting the global callback per `libusb_set_option()` currently:
1. The callback is overwritten by subsequent calls to `libusb_set_option()`.
2. After the first `libusb_init()` the callback is not set globally but for the default context only. There is no way to set the global callback through `libusb_set_option()` then.

Both issues are fixed by the attached commits. See the commit description for the details.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2327,16 +2327,14 @@ int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
 				default_context_options[option].arg.ival = arg;
 			} else if (LIBUSB_OPTION_LOG_CB == option) {
 				default_context_options[option].arg.log_cbval = log_cb;
+				libusb_set_log_cb_internal(NULL, log_cb, LIBUSB_LOG_CB_GLOBAL);
 			}
 			usbi_mutex_static_unlock(&default_context_lock);
 		}
 
 		ctx = usbi_get_context(ctx);
-		if (NULL == ctx) {
-			if (LIBUSB_OPTION_LOG_CB == option)
-				libusb_set_log_cb_internal(NULL, log_cb, LIBUSB_LOG_CB_GLOBAL);
+		if (NULL == ctx)
 			break;
-		}
 
 		switch (option) {
 		case LIBUSB_OPTION_LOG_LEVEL:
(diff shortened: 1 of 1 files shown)


larskanis:
Force-pushed with signed commits, as this seems to be a requirement.

mcuee:
> Force-pushed with signed commits, as this seems to be a requirement.

You can ignore that warning. You are not required to do that.

sonatique:
This looks good to me, thanks @larskanis

mcuee:
@tormodvolden 

Just wondering if you want to look into this for the 1.0.27 release. It seems to be a good patch to take.

larskanis:
> Just wondering if you want to look into this for the 1.0.27 release.

Makes sense. Setting the log callback is a new feature of the 1.0.27 release, but it is broken in the current stage.

tormodvolden:
I am pinging @hjelmn about this, since he introduced this new feature. Surely we'd like to fix this before 1.0.27.

tormodvolden:
I choked on the second commit message. "Was and is" - does this refer to before and after this commit? When calling libusb_set_option ? Since this change is not trivial, it would be nice if the explanation was easier to read.

tormodvolden:
Does this sound correct?

> When setting LIBUSB_OPTION_LOG_CB before the first libusb_init(), 
> the global callback is set.
> 
> However, after the first libusb_init() there is a default context, and
> setting LIBUSB_OPTION_LOG_CB would only set this default context, and
> not the global callback. There would be no way to set the global
> callback through libusb_set_option().
> 
> Therefore make sure the global log callback is set in addition to the 
> default context.
>

tormodvolden:
"... when libusb_set_option is called with a NULL ctx" that is?

Just for my understanding, is the following correct?
If libusb_init is called with a non-null context, this also becomes the default context. And calling libusb_set_option with this context will set for this context (also default context) but not touch the global callback.

tormodvolden:
@larskanis if you can agree on the explanation, I'll amend it when I commit this.

tormodvolden:
@hjelmn confirmed in another channel that the code changes look good to him.

PR #1399, opened 2023-12-22
darwin: Use atomic for in_reenumerate flag

tormodvolden:


seanm:
Thanks! This is indeed what I had in mind, but didn't have a chance to type it up because of end-of-year work craziness.

Let's see if this works.  If so, will need to see how far back these C atomics are supported on macOS, if not far, there are Apple-specific variants that could be used.

mcuee:
Unfortunately it does not fix the issue.
* https://github.com/libusb/libusb/issues/1386#issuecomment-1868176703

tormodvolden:
Thanks for testing! Maybe this still can be seen as an improvement and considered for inclusion, after 1.0.27 some time.

mcuee:
> Thanks for testing! Maybe this still can be seen as an improvement and considered for inclusion, after 1.0.27 some time.

If the PR is reviewed to be correct and fixes one existing code defect, I think we can still merge ti before 1.0.27 release, even though it does not fix issue #1386.

seanm:
I think this change is correct, and makes the code more correct, but I think there's an even bigger threading issue with nearby code. I'd suggest holding off merging this until I dig deeper...

tormodvolden:
I hope we can find a surgical fix for #1386 and soon release 1.0.27 which is otherwise well tested, then apply this and all your recent darwin merge requests afterwards.

seanm:
> I hope we can find a surgical fix for #1386 

Agreed.

>and soon release 1.0.27 which is otherwise well tested, then apply this and all your recent darwin merge requests afterwards.

Some of my patches I think should be considered for 1.0.27, I'll point them out...

tormodvolden:
@seanm @hjelmn Is this still worth merging?

seanm:
I would suggest we first get -Wthread-safety work, i.e. merge https://github.com/libusb/libusb/pull/1461

mcuee:
@tormodvolden and @seanm 
Is this still useful?

PR #1461 has been merged.

seanm:
Using an atomic int here isn't going to be any worse, but is it just papering-over a bigger bug?

I don't oppose merging this. It can't be worse.

seanm:
This introduced a warning, fixed in: https://github.com/libusb/libusb/pull/1822

PR #1407, opened 2023-12-28
Changed some http links to https

seanm:


mcuee:
This should be safe to be merged.

seanm:
> This should be safe to be merged.

I assume so, but note that it doesn't only change comments, it changes code too, notably the string in `libusb_version`

seanm:
I forget to mention. There are more http links remaining that I didn't convert, because they don't work as https. For example, this works:

http://mailing-list.libusb.info

but this does not:

https://mailing-list.libusb.info

The web server probably needs tweaking to properly redirect https too.

tormodvolden:
> it changes code too, notably the string in `libusb_version`

I think this is still safe. The "describe" field with the URL is only there for ABI compatibility, according to a code comment.

PR #1410, opened 2023-12-28
Changed remove_from_flying_list() to not lock flying_transfers_lock itself

seanm:
This is now symmetric with add_to_flying_list(), which was already requiring that *callers* lock flying_transfers_lock. Updated the only 2 callers.

This also makes the code correspond better to the big comment about locking made in 138b661f.

Also documented at the top of several functions when they require that flying_transfers_lock already be held. Verified by code review that it's actually the case.

This should not change any behaviour at all.


sonatique (approved):

tormodvolden:
I think the documentation changes should be included in 1.0.27 and the code change looks like a NOP so it should be safe too and reduce confusion (although not critical for the release).

PR #1412, opened 2023-12-30
locationID should be 32 bits, not 64

seanm:
Fortunately, this only affected a log message.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -625,8 +625,6 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) {
   struct darwin_cached_device *old_device;
 
   io_service_t device;
-  UInt64 session, locationID;
-  int ret;
 
   usbi_mutex_lock(&active_contexts_lock);
 
@@ -634,7 +632,9 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) {
     bool is_reenumerating = false;
 
     /* get the location from the i/o registry */
-    ret = get_ioregistry_value_number (device, CFSTR("sessionID"), kCFNumberSInt64Type, &session);
+    UInt64 session = 0;
+    bool ret = get_ioregistry_value_number (device, CFSTR("sessionID"), kCFNumberSInt64Type, &session);
+    UInt32 locationID = 0;
     (void) get_ioregistry_value_number (device, CFSTR("locationID"), kCFNumberSInt32Type, &locationID);
     IOObjectRelease (device);
     if (!ret)
@@ -648,7 +648,7 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) {
         if (old_device->in_reenumerate) {
           /* device is re-enumerating. do not dereference the device at this time. libusb_reset_device()
            * will deref if needed. */
-          usbi_dbg (NULL, "detected device detached due to re-enumeration. sessionID: 0x%" PRIx64 ", locationID: 0x%" PRIx64,
+          usbi_dbg (NULL, "detected device detached due to re-enumeration. sessionID: 0x%" PRIx64 ", locationID: 0x%" PRIx32,
                     session, locationID);
 
           /* the device object is no longer usable so go ahead and release it */


Youw (approved):
agree, this looks fine

seanm:
There's another problem with this code in fact... if `get_ioregistry_value_number` fails for `locationID`, then `locationID` is uninitialized and later gets logged (as garbage).  We could just init locationID to zero.  Or we could do like for `sessionID` and fail if we can't get `locationID`, but it looks like the author deliberately didn't do that...

Youw:
maybe failing would be too much
lets default it to `0` and log a warning in case if locationID is not available

I believe if locationID something is very-very wrong with the device - it won't be usable and likely everything would fail elsewhere anyway, but having just a waning here should be sufficient

mcuee:
@seanm 

Just wondering where is the Apple documentation which says LocationID is 32bits. Thanks.

Apple USB Development document is quite messy...

seanm:
> Just wondering where is the Apple documentation which says LocationID is 32bits. Thanks.

Dunno. All the other uses in libusb treat it as 32 bit. Also note the contradiction in the 3rd and 4th parameters:

`ret = get_ioregistry_value_number (device, CFSTR("sessionID"), kCFNumberSInt64Type, &session);`
 
> Apple USB Development document is quite messy...

All Apple docs are, umm, not that great, not just USB.

libusb/os/darwin_usb.c
```diff
@@ -648,7 +648,7 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) {
         if (old_device->in_reenumerate) {
           /* device is re-enumerating. do not dereference the device at this time. libusb_reset_device()
            * will deref if needed. */
-          usbi_dbg (NULL, "detected device detached due to re-enumeration. sessionID: 0x%" PRIx64 ", locationID: 0x%" PRIx64,
+          usbi_dbg (NULL, "detected device detached due to re-enumeration. sessionID: 0x%" PRIx64 ", locationID: 0x%x",
```
tormodvolden, line 651: Is it actually correct to use %x with uint32_t ? Or is PRIx32 needed (or cast to unsigned long)?

Youw: Hm, that's a good point.
To be strictly correct with `x` it has to be `unsigned int` (not `long`).
For `uint32_t` `PRIx32` is needed indeed.

And I believe this is not the only place in this file where similar issue exists: in the `darwin_get_cached_device` the `UInt32 locationID` is printed as `PRIx64` which doesn't look right too.

Please doeble-check and update in both places, since we're touching the `locationID` here.

tormodvolden: Yep, I meant unsigned long with %lx :) Is also guaranteed to work.

tormodvolden: > in the `darwin_get_cached_device` the `UInt32 locationID` is printed as `PRIx64`

You mean printed as %x , right?

tormodvolden: I'll fix up both places.

seanm: >Is it actually correct to use %x with uint32_t ? 

It is correct. Because %x is for (unsigned) int generally, and on darwin that's guaranteed to be 32 bit. That will never change because it would totally break ABI.

tormodvolden: You don't mind me changing it to PRIx32 for symmetry reasons and potential consistency with non-darwin files?

tormodvolden: Thanks, I had already changed it there and in darwin_get_cached_device, and tested on macOS (normal build).

Youw: > on darwin that's guaranteed to be 32 bit

Our of curiocity, can you point to the documentation where that is guaranteed?

seanm: https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/64bitPorting/transition/transition.html

https://developer.apple.com/documentation/xcode/application-binary-interfaces

tormodvolden:
BTW, when looking at log files I see sessionID being null,
`matching sessionID/locationID 0x0/0xd000000 against cached device with sessionID/locationID 0x0/0x80000000`
whereas a code comment says it is should never be null: 
`/* assuming sessionID != 0 normally (never seen it be 0) */`

It is just an artifact of the enumeration process and how is uses the cached device, or is something wrong?

Youw:
That's a good questions.
I've never personally seen locationID to be null (not as part of libusb - but as part of some other macOS code), unless the device is physically disconnected or malfunction (e.g. FW reboot).

PR #1413, opened 2023-12-30
Added missing mutex acquisition when manipulating active_contexts_list

seanm:
The `active_contexts_list` is supposed to be protected by the `active_contexts_lock` mutex.

Upon code review, found one place where the mutex was not acquired.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2445,10 +2445,12 @@ int API_EXPORTED libusb_init_context(libusb_context **ctx, const struct libusb_i
 	}
 
 	/* check for first init */
+	usbi_mutex_static_lock(&active_contexts_lock);
 	if (!active_contexts_list.next) {
 		list_init(&active_contexts_list);
 		usbi_get_monotonic_time(&timestamp_origin);
 	}
+	usbi_mutex_static_unlock(&active_contexts_lock);
 
 	_ctx = calloc(1, PTR_ALIGN(sizeof(*_ctx)) + priv_size);
 	if (!_ctx) {


tormodvolden:
Does this fix the segfault or thread sanitizer issues seen in #1386?

seanm:
No, but I found this working on that.

seanm:
I've added a 2nd commit, as I think I've found more missing locks.

I found these experimenting with this BTW: https://github.com/libusb/libusb/pull/1419

seanm:
https://github.com/libusb/libusb/pull/1419 is working more now... and now clang finds what I fixed in the first commit:

`libusb/core.c:2451:7 Reading variable 'active_contexts_list' requires holding mutex 'active_contexts_lock'`

it hasn't found anything else yet though...

tormodvolden:
LGTM

libusb/core.c
```diff
@@ -2463,8 +2463,12 @@ int API_EXPORTED libusb_init_context(libusb_context **ctx, const struct libusb_i
 	if (getenv("LIBUSB_DEBUG")) {
 		_ctx->debug = get_env_debug_level();
 		_ctx->debug_fixed = 1;
-	} else if (default_context_options[LIBUSB_OPTION_LOG_LEVEL].is_set) {
-		_ctx->debug = default_context_options[LIBUSB_OPTION_LOG_LEVEL].arg.ival;
+	} else {
+		usbi_mutex_static_lock(&default_context_lock);
```
sonatique, line 2467: This lock causes a (self) deadlock because of the same lock being previously acquired at line 2438.

sonatique (changes requested):
Not good to go, cause dead-lock during Init

libusb/core.c
seanm: @sonatique thanks for reviewing! Yes, you are quite right. Not sure what I was thinking here... will amend this patch.

seanm: And `-Wthread-safety` even warns about this. Doh.

I fear I accidentally committed some WIP. I have too many different libusb branches going simultaneously...

seanm: OK, I've just dropped that 1 commit.

tormodvolden:
There seems to be a deadlock happening in the macOS CI run, also without these commits, it can be seen that the job fails after 6 hours sometimes.

seanm:
> There seems to be a deadlock happening in the macOS CI run, also without these commits...

The more I play with -Wthread-safety, and review the threading, the more I'm convinced there are numerous threading issues.

But I'm taking a break on libusb, I have too many fixes in too many branches, it's getting unmanageable for me.

PR #1416, opened 2023-12-31
Fix calls to libusb_set_option() on PowerPC

aurel32:
libusb_set_option() is a variadic function, so the type of the arguments are not clearly defined. When called with LIBUSB_OPTION_LOG_LEVEL, the argument is read with va_arg() as an int, which matches the type used
when passing constants, and also most of the internal calls and the calls in the examples.
    
However the internal call site in libusb_init_context() passes the ival element of the libusb_init_option struct directly, which is of type int64_t. This causes ssues on PowerPC, likely because it is 32-bit big endian:

```
Starting test run: test_init_context_basic...
Success (0)
Starting test run: test_init_context_log_level...
Expected test_ctx->debug (0) == LIBUSB_LOG_LEVEL_ERROR (1) at init_context.c:107 Failure (1)
Starting test run: test_init_context_log_cb...
Success (0)
---
Ran 3 tests
Passed 2 tests
Failed 1 tests
Error in 0 tests
Skipped 0 tests
```

Fix that by explicitly casting the value to int, which is anyway what happens on little endian.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2492,7 +2492,7 @@ int API_EXPORTED libusb_init_context(libusb_context **ctx, const struct libusb_i
 			r = libusb_set_option(_ctx, options[i].option, options[i].value.log_cbval);
 			break;
 		default:
-			r = libusb_set_option(_ctx, options[i].option, options[i].value.ival);
+			r = libusb_set_option(_ctx, options[i].option, (int)options[i].value.ival);
 		}
 		if (LIBUSB_SUCCESS != r)
 			goto err_free_ctx;


mcuee:
Just wondering what is the OS under testing?

I am not so sure if the issue also exhibits under legacy Mac OS X for PPC.

aurel32:
The full build log is available here: https://buildd.debian.org/status/fetch.php?pkg=libusb-1.0&arch=powerpc&ver=2%3A1.0.27%7Erc1-3&stamp=1704015277&raw=0

aurel32:
It seems the patch somehow change the API, as calling `libusb_set_option()` with constants implies the argument is an `int`.

aurel32:
> It seems the patch somehow change the API, as calling `libusb_set_option()` with constants implies the argument is an `int`.

I have updated the patch in a different way.

seanm:
Hmm, maybe the problem is a bit bigger... we have:

```c
struct libusb_init_option {
  /** Which option to set */
  enum libusb_option option;
  /** An integer value used by the option (if applicable). */
  union {
    int64_t ival;
    libusb_log_cb log_cbval;
  } value;
};
```

and

```c
struct usbi_option {
  int is_set;
  union {
    int ival;
    libusb_log_cb log_cbval;
  } arg;
};
```

Doesn't seem right that `ival` is different in these two structures.

seanm:
I think that `int64_t` needs to change to `int`.

seanm:
Oh I see, `libusb_init_option` is the public API you refer to... but `usbi_option` is private, right?  So I guess it could change instead.

aurel32:
> Oh I see, `libusb_init_option` is the public API you refer to... but `usbi_option` is private, right? So I guess it could change instead.

Yes, ideally the too should match. Changing `libusb_init_option` would be the best to fix the bug, but it is part of the public API, and changing it probably requires a soname bump (it could be encapsulated in another struct, so it should not be changed).

Note however that changing `usbi_option`  does not fix the original issue, and the cast is still needed.

seanm:
> Note however that changing `usbi_option` does not fix the original issue, and the cast is still needed.

You'd also need to change `arg = va_arg(ap, int);` to `arg = va_arg(ap, int64_t);`. Then no cash should be needed.

aurel32:
> > Note however that changing `usbi_option` does not fix the original issue, and the cast is still needed.
> 
> You'd also need to change `arg = va_arg(ap, int);` to `arg = va_arg(ap, int64_t);`. Then no cash should be needed.

It was like that in the original patch, but it simply doesn't work, as passing constant default to passing an `int` value, not an ` int64_t` value, creating other issues.

seanm:
> It was like that in the original patch, but it simply doesn't work, as passing constant default to passing an `int` value, not an ` int64_t` value, creating other issues.

Do you mean a call site like `libusb_set_option(NULL, LIBUSB_OPTION_LOG_LEVEL, verbose);` passing the integer constant `LIBUSB_OPTION_LOG_LEVEL`?

aurel32:
> > It was like that in the original patch, but it simply doesn't work, as passing constant default to passing an `int` value, not an ` int64_t` value, creating other issues.
> 
> Do you mean a call site like `libusb_set_option(NULL, LIBUSB_OPTION_LOG_LEVEL, verbose);` passing the integer constant `LIBUSB_OPTION_LOG_LEVEL`?

No, I mean things like:

`libusb_set_option (NULL, LIBUSB_OPTION_LOG_LEVEL, LIBUSB_LOG_LEVEL_DEBUG);`

The compiler passes the constant `LIBUSB_LOG_LEVEL_DEBUG` as int.

seanm:
>The compiler passes the constant `LIBUSB_LOG_LEVEL_DEBUG` as int.

`LIBUSB_LOG_LEVEL_DEBUG` is a member of the `libusb_log_level` enum, which since it has only positive small numbers, is indeed an `int` (if I grok the [C rules](https://en.cppreference.com/w/c/language/enum) correctly).

(I'm glad C23 finally adopted specifying sizes to enums, which I've [just tried adding to libusb](https://github.com/libusb/libusb/pull/1417), it would help with this issue.)

>Changing libusb_init_option would be the best to fix the bug, but it is part of the public API...

Wait a sec, `libusb_init_option` is new in 1.0.27. I just checked and it's not in 1.0.26, so it's not public API yet!  There is still time to change it.

aurel32:
> Wait a sec, `libusb_init_option` is new in 1.0.27. I just checked and it's not in 1.0.26, so it's not public API yet! There is still time to change it.

Good catch! Just changing the type of the `ival` element in the `libusb_init_option` structure from `int64_t` to `int` would fix the issue. @hjelmn, any reason it was defined as int64_t and not int?

hjelmn:
Just a bug. They can either both be int or both be int64_t. Ideally we should probably not have two structs for the same thing and I can not remember why we have both. Probably can just remove the usbi struct and only use the external API struct.

hjelmn:
Though now that I look at it. The issue really is that during re-enumeration the device object might go away. I think this is probably the correct fix. The temporary was only added during an earlier version of my change that I forgot to revert anyway.

seanm:
> Probably can just remove the usbi struct and only use the external API struct.

Or vice versa, no?  `usbi_option` exists in 1.0.26, `libusb_init_option` is new and hasn't shipped. The latter needs to exist now I guess, but should be changed to match the former I think.

seanm:
> Though now that I look at it. The issue really is that during re-enumeration the device object might go away. I think this is probably the correct fix. The temporary was only added during an earlier version of my change that I forgot to revert anyway.

Was this comment for https://github.com/libusb/libusb/pull/1427 ?

aurel32:
> Just a bug. They can either both be int or both be int64_t. 

I don't think it could be `int64_t` as `libusb_set_option()` is a variadic function and thus the compiler does not do implicit conversions from int to another type for the constants.

hjelmn:
Good point. This is kind of a mess caused by libusb_set_option, right. If the calling code calls with an int64_t on big endian and we read as an int then it will similarly be off. We should make both an int and document that int must be used. Very ugly but not sure what else we can do.

hjelmn:
@seanm yes. wrong window.

seanm:
> We should make both an int

Agreed. It's already `int` here:

```
struct usbi_option {
  int is_set;
  union {
    int ival;
  } arg;
};
```

the new and unreleased `libusb_init_option` should just change to match, and before 1.0.27 ships.

tormodvolden:
So #1436 should do the job?

aurel32:
> So #1436 should do the job?

Yep, thanks!

PR #1417, opened 2023-12-31
WIP: Introduce (conditional) C23 fixed underlying types to enums

seanm:
Add opt-in fixed underlying types to selected public and private enums.

- Keep the feature disabled by default.
- Export `-DLIBUSB_ENABLE_ENUM_SIZE=1` through pkg-config for enabled Autoconf builds.
- Require C++11 or later for fixed enum syntax in C++, including the MSVC language-mode check.
- Reject unsupported enabled consumers instead of silently selecting a different ABI.
- Expand the enum annotation macro during Doxygen preprocessing.
- Exercise the enabled configuration with Clang in Linux CI.

The default primary C library build is unaffected. An enabled build uses an explicit alternate ABI. Consumers must use the generated pkg-config flags, or define the matching macro, and use a supported language mode.

Related to #1416
Related to #1904


Youw:
This will break the ABI backward compatibility.
It gets even worse, when the library is built with C23, installed into a system/etc., and then some other application is built with C11 with the same header/binary - those will be completely incompatible.

I don't think libusb can afford this change at all... At most - this can be optional config to be enabled at build time explicitly (and disabled by default) to be used by those who really need it explicitly in ther projects/application.

But I can't imagine a scenario what that is _required_ by any of the sane use-cases.

seanm:
> This will break the ABI backward compatibility.

Only if the size/sign/type is actually changed (as opposed to merely being stated explicitly).

And for private enums, we *can* change the ABI/API, can we not?

For public enums, then yes, I agree we need to be very careful about changing them

Youw:
> And for private enums, we can change the ABI/API, can we not?

Yes, that shold not harm.

> Only if the size/sign/type is actually changed (as opposed to merely being stated explicitly).

Unless stated explicitly (in C23), the `enum`s in C are backed by the `int`. (Cannot find a reference to the standard but at least one compiler with enabled optimisations confirms it: https://godbolt.org/z/ndjxzaM5h).

And as far as this PR suggests - many public-declared enums are declared to be other (smaller) than `int`.
Even explicitly specifying `uint32_t` is unsafe - on some systems `int` (so are `enum`s) is 64bit.

To be honest I don't really get the point of explicilty specifying the enum type to be `int` explicitly for any `enum`. It is just redundant verbosity to my taste.

Youw:
@tormodvolden do you happen to have an oppinion on this one?

seanm:
> Unless stated explicitly (in C23), the `enum`s in C are backed by the `int`. (Cannot find a reference to the standard but at least one compiler with enabled optimisations confirms it: https://godbolt.org/z/ndjxzaM5h).

Here's some details: https://en.cppreference.com/w/c/language/enum

> To be honest I don't really get the point of explicilty specifying the enum type to be `int` explicitly for any `enum`. It is just redundant verbosity to my taste.

It's quite helpful when converting between an enum and a raw integer type (which libusb does a lot). It helps the compiler warn about signed/unsigned mismatches, and size mismatches.

For example, `enum libusb_descriptor_type` where I added `LIBUSB_ENUM_SIZE(uint8_t)`.  It's important that the values in this enum never be bigger (or negative) because they are stored into structs like this:

```c
struct libusb_device_descriptor {
	...
	uint8_t  bDescriptorType;
	...
```

By tagging the enum in this way it 1) acts like documentation 2) the compiler can ensure values do not exceed what's expected.

I agree there are compatibility concerns. But for me, opting into a new language version (with `-std=c23`), is a good opportunity for a small compatibility break.

seanm:
How about as a first pass we just make the `LIBUSB_ENUM_SIZE` macro do nothing?

Then we can still decorate enums with `LIBUSB_ENUM_SIZE` and it will at least serve as documentation.

seanm:
FWIW: gcc 15 now defaults to C23: https://gcc.gnu.org/gcc-15/porting_to.html

Youw:
> How about as a first pass we just make the `LIBUSB_ENUM_SIZE` macro do nothing?

I vote to have an explicit build time option to enable these annotations. Disabled by default, maybe enabled by our own CI for tests/verifications/etc.

seanm:
> I vote to have an explicit build time option to enable these annotations.

I don't object to that idea, but I don't know anything about autotools. Would you know how to do it?

Youw:
sure, I know how to do so

even though I'll delegate that to AI anyway)

seanm:
> sure, I know how to do so

Thanks for working on that. (Though honestly, since I'm not a git expert, now that you've changed my PR, I'm not sure how to make further changes...)

Next we need reviews of whether the enum sizes I've chosen here are correct or not...

Youw:
>  I'm not sure how to make further changes

`git pull`

and then work as usual

---
That's not "expert", but basic git colaboration workflow. Good moment for you to practice.

Youw (approved):
I've extended the implementation to make it even more complete.
Now we should be good.

PR #1418, opened 2023-12-31
Fixed incorrect alignment in allocation in libusb_alloc_transfer

seanm:
After being suspicous of some code I was browsing, I tried changing PTR_ALIGN to align to 4096 bytes instead of just to pointer size (8 bytes), then enabled ASan then ran the xusb example, and it crashed. What ensued was a big code review of that function and related code.

- reviewed use of macros like USBI_TRANSFER_TO_LIBUSB_TRANSFER
- introduced new macros TRANSFER_PRIV_TO_USBI_TRANSFER and USBI_TRANSFER_TO_TRANSFER_PRIV to do pointer offset conversions
- introduced some temporary variables, especially for USBI_TRANSFER_TO_LIBUSB_TRANSFER results
- move some variable assignment and declaration together
- replaced a few uses of PTR_ALIGN to instead use the alignment macros

### DIFF
--- a/libusb/io.c
+++ b/libusb/io.c
@@ -1244,8 +1244,8 @@ void usbi_io_exit(struct libusb_context *ctx)
 
 static void calculate_timeout(struct usbi_transfer *itransfer)
 {
-	unsigned int timeout =
-		USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer)->timeout;
+	struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
+	unsigned int timeout = transfer->timeout;
 
 	if (!timeout) {
 		TIMESPEC_CLEAR(&itransfer->timeout);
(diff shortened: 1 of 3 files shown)


Youw:
> introduced some temporary variables

Is it for for step-by-step debugging only, or are there other theoretical/practical reasons?

seanm:
> Is it for for step-by-step debugging only, or are there other theoretical/practical reasons?

Mostly for step-by-step debugging, and to have a type written, I was getting very confused by all the different by similar structures.

tormodvolden:
It is good to have things spelled out properly but I'd like to understand where is exactly the fix here? Adding PTR_ALIGN to each term of alloc_size?

> - move some variable assignment and declaration together

This improves readability IMO, but breaks the existing code-style. We don't have an expressed policy and maybe not consistency on this anyway, so I am not against it though.

seanm:
> It is good to have things spelled out properly but I'd like to understand where is exactly the fix here?

The main thing is the calculation in `libusb_alloc_transfer`. It was using PTR_ALIGN() only once. Each of those sub-structs should have their size aligned. It didn't match this macro (which I didn't change):

```c
#define USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer)	\
	((struct libusb_transfer *)			\
	 ((unsigned char *)(itransfer)			\
	  + PTR_ALIGN(sizeof(struct usbi_transfer))))
```

Note how it uses PTR_ALIGN() on `struct usbi_transfer`, but `libusb_alloc_transfer` was not doing that.

It was totally broken and working by luck.

> This improves readability IMO, but breaks the existing code-style. We don't have an expressed policy and maybe not consistency on this anyway, so I am not against it though.

I have noticed the codebase is inconsistent about this already. I had assumed older code was C89 style, and newer code is C99 style.

tormodvolden (approved):

PR #1419, opened 2024-01-01
Add clang thread safety annotations

seanm:
See:

https://clang.llvm.org/docs/ThreadSafetyAnalysis.html

### DIFF
--- a/Xcode/common.xcconfig
+++ b/Xcode/common.xcconfig
@@ -88,6 +88,9 @@ CLANG_WARN__DUPLICATE_METHOD_MATCH = YES
 CLANG_WARN_DEPRECATED_OBJC_IMPLEMENTATIONS = YES
 GCC_WARN_UNDECLARED_SELECTOR = YES
 
+// Optionally enable thread safety warnings, which requires building as C++.
+//OTHER_CFLAGS = -Wthread-safety -x c++
+
 // Static analyzer warnings.
 CLANG_ANALYZER_NONNULL = YES
 CLANG_ANALYZER_SECURITY_FLOATLOOPCOUNTER = YES
--- a/Xcode/libusb.xcodeproj/project.pbxproj
+++ b/Xcode/libusb.xcodeproj/project.pbxproj
@@ -57,6 +57,16 @@
 		008FC0301628BC7400BC5BE2 /* listdevs.c in Sources */ = {isa = PBXBuildFile; fileRef = 008FBFE71628BA0E00BC5BE2 /* listdevs.c */; };
 		1438D77A17A2ED9F00166101 /* hotplug.c in Sources */ = {isa = PBXBuildFile; fileRef = 1438D77817A2ED9F00166101 /* hotplug.c */; };
 		1438D77F17A2F0EA00166101 /* strerror.c in Sources */ = {isa = PBXBuildFile; fileRef = 1438D77E17A2F0EA00166101 /* strerror.c */; };
+		14D9E1DF2FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E02FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E12FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E22FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E32FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E42FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E52FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E62FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E72FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
+		14D9E1E82FFD5E01005170B1 /* clang_thread_safety.h in Headers */ = {isa = PBXBuildFile; fileRef = 14D9E1DE2FFD5E01005170B1 /* clang_thread_safety.h */; };
 		2018D95F24E453BA001589B2 /* events_posix.c in Sources */ = {isa = PBXBuildFile; fileRef = 2018D95E24E453BA001589B2 /* events_posix.c */; };
 		2018D96124E453D0001589B2 /* events_posix.h in Headers */ = {isa = PBXBuildFile; fileRef = 2018D96024E453D0001589B2 /* events_posix.h */; };
 		20468D70243298C100650534 /* sam3u_benchmark.c in Sources */ = {isa = PBXBuildFile; fileRef = 20468D6E243298C100650534 /* sam3u_benchmark.c */; };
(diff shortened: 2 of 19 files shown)


libusb/libusbi.h
```diff
@@ -44,6 +44,17 @@
 #define static_assert(cond, msg) _Static_assert(cond, msg)
 #endif
 
+/* https://clang.llvm.org/docs/ThreadSafetyAnalysis.html */
+#ifdef __clang
```
Youw, line 48: I believe this is how it is suggested by `ThreadSafetyAnalysis.html`
```suggestion
#if defined(__clang__) && (!defined(SWIG))
```

Youw (commented):
shouldn't we explicitly add `-Wthread-safety` for this to kick in?
should be great to have those checks on CI if we're to add them on code level

---
generally looks good, don't see any code changes, only extra annotations

seanm:
>shouldn't we explicitly add -Wthread-safety for this to kick in?

Yes. I haven't even tried that yet, because also required is that the threading functions be annotated too.  pthread.h on my OS is not. It might be possible to annotate the wrappers in threads_posix.h though, which I will try...

>generally looks good, don't see any code changes, only extra annotations

Correct. The annotations also require that the mutex is declared before variables it protects, which is why I moved some things around.

seanm:
@mcuee not sure the macOS tag is appropriate, there's nothing macOS specific here.

tormodvolden:
I think we lost the correct commits in a rebase, can you please repush them?

seanm:
> I think we lost the correct commits in a rebase, can you please repush them?

This PR will be easier to rebase once https://github.com/libusb/libusb/pull/1461 is merged.

seanm:
Actually, wasn't hard to rebase to current master.

tormodvolden:
If this depends on the other PR, feel free to rebase on top of that one.

seanm:
@sonatique @tormodvolden With the other stuff merged (thanks!) I was finally able to rebase this.

It's very WIP still.

But you can see the idea of how functions are annotated.

Building locally I get warnings like:

```
core.c:2203:14 Writing variable 'auto_detach_kernel_driver' requires holding mutex 'dev_handle->lock' exclusively

io.c:1820:1 Mutex 'ctx->events_lock' is not held on every path through here

io.c:1845:1 Mutex 'ctx->events_lock' is still held at the end of function

io.c:1859:2 Releasing mutex 'ctx->events_lock' that was not held
```

It's been so long since I looked at this I don't recall if these are true positives, false positives, or if I just didn't finishing annotation all functions.  But these are the kinds of warnings that are possible...

seanm:
I split some of this into:

- https://github.com/libusb/libusb/pull/1682
- https://github.com/libusb/libusb/pull/1683
- https://github.com/libusb/libusb/pull/1684

which can be merged immediately (after review of course).

seanm:
I will split this into separate PRs, in the hopes of it actually being merged...

seanm:
I've moved the C++ changes to https://github.com/libusb/libusb/pull/1782 and this PR is now just for -Wthread-safety support.

seanm:
@Youw @sonatique @mcuee  with other stuff now merged, this is finally ready for pre-merge review.

I guess linux CI uses an older Clang?  I'll have to update the conditions that enable the attributes somehow...

sonatique:
@seanm Thank you for your sustained effort on this subject.

I am not at all familiar with this and I clearly lost track of the current move. As a naive reviewer, I have difficulties understanding what such a change really brings over what a modern compiler should understand automatically by itself, and it feels like a lot of added text on something quite trivial:
`
CAPABILITY("mutex") typedef pthread_mutex_t usbi_mutex_static_t;
static inline void usbi_mutex_static_lock(usbi_mutex_static_t *mutex) ACQUIRE(*mutex) NO_THREAD_SAFETY_ANALYSIS
{
	PTHREAD_CHECK(pthread_mutex_lock(mutex));
}
`
so I am obviously missing the point because I don't understand what is the information that is added by CAPABITLIY, ACQUIRES, etc. over what is already there in plain sight. 

Also: reviewing this change is difficult because it is very repetitive and missing a single misspelled word could mean that nothing will work as expected. Would you mind giving a very quick reminder about what is the purpose of the change and what reviewer should focus on?
Thank you very much!

seanm:
> ...what such a change really brings over what a modern compiler should understand automatically by itself

I'm not aware of any compiler that has compile-time checks for mutex locking errors without these kind of special annotations.

The long explanation is here: https://clang.llvm.org/docs/ThreadSafetyAnalysis.html

The short explanation is that by annotating which variables are protected by which mutexes, and which functions expect mutexes to be already-held or already-not-held, the compiler can give more warnings.

For example, in `libusb_handle_events_timeout_completed()` if I comment out the call to `libusb_unlock_event_waiters()` near the end of the function, I get the warning `warning: Mutex 'ctx->event_waiters_lock' is not held on every path through here`

This has already found bugs in libusb (since fixed).

> Also: reviewing this change is difficult because it is very repetitive and missing a single misspelled word could mean that nothing will work as expected.

True enough, but at least none of this changes codegen.  It only allows Clang to give more warnings. So it can't really break anything.

>...what reviewer should focus on? Thank you very much!

For example, checking that `GUARDED_BY` annotations are protecting the right variable with the right mutex. Sometimes existing comments make it easy, like:

```diff
-	uint32_t state_flags;   /* Protected by usbi_transfer->lock */
+	uint32_t state_flags GUARDED_BY(lock);   /* Protected by usbi_transfer->lock */
```

But sometimes it is less clear...

seanm:
With the other threading PRs seeing activity, would be good to get this merged too...

seanm:
@mcuee @sonatique not sure how to fix that CI issue... that fuzzing thing is from a different repo, right?  I guess it will need to know about the new clang_thread_safety.h file, and take a copy?

sonatique:
@seanm , as far as I can tell:
The thread-safety annotations are correct, but CI is red on every clang job (Android, Linux-clang, Fuzzing) for two build-config reasons:

1) clang-as-C fails to compile. The gate #if defined(__clang__) enables the real attributes for any clang build, but GUARDED_BY(sibling_member) needs C++ class-scope lookup. In C, clang errors use of undeclared identifier 'usb_devs_lock'. TSA is a C++-only feature (hence the -x c++ in the xcconfig). 
**Fix: add && defined(__cplusplus) to the gate.**

2) Fuzzing: threads_posix.h:28: fatal error: 'libusbi.h' file not found. The new #include "libusbi.h" is circular and breaks configs that don't have libusb/ on the include path (oss-fuzz uses -I.). 
**Fix: drop that include** (threads_posix.h is only ever included via libusbi.h, which already provides the macros.)

I pushed these changes here to validate my hypothesis: https://github.com/libusb/libusb/commit/df45829a89e6f71207f5323c5478892c179b465c

libusb/os/darwin_usb.c
```diff
@@ -1699,7 +1699,7 @@ static int darwin_get_configuration(struct libusb_device_handle *dev_handle, uin
   return LIBUSB_SUCCESS;
 }
 
-static int darwin_set_configuration(struct libusb_device_handle *dev_handle, int config) {
+static int darwin_set_configuration(struct libusb_device_handle *dev_handle, int config) REQUIRES(dev_handle->lock) {
```
sonatique, line 1702: REQUIRES(dev_handle->lock) is correct here; the gap is that libusb_set_configuration doesn't hold it on the public path. This is a known race actually: #1798 item 5, with the backend-side fix already prepared (branch stacked on #1864). No action needed in this PR; the annotation correctly documents the requirement.

libusb/os/darwin_usb.c
```diff
@@ -2311,7 +2311,8 @@ static io_service_t usb_find_interface_matching_location (const io_name_t class_
   return IOServiceGetMatchingService (darwin_default_master_port, matchingDict);
 }
 
-static int darwin_kernel_driver_active(struct libusb_device_handle *dev_handle, uint8_t interface) {
+static int darwin_kernel_driver_active(struct libusb_device_handle *dev_handle, uint8_t interface) REQUIRES(dev_handle->lock) {
```
sonatique: Actually darwin_kernel_driver_active doesn't access any dev_handle->lock-guarded state (only immutable dpriv->location + IOKit lookups), and the public libusb_kernel_driver_active calls it without the lock. Unlike set_configuration (#1798 item 5), there's no race here. The REQUIRES just over-states the contract. Suggest removing it.

Xcode/common.xcconfig
```diff
@@ -88,6 +88,9 @@ CLANG_WARN__DUPLICATE_METHOD_MATCH = YES
 CLANG_WARN_DEPRECATED_OBJC_IMPLEMENTATIONS = YES
 GCC_WARN_UNDECLARED_SELECTOR = YES
 
+// Optionally enable thread safety warnings, which requires building as C++.
+//OTHER_CFLAGS = -Wthread-safety -x c++
```
sonatique, line 92: I think that with this flag commented out and the attributes only active under -x c++, nothing in CI actually checks these annotations. Maybe worth a dedicated CI job / Xcode target that builds with -Wthread-safety -x c++ ?

Xcode/libusb.xcodeproj/project.pbxproj
```diff
@@ -274,6 +284,12 @@
 		1443EE8916417EA6007E0579 /* libusb_release.xcconfig */ = {isa = PBXFileReference; indentWidth = 4; lastKnownFileType = text.xcconfig; path = libusb_release.xcconfig; sourceTree = SOURCE_ROOT; tabWidth = 4; usesTabs = 1; };
 		145F5D8D2E60B46B004ABA5D /* windows_hotplug.h */ = {isa = PBXFileReference; indentWidth = 4; lastKnownFileType = sourcecode.c.h; path = windows_hotplug.h; sourceTree = "<group>"; tabWidth = 4; usesTabs = 1; };
 		145F5D8E2E60B46B004ABA5D /* windows_hotplug.c */ = {isa = PBXFileReference; indentWidth = 4; lastKnownFileType = sourcecode.c.c; path = windows_hotplug.c; sourceTree = "<group>"; tabWidth = 4; usesTabs = 1; };
+		1472E1592B43D66B00850BA3 /* init_context.c */ = {isa = PBXFileReference; indentWidth = 2; lastKnownFileType = sourcecode.c.c; path = init_context.c; sourceTree = "<group>"; tabWidth = 2; usesTabs = 0; };
+		1472E15A2B43D68600850BA3 /* stress_mt.c */ = {isa = PBXFileReference; indentWidth = 4; lastKnownFileType = sourcecode.c.c; path = stress_mt.c; sourceTree = "<group>"; tabWidth = 4; usesTabs = 1; };
```
sonatique, line 288: These file-reference reorderings look like incidental Xcode churn unrelated to the annotations. Could be reverted to keep the diff focused, but not a blocker.

sonatique (changes requested):

seanm:
> but CI is red on every clang job

It was earlier today, but I already fixed that in more recent pushes hours ago.  

I'll look at your other comments...

Xcode/common.xcconfig
seanm: Yes, that would be nice.  I don't know anything about github CI, but such a thing could be added in a follow up by someone that knows how...

I could make such a target in Xcode, I guess that would be handy...

The nice thing about this PR is that it doesn't affect codegen, so the worse it could do is give some false positive warnings.

Xcode/libusb.xcodeproj/project.pbxproj
seanm: Fixed.

libusb/os/darwin_usb.c
seanm: It compiles, of course, because all of this will at most cause warnings.

This looks like a missing lock to me.

`darwin_set_configuration` seems (to me) to require `dev_handle->lock` because that's the lock for `claimed_interfaces`, which `darwin_set_configuration` uses.

A followup PR could split the function (create a `_locked` variant) like you did in https://github.com/libusb/libusb/pull/1864/changes.

sonatique: Sorry, I commented too quickly. Please see revised comment.

seanm: OK great!

Yeah, function pointers are definitely pessimizing for static analysis. :(

libusb/os/darwin_usb.c
seanm: Yeah, I think you're right. It's been so long, I'm not sure why I added it...

Xcode/common.xcconfig
seanm: Actually, maybe a target in the Xcode project is not so great, because really you want this not just for the library, but for the other targets too (like test code, etc.)

Probably best is just additional CI cases using `-Wthread-safety -x c++` (on both macOS & linux).

sonatique: OK, let's postpone this

libusb/os/darwin_usb.c
```diff
@@ -2312,6 +2312,7 @@ static io_service_t usb_find_interface_matching_location (const io_name_t class_
 }
 
 static int darwin_kernel_driver_active(struct libusb_device_handle *dev_handle, uint8_t interface) {
+
```
sonatique: this blank line may be removed..

sonatique (approved):

sonatique:
@seanm : this seems to be safe to merge, though there is only my approval.

libusb/os/darwin_usb.c
sonatique: Note: the fix is actually now in #1864.

sonatique (approved):

libusb/os/darwin_usb.c
seanm: Done.

seanm:
> @seanm : this seems to be safe to merge, though there is only my approval.

Well, after 2.5 years, others have had plenty of chance to comment. :)

sonatique:
You have a point @seanm

PR #1421, opened 2024-01-02
Some minor misc Xcode improvements

seanm:


seanm:
Does libusb have standard about tabs vs space, and indent amount? (these can be specified in the Xcode project, and I want to get them right).

Xcode/debug.xcconfig
```diff
@@ -27,3 +27,6 @@ GCC_OPTIMIZATION_LEVEL = 0
 
 // 
 GCC_PREPROCESSOR_DEFINITIONS = $(inherited) DEBUG=1
+
+// No need for Universal Binaries in debug.
+ONLY_ACTIVE_ARCH = YES
```
Youw, line 32: why not?

Xcode/common.xcconfig
```diff
@@ -26,6 +26,13 @@ ALWAYS_SEARCH_USER_PATHS = NO
 // Enable weak references for Objective-C
 CLANG_ENABLE_OBJC_WEAK = YES
 
+// Allocate even uninitialized global variables in the data section of the object file, rather than generating them as common blocks. This has the effect that if the same variable is declared (without 'extern') in two different compilations, you will get an error when you link them.
+GCC_NO_COMMON_BLOCKS = YES
+
+// Keep private symbols private.  The first setting is -fvisibility=hidden, the second is -fvisibility-inlines-hidden.
+GCC_SYMBOLS_PRIVATE_EXTERN = YES
+GCC_INLINES_ARE_PRIVATE_EXTERN = YES
```
Youw, line 34: the option says `GCC`
would it be applied if CLANG is used?

seanm: Yes. The names are historical, think of "GCC" meaning "c compiler" here.

Xcode/debug.xcconfig
seanm: It doubles build time to compile for both architectures, and generally one is going to debug on the native architecture (not under Rosetta emulation). 

Note that I've merely moved this existing setting from one file to another. It's not something new I'm enabling/disabling.

seanm:
@tormodvolden would be nice to get this one merged. As I move between my various branches, and perform global searches in Xcode, I'm often missing things...

tormodvolden:
Yeah, this is not release-critical but OTOH it seems pretty safe, and the CI XCode build is happy too. We should prepend "xcode:" to the commit summaries, but I can do this when I merge it.

Youw (approved):

hjelmn:
> Does libusb have standard about tabs vs space, and indent amount? (these can be specified in the Xcode project, and I want to get them right).

We kind of have been doing things ad-hoc up until now. I had proposed standardizing it years ago to eliminate tabs (the linux code uses tabs). The darwin backend uses a different format (see the emacs blob in darwin_usb.c). We should probably agree on a standard before we proceed but all I have to say is I hate tabs in source files. They are just pain awful :).

PR #1422, opened 2024-01-02
Replaced old atoi() with better strtol()

seanm:
atoi() gives no opportunity for error checking, strtol() does. Though I did not actually improve error checking...

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2380,14 +2380,19 @@ int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
  */
 static enum libusb_log_level get_env_debug_level(void)
 {
+	enum libusb_log_level level = LIBUSB_LOG_LEVEL_NONE;
 	const char *dbg = getenv("LIBUSB_DEBUG");
-	enum libusb_log_level level;
 	if (dbg) {
-		int dbg_level = atoi(dbg);
+		char *end = NULL;
+		long dbg_level = strtol(dbg, &end, 10);
+		if (dbg == end ||
+			*end != '\0' ||
+			dbg_level < LIBUSB_LOG_LEVEL_NONE ||
+			dbg_level > LIBUSB_LOG_LEVEL_DEBUG) {
+			usbi_warn(NULL, "LIBUSB_DEBUG is invalid or out of range; clamping");
+		}
 		dbg_level = CLAMP(dbg_level, LIBUSB_LOG_LEVEL_NONE, LIBUSB_LOG_LEVEL_DEBUG);
 		level = (enum libusb_log_level)dbg_level;
-	} else {
-		level = LIBUSB_LOG_LEVEL_NONE;
 	}
 	return level;
 }
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -29,6 +29,7 @@
 #include <setupapi.h>
 #include <ctype.h>
 #include <stdio.h>
+#include <stdlib.h>
 
 #include "libusbi.h"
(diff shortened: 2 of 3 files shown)


Youw:
> Though I did not actually improve error checking...

I don't see any actual benefin otherwise.
Is there?
Unless you're planning more changes on top of it.

seanm:
> I don't see any actual benefin otherwise.

There's no big benefit. It's just a little cleanup to remove an obsolete function.

tormodvolden:
atoi is not obsolete, it is just not recommended any longer :)  strtol is better /if/ you add error checking.

seanm:
> strtol is better /if/ you add error checking

Draft error checking now added.

libusb/core.c
```diff
@@ -2381,14 +2381,12 @@ int API_EXPORTEDV libusb_set_option(libusb_context *ctx,
  */
 static enum libusb_log_level get_env_debug_level(void)
 {
+	enum libusb_log_level level = LIBUSB_LOG_LEVEL_NONE;
 	const char *dbg = getenv("LIBUSB_DEBUG");
-	enum libusb_log_level level;
 	if (dbg) {
-		int dbg_level = atoi(dbg);
+		long dbg_level = strtol(dbg, NULL, 10);
 		dbg_level = CLAMP(dbg_level, LIBUSB_LOG_LEVEL_NONE, LIBUSB_LOG_LEVEL_DEBUG);
```
tormodvolden, line 2388: This silently sanitizes it, but maybe a warning should be shown to the user?

seanm: Done.

libusb/os/windows_winusb.c
```diff
@@ -1237,7 +1241,12 @@ static bool get_dev_port_number(HDEVINFO dev_info, SP_DEVINFO_DATA *dev_info_dat
 			NULL, (PBYTE)buffer, sizeof(buffer), NULL)) {
 		// Check for the required format.
 		if (strncmp(buffer, "Port_#", 6) == 0) {
-			*port_nr = atoi(buffer + 6);
+			start = buffer + 6;
+			port = strtoll(start, &end, 10);
+			if (port < 0 || port > 4294967295 || end == start || *end != '\0') {
```
tormodvolden, line 1246: Why not MAX_INT or something here? And can there be a possibility of such high port numbers anyway? From the above comment I wouldn't expect anything above 0xffff.

seanm: >Why not MAX_INT or something here? 

Simply because I couldn't find such a constant for the maximum `DWORD`. Do you know one? The point of this line is to make 100% sure the cast a couple of lines later is safe:

```
*port_nr = (DWORD)port;
```

>And can there be a possibility of such high port numbers anyway? 

`strtoll` returns `long long` so I was being maximally paranoid.

>From the above comment I wouldn't expect anything above 0xffff.

In fact the comment is "The string *may* have a format similar to..." with "may" emphasized, as if in doubt. If `pSetupDiGetDeviceRegistryPropertyA` puts some crazyness in the buffer, who knows...

But if you're sure 0xFFFF is the max valid, we could tighten the check.

Again, I'm only trying to get rid of old `atoi` here. The old code was not limiting to 0xFFFF.

tormodvolden: > Simply because I couldn't find such a constant for the maximum `DWORD`. Do you know one?

DWORD is typedef unsigned long, so ULONG_MAX.

seanm: Done.

tormodvolden:
Thanks!

PR #1423, opened 2024-01-02
Increased usbi_get_tid() from int to long

seanm:
This function has different implementations on every OS, but for some, like macOS, it was truncating from 64 to 32 bit by casting to int. So increased its size from int to long.

### DIFF
--- a/libusb/os/threads_posix.c
+++ b/libusb/os/threads_posix.c
@@ -22,6 +22,7 @@
 #include "libusbi.h"
 
 #include <errno.h>
+#include <limits.h>
 #if defined(__ANDROID__)
 # include <unistd.h>
 #elif defined(__HAIKU__)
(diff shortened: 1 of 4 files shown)


libusb/os/threads_posix.c
```diff
@@ -79,47 +79,47 @@ int usbi_cond_timedwait(pthread_cond_t *cond,
...
 #elif defined(__APPLE__)
 #ifdef HAVE_PTHREAD_THREADID_NP
 	uint64_t thread_id;
 
 	if (pthread_threadid_np(NULL, &thread_id) == 0)
-		tid = (int)thread_id;
+		tid = (unsigned long)thread_id;
 	else
-		tid = -1;
+		tid = ULONG_MAX;
```
Youw, line 99: are there any places in the code where the return value would be checked for `-1`?

Youw: (except below in this very function)

Youw: hm, I guess line 121 answeres my question
no other places are possible

Youw (approved):
This looks good to me

seanm:
@tormodvolden this could be considered for merging too.  The truncation is plainly wrong on macOS, and this code is only used for logging, so probably safe to change.

tormodvolden:
Since we only use this value for differentiating threads in the log output, and there only display 32 bits of it anyway, I think this can safely wait.

seanm:
> Since we only use this value for differentiating threads in the log output, and there only display 32 bits of it anyway, I think this can safely wait.

@tormodvolden with the release now out, can this be merged?

libusb/os/threads_posix.c
```diff
@@ -79,47 +80,47 @@ int usbi_cond_timedwait(pthread_cond_t *cond,
 		return LIBUSB_ERROR_OTHER;
 }
 
-unsigned int usbi_get_tid(void)
+unsigned long usbi_get_tid(void)
 {
-	static _Thread_local unsigned int tl_tid;
-	int tid;
+	static _Thread_local unsigned long tl_tid;
+    unsigned long tid;
```
tormodvolden, line 86: I fixed a whitespace error :)

Thanks for the patch!

PR #1426, opened 2024-01-04
Fixed -Wswitch warnings by including all enum values in switch

seanm:
No change in behaviour.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2491,6 +2491,11 @@ int API_EXPORTED libusb_init_context(libusb_context **ctx, const struct libusb_i
 		case LIBUSB_OPTION_LOG_CB:
 			r = libusb_set_option(_ctx, options[i].option, options[i].value.log_cbval);
 			break;
+
+		case LIBUSB_OPTION_LOG_LEVEL:
+		case LIBUSB_OPTION_USE_USBDK:
+		case LIBUSB_OPTION_NO_DEVICE_DISCOVERY:
+		case LIBUSB_OPTION_MAX:
 		default:
 			r = libusb_set_option(_ctx, options[i].option, options[i].value.ival);
 		}


mcuee:
@tormodvolden 

This should be good to go.

tormodvolden:
There should be no reason for keeping "default" now, but I guess the compiler will complain if we remove it?

libusb/core.c
```diff
@@ -2491,6 +2491,11 @@ int API_EXPORTED libusb_init_context(libusb_context **ctx, const struct libusb_i
 		case LIBUSB_OPTION_LOG_CB:
 			r = libusb_set_option(_ctx, options[i].option, options[i].value.log_cbval);
 			break;
+
+		case LIBUSB_OPTION_LOG_LEVEL:
+		case LIBUSB_OPTION_USE_USBDK:
+		case LIBUSB_OPTION_NO_DEVICE_DISCOVERY:
+		case LIBUSB_OPTION_MAX:
 		default:
```
Youw, line 2499: since we've listed all of the options, probably best to remote the `default`
```suggestion
```

seanm:
> There should be no reason for keeping "default" now, but I guess the compiler will complain if we remove it?

Some compilers complain if there's no default. Some complain if all enum values are covered and there *is* a default. It's hard to satisfy all.

Keeping the default is safer because then for sure codegen does not change here.  It's not impossible that some other value is stored there. `enum libusb_option foo = 8634;` does compile in C!

Youw:
> Some complain if all enum values are covered and there is a default.

True. It is not about the compiler specifically, but more about warning options clang/gcc supports.

> Some compilers complain if there's no default.

Never seen such things. If all enum cases listed and no default - it always worked for me. I vote for using such aproach and having a warning if not all enum cases handled by default.
That is the least bug-prone aproach as per my experience.

seanm:
> Never seen such things.

https://clang.llvm.org/docs/DiagnosticsReference.html#wswitch-default

>If all enum cases listed and no default - it always worked for me. I vote for using such aproach and having a warning if not all enum cases handled by default. That is the least bug-prone aproach as per my experience.

In C99, enums are basically `int`, which means it's not impossible to have a value stored in there that's not in the enumeration list. So having the `default` is safer.

There's even a reasonable scenario where it matters:

Let's say future libusb introduces a new value to that enum (likely to happen). Let's say `LIBUSB_OPTION_COOL = 137`. Some app then calls `libusb_set_option(NULL, LIBUSB_OPTION_COOL, ...)`.  Let's say the app links dynamically to libusb.  If future-libusb is found at linktime, this option will work.  If libusb 1.0.27 is found at linktime, what happens with that switch statement? You need the `default`.

tormodvolden:
That is true, but shouldn't we in this case warn about e.g. "unhandled option %d" instead of calling libusb_set_option() with this unknown value?

seanm:
> That is true, but shouldn't we in this case warn about e.g. "unhandled option %d" instead of calling libusb_set_option() with this unknown value?

Yeah probably.

This PR is only trying to silence the compiler warnings, without changing behaviour. Which is why I'm arguing against removing the `default`, because that would change behaviour.

Youw (approved):
OK, thats a good point

PR #1427, opened 2024-01-05
darwin: Reverted seemingly harmless introduction of temporary variable

seanm:
This change was made in 13a6953379fa8c94bb2476dc460439d4ba2b7c17 and this partly reverts some of it.

This change would be harmless if there was no multithreading happening, but there is.

On the hotplug background thread, darwin_devices_detached() is called, which in turn calls Release() on the device, which frees the memory, and then sets old_device->device to NULL. Shortly after, darwin_devices_attached() is called (because the kernel driver is reattaching?) and this calls darwin_get_cached_device(), which sets ->device to something new (and not NULL).

Meanwhile, back on the main thread, darwin_reenumerate_device() is running and had cached ->device in the seemingly harmless temporary variable.  But that thing was already deallocated on the other thread!

Re-reading it from the structure makes you more likely to get the value you want.

I believe there is a larger unfixed issue here, and that some mutex needs to be added somewhere. But this is a short term fix at least.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1484,12 +1484,11 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
 static int darwin_open (struct libusb_device_handle *dev_handle) {
   struct darwin_device_handle_priv *priv = usbi_get_device_handle_priv(dev_handle);
   struct darwin_cached_device *dpriv = DARWIN_CACHED_DEVICE(dev_handle->dev);
-  usb_device_t darwin_device = dpriv->device;
   IOReturn kresult;
 
   if (0 == dpriv->open_count) {
     /* try to open the device */
-    kresult = (*darwin_device)->USBDeviceOpenSeize (darwin_device);
+    kresult = (*dpriv->device)->USBDeviceOpenSeize (dpriv->device);
     if (kresult != kIOReturnSuccess) {
       usbi_warn (HANDLE_CTX (dev_handle), "USBDeviceOpen: %s", darwin_error_str(kresult));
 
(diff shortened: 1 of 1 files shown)


mcuee:
Unfortunately this does not fix Issue #1386
* https://github.com/libusb/libusb/issues/1386#issuecomment-1878366360

mcuee:
> I believe there is a larger unfixed issue here, and that some mutex needs to be added somewhere.

Looks like we need more work to fix #1386.

mcuee:
The latest commit fixed my issues and the tests are good.
* https://github.com/libusb/libusb/issues/1386#issuecomment-1879394347

Youw:
> I believe there is a larger unfixed issue here, and that some mutex needs to be added somewhere. But this is a short term fix at least.

Sounds like so.
Technically it doesn't fix the multithreading issue, and theoretically only hides it deeper.
Maybe it is best to try to introduce a lock before assigning the temp var inestaed (and unlock at the end of function/right after last usage of that var). So that one should not use a mutex each time a `device` is being accesses?

seanm:
>Sounds like so. Technically it doesn't fix the multithreading issue, and theoretically only hides it deeper. 

I think so. Meaning it is as good as 1.0.26 was. :)

>Maybe it is best to try to introduce a lock before assigning the temp var inestaed (and unlock at the end of function/right after last usage of that var). So that one should not use a mutex each time a `device` is being accesses?

I think after 1.0.27 is out, we can finish the work I started in https://github.com/libusb/libusb/pull/1419 and also do testing with Thread Sanitizer.

tormodvolden:
I think this PR is the most pressing to get into master soon so that it can bake a bit there before release. Are you happy with this commit message?

Youw:
I'd take the thumb-up reaction as a "yes".

mcuee:
> I'd take the thumb-up reaction as a "yes".

Yes I think this is the most important PR to be merged now to prepare for 1.0.27 release.

hjelmn:
Yeah, let's go forward with this to not regress. I need to think a bit about the underlying issue and how best to fix it.

Youw (approved):

hjelmn:
Though now that I look at it. The issue really is that during re-enumeration the device object might go away. I think this is probably the correct fix. The temporary was only added during an earlier version of my change that I forgot to revert anyway.

PR #1430, opened 2024-01-06
darwin: fixed error checking regression

seanm:
13a6953379 slightly/subtly made an incorrect change to error checking. Before that commit, we had:

```
  if (kIOReturnSuccess != kresult || !plugInInterface) {
    return NULL;
  }
```

Which was correct. The commit changed the function signature. Instead of returning the pointer directly, it now returns an error code directly, and the pointer by reference.  The above block became:

```
  if (kIOReturnSuccess != kresult || !plugInInterface) {
    return darwin_to_libusb(kresult);
  }
```

But if kresult is somehow kIOReturnSuccess but plugInInterface is NULL (probably impossible), then we'd return LIBUSB_SUCCESS but a NULL pointer, which is a nonsense combination.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -573,9 +573,13 @@ static int darwin_device_from_service (struct libusb_context *ctx, io_service_t 
     nanosleep(&(struct timespec){.tv_sec = 0, .tv_nsec = 1000}, NULL);
   }
 
-  if (kIOReturnSuccess != kresult || !plugInInterface) {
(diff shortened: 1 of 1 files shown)


seanm:
@tormodvolden @mcuee because of https://github.com/libusb/libusb/issues/1386 I reviewed https://github.com/libusb/libusb/commit/13a6953379fa8c94bb2476dc460439d4ba2b7c17 very carefully, and also found this small regression.

Youw (approved):
seem to be right

PR #1431, opened 2024-01-08
(test only) collection of PRs from Sean

tormodvolden:

PR #1432, opened 2024-01-09
Fixed small buffer overread

seanm:
Calculate the correct source buffer size and iterate using it.

Also use separate indexes to index the source and destination buffers.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -1148,7 +1148,7 @@ int API_EXPORTED libusb_get_string_descriptor_ascii(libusb_device_handle *dev_ha
 	uint8_t desc_index, unsigned char *data, int length)
 {
 	union usbi_string_desc_buf str;
-	int r, si, di;
+	int r;
 	uint16_t langid, wdata;
 
 	/* Asking for the zero'th index is special - it returns a string
@@ -1184,20 +1184,26 @@ int API_EXPORTED libusb_get_string_descriptor_ascii(libusb_device_handle *dev_ha
 	else if ((str.desc.bLength & 1) || str.desc.bLength != r)
 		usbi_warn(HANDLE_CTX(dev_handle), "suspicious bLength %u for string descriptor (read %d)", str.desc.bLength, r);
 
-	di = 0;
-	for (si = 2; si < str.desc.bLength; si += 2) {
-		if (di >= (length - 1))
-			break;
+	/* stop one byte before the end of the destination, to leave room for null termination. */
+	int dest_max = length - 1;
 
-		wdata = libusb_le16_to_cpu(str.desc.wData[di]);
+	/* The source buffer is this long, see usbi_string_descriptor. */
+	int src_max = (str.desc.bLength - 1 - 1) / 2;
+
+	/* Ensure that we neither read nor write more than whichever buffer is smaller. */
+	int idx_max = MIN(dest_max, src_max);
+
+	int idx;
+	for (idx = 0; idx < idx_max; ++idx) {
+		wdata = libusb_le16_to_cpu(str.desc.wData[idx]);
 		if (wdata < 0x80)
-			data[di++] = (unsigned char)wdata;
+			data[idx] = (unsigned char)wdata;
 		else
-			data[di++] = '?'; /* non-ASCII */
+			data[idx] = '?'; /* non-ASCII */
 	}
 
-	data[di] = 0;
-	return di;
+	data[idx] = 0; /* null-terminate string */
+	return idx;
 }
 
 static int parse_iad_array(struct libusb_context *ctx,


libusb/descriptor.c
```diff
@@ -1185,18 +1185,20 @@ int API_EXPORTED libusb_get_string_descriptor_ascii(libusb_device_handle *dev_ha
 		usbi_warn(HANDLE_CTX(dev_handle), "suspicious bLength %u for string descriptor (read %d)", str.desc.bLength, r);
 
 	di = 0;
```
sonatique, line 1187: I would suggested dropping variable `di` since it is now 1:1 with `si` and only use for breaking the loop. Running the loop only while `si`/`di` is smaller than `length - 1` should be done by properly setting `si_max`

tormodvolden:
[continued from previous PR #1409]
> > OK, so what you are saying is that if the device erroneously returns a string descriptor with an odd bLength, we would be reading a byte too much from the descriptor buffer, and if bLength=255 also beyond the buffer?
> 
> Yes. It came about trying to figure out the count of `usbi_string_descriptor.wData`. Also, you can substitute 'maliciously' for 'erroneously'. :) In case it wasn't clear, the asserts in my WIP _are_ triggered.
> 
> > We are already issuing a warning on odd bLength
> 
> Correct. But it's just a warning and doesn't change the buffer read size.
> 
> > maybe we should just round bLength down to nearest even number before decoding the 16-bit characters? Well, maybe that's what your si_max is doing? :)
> 
> Exactly.
> 
> > Do we need all the size_of's? Is it even relevant what the storage size is, if we index them properly?
> 
> You mean `int si_max = (str.desc.bLength - sizeof(uint8_t) - sizeof(uint8_t)) / sizeof(uint16_t);`? Well, we could write `int si_max = (str.desc.bLength - 2) / 2;` if you prefer, but I thought the former was more explanatory, as it corresponds to the fields of the sturcture:
> 
> ```c
> struct usbi_string_descriptor {
> 	uint8_t  bLength;
> 	uint8_t  bDescriptorType;
> 	uint16_t wData[LIBUSB_FLEXIBLE_ARRAY] LIBUSB_COUNTED_BY((bLength - 2) / 2);
> } LIBUSB_PACKED;
> ```
> 
> Since compilers don't support it yet, I'm not sure how complicated the expressions within the 'counted_by' attribute can be...
> 

I see what you mean, thinking in terms of the C struct. Personally I don't think about us reading from a C struct here, but reading the array of bytes in the received descriptor. Obviously the two have to match. But I don't find it make the code easier to read nor more robust to "formalize" it with the sizeof's here. Actually, for the sake of fixing the bug for the release, without improving the code further, I would prefer the variant I am testing out here:  https://github.com/libusb/libusb/pull/1431/commits/4eb3fc3170330cca062d82d1a4fd6dbf2fba5fd8 It just rounds down the (byte) length to an even number, using a common idiom.

> > si and di are definitely source respectively destination index.
> 
> Thanks for confirming.
> 
> > di was used when reading from the source buffer because the latter is an array of 16-bit characters, as noted in a comment since it is a bit counter-intuitive.
> 
> Yeah. The source and destination buffers need not be the same size though, so I think it's much safer that they be individually indexed.

libusb/descriptor.c
```diff
@@ -1185,18 +1185,20 @@ int API_EXPORTED libusb_get_string_descriptor_ascii(libusb_device_handle *dev_ha
...
-	for (si = 2; si < str.desc.bLength; si += 2) {
+	int si_max = (str.desc.bLength - sizeof(uint8_t) - sizeof(uint8_t)) / sizeof(uint16_t);
```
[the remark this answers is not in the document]
tormodvolden: It is intentional to match the type inside usbi_string_descriptor, see the comment copied from the older PR.

seanm:
> Actually, for the sake of fixing the bug for the release, without improving the code further, I would prefer the variant I am testing out here: [4eb3fc3](https://github.com/libusb/libusb/commit/4eb3fc3170330cca062d82d1a4fd6dbf2fba5fd8) It just rounds down the (byte) length to an even number

I'm really not a fan of continuing to index the *source* buffer with the *destination* index. This part:

`wdata = libusb_le16_to_cpu(str.desc.wData[di]);`

It just looks like a mistake. And in fact there ***was*** a mistake there!!  Why?  Because it's mixing things up, and trying to be too clever.  It is safer to know and keep separate:

1. the index into the source buffer (`si`)
2. the index into the destination buffer (`di`)
3. the count of the source buffer (`si_max`)
4. the count of the destination buffer (`length`)

With my formulation, these 4 things are all kept separate and understandable. Yes, the length of the source buffer is a bit uglier to compute, but once it's computed correctly, you'll never overrun your buffer accidentally.

If it's all those sizeofs you dislike, how about just:

```diff
-int si_max = (str.desc.bLength - sizeof(uint8_t) - sizeof(uint8_t)) / sizeof(uint16_t);
+int si_max = (str.desc.bLength - 2) / 2; // remove space used by bLength & bDescriptorType, and count in int16s
```

libusb/descriptor.c
seanm: There is no guarantee or requirement that the source and destination buffers are the same size.  Buffer overruns are disastrously common in C. It's safer and easier to reason about preventing them when you know the size of each buffer and guard each buffer with its corresponding size and separate index. Trying to cleverly reduce down to one index makes it harder to reason about.

sonatique: Thanks. My suggestion was actually to replace current `si_max `with someting like `Max(length -1,  (str.desc.bLength - sizeof(uint8_t) - sizeof(uint8_t)) / sizeof(uint16_t))` to express that we'll loop until we reach the first smallest boundary and also use only one loop variable instead of two.
I feel the code would look more "common" and easy to understand / modify, but I guess that is more a matter of taste.

seanm: Ah, thanks for clarifying!

But you've just overrun the buffer again. :)  You want Min(), not Max(). :)  One might argue that this is again trying to be too clever, and leads to mistakes.

But with that corrected, I'd be ok with that formulation. How about you @tormodvolden ?  If so, I'll update this PR.

sonatique: Damn' ! Believe me: I clearly wanted to write `Min` ;-) Anyway, that is the only way my proposal make, so you autocorrected it, great.

tormodvolden: Yes, I like the Min() version too. Then it is clear from the start how long the for loop goes, without having to break out of it conditionally in the middle. And we can use `i` instead of `si` :)

tormodvolden:
Yes, I actually prefer -2 / 2 with a comment instead of sizeofs.

libusb/descriptor.c
seanm: OK, see updated commit.

PR #1434, opened 2024-01-09
Added a clang-tidy config file

seanm:
All checks are enabled except for those that cause any warning.

This is a starting point, some of the currently-supressed warnings can be fixed hereafter.


Youw:
I think it is best to introduce things like this along with the CI script to run/check it.
It is not implicitly used by any of the existing tools, right?

Youw:
And I agree - a very good starting point.

seanm:
> I think it is best to introduce things like this along with the CI script to run/check it.

Agreed. But I don't know anything about github CI, hopefully someone else can integrate this.

At least this file allows developers to use clang-tidy locally.

>It is not implicitly used by any of the existing tools, right?

Correct; as far as I know.

hjelmn:
Do we want to take this one step further and add clang-format as a prerequisite?

seanm:
> Do we want to take this one step further and add clang-format as a prerequisite?

Did you mean clang-tidy? clang-format is an orthogonal discussion. I had started work on clang-format, but waiting reply here: https://github.com/libusb/libusb/pull/1421#issuecomment-1873605415

hjelmn:
I mean both. I kind of want to set minimum standards on commits.

seanm:
> I mean both. I kind of want to set minimum standards on commits.

I'm all for clang-format too. But it's unrelated to clang-tidy and this PR. Let's move the discussion here: https://github.com/libusb/libusb/issues/1442

seanm:
@tormodvolden for example, if this was merged to master, then I (and others) could be using clang-tidy to check our work in other branches.

Youw (approved):
#1457

seanm:
@tormodvolden @hjelmn is there any reason to not just merge this? It can't break anything else, it's just a config file...

PR #1439, opened 2024-01-21
Cleanup of darwin's usbi_backend structure

seanm:
No real change.

- rearranged elements to be initialized in the same order as they appear in the struct declaration
- explicitly set values to NULL so that global searches for things like `wrap_sys_device` easily reveal that the function is NULL on Darwin


Youw (approved):
I like this change, even though it is purely cosmetic.

tormodvolden:
Makes sense to me. @hjelmn I'll push this unless you have comments to it.

PR #1443, opened 2024-01-23
WIP: add a .clang-format config file

seanm:


Youw:
Can you make another draft/PR with formatting already applied?
(and include `.git-blame-ignore-revs`, if feasible)

seanm:
Yes, I can do that... but I'll wait for some more feedback from others about what options the project wants.

In my local copy, I do diffs using [Araxis Merge](https://www.araxis.com/merge/index.en) which has some nice options to ignore only leading whitespace, trailing whitespace, inner whitespace, etc. With my current .clang-format file, if I ignore leading and trailing whitespace, there aren't very many changes in fact.

PR #1448, opened 2024-01-24
core: exit does not dereference device pointers when logging

sonatique:
Device pointers that wrongly remain in the referenced device list could point to already freed memory. Dereferencing them for the purpose of logging devices related data could lead to segmentation faults or inaccurate information

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2578,8 +2578,7 @@ err_free_ctx:
 void API_EXPORTED libusb_exit(libusb_context *ctx)
 {
 	struct libusb_context *_ctx;
-	struct libusb_device *dev;
-
+	
 	usbi_mutex_static_lock(&default_context_lock);
 
 	/* if working with default context, only actually do the deinitialization
@@ -2626,11 +2625,8 @@ void API_EXPORTED libusb_exit(libusb_context *ctx)
 
 	usbi_io_exit(_ctx);
 
-	for_each_device(_ctx, dev) {
-		usbi_warn(_ctx, "device %d.%d still referenced",
-			dev->bus_number, dev->device_address);
-		DEVICE_CTX(dev) = NULL;
-	}
+	if (!list_empty(&_ctx->usb_devs))
+		usbi_warn(_ctx, "application left some devices referenced");
 
 	if (!list_empty(&_ctx->open_devs))
 		usbi_warn(_ctx, "application left some devices open");


mcuee:
AppVeyor CI build failed because of a warning treated as error.

```
C:\projects\libusb\libusb\core.c(2581,24): error C2220: the following warning is treated as an error [C:\projects\libusb\msvc\libusb_static.vcxproj]
         	struct libusb_device *dev;
         	                      ^
         
     2>C:\projects\libusb\libusb\core.c(2581,24): warning C4101: 'dev': unreferenced local variable [C:\projects\libusb\msvc\libusb_static.vcxproj]
         	struct libusb_device *dev;
         	                      ^
         
         descriptor.c
       ClCompile:
```

mcuee:
Linux umock test failed which is very strange. I have re-run it but it still failed, should not be related to this PR.

```
+ echo ''
+ echo 'Running umockdev tests ...'
Running umockdev tests ...
+ tests/umockdev
TAP version 13
# random seed: R02Sfc551fa146179a46fb9bb6d0a92934a6
1..10
# Start of libusb tests
# DEBUG: umockdev.vala:123: Created udev test bed /tmp/umockdev.5BT4H2
# DEBUG: umockdev.vala:1297: parsing device description for /devices/usb1
# DEBUG: umockdev.vala:1366: creating device /devices/usb1 (subsystem usb)
# DEBUG: umockdev.vala:[1411](https://github.com/libusb/libusb/actions/runs/7644366026/job/20835487537?pr=1448#step:12:1412): create_node_for_device: creating file device /tmp/umockdev.5BT4H2/dev/bus/usb/001/001
# DEBUG: umockdev.vala:771: umockdev_testbed_uevent: lazily initializing uevent_sender
# DEBUG: umockdev.vala:775: umockdev_testbed_uevent: sending uevent add for device /sys/devices/usb1
[ 0.000920] [00002020] libusb: debug [libusb_get_device_list]
[ 0.000946] [00002020] libusb: debug [libusb_get_device_descriptor]
[ 0.000953] [00002020] libusb: debug [libusb_open] open 1.1
[ 0.001452] [00002020] libusb: debug [usbi_add_event_source] add fd 13 events 4
[ 0.001479] [00002020] libusb: debug [libusb_close]
[ 0.001485] [00002020] libusb: debug [usbi_remove_event_source] remove fd 13
[ 0.001512] [00002020] libusb: debug [libusb_get_device_list]
[ 0.001519] [00002020] libusb: debug [libusb_get_device_descriptor]
[ 0.001523] [00002020] libusb: debug [libusb_open] open 1.1
[ 0.002219] [00002020] libusb: debug [usbi_add_event_source] add fd 13 events 4
[ 0.002236] [00002020] libusb: debug [libusb_close]
[ 0.002241] [00002020] libusb: debug [usbi_remove_event_source] remove fd 13
[ 0.002257] [00002020] libusb: debug [libusb_get_device_list]
# DEBUG: umockdev.vala:150: Removing test bed /tmp/umockdev.5BT4H2
[ 0.002264] [00002020] libusb: debug [libusb_exit]
[ 0.002268] [00002020] libusb: debug [libusb_unref_device] destroy device 1.1
[ 0.002341] [00002020] libusb: debug [usbi_remove_event_source] remove fd 10
[ 0.002354] [00002020] libusb: debug [usbi_remove_event_source] remove fd 9
# GLib-DEBUG: unsetenv() is not thread-safe and should not be used after threads are created
ok 1 /libusb/open-close
# GLib-DEBUG: setenv()/putenv() are not thread-safe and should not be used after threads are created
# DEBUG: umockdev.vala:123: Created udev test bed /tmp/umockdev.QSX4H2
# DEBUG: umockdev.vala:1297: parsing device description for /devices/usb1
# DEBUG: umockdev.vala:1366: creating device /devices/usb1 (subsystem usb)
# DEBUG: umockdev.vala:1411: create_node_for_device: creating file device /tmp/umockdev.QSX4H2/dev/bus/usb/001/001
# DEBUG: umockdev.vala:771: umockdev_testbed_uevent: lazily initializing uevent_sender
# DEBUG: umockdev.vala:775: umockdev_testbed_uevent: sending uevent add for device /sys/devices/usb1
[ 0.007253] [00002020] libusb: error [usbi_get_context] API misuse! Using non-default context as implicit default.
-bash: line 29:  8224 Segmentation fault      (core dumped) tests/umockdev
+ '[' 139 -eq 0 ']'
+ exit 1
Error: Process completed with exit code 1.
```

sonatique:
> AppVeyor CI build failed because of a warning treated as error.
> 
> ```
> C:\projects\libusb\libusb\core.c(2581,24): error C2220: the following warning is treated as an error [C:\projects\libusb\msvc\libusb_static.vcxproj]
>          	struct libusb_device *dev;
>          	                      ^
>          
>      2>C:\projects\libusb\libusb\core.c(2581,24): warning C4101: 'dev': unreferenced local variable [C:\projects\libusb\msvc\libusb_static.vcxproj]
>          	struct libusb_device *dev;
>          	                      ^
>          
>          descriptor.c
>        ClCompile:
> ```

ouch sorry, I was too quick on this one, I will correct this

sonatique:
@mcuee I have an explanation for the seg fault here. Things are quite entangled as far as I can say.
So: it is the removal of line 2632 `DEVICE_CTX(dev) = NULL;` which causes the crash later.
Here is why:
In umockdev `test_implicit_default` function, the we have 
```
libusb_init_context(/*ctx=*/NULL, 
ibusb_get_device_list(NULL
libusb_exit(NULL);
libusb_free_device_list(devs, TRUE);
```
Here, the code intentionally leave a device in the usb_devs list of the default context when calling `libsub_exit`, I understand that this is part of the test.
master code enumerate this list and do `DEVICE_CTX(dev) = NULL;` for the device in the list. 
Before this call` dev->ctx` points to the default context, after it it points to NULL.
`libusb_exit `then free the default context.
Later, in `libusb_free_device_list `(list that also contains the same device) there is a call to `libusb_unref_device ` on this device (whose context is == NULL, as explained) because `unref_devices `parameter is TRUE.
Since `refcnt` reaches 0 the device has to be destroyed. This trigger a usbi_dbg log at line 1299 (core.c) ; `usbi_dbg(DEVICE_CTX(dev), "destroy device %d.%d"`. Since `DEVICE_CTX(dev)` is NULL, the log use no context, everything is fine.

Now, with my PR: we don't do `DEVICE_CTX(dev) = NULL;` anymore, because, since `dev` can be freed already at this point (remember that if we are here it is because there is an error leaving a potentially freed device in the list), we don't want to risk to seg fault by writing to memory we don't own anymore. Therefore, i the case of umockdev, we still have the device (the one left inside the list) with a context pointing to the default context, which is freed a few lines later.
So, I guess you got it: when we reach the `usbi_dbg `log in `libusb_unref_device`, we take the context from the device, which is not NULL but points to freed memory. `ubsi_dbg `access it and causes the seg fault.  

So, well, well, well. umockdev does many "wrong" things on purpose. Accessing the list of device, hence the devices (even for freeing the list) after calling `libusb_exit `is explicitly disallowed by the comment in `libusb_exit`: "Don't bother with locking after this point because **unless there is an application bug, nobody will be accessing the context**" and umockdev access the context after this point (via the device) because it calls a libusb function (`libusb_free_device_list`) after exit, when creating a error condition on purpose.  This code is working in master because the dangling device is not freed. If it would have been freed (we can imagine other/new tests doing it) then we would have get the seg fault in `libusb_exit` when printing the dangling device info.

And all this happens only when logging with Debug level.

If I replace `usbi_dbg(DEVICE_CTX(dev), "destroy device %d.%d"` with `usbi_dbg(NULL, "destroy device %d.%d"` we of course no longer have the seg fault, but umockdev fails early during another test because it expects to see the "libusb_unref_device's destroy device" log message and cannot see it because it is not logged within the correct context.

So, hum: it seems quite trick to do everything right at once. I am still convinced that it is not good to access dev pointers directly (for nulling the context  and/or logging) in the check loop inside `libusb_exit` because in case of error we cannot be sure of anything, including whether those pointers are valid or not. But at the same time, having umockdev challenging the code is a good thing.

I would favor keeping my PR and changing umockdev behavior, but I am not sure. Do you have any idea about what is the best thing to do?

mcuee:
I need some time to digest why umockdev fails but I agree with what this PR tries to do. 

umockdev fails anyway without this PR when logging is enabled.
* https://github.com/libusb/libusb/issues/1449

mcuee:
@tormodvolden 

Please take a look at the points raised by @sonatique. Thanks.

sonatique:
Hello @mcuee @tormodvolden ,
I slept on this and woke up with the following thought: unless what I understood at the beginning, I now think things are like this:

- a user-level application can only "forget" **not** freed device pointers in the context list. There is/shall be no way using the public API to end-up in situations where some freed device pointers are left in the list when exiting the context: this list is internal and only manipulated by libusb-1.0 internal implementations details. So we can assume it is safe to dereference pointers as is it done now. umockdev is a "user application" so whatever it does shall never seg fault. It is the case with master and everything both inside libusb-1.0 lib a and examples/tests assume this.

- having freed device pointers in the list indicate an internal implementation error, that shall never be released: we shall detect seg fault caused by dereferencing during testing and obviously correct issues before releasing / merging. We are in this situation with my windows hotplug PR in some (very) corner cases. This is obviously a show stopper as long as it is not understood and corrected.

Conclusion: we should decline this current PR aimed at changing libusb_exit code and I should go back fixing windows hotplug.

mcuee:
I think we can keep this PR open and let @tormodvolden and @hjelmn take a look first.

As for master, the crash issue with umockdev still exists, which may not be related to this issue. But it does  cause crash as well.
https://github.com/libusb/libusb/issues/1449

mcuee:
> * having freed device pointers in the list indicate an internal implementation error, that shall never be released: we shall detect seg fault caused by dereferencing during testing and obviously correct issues before releasing / merging. We are in this situation with my windows hotplug PR in some (very) corner cases. This is obviously a show stopper as long as it is not understood and corrected.

Not so sure if the dicussion here is relevant here or not.
* https://github.com/libusb/libusb/issues/1287#issuecomment-1622583938

@tormodvolden wrote of the following.
> libusb_get_device_list() returns a list of devices structures. It refcounts each of the devices. Maybe some of the real/underlying devices that they are associated with don't exist any longer after the call, but the devices structures still exist as long as they are referenced.
>
> libusb_get_device_list() works perfectly in pair with libusb_free_device_list(, 1) because the former has refcounted all the devices in the list, and the latter unrefs them. Unless something else in between has messed up the reference counting, all devices will be released.

tormodvolden:
> Conclusion: we should decline this current PR aimed at changing libusb_exit code and I should go back fixing windows hotplug.

I'll mark this "Draft" for now.

Issue #1449, opened 2024-01-25
umockdev fails when built with --enable-debug-log


sonatique:
When I do
```
./configure
make
./tests/umockdev
```
I get a successful run (no ERROR) with the following output:
```
# random seed: R02Sb1078e9ba8805394a1051f93621bf7c1
1..10
# Start of libusb tests
# GLib-DEBUG: setenv()/putenv() are not thread-safe and should not be used after threads are created
# DEBUG: umockdev.vala:110: Created udev test bed /tmp/umockdev.KYRBI2
# DEBUG: umockdev.vala:1294: parsing device description for /devices/usb1
# DEBUG: umockdev.vala:1363: creating device /devices/usb1 (subsystem usb)
# DEBUG: umockdev.vala:1408: create_node_for_device: creating file device /tmp/umockdev.KYRBI2/dev/bus/usb/001/001
# DEBUG: umockdev.vala:765: umockdev_testbed_uevent: lazily initializing uevent_sender
# DEBUG: umockdev.vala:769: umockdev_testbed_uevent: sending uevent add for device /sys/devices/usb1
[ 0.000749] [00354714] libusb: debug [libusb_get_device_list]
[ 0.000781] [00354714] libusb: debug [libusb_get_device_descriptor]
[ 0.000805] [00354714] libusb: debug [libusb_open] open 1.1
[ 0.001385] [00354714] libusb: debug [usbi_add_event_source] add fd 13 events 4
[ 0.001436] [00354714] libusb: debug [libusb_close]
[ 0.001456] [00354714] libusb: debug [usbi_remove_event_source] remove fd 13
[ 0.001498] [00354714] libusb: debug [libusb_get_device_list]
[ 0.001520] [00354714] libusb: debug [libusb_get_device_descriptor]
[ 0.001535] [00354714] libusb: debug [libusb_open] open 1.1
[ 0.001853] [00354714] libusb: debug [usbi_add_event_source] add fd 13 events 4
[ 0.001878] [00354714] libusb: debug [libusb_close]
[ 0.001897] [00354714] libusb: debug [usbi_remove_event_source] remove fd 13
[ 0.001924] [00354714] libusb: debug [libusb_get_device_list]
[ 0.001942] [00354714] libusb: debug [libusb_exit]
[ 0.001955] [00354714] libusb: debug [libusb_unref_device] destroy device 1.1
[ 0.002034] [00354714] libusb: debug [usbi_remove_event_source] remove fd 10
[...]

PR #1450, opened 2024-01-28
umockdev: immediately abort when ENABLE_DEBUG_LOGGING is defined

sonatique:
Umockdev tests rely on being able to set per context callbacks for validating functions behaviors. As per documentation:"If ENABLE_DEBUG_LOGGING is defined then per context callback function will never be called." In this case, Umockdev will then incorrectly fails most of the tests, providing erroneous results.

With this commit, Umockdev will immediately exit with an error message when compiled with ENABLE_DEBUG_LOGGING defined.

### DIFF
--- a/tests/umockdev.c
+++ b/tests/umockdev.c
@@ -1120,6 +1120,11 @@ test_hotplug_add_remove(UMockdevTestbedFixture * fixture, UNUSED_DATA)
 int
 main(int argc, char **argv)
 {
+#if defined ENABLE_DEBUG_LOGGING
+	fprintf(stderr, "ENABLE_DEBUG_LOGGING is defined and UMockdev doesn't support it since tests rely on per context callbacks, aborting.\n");
+	exit(-1);
+#endif
+
 	g_test_init(&argc, &argv, NULL);
 
 	g_test_add("/libusb/open-close", UMockdevTestbedFixture, NULL,


tormodvolden:
It might be better to simply disable umockdev in configure.ac if `--enable-debug-log`, the same way it is already done for `--disable-log`. For instance:
```
--- a/configure.ac
+++ b/configure.ac
@@ -383,7 +383,7 @@ AC_ARG_ENABLE([tests-build],
 
 AM_CONDITIONAL([BUILD_EXAMPLES], [test "x$build_examples" != xno])
 AM_CONDITIONAL([BUILD_TESTS], [test "x$build_tests" != xno])
-AM_CONDITIONAL([BUILD_UMOCKDEV_TEST], [test "x$ac_have_umockdev" = xyes -a "x$log_enabled" != xno])
+AM_CONDITIONAL([BUILD_UMOCKDEV_TEST], [test "x$ac_have_umockdev" = xyes -a "x$log_enabled" != xno -a "x$debug_log_enabled" != xno])
 AM_CONDITIONAL([CREATE_IMPORT_LIB], [test "x$create_import_lib" = xyes])
 AM_CONDITIONAL([OS_DARWIN], [test "x$backend" = xdarwin])
 AM_CONDITIONAL([OS_HAIKU], [test "x$backend" = xhaiku])

```

sonatique:
@tormodvolden OK yes, why not, makes sense as well. I tend to prefer to have things existing but telling something rather than not existing at all, since I think it makes it easier to understand the problem in the first place, and I just implemented what we agreed on the discussion: https://github.com/libusb/libusb/issues/1449 .
But in this case, your solution is most probably better given umockdev compilation is already conditioned to some defines being (not) defined, so it's just a matter of complementing it.
Given it is a different approach I created another PR: https://github.com/libusb/libusb/pull/1451 , and discarded this one.

PR #1451, opened 2024-01-29
umockdev: do not build when ENABLE_DEBUG_LOGGING is defined

sonatique:
Umockdev tests rely on being able to set per context callbacks for validating functions behaviors. As per documentation: "If ENABLE_DEBUG_LOGGING is defined then per context callback function will never be called." When this is this case, Umockdev will therefore incorrectly fails, providing erroneous results, such that it is better not to build it at all.


tormodvolden:
I had this already queued up and tested locally, so I should have told you there was no need for a PR. But thanks for confirming the approach.

PR #1452, opened 2024-01-30
hotplug_exit: removes all parents with a single child remaining in parent tree + lock device list

sonatique:
1) Removes all parents with a single child remaining in parent tree
This ensure that no parents of the direct parent of the device being considered
are left in the list, when appearing before their child in the processing order
of the context device list.

2) mutex protection of context device list while cleaning it

This prevents concurrent access to the list by the event background
thread, which could still be processing a previous hotplug event and
having to modify the list

Note: this PR is based on to other PR given these other fixes are needed for proper testing. Please only consider the last commit as part of this PR.

### DIFF
--- a/libusb/hotplug.c
+++ b/libusb/hotplug.c
@@ -214,6 +214,7 @@ void usbi_hotplug_exit(struct libusb_context *ctx)
 		free(msg);
 	}
 
+	usbi_mutex_lock(&ctx->usb_devs_lock); /* hotplug thread might still be processing an already triggered event, possibly accessing this list as well */
 	/* free all discovered devices */
 	for_each_device_safe(ctx, dev, next_dev) {
 		/* remove the device from the usb_devs list only if there are no
@@ -222,11 +223,12 @@ void usbi_hotplug_exit(struct libusb_context *ctx)
 		if (usbi_atomic_load(&dev->refcnt) == 1) {
 			list_del(&dev->list);
 		}
-		
+
 		usbi_recursively_remove_parents(dev, next_dev);
-		
+
 		libusb_unref_device(dev);
 	}
+	usbi_mutex_unlock(&ctx->usb_devs_lock);
 
 	usbi_mutex_destroy(&ctx->hotplug_cbs_lock);
 }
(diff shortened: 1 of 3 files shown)


mcuee:
This does not sore out issue for #1445 under macOS, but at least it does not seem to cause regressions under macOS.

* https://github.com/libusb/libusb/issues/1445#issuecomment-1918149420

libusb/hotplug.c
```diff
@@ -215,6 +215,7 @@ void usbi_hotplug_exit(struct libusb_context *ctx)
 	}
 
 	/* free all discovered devices. due to parent references loop until no devices are freed. */
```
tormodvolden, line 217: Not directly affecting your PR, but I notice this comment seems outdated (already since 87a8578). It used to be a do-while loop.

tormodvolden:
Just a nit on your commit summary "removes ...", preferably imperative instead of present tense, and it could be shorter and leave the details to the commit message, like "Remove parents recursively".

libusb/hotplug.c
sonatique: I simplified the comment

sonatique:
> Just a nit on your commit summary "removes ...", preferably imperative instead of present tense, and it could be shorter and leave the details to the commit message, like "Remove parents recursively".

How does it look now? Thanks

tormodvolden:
Thanks, now only whitespace nits :)
```
$ git diff --check HEAD^^
libusb/hotplug.c:226: trailing whitespace.
+               
libusb/hotplug.c:228: trailing whitespace.
+               

```

tormodvolden:
@hjelmn Are you fine with these changes?

mcuee:
@hjelmn 

Ping. Thanks.

tormodvolden:
We see that the hotplug_exit mutex is fixing a number of observed bugs, so I am not going to wait longer with this. I hope @hjelmn will review this closely before next release. In the meantime it will get extended testing as well.

I also see no issues with the recursive parents removal. It might only be needed on the Windows backend, but that doesn't affect the correctness of it.

tormodvolden:
Note that your whitespace fixes were squashed into the wrong commits, but I fixed them up.

PR #1456, opened 2024-02-02
Added an API call to check the connection status of a device handle

Quantx:
Aside from hot-plug events there is no good way to quickly check if a device handle is still attached to the system. I believe the only way to currently check if a device handle is still attached in a non-blocking manner is by attempting to claim an already claimed interface. This new API call adds a convenient method by which to check this.

```
int libusb_check_connected(libusb_device_handle * dev_handle)
```

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -372,6 +372,7 @@ if (cfg != desired)
   * - libusb_attach_kernel_driver()
   * - libusb_bulk_transfer()
   * - libusb_cancel_transfer()
+  * - libusb_check_connected()
   * - libusb_claim_interface()
   * - libusb_clear_halt()
   * - libusb_close()
@@ -1436,6 +1437,23 @@ int API_EXPORTED libusb_open(libusb_device *dev,
 	list_add(&_dev_handle->list, &ctx->open_devs);
 	usbi_mutex_unlock(&ctx->open_devs_lock);
 	*dev_handle = _dev_handle;
+
+	return 0;
+}
+
+/** \ingroup libusb_dev
+ * Checks if a device handle is still connected to the system.
+ *
+ * This is a non-blocking function; no requests are sent over the bus.
+ *
+ * \param dev_handle device handle to be checked.
+ * \returns 0 if the device is connected
+ * \returns \ref LIBUSB_ERROR_NO_DEVICE if the device has been disconnected
+ */
+int API_EXPORTED libusb_check_connected(libusb_device_handle * dev_handle)
+{
+	if (!usbi_atomic_load(&dev_handle->dev->attached))
+		return LIBUSB_ERROR_NO_DEVICE;
 
 	return 0;
 }
--- a/libusb/libusb-1.0.def
+++ b/libusb/libusb-1.0.def
@@ -10,6 +10,8 @@ EXPORTS
   libusb_bulk_transfer@24 = libusb_bulk_transfer
   libusb_cancel_transfer
   libusb_cancel_transfer@4 = libusb_cancel_transfer
+  libusb_check_connected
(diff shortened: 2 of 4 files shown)


seanm:
This kind of API is prone to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) problems, so I'm not sure it's advisable to add.  With multithreading and hotplug, the answer it gives can be incorrect an instant later.

libusb/core.c
```diff
@@ -1440,6 +1441,23 @@ int API_EXPORTED libusb_open(libusb_device *dev,
 	return 0;
 }
 
+/** \ingroup libusb_dev
+ * Checks if a device handle is still connected to the system.
+ *
+ * This is a non-blocking function; no requests are sent over the bus.
+ *
+ * \param dev_handle device to be checked check.
+ * \returns 0 on success
```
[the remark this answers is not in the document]
Quantx: Done

Quantx:
The TOCTOU issue is addressed by the fact that all other API calls will also report `LIBUSB_ERROR_NO_DEVICE` if the device was disconnected between a request to `libusb_check_connected()` and the subsequent API call. The reason I want to see this API call added is because there is currently no way to synchronously check whether a device handle is still connected without also causing some other side effect. While you can create a mechanism with hot-plug events to achieve a similar result, I think that is a lot of work to acquire a piece of information that could easily be returned to the user by this API call. Furthermore, on platforms where hot-plug events are not supported, there is currently no dedicated API call to detect whether or not a device is still connected.

seanm:
Yeah, sounds reasonable enough.

Could you squash your commits?

Quantx:
I've squashed it

libusb/core.c
```diff
@@ -1440,6 +1441,23 @@ int API_EXPORTED libusb_open(libusb_device *dev,
 	return 0;
 }
 
+/** \ingroup libusb_dev
+ * Checks if a device handle is still connected to the system.
+ *
+ * This is a non-blocking function; no requests are sent over the bus.
+ *
+ * \param dev_handle device handle to be checked.
+ * \returns 0 on if the device is connected
```
seanm, line 1450: I think the "on" is leftover from the previous edit?

Quantx: Fixed

seanm (approved):

mcuee:
It may be good to add this to one of the existing examples as well to make it easy to test the new function.

Quantx:
I've added a call to `libusb_check_connected()` inside one of the basic open/close tests.

Quantx:
If there are no further comments or concerns regarding this PR can it please be merged?

mcuee:
> If there are no further comments or concerns regarding this PR can it please be merged?

We will need to have the review and approval of the current active developers, basically @tormodvolden and/or @hjelmn.

seanm (approved):

maybetaken:
Kernel will report POLLHUP event from the disconnected device. If libusb tries to communicate with a POLLHUP device, it will hang at ioctl for some milliseconds until POLLERR reports from the kernel, libusb will remove event. But currently, libusb only handles POLLERR. I think if your want to get a quick notice that device disconnected is deal with POLLHUP event and it will send to LIBUSB_TRANSFER_NO_DEVICE instead of quiring this from the user application side.

minicarpet (approved):

minicarpet:
This PR is small and could be merged, previous comment could be an other PR to speed up device disconnection event

mcuee:
Need to resolve the conflict.

mcuee:
@sonatique 

Please review this PR as well. Thanks.

PR #1458, opened 2024-02-02
WIP stress_mt rework

tormodvolden:
Just showing what I am working on, in case someone else is thinking in this line.

I also has some more work uncommitted to test enumeration without opening devices (so a third option) but it doesn't compile yet.


tormodvolden:
The third option stuff now compiles but I haven't verified the logic yet.

PR #1428, opened 2024-01-06
Add support for SuperSpeed+ Capability Descriptors

fabiensanglard:
See specs in USB 3.1 specs in section:

9.6.2.5 SuperSpeedPlus USB Device Capability

Fix issue #1429

PR #1460, opened 2024-02-05
Replaced parse_descriptor() function, fixed some overruns

seanm:


Youw:
To my personal taste - I like this (explicit) way better.
Although, I don't have an oppinion if the change is inline with the whole design of this part of the code.

I've reviewed it, and the change looks good.
I also found where the improvements are, but would be great to describe those explicitly in the PR description section.

Youw (approved):

Youw:
Oh, I didn't see the commit message(s).
Yes, that would suffice, but I believe  each PR is merged into master as a single commit, so those commit messages would have been lost, unless someone explicitly combines them into a single message for a single (final) commit.

NOTE: libusb does not use squash-merge functionality github provides.

seanm:
>I believe each PR is merged into master as a single commit

Really?!?! Why?

That really reduces the usefulness of `git bisect` for example.

Youw:
> Really?!?! Why?

A very simple and verified over time philosophy: a single PR should contain a single independent change and solve a single "problem".
This worked very well for many years, not only for libusb.

If a PR is more than one commit - those are either not independent or should be split into separate PRs.

seanm:
> A very simple and verified over time philosophy: a single PR should contain a single independent change and solve a single "problem".

This PR does solve a single "problem".  The problem of `parse_descriptor()` being broken.  This single problem is divided into smaller subproblems.

If the official libusb policy is indeed 1 commit per PR, then it really should be documented in the `HACKING` file (or somewhere) because it is very unusual.

tormodvolden:
No, there is no rule saying a PR gets squashed. We will certainly want to squash fixup commits and other WIP artifacts, but if each commit is an independent change they go in one by one. Sometimes it is for instance good to separate preparatory commits from those commits that really make a change, it makes it easier to review up front, and to understand and verify afterwards.

tormodvolden:
And if and when we squash commits, no commit messages gets lost. When you use "squash" in git rebase it will add all commit messages, then you edit it before committing.

Youw:
Alright, looks like my observatory assumption is not correct. Some habbit of mine to generalize things I see consistently.
Maybe I haven't been around long-enough to see PRs with more that one commit merged with rebase-merge strategy.

But in any case I find it way more convenient to see the summary of the PR upfront, rather than having to look for individual commit messages. Maybe it's just my habbit.

tormodvolden:
After a quick look at these, I see no need to squash them. OK, they all improve parse_descriptor, but in different ways. Each of them would have made sense without the others.

A bunch of easy-to-review commits is better than huge commits where you have to look back and forth to see what pieces play together. And the larger the commit the more sure there will be something wrong that slips in and is missed in a review. I think most regressions I have found are from mega-commits where too many things were changed at a time.

Youw:
> A bunch of easy-to-review commits is better than huge commits

> Each of them would have made sense without the others

To my experience - that's a good reason to have separate PRs, not just separate commits.
I'm often seen cases when a change request or discussions to one of the commits would delay the merge of the others, not to mention potential inconvenience with interractive rebases/etc., to make changes only to some part of a multi-commit PR.

tormodvolden:
> But in any case I find it way more convenient to see the summary of the PR upfront, rather than having to look for individual commit messages. Maybe it's just my habbit.

If you click on the "..." after each commit summary on this web page (e.g. under "seanm added 6 commits"), there is a drop-down with the full commit message,.

Most of the time it is the opposite problem, people write a lot in the PR description (which doesn't get into the repo) while their commit messages are poor or empty.

tormodvolden:
> > Each of them would have made sense without the others
> 
> To my experience - that's a good reason to have separate PRs, not just separate commits.

Sometimes they depend on each other to avoid conflicts, and it is just a saner way to work both for coder and reviewer. They fit together logically so they are best treated in the same PR. There is a risk that discussion about one commit delays the others, but it is rarely a big problem. The committer can in such cases also choose to cherry-pick the easy ones, and leave the questionable commits to be fixed up (and rebased if needed).

Youw:
[This](https://github.com/libusb/libusb/pull/1460/commits/31781b7bd93c86a1b3967c8d1d9014132df8a0a0) commit, apparently introduces an issue:
```
C:\projects\libusb\libusb\descriptor.c(276,24): error C2220: the following warning is treated as an error [C:\projects\libusb\msvc\libusb_static.vcxproj]
         			ifp->extra_length = len;
         			                    ^
         
     2>C:\projects\libusb\libusb\descriptor.c(276,24): warning C4244: '=': conversion from 'intptr_t' to 'int', possible loss of data [C:\projects\libusb\msvc\libusb_static.vcxproj]
         			ifp->extra_length = len;
         			                    ^
         
     2>C:\projects\libusb\libusb\descriptor.c(427,[28](https://ci.appveyor.com/project/dickens/libusb/builds/49109043/job/or3trscp22wbflua#L28)): warning C4244: '+=': conversion from 'intptr_t' to 'int', possible loss of data [C:\projects\libusb\msvc\libusb_static.vcxproj]
         			config->extra_length += len;
         			                        ^
```
At least on MSVC x64 compilers.

libusb/descriptor.c
```diff
@@ -127,7 +126,7 @@ static int parse_endpoint(struct libusb_context *ctx,
 
 	/* Copy any unknown descriptors into a storage area for drivers */
 	/*  to later parse */
-	len = (int)(buffer - begin);
+	intptr_t len = (intptr_t)(buffer - begin);
```
Youw, line 129: now, when I looke a this - what's the point of using `intptr_t` and not using `int` as it was originally?

seanm: Admittedly, not much.

The old cast is theoretically truncating (64 bit pointers, 32 bit ints), but of course they should never be so far apart.

This way is just slightly more correct. `len` is now the correct type, and at least the check on the next line (`if (len <= 0)` is therefore more correct too in the face of overflow.

Youw: When we casted to `int` right away, in case of overflow there was a chance it would be cut off by the `<0` check.
If we're to make it property right, then in addition to `<0` check we should also check for (at least) `INT_MAX`.

_Unless_ it is already checked elsewhere for some othe constant (e.g. something usb-specific), in which case casting it to `int` instead of `intptr_t` would make more elegant implementation.

hjelmn (approved):

libusb/descriptor.c
tormodvolden: parse_configuration() or its sole caller raw_desc_to_config() is called with an _int_ size. raw_desc_to_config() is only called from libusb_get_active_config_descriptor() where the size is taken from wTotalLength which therefore cannot be larger than 65535 ever.

tormodvolden:
The last commit here is marked WIP, I guess the others are ready to go?

seanm:
> The last commit here is marked WIP, I guess the others are ready to go?

Let's figure out the WIP one before merging... Let me see if I remember what I was thinking those weeks ago....

seanm:
Ah yes... The old code is this:

```c
		// Second pass: Iterate through desc list, fill IAD structures
		consumed = 0;
		i = 0;
		while (consumed < size) {
			header.bLength = buffer[0];
			header.bDescriptorType = buffer[1];
			if (header.bDescriptorType == LIBUSB_DT_INTERFACE_ASSOCIATION) {
				iad[i].bLength = buffer[0];
				iad[i].bDescriptorType = buffer[1];
				iad[i].bFirstInterface = buffer[2];
				iad[i].bInterfaceCount = buffer[3];
				iad[i].bFunctionClass = buffer[4];
				iad[i].bFunctionSubClass = buffer[5];
				iad[i].bFunctionProtocol = buffer[6];
				iad[i].iFunction = buffer[7];
				i++;
			}
			
			buffer += header.bLength;
			consumed += header.bLength;
		}
```

We know the length of `buffer` is `size` because it's passed as a parameter to this function.

`buffer` is incremented here by `header.bLength` (aka `buffer[0]`). But if the contents of `buffer[0]` are unexpected/garbage, then it seems to me `buffer` could get incremented too far...

libusb/descriptor.c
```diff
@@ -1270,12 +1270,12 @@ static int parse_iad_array(struct libusb_context *ctx,
...
 		// Second pass: Iterate through desc list, fill IAD structures
-		consumed = 0;
+		int remaining = size;
 		i = 0;
-		while (consumed < size) {
+		do {
 			header.bLength = buffer[0];
 			header.bDescriptorType = buffer[1];
-			if (header.bDescriptorType == LIBUSB_DT_INTERFACE_ASSOCIATION) {
+			if (header.bDescriptorType == LIBUSB_DT_INTERFACE_ASSOCIATION && (remaining >= 8)) {
```
tormodvolden, line 1278: Why not sizeof() here as well, instead of "magic" 8?

seanm: I took a look, but not sure what I could substitute... the 8 is because the next lines read elements 0 through 7.  Using `sizeof libusb_interface_association_descriptor` doesn't feel right, because that's the destination side, not the source side.  And also that struct is not `LIBUSB_PACKED`.

tormodvolden: If it is clear it should be 8, we can still write 8 instead of an ever so correct, formal compiler abstraction, but we could consider a small comment like a valid IAD descriptor must have 8 bytes. We also define many descriptor length macros in libusb.h, maybe that is over the top here. I guess I got hung up in the inconsistency where above we use a sizeof instead of 2 (I see it more like any descriptor must be long enough to contain its own header otherwise we cannot localize the next, rather than having to fill up a struct if copied).

seanm: `sizeof (libusb_interface_association_descriptor)` is tempting but incorrect, because a compiler is free to add padding to a struct.

I like the idea of a new constant, see new updated commit.  How shall we name/spell them?

tormodvolden: For consistency it should be the name of the descriptor type macro + `_SIZE` so LIBUSB_DT_INTERFACE_ASSOCIATION_SIZE even if it's long. DT means descriptor type, so LIBUSB_DT_HEADER doesn't rhyme since it is not specific to a type. We can make something like a  LIBUSB_SMALLEST_DESCRIPTOR_SIZE and set it to 2, or even 3, unless there is a data-less descriptor type of size = 2 (unknown to me). Theoretically even a type-less descriptor of size = 1 would be parseable but probably not something that we should accept as valid input. Unless the same comparison is done elsewhere, so the macro can be reused, maybe it is better to just write 2 and leave a comment about why?

seanm: Done.

seanm: I've updated the commit message on the last commit, notably removing the WIP.

libusb/descriptor.c
tormodvolden: What was the conclusion here, is the change pointless?

libusb/libusb.h
```diff
@@ -335,6 +335,8 @@ enum libusb_descriptor_type {
 #define LIBUSB_DT_SS_ENDPOINT_COMPANION_SIZE	6
 #define LIBUSB_DT_BOS_SIZE			5
 #define LIBUSB_DT_DEVICE_CAPABILITY_SIZE	3
+#define LIBUSB_DT_INTERFACE_ASSOCIATION_SIZE			8
+#define LIBUSB_SMALLEST_DESCRIPTOR_SIZE			2	/* length & desc type, like usbi_descriptor_header */
```
tormodvolden, line 339: Do we want this one in the public header?

libusb/descriptor.c
seanm: My opinion is still that the change is a small improvement.

libusb/libusb.h
seanm: I have no strong feeling. I'll put them wherever you want.

libusb/descriptor.c
tormodvolden: We already use DESC_HEADER_LENGTH for exactly this in other places in the same file so no need for LIBUSB_SMALLEST_DESCRIPTOR_SIZE.

seanm: Done.

seanm:
> BTW, I will merge #1428 first, and this one will need a small rework afterwards.

I've rebased and pushed.

tormodvolden:
"Fixed potential buffer overread" still talks about WIP.

Here as well as in your other PR it would be good to see in the commit summary what part of the code you are changing. We have established a prefix convention that everybody is using. When reviewing or perusing the git log it has great value e.g. to be able to identify changes in core code vs in examples.

seanm:
> "Fixed potential buffer overread" still talks about WIP.

Yeah, would appreciate thoughts on that. I currently call `usbi_warn` and return, is that good?

> Here as well as in your other PR it would be good to see in the commit summary what part of the code you are changing. We have established a prefix convention that everybody is using. When reviewing or perusing the git log it has great value e.g. to be able to identify changes in core code vs in examples.

Any docs on that? I see nothing in `HACKING`.

tormodvolden:
Just look at the git log, i.e. `git log --oneline`

libusb/descriptor.c
```diff
@@ -84,6 +84,11 @@ static int parse_endpoint(struct libusb_context *ctx,
 	endpoint->wMaxPacketSize = ReadLittleEndian16(&buffer[4]);
 	endpoint->bInterval = buffer[6];
 	if (header->bLength >= LIBUSB_DT_ENDPOINT_AUDIO_SIZE) {
+		if (size < LIBUSB_DT_ENDPOINT_AUDIO_SIZE) {
```
tormodvolden, line 86: We already check on line 74 that size >= header->bLength, so if header->bLength >= x then size must be >= x as well.

seanm: Yeah, you're right. My previous reasoning seems totally wrong. This commit is therefore not needed.

seanm:
> Just look at the git log, i.e. `git log --oneline`

Sure, but if this is required of commit messages, shouldn't it be documented in HACKING, which already discusses commit messages:

"Commit messages should be formatted to 72 chars width and have a
free-standing summary line. See for instance "Commit Guidelines" on
https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project
or https://cbea.ms/git-commit/ about how to make well-formed commit
messages.

Put detailed information in the commit message itself, which will end
up in the git history."

libusb/libusb.h
```diff
@@ -335,6 +335,7 @@ enum libusb_descriptor_type {
 #define LIBUSB_DT_SS_ENDPOINT_COMPANION_SIZE	6
 #define LIBUSB_DT_BOS_SIZE			5
 #define LIBUSB_DT_DEVICE_CAPABILITY_SIZE	3
+#define LIBUSB_DT_INTERFACE_ASSOCIATION_SIZE			8
```
tormodvolden, line 338: Too many tabs, "8" should be aligned with the rest (with tab stop width 8).

seanm: Ha, to me they all look unaligned because the Xcode project is configured for 4 space tabs.

We really should proceed with https://github.com/libusb/libusb/pull/1443 so that human effort reviewing and fixing trivialities like this can be automated.

Anyway, fixed it.

tormodvolden: Thanks. I'd recommend using "git diff" before or "git show" after committing, it exposes such whitespace errors easily and clearly and also lets you review other changes and commit messages from another angle. QED GUIs fail miserably on this.

tormodvolden:
Another tip: To spellcheck your commit messages you can use for instance:
`git log origin/master.. | grep -v ^commit | aspell -a --dont-suggest | grep '^#'`

libusb/descriptor.c
```diff
@@ -30,54 +30,17 @@
...
-			sp += 16;
-			dp += 16;
-			break;
-		}
-	}
+static inline uint32_t ReadLittleEndian32(const uint8_t p[4])
+{
+	return (uint32_t)((uint32_t)(p[3] << 24) |
+					  (uint32_t)(p[2] << 16) |
+					  (uint32_t)(p[1] << 8) |
```
tormodvolden, line 42: Shouldn't the cast/promotion be done before the shifting?

tormodvolden: Cast before shifting is not needed because of implicit promotion: https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Shift-Operations.html (as long as we don't care about 16-bit platforms!).
I wouldn't think casting before bitwise or'ing is needed either.

seanm:
> Another tip: To spellcheck your commit messages you can use for instance: `git log origin/master.. | grep -v ^commit | aspell -a --dont-suggest | grep '^#'`

Spellings fixed.

libusb/descriptor.c
seanm: Indeed. I have restored the casting to be as it was in the macro.

tormodvolden: Why can't we just write
`return (p[3] << 24) | (p[2] << 16) | (p[1] << 8) | p[0];`
?

libusb/descriptor.c
```diff
@@ -1043,7 +1045,14 @@ int API_EXPORTED libusb_get_ssplus_usb_device_capability_descriptor(
 
 	// We can only parse the non-variable size part of the SuperSpeedPlus descriptor. The attributes
 	// have to be read "manually".
```
tormodvolden, line 1044: This comment doesn't make much sense now that we are not "parsing" any longer.

libusb/descriptor.c
seanm: That would implicitly promote to *signed* int (IIUC), and it seems safer to keep everything unsigned because [shifting a 1 into the sign bit is technically UB](https://blog.regehr.org/archives/738). Besides, it was already like that, and presumably correct as it was.

libusb/descriptor.c
seanm: Please give me the comment text you'd like, and I'll update it.

libusb/descriptor.c
tormodvolden: I am fine with the change. But since `len` is now intptr_t there is no need for the cast, it is implicit on assignment.

libusb/descriptor.c
tormodvolden: Thanks for the link, interesting. I knew shifting right can be tricky but that shifting left is "undefined behavior" was new. In any case, for GCC there is no ambiguity: "Note: according to the C standard, shifting of signed values isn’t guaranteed to work properly when the value shifted is negative, or becomes negative during the operation of shifting left. However, only pedants have a reason to be concerned about this; only computers with strange shift instructions could plausibly do this wrong. In GNU C, the operation always works as expected" (https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Shift-Caveats.html)

Actually there would be no -Wconversion warning without the casts either. But I agree to leave them in for correctness. The final cast is not needed though, it is implicit in return statements.

libusb/descriptor.c
tormodvolden: I was hoping you would confirm my interpretation of the comment :) I think by "parsing" it means using the old parse_descriptor function with a fixed conversion pattern. Now that the function is taken away, we are reading "manually" in any case. So I think the whole comment can be removed.

tormodvolden:
> So, is it that you want me to add the prefix `core: ` to all these commits?

We usually use "descriptor:" for changes in the descriptor parsing.

> Sure, but if this is required of commit messages, shouldn't it be documented in HACKING

It is not straightforward to define formal rules for this, it is a question of taste and style, and above all usefulness so it might change over time. The idea is that someone when perusing the git log, can quickly identify what is of interest, or maybe even more important, what is not of interest. So e.g. a Linux distro maintainer wanting to see what has changes since last release can skip all commits like xcode, windows, etc. Or e.g. someone else doesn't care about examples and tests. But true, we could mention something in HACKING.

For the casual contributor, I fix up this when I merge it anyway. But since you are a very productive contributor it would good if you get into this. BTW did I mention imperative commit messages? :) That is documented in the docs specified in HACKING.

libusb/descriptor.c
seanm: I've removed the cast, and also changed to `ptrdiff_t`, which is more correct.

seanm:
> We usually use "descriptor:" for changes in the descriptor parsing.

Done.

libusb/descriptor.c
seanm: >The final cast is not needed though, it is implicit in return statements.

C does lots of things implicitly, but I often find it's better to be explicit.

On a system where `int` is 64 bit (of which there are probably none that libusb supports) you might get a warning about potential truncation by *implicitly* converting from 64 to 32 bit.

But I just want this merged, so will remove it if you insist...

tormodvolden:
Thanks! I just fixed up a few typos in the commit messages.

PR #1461, opened 2024-02-05
Prepatory work for supporting -Wthread-safety

seanm:


seanm:
This is prep work for https://github.com/libusb/libusb/pull/1419.

It is ready for reviewing and merging.

seanm:
I should add: -Wthread-safety only works fully in C++. Luckily, libusb can be made to (optionally) build as C++ with these modifications.

I'm not proposing to change the language libusb builds as by default. I'm proposing it be both valid C and valid C++ so that we can use -Wthread-safety fully.

libusb/os/darwin_usb.c
```diff
@@ -570,7 +570,8 @@ static int darwin_device_from_service (struct libusb_context *ctx, io_service_t
     usbi_dbg (ctx, "set up plugin for service retry: %s", darwin_error_str (kresult));
 
     /* sleep for a little while before trying again */
-    nanosleep(&(struct timespec){.tv_sec = 0, .tv_nsec = 1000}, NULL);
+    struct timespec delay = {.tv_sec = 0, .tv_nsec = 1000};
```
tormodvolden, line 573: Can it be const?

tormodvolden: (On Linux, it /should/ be const.)

libusb/os/darwin_usb.c
```diff
@@ -1666,8 +1668,8 @@ static const struct libusb_interface_descriptor *get_interface_descriptor_by_num
   return NULL;
 }
 
-static enum libusb_error get_endpoints (struct libusb_device_handle *dev_handle, uint8_t iface) {
```
tormodvolden, line 1669: Why are some function changed to return int instead of enum?

libusb/core.c
```diff
@@ -2467,14 +2467,14 @@ int API_EXPORTED libusb_init_context(libusb_context **ctx, const struct libusb_i
 	list_init(&_ctx->open_devs);
 
 	/* apply default options to all new contexts */
-	for (enum libusb_option option = 0 ; option < LIBUSB_OPTION_MAX ; option++) {
```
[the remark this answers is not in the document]
seanm: The old way, two errors:

`core.c:2470:26 Cannot initialize a variable of type 'enum libusb_option' with an rvalue of type 'int';
`core.c:2470:74 Cannot increment expression of enum type 'enum libusb_option'`

libusb/descriptor.c
```diff
@@ -157,7 +157,7 @@ static int parse_endpoint(struct libusb_context *ctx,
 		return LIBUSB_ERROR_NO_MEM;
 
 	memcpy(extra, begin, len);
-	endpoint->extra = extra;
+	endpoint->extra = (unsigned char *)extra;
```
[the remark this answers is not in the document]
seanm: Done.

libusb/os/darwin_usb.c
seanm: Done.

libusb/os/darwin_usb.c
seanm: To pacify the C++ compiler, which is more type strict and does not allow such freewheeling conversions between enums and integers.

hjelmn: Weird. Does it have to use the C++ compiler. There will be issues if it does because C++ is not a superset of C and many C11 features do not work in C++20.

seanm: >Does it have to use the C++ compiler.

`-Wthread-safety` partly works in C. The big issue is: https://github.com/llvm/llvm-project/issues/20777

>C++ is not a superset of C 

Indeed, but it mostly is. At least enough for current libusb master.

Again, I'm not wanting to actually switch libusb's build system to C++.  I'm only saying it's handy to do so occasionally, to turn -Wthread-safety on occasionally, like as check before cutting a new release.

hjelmn: Hah, they have been trying to figure out what to do for 10 years?! I see some reasonable options but little movement. There is no technical reason why thread analysis should not work with C.

seanm: >Hah, they have been trying to figure out what to do for 10 years?!

Exactly, alas. So to get the full abilities of -Wthread-safety, we need to build libusb as C++ (or start patching clang). Hence this PR. :)

seanm:
@tormodvolden would be nice to get this merged... I really think libusb could benefit from -Wthread-safety, it has found bugs for us already...

tormodvolden:
Is this replacing #1419 ?

seanm:
> Is this replacing #1419 ?

No. This is prep for for #1419.  #1419 will add the actual annotations themselves.

libusb/os/darwin_usb.c
tormodvolden: The above 10-year old llvm bug was fixed last week, hurray!

seanm: No way!?  Nice.

Still, it'll be quite a while before such a compiler is widespread, so I still think we should merge this.

tormodvolden:
> > Is this replacing #1419 ?
> 
> No. This is prep for for #1419. #1419 will add the actual annotations themselves.

The reason I asked is that the two PRs have the same 5 commits. Only hashes are different.

seanm:
> The reason I asked is that the two PRs have the same 5 commits. Only hashes are different.

This one I split out to only support C++ and not include actual annotations yet.

tormodvolden:
So you will add the annotations to the other PR then?

seanm:
> So you will add the annotations to the other PR then?

Yes, eventually. I've stopped working on it until this PR is merged. Because if this isn't merged first, it's work for nothing.  If this is merged I can dust off that work and try again. (I worked on it over the xmas break, and haven't looked at it since.)

seanm:
So @tormodvolden @hjelmn, it was a year ago now, during last Christmas vacation, that I started this work. :)

Any thoughts on merging this?  Are there just not enough hours in the day to review, or are changes need?

seanm:
@tormodvolden with 1.0.28 out, and us having known multithreading issues in `stress_mt`, can this be merged now?

tormodvolden:
There are parts of this I would take easily because they are correctness fixes valid for C as well. However I feel some of this like all the casting doesn't make the C code easier to read. And since llvm can do -Wthread-safety on C code now, it seems less important to make it C++ compatible. Or can you still do better thread analysis on C++ than on C?

seanm:
> There are parts of this I would take easily because they are correctness fixes valid for C as well.

All the commits except [Added many explicit cast for C++ compatibility](https://github.com/libusb/libusb/pull/1461/commits/15926bf86bff8741aa33a9817bfca8ddc67247c6) ?

If so I'll split that commit to a different PR if you'll merge the rest...

seanm:
> All the commits except [Added many explicit cast for C++ compatibility](https://github.com/libusb/libusb/pull/1461/commits/15926bf86bff8741aa33a9817bfca8ddc67247c6) ?

I've removed the commit with the casts, hoping that's what you meant...

tormodvolden:
Thanks, I will take a look again (after 1.0.29).

seanm:
@sonatique could you review this too please?

tormodvolden:
You didn't comment on the need for making libusb C++ compatible, in light of llvm now being able to do -Wthread-safety on C. If I understand correctly we can add the thread annotations and do thread analysis without the changes in this PR?

libusb/os/darwin_usb.c
```diff
@@ -2171,7 +2171,7 @@ static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, b
 static int darwin_reset_device (struct libusb_device_handle *dev_handle) {
```
tormodvolden: Why does this one return int, although it uses e.g. darwin_to_libusb?
E.g. darwin_detach_kernel_driver should also be made return enum, right?

tormodvolden:
The three first commits are IMO no-brainers, I just have some questions (and need to look closer) on the darwin part of the fourth.

seanm:
> You didn't comment on the need for making libusb C++ compatible

I did, I said I dropped that commit from this PR.

> in light of llvm now being able to do -Wthread-safety on C. If I understand correctly we can add the thread annotations and do thread analysis without the changes in this PR?

A very bleeding edge llvm, which Xcode doesn't yet include and thus I have not tried.

libusb/os/darwin_usb.c
seanm: Hmmm, indeed. Well, I haven't looked at it in months, so I don't recall...

seanm: Ah, I know why. It's because `darwin_reset_device`, despite being `static` is sorta a public API, it's function signature must match what's in `struct usbi_os_backend`, and everything there uses raw `int`.

It's maybe hopeless without changing public declarations too.

I guess I'll drop this commit also, so at least *something* can be merged.

seanm: OK it's dropped.

sonatique (approved):

sonatique:
@seanm : I reviewed the changes and have no objections to merge them, I see them as improvements, so I would recommend immediate merge.

However, I am a bit lost regarding the goal of the PR. Reading the discussion I understand that originally the goal was to make the whole libusb code compatible with C++ compilation as a preliminary step toward supporting -Wthread-safety (by the way having the compilation C++ compatible could be a goal per se).

Since then you dropped some commits and therefore the changes that are left are not sufficient to make libusb fully  C++ compatible.

So is this some sort of preliminary steps for the next preliminary steps ? ;-)

Are you still planning to ensure C++ compatibility as a  pre-requisite for enabling -Wthread-safety?

libusb/os/darwin_usb.c
tormodvolden: OK, thanks! Were the changes in core.c in that commit depending on the darwin changes?

seanm:
> So it this some sort of preliminary steps for the next preliminary steps ? ;-)

Correct.

> Are you still planning to ensure C++ compatibility as a pre-requisite for enabling -Wthread-safety?

Yes, in another PR.

PR #1475, opened 2024-03-20
windows: Restore behaviour of skipping malformed device GUIDs

hartcw:
Previously when getting the device list, a malformed device GUID would be explicitly ignored and skipped, allowing the operation to complete. A recent change to winusb_get_device_list() accidently changed this behaviour, so this scenario instead caused an early exit with error code LIBUSB_ERROR_NO_MEM.

The bug was introduced with commit:
https://github.com/libusb/libusb/commit/fdab67b14ac32891158dc210636b622c42e24e22

@tormodvolden

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1572,7 +1572,6 @@ static int get_guid(struct libusb_context *ctx, char *dev_id, HDEVINFO *dev_info
 		usbi_warn(ctx, "device '%s' has malformed DeviceInterfaceGUID string '%s', skipping", dev_id, guid);
 		free(*if_guid);
 		*if_guid = NULL;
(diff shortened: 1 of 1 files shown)


tormodvolden:
Thanks, this looks correct. Out of interest, what are these devices (or drivers) and what do these GUIDs look like?

tormodvolden:
@hartcw, what are these devices (or drivers) and what do these GUIDs look like?

seanm:
Could you please amend the commit message to contain the SHA of when this was accidentally broken (I know you've put it in github comments, but it would be useful in the commit message).

tormodvolden:
I am happy to amend the commit message when I merge it, but I would still like to know what these "malformed GUIDs" are about.

PR #1479, opened 2024-04-09
Fixed most all of the reasonable clang-tidy warnings

seanm:


libusb/descriptor.c
```diff
@@ -1168,9 +1168,7 @@ int API_EXPORTED libusb_get_string_descriptor_ascii(libusb_device_handle *dev_ha
 	r = libusb_get_string_descriptor(dev_handle, 0, 0, str.buf, 4);
 	if (r < 0)
 		return r;
-	else if (r != 4 || str.desc.bLength < 4)
-		return LIBUSB_ERROR_IO;
-	else if (str.desc.bDescriptorType != LIBUSB_DT_STRING)
+	else if (r != 4 || str.desc.bLength < 4 || str.desc.bDescriptorType != LIBUSB_DT_STRING)
```
Youw, line 1171: what was the warning here?
I personally find the "before" more descriptive/readable

libusb/descriptor.c
```diff
@@ -1179,9 +1177,7 @@ int API_EXPORTED libusb_get_string_descriptor_ascii(libusb_device_handle *dev_ha
 	r = libusb_get_string_descriptor(dev_handle, desc_index, langid, str.buf, sizeof(str.buf));
 	if (r < 0)
 		return r;
-	else if (r < DESC_HEADER_LENGTH || str.desc.bLength > r)
-		return LIBUSB_ERROR_IO;
-	else if (str.desc.bDescriptorType != LIBUSB_DT_STRING)
+	else if (r < DESC_HEADER_LENGTH || str.desc.bLength > r || str.desc.bDescriptorType != LIBUSB_DT_STRING)
```
Youw, line 1180: ditto

libusb/descriptor.c
seanm: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/branch-clone.html

tormodvolden:
I appreciate the systematic approach, and I understand the motivation to silence clang-tidy, but I would prefer the commit summary to say what change is done*, or how it makes the code better, independent of clang-tidy. Silencing clang-tidy is of course a nice side-effect :) And it is important to note if some commits only affects examples or some platform, it makes it much easier for someone to peruse the git log and skip things that are not of their interest. If small trivial fixes to an example file can be grouped together that is fine too.

*) Generally it should express the intention do something, not what one thinks has been done. It is not always the same :)

E.g.:
<- Fixed all clang-tidy bugprone-unsafe-functions warnings
<- Replaced rewind, which gives no error result, with fseek, which does.

-> examples/ezusb: Replace rewind, which gives no error result, with fseek, which does  
-> This fixes the last clang-tidy bugprone-unsafe-functions warning.

seanm:
> ...I would prefer the commit summary to say what change is done...

Some of them do, but yeah ok I'll improve all the commit messages with more detail.

tormodvolden:
All the _summaries_ (first line) say "Fixed all clang-tidy...".

seanm:
> All the _summaries_ (first line) say "Fixed all clang-tidy...".

Oh missed that, it's the summaries you don't like... I guess I'm so familiar with their names that it seems clear to me.  I'll reword them...

tormodvolden:
> > All the _summaries_ (first line) say "Fixed all clang-tidy...".
> 
> Oh missed that, it's the summaries you don't like... I guess I'm so familiar with their names that it seems clear to me. I'll reword them...

No, you don't need to reword the names of the clang-tidy keywords if that is what you mean. But they should be relegated to the message, not the summary. Please see my example above.

seanm:
OK done, hopefully you like it better now.

seanm:
I added those prefix things you like

seanm:
@tormodvolden hopefully we can get this one merged next?

seanm:
@tormodvolden friendly ping... any other final tweaks needed?

tormodvolden:
Thanks a lot for the clean-up!

tormodvolden:
@anotheruserofgithub Thanks for the review! I think you are correct.

Issue #1387, opened 2023-12-20
Haiku OS: build fails due to stdatomic.h issue


tormodvolden:
For Haiku OS, the build issue is still there, similar to 1.0.26 release. gcc version is 11.2.0.
https://www.haiku-os.org/get-haiku/r1beta4/
* https://github.com/libusb/libusb/issues/1099#issuecomment-1080067439

<details>
<summary> build log -- failed under Haiku OS R1 Beta4 </summary>

```
~/Desktop/build/libusb-1.0.27-rc1> make -i
make  all-recursive
make[1]: Entering directory '/boot/home/Desktop/build/libusb-1.0.27-rc1'
Making all in libusb
make[2]: Entering directory '/boot/home/Desktop/build/libusb-1.0.27-rc1/libusb'
  CXX      os/haiku_usb_backend.lo
cc1plus: warning: '-Werror=' argument '-Werror=implicit-function-declaration' is not valid for C++
cc1plus: warning: '-Werror=' argument '-Werror=implicit-int' is not valid for C++
cc1plus: warning: '-Werror=' argument '-Werror=missing-prototypes' is not valid for C++
cc1plus: warning: '-Werror=' argument '-Werror=strict-prototypes' is not valid for C++
In file included from ./libusbi.h:106,
                 from os/haiku_usb.h:25,
                 from os/haiku_usb_backend.cpp:27:
/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/11.2.0/include/stdatomic.h:40:9: error: '_Atomic' does not name a type
   40 | typedef _Atomic _Bool atomic_bool;
      |         ^~~~~~~
/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/11.2.0/include/stdatomic.h:41:9: error: '_Atomic' does not name a type
   41 | typedef _Atomic char atomic_char;
      |         ^~~~~~~
/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/11.2.0/include/stdatomic.h:42:9: error: '_Atomic' does not name a type
   42 | typedef _Atomic signed char atomic_schar;
      |         ^~~~~~~
/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/11.2.0/include/stdatomic.h:43:9: error: '_Atomic' does not name a type
   43 | typedef _Atomic unsigned char atomic_uchar;
      |         ^~~~~~~
[...]

PR #1482, opened 2024-04-18
Fix build for Haiku.

pulkomandy:
Fix the issue discussed in #1387

The Haiku build uses C++ files, and gcc's stdatomic.h does not define the needed functions when building as a C++11 header (it works fine in C, or when using C++23, or when using clang as a compiler).

Fix the build by using the gcc builtins (as it used to be done) specifically for Haiku.

It looks like I took too long to write that patch and I missed the 1.0.27 release. Sorry about that.

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -103,12 +103,25 @@ typedef volatile LONG usbi_atomic_t;
 #define usbi_atomic_inc(a)	InterlockedIncrement((a))
 #define usbi_atomic_dec(a)	InterlockedDecrement((a))
 #else
+#ifdef __HAIKU__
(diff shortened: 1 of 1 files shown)


seanm (approved):

Youw (approved):
I don't have haiku to verify, but the change looks OK.

tormodvolden:
So if this workaround is only needed for GCC and not clang, should we change it to `#if defined(HAIKU) && !defined(__clang__)` ?

pulkomandy:
I'm not set up to actually test building with clang on Haiku, but I think what you propose should work.

tormodvolden:
I added __GNUC__ for completeness, and amended the comment.

seanm (approved):

tormodvolden:
Thanks for the patch and reviews, guys!

PR #1497, opened 2024-05-12
Avoid --Wconversion warnings

tormodvolden:
This allows building without warnings even with -Wconversion added to CFLAGS. At the moment, this covers the common code  and not the platform-specific files.

Discussion is (as always) welcome. I have mixed feelings about some of this because it makes the code more difficult to read for humans (or those wired like me, I guess).

Especially the enum stuff seems stupid, the constant are specified as unsigned int, but enums are always signed int (before C23) so they still need to be cast back. Using const unsigned int instead of enums would be an option, since these are private APIs.

@fabiensanglard you are particularly welcome to look at the libusb_get_ssplus_usb_device_capability_descriptor() changes.


tormodvolden:
The last commit shows the alternative of using macros instead of an enum. The same could be done with the itransfer->state_flags like USBI_TRANSFER_IN_FLIGHT.

tormodvolden:
I started on linux_usbfs, but there is a lot more to do.

libusb/os/linux_usbfs.c
```diff
@@ -182,7 +182,7 @@ static int dev_has_config0(struct libusb_device *dev)
 	return 0;
 }
 
-static int get_usbfs_fd(struct libusb_device *dev, mode_t mode, int silent)
+static int get_usbfs_fd(struct libusb_device *dev, int mode, int silent)
```
Youw, line 185: this one I'd actually call a step back in readability in general

Youw: maybe a `typedef` for readability?

tormodvolden: Well, whereas for the other commits it is a lot of sprinkling on casts, this one actually corrects a wrong type. The type should be int and not mode_t. Did you see the commit message? Would you prefer a better variable name?

Youw: > Did you see the commit message?

I must confess - I didn't (now I did).

> Would you prefer a better variable name?

Maybe a `open_flags`? That would be as straight-forward as possible, since it is getting passed into `open` as `flags` argument.

tormodvolden: To follow the terminology of the open() manual page I think `access_mode` is a good choice. It is only used to carry the access mode here, and no other flags.

Youw: sure, sounds good

libusb/libusbi.h
```diff
@@ -471,25 +471,24 @@ static inline struct libusb_context *usbi_get_context(struct libusb_context *ctx
...
+#define USBI_EVENT_HOTPLUG_MSG_PENDING     (1U << 3)
 
-	/* One or more completed transfers are pending */
-	USBI_EVENT_TRANSFER_COMPLETED = 1U << 4,
+/* One or more completed transfers are pending */
+#define USBI_EVENT_TRANSFER_COMPLETED      (1U << 4)
 
-	/* A device is in the process of being closed */
-	USBI_EVENT_DEVICE_CLOSE = 1U << 5,
-};
```
seanm, line 492: This seems a backwards change. enums should be encouraged over #defines, not the other way around.

Better would be to do this: https://github.com/libusb/libusb/pull/1417/commits/6b52b8e1508987e906cfddf5573b703cebc9668c

fabiensanglard:
The changes to `libusb_get_ssplus_usb_device_capability_descriptor` look fine to me.

libusb/libusbi.h
tormodvolden: Hmm, I don't think we should bump compiler requirements for something so cosmetic. I understand that the suggestion doesn't strictly require C23 but it wouldn't be tested without.

There are certainly reasons for encouraging enums (especially if they had types) over defines in many cases but I am not sure it applies to us here.

As I said initially, I have mixed feelings about all this because we try everything to please compilers but likely make the code less readable. Maybe the best way is just to keep -Wconversion out :)

tormodvolden:
I merged the two actual fixes for linux. For the rest I am in no hurry. It was more an experiment to see what would be needed to become -Wconversion clean, it become quickly an obsession. -Wconversion demonstratively has a value, as it found the two Linux errors, but is all of it worth all the clutter?

libusb/libusbi.h
seanm: > Hmm, I don't think we should bump compiler requirements for something so cosmetic.

To be clear, I'm not advocating bumping the minimum compiler requirements. As you see in 6b52b8e1508987e906cfddf5573b703cebc9668c, the feature can be wrapped in a macro like we do already for `LIBUSB_DEPRECATED_FOR`. 

>I understand that the suggestion doesn't strictly require C23 but it wouldn't be tested without.

Well sure, but adding a C23 variant to CI shouldn't be too hard.

> There are certainly reasons for encouraging enums (especially if they had types) over defines in many cases but I am not sure it applies to us here.

I have found it very helpful as documenation, especially where those enumerated values are converted in/out of raw uintX_t struct fields.

> As I said initially, I have mixed feelings about all this because we try everything to please compilers but likely make the code less readable. 

I find this change *more* readable because now I, the reader, know the size and range of the thing:

```diff
-enum libusb_endpoint_direction {
+enum libusb_endpoint_direction LIBUSB_ENUM_SIZE(uint8_t) {
```

>Maybe the best way is just to keep -Wconversion out :)

Indeed I have found -Wconversion something that can find problems, but is often too noisy.  I haven't reviewed all this PR yet, but will...

PR #1499, opened 2024-05-22
Fix typo mantisa -> mantissa

fabiensanglard:
The correct spelling is "mantisSa"

### DIFF
--- a/examples/xusb.c
+++ b/examples/xusb.c
@@ -770,7 +770,7 @@ static void print_sublink_speed_attribute(struct libusb_ssplus_sublink_attribute
 	static const char exponent[] = " KMG";
 	printf("                  id=%u speed=%u%cbs %s %s SuperSpeed%s",
 		ss_attr->ssid,
-		ss_attr->mantisa,
+		ss_attr->mantissa,
 		(exponent[ss_attr->exponent]),
 		(ss_attr->type == LIBUSB_SSPLUS_ATTR_TYPE_ASYM)? "Asym" : "Sym",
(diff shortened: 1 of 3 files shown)


fabiensanglard:
This is a change in the API but these structure were merged recently and have not been published in a stable release. I think it is fine to fix the typo.

fabiensanglard:
@mcuee and @tormodvolden please take a look

mcuee:
> This is a change in the API but these structure were merged recently and have not been published in a stable release. I think it is fine to fix the typo.

I agree with you. But I will let @tormodvolden to decide.

Youw (approved):
LGTM.
This hasn't been released yet and just recently added. Better to fix now, while it doesn't impact too many users (if any).

seanm (approved):

tormodvolden:
Yes, there is no problem fixing up the API before it has been published. Thanks for taking a double look!

We did a less than stellar job at reviewing this. The spelling was not even consistent with the code comments. Curiously, codespell wouldn't have caught "mantisa" even in a comment. It does find these though, introduced in the same commit:
```
$ codespell libusb
libusb/descriptor.c:1008: ony ==> only, on, one
libusb/descriptor.c:1028: re-use ==> reuse
```
Further, maybe we should spell superspeed+/SuperSpeed+/SuperSpeedPlus consistently, also for better searchability?

Apart from certain platform files in os/* we use classic style /* */ for comments and not the C++ style // up to now almost consistently, so it we shouldn't diverge here.

Can you please fix up these as well? In the meantime I'll merge the present API change, it is worth having it separately from cosmetic changes in any case.

fabiensanglard:
@tormodvolden: All done in PR #1502 1502

PR #1500, opened 2024-05-24
Emscripten: Fix Uncaught TypeError on Firefox

srcejon:
If libusb_get_device_list() is called using the Emscripten port of libusb when running in the Firefox browser, an application will be terminated with:

    Uncaught TypeError: objHandle is undefined

This is because Firefox doesn't support the navigator.usb object, used by getDeviceList() in emscripten_webusb.cpp.

This patch checks if the object is undefined before using it and will return an error instead.

### DIFF
--- a/libusb/os/emscripten_webusb.cpp
+++ b/libusb/os/emscripten_webusb.cpp
@@ -586,12 +586,17 @@ unsigned long getDeviceSessionId(val& web_usb_device) {
 }
 
 val getDeviceList(libusb_context* ctx, discovered_devs** devs) {
+	// Check if browser supports USB
+	val navigator_usb = val::global("navigator")["usb"];
+	if (navigator_usb == val::undefined()) {
+		co_return (int) LIBUSB_ERROR_NOT_SUPPORTED;
+	}
 	// C++ equivalent of `await navigator.usb.getDevices()`. Note: at this point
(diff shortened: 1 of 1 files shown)


tormodvolden:
@RReverser can you please review this?

libusb/os/emscripten_webusb.cpp
```diff
@@ -586,6 +586,11 @@ unsigned long getDeviceSessionId(val& web_usb_device) {
 }
 
 val getDeviceList(libusb_context* ctx, discovered_devs** devs) {
+	// Check if browser supports USB
+	val nav_usb = val::global("navigator")["usb"];
```
RReverser, line 590: Nit: since you already got this object anyway, can you reuse it where we need `navigator.usb` for the actual call below?

Also I'm a fan of longer names, they don't cost us anything, so I'd just call it `navigator_usb`.

RReverser (changes requested):
Good idea, just minor nits.

RReverser (approved):

tormodvolden:
Yes, looks good, thanks. Here in the web interface you can see your single commit  under the commits tab, and the "Compare" button next to your forced push shows there were no changes from the old head.

tormodvolden:
Thanks for your contribution!

PR #1507, opened 2024-06-03
xusb: Define proper exit status 

RinHizakura:
Can we consider to propagate the return value from test_device() to main()? In this way, we can easily check the return code directly like `echo $?` in Linux to know whether the test is failed.

### DIFF
--- a/examples/xusb.c
+++ b/examples/xusb.c
@@ -1077,9 +1077,25 @@ static int test_device(uint16_t vid, uint16_t pid)
 	return 0;
 }
 
+static void display_help(const char *progname)
+{
+	printf("usage: %s [-h] [-d] [-i] [-k] [-b file] [-l lang] [-j] [-x] [-s] [-p] [-w] [vid:pid]\n", progname);
+	printf("   -h      : display usage\n");
+	printf("   -d      : enable debug output\n");
+	printf("   -i      : print topology and speed info\n");
+	printf("   -j      : test composite FTDI based JTAG device\n");
+	printf("   -k      : test Mass Storage device\n");
+	printf("   -b file : dump Mass Storage data to file 'file'\n");
+	printf("   -p      : test Sony PS3 SixAxis controller\n");
+	printf("   -s      : test Microsoft Sidewinder Precision Pro (HID)\n");
+	printf("   -x      : test Microsoft XBox Controller Type S\n");
+	printf("   -l lang : language to report errors in (ISO 639-1)\n");
+	printf("   -w      : force the use of device requests when querying WCID descriptors\n");
+	printf("If only the vid:pid is provided, xusb attempts to run the most appropriate test\n");
+}
+
 int main(int argc, char** argv)
 {
-	bool show_help = false;
 	bool debug_mode = false;
 	const struct libusb_version* version;
 	int j, r;
@@ -1096,7 +1112,12 @@ int main(int argc, char** argv)
 	if (((uint8_t*)&endian_test)[0] == 0xBE) {
 		printf("Despite their natural superiority for end users, big endian\n"
 			"CPUs are not supported with this program, sorry.\n");
-		return 0;
+		return EXIT_FAILURE;
+	}
+
+	if ((argc == 1) || (argc > 7)) {
+		display_help(argv[0]);
+		return EXIT_FAILURE;
 	}
 
 	if (argc >= 2) {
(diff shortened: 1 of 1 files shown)


Youw:
I'd also change the `test_device` to return `0`/`1` (instead of `-1`), or even `EXIT_SUCCESS`/`EXIT_FAILURE` named constants.
Negative exit codes usually means exit by interruption.

tormodvolden:
Yes, a proper exit status would be a good improvement. In general I'd suggest letting internal functions return and propagate libusb error codes, but then in main(), also on line 1216, something like if (r<0) return EXIT_FAILURE.

Youw:
That sounds even better

RinHizakura:
I make a change according to the comment, but please let me know if this is what you want. Thanks!

examples/xusb.c
```diff
@@ -1096,7 +1096,7 @@ int main(int argc, char** argv)
 	if (((uint8_t*)&endian_test)[0] == 0xBE) {
 		printf("Despite their natural superiority for end users, big endian\n"
 			"CPUs are not supported with this program, sorry.\n");
-		return 0;
+		return EXIT_SUCCESS;
```
tormodvolden, line 1099: Even though we returned "success" here it should return a failure in this case.

examples/xusb.c
```diff
@@ -1225,7 +1225,8 @@ int main(int argc, char** argv)
 			printf("Invalid or unsupported locale '%s': %s\n", error_lang, libusb_strerror((enum libusb_error)r));
 	}
 
-	test_device(VID, PID);
+	if (test_device(VID, PID) < 0)
+		return EXIT_FAILURE;
 
 	libusb_exit(NULL);
```
tormodvolden, line 1231: We should call libusb_exit() even if test_device fails.

examples/xusb.c
```diff
@@ -1199,7 +1199,7 @@ int main(int argc, char** argv)
 		printf("   -l lang : language to report errors in (ISO 639-1)\n");
 		printf("   -w      : force the use of device requests when querying WCID descriptors\n");
 		printf("If only the vid:pid is provided, xusb attempts to run the most appropriate test\n");
-		return 0;
+		return EXIT_SUCCESS;
```
RinHizakura, line 1202: How about this one? I feel `show_help` should return `EXIT_SUCCESS` and `(argc == 1) || (argc > 7)` has to be `EXIT_FAILURE`. Is this expected by most users?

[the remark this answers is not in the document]
tormodvolden: Yes, ideally success on -h, otherwise failure. Maybe it can make sense to factor out a display_help() function.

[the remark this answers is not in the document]
RinHizakura: Change accordingly. Please check!

examples/xusb.c
```diff
@@ -1077,9 +1077,27 @@ static int test_device(uint16_t vid, uint16_t pid)
...
+	printf("   -k      : test Mass Storage device\n");
+	printf("   -b file : dump Mass Storage data to file 'file'\n");
+	printf("   -p      : test Sony PS3 SixAxis controller\n");
+	printf("   -s      : test Microsoft Sidewinder Precision Pro (HID)\n");
+	printf("   -x      : test Microsoft XBox Controller Type S\n");
+	printf("   -l lang : language to report errors in (ISO 639-1)\n");
+	printf("   -w      : force the use of device requests when querying WCID descriptors\n");
+	printf("If only the vid:pid is provided, xusb attempts to run the most appropriate test\n");
+
+	exit(exit_status);
```
tormodvolden, line 1096: I would prefer you let this function simply return (otherwise it will be a display_help_and_exit function. Then in main() just return (with success or failure) after having called display_help().

examples/xusb.c
```diff
@@ -1077,9 +1077,27 @@ static int test_device(uint16_t vid, uint16_t pid)
 	return 0;
 }
 
+static void display_help(char *exe, int exit_status)
```
tormodvolden, line 1080: I would have called it progname instead of "exe".

tormodvolden: And it can be `const char *` I believe.

examples/xusb.c
tormodvolden: Your original way would make very much sense in case exit_status was determined inside display_help(). But since the calling main() knows the exit_status in each case, it is tidier to let main() also return directly.

tormodvolden (approved):
Thanks, your changes look perfect to me. The exit status handling seems very clean now.

Youw (approved):

tormodvolden:
Thanks for your contribution!

PR #1509, opened 2024-06-11
windows: Fix redefinition errors in some build environments

oleavr:
By including winsock2.h before windows.h.


Youw:
Add a comment explaining that it has to stay in that order an why

oleavr:
> Add a comment explaining that it has to stay in that order an why

Good point, thanks! Just updated the commit.

mcuee:
I have never been a fan of Cygwin but it is good that people are providing PRs for Cygwin.

Youw (approved):
LGTM

seanm (approved):

tormodvolden:
This change only affects non-Cygwin Window builds, doesn't it? To be more precise and avoid confusion, can you please rephrase "some build environments"? Are these redefinition errors visible on any of our CI runs?

mcuee:
> This change only affects non-Cygwin Window builds, doesn't it? To be more precise and avoid confusion, can you please rephrase "some build environments"? Are these redefinition errors visible on any of our CI runs?

Good point.

I wrongly assumed the fix is for Cygwin, but actually it is not for Cygwin but for non-Cygwin.

I never see such issues with MSYS2 mingw build environment and MSVC. So I tend to think this PR may not be correct.

tormodvolden:
If I understand https://stackoverflow.com/questions/9153911/is-there-a-difference-between-winsock-h-and-winsock2-h correctly, windows.h will include winsock.h and then winsock2.h cannot be included afterwards because it will try to redefine stuff. But if winsock2.h is included first, any following winsock.h will not be parsed and everything is fine.

So if a program includes winsock2.h before libusb.h there shouldn't be any problem, right? Is that why we haven't heard complaints before?

From looking up commit 0ded9c12 , commit eea7ebe6c , commit 2442719 it seems winsock was introduced only for WinCE and for WDK. I don't think we support WinCE any longer, and WDK I have no idea about. Do we need to include winsock*.h at all? Is this about WinCE?

mcuee:
> From looking up commit [0ded9c1](https://github.com/libusb/libusb/commit/0ded9c126e9b48b5ed217e60d392cfb882cbc351) , commit [eea7ebe](https://github.com/libusb/libusb/commit/eea7ebe6ca5645bc3a80b8f3c9a1627cf6884ee9) , commit [2442719](https://github.com/libusb/libusb/commit/244271931a782028fd9bc1cdcdb12200f65cf8d9) it seems winsock was introduced only for WinCE and for WDK. I don't think we support WinCE any longer, and WDK I have no idea about. Do we need to include winsock*.h at all? Is this about WinCE?

We no longer support WinCE and old WDK build. So maybe we can totally remove winsock*.h.

seanm:
So is this no longer needed since https://github.com/libusb/libusb/pull/1657 was merged?

mcuee:
I believe so.

PR #1512, opened 2024-06-17
Reintroduce RAW_IO support in WinUSB backend

HannesFranke-smartoptics:
This is a continuation of PR #1297 which also supports composite devices, using the same mechanism as other functions which support composite devices, e.g. `libusb_submit_transfer()`.

It provides RAW_IO support by introducing three new API functions:
```c
int libusb_endpoint_supports_raw_io(libusb_device_handle* dev_handle, uint8_t endpoint);
int libusb_endpoint_set_raw_io(libusb_device_handle* dev_handle, uint8_t endpoint, int enable);
int libusb_get_max_raw_io_transfer_size(libusb_device_handle *dev_handle, uint8_t endpoint);
```

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -382,6 +382,8 @@ if (cfg != desired)
   * - libusb_detach_kernel_driver()
   * - libusb_dev_mem_alloc()
   * - libusb_dev_mem_free()
+  * - libusb_endpoint_set_raw_io()
+  * - libusb_endpoint_supports_raw_io()
   * - libusb_error_name()
   * - libusb_event_handler_active()
   * - libusb_event_handling_ok()
@@ -419,6 +421,7 @@ if (cfg != desired)
   * - libusb_get_iso_packet_buffer_simple()
   * - libusb_get_max_alt_packet_size()
   * - libusb_get_max_iso_packet_size()
+  * - libusb_get_max_raw_io_transfer_size()
   * - libusb_get_max_packet_size()
   * - libusb_get_next_timeout()
   * - libusb_get_parent()
@@ -2202,6 +2205,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
 
 	dev_handle->auto_detach_kernel_driver = enable;
 	return LIBUSB_SUCCESS;
+}
+
+/** \ingroup libusb_dev
+ * Check if the endpoint supports RAW_IO.
+ * \param dev_handle a device handle
+ * \param endpoint the endpoint to check
+ *
+ * \returns 1 if the endpoint supports RAW_IO
+ * \returns 0 if the endpoint does not support RAW_IO
+ * \returns a LIBUSB_ERROR code on failure
+ *
+ * Only endpoints using the WinUSB driver support RAW_IO,
+ * for all other backends/drivers this function will return 0.
+ *
+ * Fails if the interface the endpoint belongs to isn't claimed,
+ * \see libusb_claim_interface().
+ *
+ * \see libusb_endpoint_set_raw_io()
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
+ */
+int API_EXPORTED libusb_endpoint_supports_raw_io(libusb_device_handle* dev_handle,
+	uint8_t endpoint)
+{
+	if (usbi_backend.endpoint_supports_raw_io == NULL)
+	{
+		return 0;
+	}
+
+	// If the `endpoint_supports_raw_io` function is present, these two should be too:
+	assert(usbi_backend.endpoint_set_raw_io != NULL);
+	assert(usbi_backend.get_max_raw_io_transfer_size != NULL);
+
+	return usbi_backend.endpoint_supports_raw_io(dev_handle, endpoint);
+}
+
+/** \ingroup libusb_dev
+ * Enable/disable RAW_IO for an endpoint on an open device.
+ *
+ * Only endpoints using the WinUSB driver support RAW_IO,
+ * for all other backends/drivers this function will return an error code.
+ *
+ * Using RAW_IO can greatly improve USB throughput by directly passing
+ * transfer requests to the underlying USB driver instead of queuing them
+ * in WinUSB. This can be particularly useful for high-throughput devices
+ * like cameras or oscilloscopes.
+ *
+ * Transfers submitted to the endpoint while RAW_IO is enabled will fail unless
+ * they adhere to the following rules :
+ * - The buffer length must be a multiple of the maximum endpoint packet size
+ *   \see libusb_get_max_packet_size.
+ * - The buffer length must be less than or equal to the value returned by
+ *   \ref libusb_get_max_raw_io_transfer_size.
+ *
+ * This option should not be changed when any transfer is in progress on
+ * the specified endpoint.
+ *
+ * Fails if the interface the endpoint belongs to isn't claimed,
+ * \see libusb_claim_interface().
+ *
+ * \param dev_handle a device handle
+ * \param endpoint the endpoint to set RAW_IO for
+ * \param enable 1 to enable RAW_IO, 0 to disable it
+ *
+ * \returns \ref LIBUSB_SUCCESS on success
+ * \returns \ref LIBUSB_ERROR_NOT_SUPPORTED if the backend does not support RAW_IO.
+ * \returns another LIBUSB_ERROR code on other failure
+ *
+ * \see libusb_endpoint_supports_raw_io()
+ * \seealso https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
+ */
+int API_EXPORTED libusb_endpoint_set_raw_io(libusb_device_handle* dev_handle,
+	uint8_t endpoint, int enable)
+{
+	if (!usbi_backend.endpoint_set_raw_io)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	return usbi_backend.endpoint_set_raw_io(dev_handle, endpoint, enable);
+}
+
+/** \ingroup libusb_dev
+ * Retrieve the maximum transfer size in bytes supported for WinUSB RAW_IO
+ * for an inbound bulk or interrupt endpoint on an open device.
+ *
+ * Returns a maximum transfer size in bytes, or a negative error code.
+ * If the backend does not support WinUSB RAW_IO, returns
+ * \ref LIBUSB_ERROR_NOT_SUPPORTED.
+ *
+ * Fails if the interface the endpoint belongs to isn't claimed,
+ * \see libusb_claim_interface().
+ *
+ * \see libusb_endpoint_set_raw_io()
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
+ */
+int API_EXPORTED libusb_get_max_raw_io_transfer_size(
+	libusb_device_handle *dev_handle, uint8_t endpoint)
+{
+	if (!usbi_backend.get_max_raw_io_transfer_size)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	return usbi_backend.get_max_raw_io_transfer_size(
+		dev_handle, endpoint);
 }
 
 /** \ingroup libusb_lib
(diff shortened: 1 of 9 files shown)


libusb/core.c
```diff
@@ -2204,6 +2207,108 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+ *
+ * \see libusb_endpoint_set_raw_io()
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
+ */
+int API_EXPORTED libusb_endpoint_supports_raw_io(libusb_device_handle* dev_handle,
+	uint8_t endpoint)
+{
+	if (usbi_backend.endpoint_supports_raw_io == NULL ||
+		usbi_backend.endpoint_set_raw_io == NULL)
```
tormodvolden, line 2233: Why the check for usbi_backend.endpoint_set_raw_io?

martinling: Presumably because calling `libusb_endpoint_supports_raw_io` is supposed to guarantee that it's OK to call `libusb_endpoint_set_raw_io`. If the function pointer for the latter isn't set, then the RAW_IO support isn't complete.

I think the cross-check makes sense, and it should probably also check that `usbi_backend.get_max_raw_io_transfer_size` is set too.

HannesFranke-smartoptics: Agreed, shows the intent much better.

JVital2013:
Hello!

First of: thanks for this PR. It's helped a lot in my testing, and I wanted to share an interesting tidbit I discovered.

I work on software that interfaces with many software defined radios (SDRs) that receive signals, converts it to I/Q data, and sends it to the computer via libusb. Most entry level SDRs work the same way: the SDR's library fills/submits some number of bulk transfers to the device (usually 4-16, depending on the device). Once I/Q data is returned, the lib re-submits the transfer for more data and the cycle continues. As these libraries are written right now, they do not use `RAW_IO` in any way.

This works fine as-is on many Windows machine, but with one major outlier: the new Snapdragon X laptops. I tested two models (the HP Omnibook X and Dell Inspiron 14 Plus 7441). All the SDRs I tested would drop I/Q samples on both machines, apparently because the bulk transfer buffers were not being handled quickly enough.

After locally implementing the new functions presented in this PR into libairspy, librtlsdr, and libhackrf, my issues completely went away on Windows ARM64 machines. I'm not 100% sure why the transfer buffers were going too slow in the first place, since these libraries typically work on much, much slower machines. I'm guessing it's some inefficiency somewhere in WinUSB on these devices. Regardless, `RAW_IO` may be the right way to go for SDRs that use bulk transfer anyway.

All that to say: this PR seems to work for me! Hopefully we can see it merged in the near future.

martinling:
@JVital2013, thanks for all that testing of this!

I'm one of the HackRF developers, and this use case is exactly why I was working on this API in the earlier PRs. We'll add RAW_IO support to upstream libhackrf as soon as it's available in libusb - if you already have those changes ready, then a PR from you would be very welcome.

> I'm not 100% sure why the transfer buffers were going too slow in the first place, since these libraries typically work on much, much slower machines.

The problem isn't related to the speed of the machine but rather to latency.

Without RAW_IO, the host will only submit one operation to the host controller at a time, so there is some latency between an inbound transfer completing on the bus, and the OS submitting the next operation so that the host controller can start polling the device again.

On the HackRF we only have 32KB of suitable DMA-capable RAM that we can buffer samples in between receiving them from the ADC and sending them over USB. At 40MB/s that's less than a millisecond of runway, so it's essential that the host keeps polling us continuously or samples will have to be dropped. Other capture devices have similar constraints.

With RAW_IO enabled, the host is able to queue up the operations on the host controller, so it's always polling us and we can keep data flowing out of the buffer.

The reason that the problem is worse on those Snapdragon X machines is probably due to SMI interrupts or similar things tying up the system, increasing the worst-case time between a transfer completing and a new one being started.

And yes, in theory this is exactly the use case for isochronous transfers, but those are only allowed to use 80% of the bus bandwidth, so a continous bulk transfer is the only way to actually push the data rate we need.

martinling (approved):
I believe this is ready for merge.

- The choice of API has already been subject to a great deal of discussion and prototyping, during which several other proposals were made and rejected.

- The implementation issues with composite devices have been fixed.

- The code has now been thoroughly tested by @JVital2013, who has integrated it into multiple downstream projects and tested them on multiple USB devices and Windows systems.

- This feature is essential for performance and reliability with many types of devices, and the lack of it has caused problems for many years.

Let's finish this.

whitequark:
The same problem that is affecting HackRF is also affecting [Glasgow Interface Explorer](https://glasgow-embedded.org/), for the same reasons. I'm eagerly awaiting the moment when I can enable RAW_IO.

JVital2013 (approved):

charlie-rfcreations:
Is there something stopping this from being merged? We would really like this feature so we can stop using a fork.

libusb/core.c
```diff
@@ -2204,6 +2207,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+
+/** \ingroup libusb_dev
+ * Enable/disable RAW_IO for an endpoint on an open device.
+ *
+ * Only endpoints using the WinUSB driver support RAW_IO,
+ * for all other backends/drivers this function will return an error code.
+ *
+ * Using RAW_IO can greatly improve USB throughput by directly passing
+ * transfer requests to the underlying USB driver instead of queuing them
+ * in WinUSb. This can be particularly useful for high-throughput devices
```
seanm, line 2252: nit: WinUBb -> WinUSB

libusb/core.c
```diff
@@ -2204,6 +2207,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+	uint8_t endpoint, int enable)
+{
+	if (!usbi_backend.endpoint_set_raw_io)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	return usbi_backend.endpoint_set_raw_io(dev_handle, endpoint, enable);
+}
+
+/** \ingroup libusb_dev
+ * Retrieve the maximum transfer size supported for WinUSB RAW_IO for an inbound
```
seanm, line 2291: Size in what units? bytes? Best to say explicitly.

libusb/libusbi.h
```diff
@@ -1353,6 +1353,37 @@ struct usbi_os_backend {
...
+	 *
+	 * Return:
+	 * - 0 on success
+	 * - LIBUSB_ERROR_NOT_SUPPORTED if RAW_IO is not supported by the endpoint
+	 * - another LIBUSB_ERROR code on other failure
+	 */
+	int (*endpoint_set_raw_io)(struct libusb_device_handle* dev_handle,
+		uint8_t endpoint, int enable);
+
+	/* Retrieve the maximum transfer size supported for WinUSB RAW_IO for an
```
seanm, line 1376: Here too: state bytes, bits, kilobytes, or whatever units.

seanm (commented):
I don't know Windows, or even USB really, but this looks like nice work to me!  I noted a couple of small things.

mcuee:
@sonatique 

Please help to review this PR when you get a chance. Thanks.

mcuee:
> I don't know Windows, or even USB really, but this looks like nice work to me! I noted a couple of small things.

@seanm 

Just wondering if you still have some comments about this PR. If not, you can approve the PR.

libusb/core.c
```diff
@@ -2204,6 +2207,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+ *
+ * Only endpoints using the WinUSB driver support RAW_IO,
+ * for all other backends/drivers this function will return 0.
+ *
+ * Fails if the interface the endpoint belongs to isn't claimed,
+ * \see libusb_claim_interface().
+ *
+ * \see libusb_endpoint_set_raw_io()
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
```
seanm, line 2227: I guess this version number needs updating. In several places...

sonatique: Yes, shall we confidently put 1.0.30? I would say yes.

libusb/core.c
```diff
@@ -2204,6 +2207,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+ * \param enable 1 to enable RAW_IO, 0 to disable it
+ *
+ * \returns \ref LIBUSB_SUCCESS on success
+ * \returns \ref LIBUSB_ERROR_NOT_SUPPORTED if the backend does not support RAW_IO.
+ * \returns another LIBUSB_ERROR code on other failure
+ *
+ * \see libusb_endpoint_supports_raw_io()
+ * \seealso https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
```
sonatique, line 2279: Pease update to 1.0.30

libusb/core.c
```diff
@@ -2204,6 +2207,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+ * Returns a maximum transfer size in bytes, or a negative error code.
+ * If the backend does not support WinUSB RAW_IO, returns
+ * \ref LIBUSB_ERROR_NOT_SUPPORTED.
+ *
+ * Fails if the interface the endpoint belongs to isn't claimed,
+ * \see libusb_claim_interface().
+ *
+ * \see libusb_endpoint_set_raw_io()
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
```
sonatique, line 2303: Please update to 1.0.30

libusb/core.c
```diff
@@ -2204,6 +2207,111 @@ int API_EXPORTED libusb_set_auto_detach_kernel_driver(
...
+ * for all other backends/drivers this function will return 0.
+ *
+ * Fails if the interface the endpoint belongs to isn't claimed,
+ * \see libusb_claim_interface().
+ *
+ * \see libusb_endpoint_set_raw_io()
+ *
+ * Since version 1.0.28, \ref LIBUSB_API_VERSION >= 0x0100010B
+ */
+int API_EXPORTED libusb_endpoint_supports_raw_io(libusb_device_handle* dev_handle,
```
sonatique, line 2229: I know there have been a lot of discussion about the API, and I don't want to delay the merging of this nice PR more than necessary, but I am wondering why is this function needs to be on the public API, given that calling libusb_endpoint_set_raw_io on an endpoint that does not support it will return LIBUSB_ERROR_NOT_SUPPORTED anyway, so it seams that libusb_endpoint_supports_raw_io(dev, ep) is equivalent to libusb_endpoint_set_raw_io(dev, ep, 0) for instance.

I am also wondering about why the choice of 0 and 1 values instead of say, LIBUSB_ERRROR_NOT_SUPPORTED and LIBUSB_SUCCESS for instance.

sonatique (changes requested):

libusb/core.c
martinling: I answered that that question previously in [this comment](https://github.com/libusb/libusb/pull/1297#issuecomment-1882107586).

libusb/core.c
seanm: I think it's mergeable.

libusb/core.c
sonatique: Ok, thank you and sorry for missing this message. 

I understand and agree somehow. I tend to prefer less functions in API but I also acknowledge that your choice enables clearer and simpler client code.

libusb/core.c
sonatique: @seanm @HannesFranke-smartoptics : just regarding version numbers: what do you prefer:
a) I change it to 1.30 and merge
b) @HannesFranke-smartoptics change it (and rebase) then I merge
c) we merge as is and try not to forget to adapt the version numbers next time we do a release?

seanm: Personally I prefer (b), if he has not given up after over a year. Otherwise (a).

sonatique: > Personally I prefer (b), if he has not given up after over a year. Otherwise (a).

@seanm : OK, I'll do plan a) if @HannesFranke-smartoptics doesn't show up within a reasonable amount of time.

sonatique: @seanm @tormodvolden : please validate commit https://github.com/libusb/libusb/commit/2010b4f0f4b7f58847c73a6b340efbb19d31ef90 . 
This is what I plan to apply to this PR before merging. I prefer to have somebody validating the way I did it before merging it to master.

seanm: lgtm

tormodvolden: Yes, absolutely, please amend this to 1.0.30 before merging. I am sure @HannesFranke-smartoptics has no objection to that fixup.

martinling:
Glad to see this merged, thanks to all involved.

PR #1513, opened 2024-06-17
xusb minor fixes: Restore LIBUSB_DEBUG check, check for vid:pid

tormodvolden:


mcuee:
@tormodvolden 

Please help to resolve the conflict and update this PR when you got the time. Thanks.

tormodvolden:
Done. Sorry if I miss out on some @ mentions, I am subscribed to all libusb activity so I am swamped with notifications due to the high activity, I will need to tune my subscriptions and filters :)

examples/xusb.c
```diff
@@ -1207,12 +1204,14 @@ int main(int argc, char** argv)
...
+					unsigned int tmp_vid, tmp_pid;
+
 					if (sscanf(argv[j], "%x:%x" , &tmp_vid, &tmp_pid) != 2) {
 						printf("   Please specify VID & PID as \"vid:pid\" in hexadecimal format\n");
 						return EXIT_FAILURE;
 					}
-					VID = (uint16_t)tmp_vid;
-					PID = (uint16_t)tmp_pid;
+					vid = (uint16_t)tmp_vid;
+					pid = (uint16_t)tmp_pid;
```
seanm, line 1214: Maybe we don't care because this is example code, but these casts could truncate, and then you'd have garbage vid/pid.

Since you're already bothering the check the return value of sscanf, maybe check the range before casting too.

seanm (changes requested):

examples/xusb.c
tormodvolden: Neither the casts nor the sscanf check are modfied in this PR, it only changes the variable names to be lowercase. If you ignore white-space changes (git show -w) this is maybe more obvious.

If you think an input validation is important we can add this, but it wasn't in the original scope of the PR. VID/PIDs are always 16-bit but if a user e.g. wrongly enters a large decimal number it will be truncated.

seanm: > if a user e.g. wrongly enters a large decimal number it will be truncated.

Yes, without warning or explanation.

But indeed github's whitespace-sensitive change made me think this was something you changed.

I don't feel strongly about it as I don't use this code myself.

PR #1540, opened 2024-07-29
core: removed some useless consts

seanm:


seanm:
@tormodvolden friendly ping. This PR is trivial and safe.

seanm:
@tormodvolden with 1.0.28 out, can this triviality finally be merged?

mcuee:
@seanm 

Just wondering if this will cause any issue as it changes the core `libusb.h` file.

Is this change really necessary?

mcuee (approved):

tormodvolden:
> > Just wondering if this will cause any issue as it changes the core `libusb.h` file.
> 
> I don't see how...
> 
> > Is this change really necessary?
> 
> It cleans up a code smell. The `const` does nothing in these cases.

Could it cause a new warning in some users' code? Just to know in advance and potentially include in release notes.

seanm:
> Could it cause a new warning in some users' code?

I mean, theoretically a compiler / linter can warn about anything it wants. :)

I guess it's possible someone really wants the libusb_version struct to be immutable after creation. It's not a pattern I've seen much, but I could drop that part if you want.

tormodvolden (approved):
Upon review of this, I agree to all the changes. For `static inline uint16_t libusb_cpu_to_le16(const uint16_t x)` x is passed as value, so the const is pointless.
For the libusb_version struct, it doesn't make sense to declare individual members as const, either the whole struct is const or it is not. What is important is that the version value is non-modifiable (so that it e.g. can be located in a read-only segment, or that the compiler complains if code tries to modify it) and this is taken care of by `static const struct libusb_version libusb_version_internal` and `const struct libusb_version * LIBUSB_CALL libusb_get_version(void);`

I will merge this after the fix-up release.

seanm:
> For `static inline uint16_t libusb_cpu_to_le16(const uint16_t x)` x is passed as value, so the const is pointless.

Exactly.

> For the libusb_version struct, it doesn't make sense to declare individual members as const, either the whole struct is const or it is not. 

In fact all the members are const. So the existing construct would prevent modifying fields after the fact, like this:

https://godbolt.org/z/56hcMYo14

I had not considered that. Maybe that's why it was done. Personally I've not seen this pattern much in C.

tormodvolden:
Yes, but I don't think we need to prevent the user from using the struct read-write if they somehow can find a use for it for their own variables. It is enough for us that libusb_get_version() points to a read-only instance. So I would be fine with your change.

seanm:
OK, fine with me. Can we merge now?

tormodvolden:
Maybe we should rephrase the comment a bit, as you have demonstrated, the const is not useless, and it does change things, but it is maybe overly restrictive.

Re "now", as part of the release management, I usually I wait a bit after a release before I merge new things, certainly if they are not important. We know that any seemingly trivial thing can turn out more complex.

seanm:
> Maybe we should rephrase the comment a bit, as you have demonstrated, the const is not useless, and it does change things, but it is maybe overly restrictive.

Updated.

> Re "now", as part of the release management, I usually I wait a bit after a release before I merge new things, certainly if they are not important. We know that any seemingly trivial thing can turn out more complex.

Considering how small 1.0.28 to 1.0.29 was, and how conservative this change is, it seems plenty safe to merge.

tormodvolden:
Interesting too, if the original use of const should, for consistency, protect "rc" and "desc" from changing too, they would need an extra `*const`, otherwise they are mutable pointers (but to immutable chars). So the original struct was a mix of mutable and immutable members. This is C-wise actually fine, and potentially useful in some cases, in contrast to what I said above. But if at least one member is const, you cannot use the struct with assignments any longer. OTOH I guess that as soon as one member is non-const, the whole struct is stored in RAM without fine-grained memory protection, so that the originally const elements can be manipulated by casting (pointers) to a non-const struct.

https://www.gnu.org/software/c-intro-and-ref/manual/html_node/const-Fields.html

sonatique (approved):

mcuee:
@sonatique 

This PR is now safe to merge. You may want to try to practice merging the PR. Please take note of the things mentioned in the Wiki.
https://github.com/libusb/libusb/wiki/Maintainers'-corner

You can also check with @tormodvolden if you have some doubts.

I am basically not a programmer and not so familiar with git, so I do not touch this area. Rather I focus on testing and technical support side.

tormodvolden:
@sonatique Just study the git history to see what well-formed commits look like. I usually fix up commit messages for style and spelling, i.e. capitalization and imperative form, and add Fixes #bug number and Closes #PR number when appropriate. And double-check the nano version :)

tormodvolden:
@sonatique thanks for the review. I went ahead and merged since I was cleaning out low-hanging fruit.
@seanm Thanks for your contribution!

sonatique:
@mcuee : thanks for offering me to merge. It would have been a pleasure but unfortunately I missed your comment and saw it too late. Next time!

Issue #1442, opened 2024-01-22
Agree on formatting style, and then create .clang-format config file to enforce it


seanm:
We should agree on a formatting style.

Some files intent with tabs, some with spaces.  Some put braces one way, some another.

Once we agree on a style, we can then autoformat using the [clang-format](https://clang.llvm.org/docs/ClangFormat.html) tool. There are even tools that examine a codebase and generate a preliminary clang-format config file based on the predominant style.

seanm:
I started some work here: https://github.com/libusb/libusb/pull/1443

First thing to decide on: tabs vs spaces :)

Youw:
I think we will not be able to get a single `.clang-format` for the entire project. At least not at the beginning.
We need to figure out if it is possible to have several flavours of it at least per-platform.
So we can select "tabs vs spaces" at least per-plaform based on how it is mostly used right now.

Youw:
Second thing: I think we should come up with a `.clang-format`, which, when applied - produces (reasonably) the smallest diff in the existing code.

Youw:
> Why is that desirable?

To preserve git history - it matters:
![image](https://github.com/libusb/libusb/assets/5939659/acec50c8-0e30-4e81-8262-977a99cbe38c)

If we apply `.clang-format` to all existing code - _a lot_ of immediate blame/history will be hidden under something like "applying code style".
If we agree but don't apply format to all existing code - any change to places where something (e.g. tags vs spaces) is different would introduce lots of weird inconsistencies. I've seen that - looks terrible.

> I did that already,

I meant per-backend, or even per-file.

seanm:
> If we apply `.clang-format` to all existing code - _a lot_ of immediate blame/history will be hidden under something like "applying code style".

Which is why `git blame` has `--ignore-rev` and `--ignore-revs-file`. We can create a file listing all revisions that should be ignored in blame and even put that in the `.gitconfig` file.

tormodvolden:
> > Second thing: I think we should come up with a `.clang-format`, which, when applied - produces (reasonably) the smallest diff in the existing code.
> 
> I did that already, using https://github.com/mikr/whatstyle. The result is the first commit here.

When doing this I think some oddballs in os/* should be taken out. Maybe that whole folder.

tormodvolden:
> > We need to figure out if it is possible to have several flavours of it at least per-platform.
> 
> Why is that desirable? Such a small project could surely have a uniform style.

[...]

PR #1541, opened 2024-07-29
descriptor: fixed clang -Wimplicit-int-conversion warnings

seanm:
For the 16 bit case especially, the result of the `or` is implicitly promoted to `int`, then when returned was warning:

warning: implicit conversion loses integer precision: 'int' to 'uint16_t' (aka 'unsigned short') [-Wimplicit-int-conversion]

Added more casts to shut it up.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -32,16 +32,16 @@
 
 static inline uint16_t ReadLittleEndian16(const uint8_t p[2])
 {
-	return (uint16_t)p[1] << 8 |
-	       (uint16_t)p[0];
+	return (uint16_t)((uint16_t)p[1] << 8 |
+			  (uint16_t)p[0]);
 }
 
 static inline uint32_t ReadLittleEndian32(const uint8_t p[4])
 {
-	return (uint32_t)p[3] << 24 |
-	       (uint32_t)p[2] << 16 |
-	       (uint32_t)p[1] << 8 |
-	       (uint32_t)p[0];
+	return (uint32_t)((uint32_t)p[3] << 24 |
+			  (uint32_t)p[2] << 16 |
+			  (uint32_t)p[1] << 8 |
+			  (uint32_t)p[0]);
 }
 
 static void clear_endpoint(struct libusb_endpoint_descriptor *endpoint)


libusb/descriptor.c
```diff
@@ -32,16 +32,16 @@
 
 static inline uint16_t ReadLittleEndian16(const uint8_t p[2])
 {
-	return (uint16_t)p[1] << 8 |
-	       (uint16_t)p[0];
+	return (uint16_t)((uint16_t)p[1] << 8 |
+					  (uint16_t)p[0]);
```
Youw, line 36: Indentation this way is a bit off, I'd rather see:
```suggestion
	return (uint16_t)((uint16_t)p[1] << 8 |
	                  (uint16_t)p[0]);
```

libusb/descriptor.c
```diff
@@ -32,16 +32,16 @@
...
 static inline uint32_t ReadLittleEndian32(const uint8_t p[4])
 {
-	return (uint32_t)p[3] << 24 |
-	       (uint32_t)p[2] << 16 |
-	       (uint32_t)p[1] << 8 |
-	       (uint32_t)p[0];
+	return (uint32_t)((uint32_t)p[3] << 24 |
+					  (uint32_t)p[2] << 16 |
+					  (uint32_t)p[1] << 8 |
+					  (uint32_t)p[0]);
```
Youw, line 44: yeah, I usually hate mixed indentation, but this is exactly the place for it:
```suggestion
	return (uint32_t)((uint32_t)p[3] << 24 |
	                  (uint32_t)p[2] << 16 |
	                  (uint32_t)p[1] << 8 |
	                  (uint32_t)p[0]);
```

libusb/descriptor.c
seanm: Ah, the Xcode project has this file set for tab=4, but from the emacs configuration directives at the top of the file, it should be tab=8 it seems... will fix.

God I wish we could just decide on a clang-format: https://github.com/libusb/libusb/issues/1442

seanm: Updated Xcode project: https://github.com/libusb/libusb/pull/1543

Youw: The thing is - you can never guess the "correct" tab size. All my editors/IDEs are using 4 spaces for tab, and GitHub uses 8.

That's exactly why in cases like this we have to use mixed spaces - tabs for the beginning-of-line indentation, and then spaces to align the "content" of the line, like I suggested in my comments.

seanm: Well, in this cause I could guess because the file starts with:

`/* -*- Mode: C; indent-tabs-mode:t ; c-basic-offset:8 -*- */`

PR #1542, opened 2024-07-29
core: add LIBUSB_PACKED_PUSH / LIBUSB_PACKED_POP for portable struct padding

seanm:
…padding

GCC/Clang support MSVC syntax for struct packing, so we can use that everywhere instead of having two different techniques.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -81,10 +81,23 @@ typedef SSIZE_T ssize_t;
 #define LIBUSB_DEPRECATED_FOR(f)
 #endif /* __GNUC__ */
 
+/* Deprecated, use LIBUSB_PACKED_PUSH/LIBUSB_PACKED_POP instead. */
 #if defined(__GNUC__)
 #define LIBUSB_PACKED __attribute__ ((packed))
 #else
 #define LIBUSB_PACKED
+#endif /* __GNUC__ */
+
+#if defined(__GNUC__)
+#define LIBUSB_PACKED_PUSH _Pragma("pack(push, 1)")
+#define LIBUSB_PACKED_POP _Pragma("pack(pop)")
+#elif defined(_MSC_VER) || defined(__WATCOMC__)
+#define LIBUSB_PACKED_PUSH __pragma("pack(push, 1)")
+#define LIBUSB_PACKED_POP __pragma("pack(pop)")
+#else
+#warning Compiler not recognised; therefore not possible to pack structs.
+#define LIBUSB_PACKED_PUSH
+#define LIBUSB_PACKED_POP
 #endif /* __GNUC__ */
 
 /** \def LIBUSB_CALL
@@ -1157,9 +1170,7 @@ struct libusb_platform_descriptor {
 
 /** \ingroup libusb_asyncio
  * Setup packet for control transfers. */
-#if defined(_MSC_VER) || defined(__WATCOMC__)
-#pragma pack(push, 1)
-#endif
+LIBUSB_PACKED_PUSH
 struct libusb_control_setup {
 	/** Request type. Bits 0:4 determine recipient, see
 	 * \ref libusb_request_recipient. Bits 5:6 determine type, see
@@ -1184,10 +1195,8 @@ struct libusb_control_setup {
 
 	/** Number of bytes to transfer */
 	uint16_t wLength;
-} LIBUSB_PACKED;
-#if defined(_MSC_VER) || defined(__WATCOMC__)
-#pragma pack(pop)
-#endif
+};
+LIBUSB_PACKED_POP
 
 #define LIBUSB_CONTROL_SETUP_SIZE (sizeof(struct libusb_control_setup))
 
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -83,10 +83,26 @@ typedef SSIZE_T ssize_t;
 #define LIBUSB_DEPRECATED_FOR(f)
 #endif /* __GNUC__ */
 
+/* Deprecated, use LIBUSB_PACKED_PUSH/LIBUSB_PACKED_POP instead. */
 #if defined(__GNUC__)
 #define LIBUSB_PACKED __attribute__ ((packed))
 #else
 #define LIBUSB_PACKED
+#endif /* __GNUC__ */
+
+#if defined(__GNUC__)
+#define LIBUSB_PACKED_PUSH _Pragma("pack(push, 1)")
+#define LIBUSB_PACKED_POP _Pragma("pack(pop)")
+#elif defined(_MSC_VER)
+#define LIBUSB_PACKED_PUSH __pragma(pack(push, 1))
+#define LIBUSB_PACKED_POP __pragma(pack(pop))
+#elif defined(__WATCOMC__)
+#define LIBUSB_PACKED_PUSH pragma pack(__push, 1)
+#define LIBUSB_PACKED_POP pragma pack(__pop, 1)
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)
+#warning Compiler not recognised; therefore not possible to pack structs as required.
+#define LIBUSB_PACKED_PUSH
+#define LIBUSB_PACKED_POP
 #endif /* __GNUC__ */
 
 /** \def LIBUSB_CALL
@@ -1160,9 +1176,7 @@ struct libusb_platform_descriptor {
 
 /** \ingroup libusb_asyncio
  * Setup packet for control transfers. */
-#if defined(_MSC_VER) || defined(__WATCOMC__)
-#pragma pack(push, 1)
-#endif
+LIBUSB_PACKED_PUSH
 struct libusb_control_setup {
 	/** Request type. Bits 0:4 determine recipient, see
 	 * \ref libusb_request_recipient. Bits 5:6 determine type, see
@@ -1187,10 +1201,9 @@ struct libusb_control_setup {
 
 	/** Number of bytes to transfer */
 	uint16_t wLength;
-} LIBUSB_PACKED;
-#if defined(_MSC_VER) || defined(__WATCOMC__)
-#pragma pack(pop)
-#endif
+};
+LIBUSB_PACKED_POP
+static_assert(sizeof(struct libusb_control_setup) == 8, "struct packing failed");
 
 #define LIBUSB_CONTROL_SETUP_SIZE (sizeof(struct libusb_control_setup))
 
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -659,16 +659,15 @@ enum usbi_transfer_timeout_flags {
 	 ((unsigned char *)(transfer)			\
 	  - PTR_ALIGN(sizeof(struct usbi_transfer))))
 
-#ifdef _MSC_VER
-#pragma pack(push, 1)
-#endif
-
 /* All standard descriptors have these 2 fields in common */
+LIBUSB_PACKED_PUSH
 struct usbi_descriptor_header {
 	uint8_t  bLength;
 	uint8_t  bDescriptorType;
-} LIBUSB_PACKED;
+};
+LIBUSB_PACKED_POP
 
+LIBUSB_PACKED_PUSH
 struct usbi_device_descriptor {
 	uint8_t  bLength;
 	uint8_t  bDescriptorType;
@@ -684,8 +683,10 @@ struct usbi_device_descriptor {
 	uint8_t  iProduct;
 	uint8_t  iSerialNumber;
 	uint8_t  bNumConfigurations;
-} LIBUSB_PACKED;
+};
+LIBUSB_PACKED_POP
 
+LIBUSB_PACKED_PUSH
 struct usbi_configuration_descriptor {
 	uint8_t  bLength;
 	uint8_t  bDescriptorType;
@@ -695,8 +696,10 @@ struct usbi_configuration_descriptor {
 	uint8_t  iConfiguration;
 	uint8_t  bmAttributes;
 	uint8_t  bMaxPower;
-} LIBUSB_PACKED;
+};
+LIBUSB_PACKED_POP
 
+LIBUSB_PACKED_PUSH
 struct usbi_interface_descriptor {
 	uint8_t  bLength;
 	uint8_t  bDescriptorType;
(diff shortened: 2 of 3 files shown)


seanm:
@tormodvolden friendly ping

seanm:
@tormodvolden with 1.0.28 out, can we please merge this?

seanm:
So those warnings were there before?  Meaning this doesn't fix them, but doesn't add any warnings either?

mcuee:
> So those warnings were there before? Meaning this doesn't fix them, but doesn't add any warnings either?

That is right. So I am fine with this PR myself. Needs the review from @tormodvolden and @sonatique.

mcuee (approved):

seanm:
@tormodvolden could you review this one too?

tormodvolden (approved):

tormodvolden (changes requested):
Hang on, this seems to break Windows CI on VS2017, VS2015, VS2013

seanm:
Hmm, maybe it really wants the semicolon after the pragma... will try...

seanm:
No, that's not it... 

Maybe it's because I used `_Pragma`, which is apparently C99...

seanm:
Looks like C99 `_Pragma` is supported by MS only if using C11 mode or higher: https://learn.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=msvc-170

mcuee:
Looks like this PR does not help the MSYS2 clang64 situation.
https://github.com/libusb/libusb/actions/runs/25385278334/job/74444519517?pr=1542

```
  CC       os/windows_common.lo
In file included from ../../libusb/os/windows_common.c:30:
../../libusb/os/windows_common.h:163:10: warning: the current #pragma pack alignment value is modified in the included file [-Wpragma-pack]
  163 | #include <pshpack1.h>
      |          ^
C:/a/_temp/msys64/clangarm64/include/pshpack1.h:7:9: note: previous '#pragma pack' directive that modifies alignment is here
    7 | #pragma pack(push,1)
      |         ^
In file included from ../../libusb/os/windows_common.c:30:
../../libusb/os/windows_common.h:193:10: warning: the current #pragma pack alignment value is modified in the included file [-Wpragma-pack]
  193 | #include <poppack.h>
      |          ^
note: previous '#pragma pack' directive that modifies alignment is here
  CC       os/windows_usbdk.lo
2 warnings generated.
  CC       os/windows_winusb.lo
  CC       os/events_windows.lo
  CC       os/threads_windows.lo
In file included from ../../libusb/os/windows_usbdk.c:30:
In file included from ../../libusb/os/windows_usbdk.h:27:
../../libusb/os/windows_common.h:163:10: warning: the current #pragma pack alignment value is modified in the included file [-Wpragma-pack]
  163 | #include <pshpack1.h>
      |          ^
C:/a/_temp/msys64/clangarm64/include/pshpack1.h:7:9: note: previous '#pragma pack' directive that modifies alignment is here
    7 | #pragma pack(push,1)
      |         ^
In file included from ../../libusb/os/windows_winusb.c:35:
../../libusb/os/windows_winusb.h:712:10: warning: the current #pragma pack alignment value is modified in the included file [-Wpragma-pack]
  712 | #include <poppack.h>
      |          ^
note: previous '#pragma pack' directive that modifies alignment is here
../../libusb/os/windows_winusb.c:1020:10: warning: enumeration value 'LIBUSB_SPEED_SUPER_PLUS_X2' not explicitly handled in switch [-Wswitch-enum]
 1020 |         switch (speed) {
      |                 ^~~~~
../../libusb/os/windows_winusb.c:2293:10: warning: enumeration value 'LIBUSB_DEVICE_STRING_COUNT' not explicitly handled in switch [-Wswitch-enum]
 2293 |         switch (string_type) {
      |                 ^~~~~~~~~~~
In file included from ../../libusb/os/events_windows.c:24:
../../libusb/os/windows_common.h:163:10: warning: the current #pragma pack alignment value is modified in the included file [-Wpragma-pack]
  163 | #include <pshpack1.h>
      |          ^
C:/a/_temp/msys64/clangarm64/include/pshpack1.h:7:9: note: previous '#pragma pack' directive that modifies alignment is here
    7 | #pragma pack(push,1)
      |         ^
In file included from ../../libusb/os/events_windows.c:24:
../../libusb/os/windows_common.h:193:10: warning: the current #pragma pack alignment value is modified in the included file [-Wpragma-pack]
  193 | #include <poppack.h>
      |          ^
note: previous '#pragma pack' directive that modifies alignment is here
2 warnings generated.
10 warnings generated.
```

seanm:
> Looks like this PR does not help the MSYS2 clang64 situation.

Yes, you said so a year ago too. Nothing has changed since.

seanm:
@sonatique @Youw well, it took a very long time, but I think this is finally correct and ready to review/merge!

libusb/libusbi.h
```diff
@@ -647,16 +647,15 @@ enum usbi_transfer_timeout_flags {
...
 /* All standard descriptors have these 2 fields in common */
+LIBUSB_PACKED_PUSH
 struct usbi_descriptor_header {
 	uint8_t  bLength;
 	uint8_t  bDescriptorType;
-} LIBUSB_PACKED;
+};
+LIBUSB_PACKED_POP
 
+LIBUSB_PACKED_PUSH
```
Youw, line 662: maybe this is a matter of taste, but I don't think this is necessary:
```suggestion

```
here and in all other places

libusb/libusb.h
```diff
@@ -83,12 +83,25 @@ typedef SSIZE_T ssize_t;
...
 #if defined(__GNUC__)
 #define LIBUSB_PACKED __attribute__ ((packed))
 #else
 #define LIBUSB_PACKED
 #endif /* __GNUC__ */
 
+#if defined(__GNUC__)
+#define LIBUSB_PACKED_PUSH _Pragma("pack(push, 1)")
+#define LIBUSB_PACKED_POP _Pragma("pack(pop)")
+#elif defined(_MSC_VER) || defined(__WATCOMC__)
```
Youw: **[P1] Preserve Watcom C header compatibility**

`__WATCOMC__` cannot use MSVC's `__pragma(pack(...))` form. [Open Watcom documents `__pragma` as a C++ modifier that requires a string naming an auxiliary pragma](https://github.com/open-watcom/open-watcom-v2/blob/340c139c15c3fa576c3a742128a10be21bb876d5/docs/doc/cg/cpwcc.gml#L1993-L2001). This macro expands to invalid C before `libusb_control_setup`. The base header intentionally uses direct `#pragma pack` for Watcom; that support was added in [#904](https://github.com/libusb/libusb/pull/904). Preserve that direct pragma path, or use another form verified with the Watcom C compiler. A Watcom public-header compile check will prevent this regression.

libusb/libusb.h
```diff
@@ -83,12 +83,25 @@ typedef SSIZE_T ssize_t;
...
+#define LIBUSB_PACKED_PUSH __pragma(pack(push, 1))
+#define LIBUSB_PACKED_POP __pragma(pack(pop))
+#else
+#warning Compiler not recognised; therefore not possible to pack structs.
```
Youw: **[P1] Keep the installed header silent for the fallback case**

The base header lets other compilers include `libusb.h` with `LIBUSB_PACKED` empty. This new `#warning` makes the same include fail when warnings are errors. It is also not portable before C23/C++23. In a bounded GCC probe that simulated the fallback with `-U__GNUC__ -Werror=cpp`, the base include passed and this head failed at this line. The fallback macros remain empty, so remove the diagnostic to preserve source compatibility.

```suggestion

```

Youw (commented):
Codex found two backward-compatibility blockers in the installed public header. Both findings are in inline comments.

Positive checks:

- `libusb_control_setup` keeps size 8, alignment 1, and offsets 0/1/2/4/6 at the base and head revisions with GCC 11.
- The header restores a caller's `#pragma pack(push, 4)` state.
- C99, C11, C++98, and C++20 compile checks pass for the GNU path.
- `git diff --check` passes. All current CI checks pass.
- Current CI does not test Watcom or another compiler that reaches the fallback branch.

For the smallest patch, preserve the direct Watcom pragma path, remove the fallback warning, and use one push/pop pair around the contiguous internal descriptor group as the existing inline thread suggests.

libusb/libusbi.h
seanm: I thought it preferable to treat each struct individually.

libusb/libusb.h
seanm: Better to fail loudly than silently compile with non-packed structs, which would break all kinds of stuff in hard to find ways.

libusb/libusb.h
seanm: Looks like we could do `#pragma pack(__push,4)` with Watcom:

https://github.com/open-watcom/open-watcom-v2/blob/340c139c15c3fa576c3a742128a10be21bb876d5/bld/watcom/h/pushpck4.h#L34

libusb/libusb.h
Youw: This is debatable. We used to allow compilation where pragma/pack wasn't available, now we do not.
But even when we did - was that implementation/binary really compliant to libusb standard?

Youw: It could be compliant; it was not guaranteed.

On the usual ABI, this specific struct has size 8 and offsets `0/1/2/4/6` without packing. Codex verified that layout in the old fallback path. Its alignment is 2 instead of 1, but [the public helper already requires the buffer to be aligned to at least 2 bytes](https://github.com/libusb/libusb/blob/37ebb738a55ceb2815f7f472852ada002e5d6aeb/libusb/libusb.h#L1935-L1940). The wire bytes and documented use are therefore valid on such an ABI.

An ABI that inserts padding would not be compliant. `LIBUSB_CONTROL_SETUP_SIZE` would change, and some backends would send the wrong bytes. Explicit packing is useful for that reason.

The `#warning` tests the compiler name, not the actual layout, and it does not stop a normal build. A compile-time layout check, including `sizeof(struct libusb_control_setup) == 8`, would fail deterministically for an incompatible layout while keeping compatible fallback consumers working.

libusb/libusbi.h
Youw: I'd rather prefer having an explicit compile-time size requirement for each struct (which is indirectly enforced by the `pack` directive)

libusb/libusb.h
Youw: `#pragma pack(__push,1)`:
https://github.com/open-watcom/open-watcom-v2/blob/340c139c15c3fa576c3a742128a10be21bb876d5/bld/watcom/h/pushpck1.h#L34

Right, the point is - it uses `#pragma pack(__push`, not `__pragma(pack(push` syntax

libusb/libusb.h
Youw: I agree that a better way would be to have a compile time check for the struct size, e.g:
```c
#if sizeof(struct libusb_control_setup) != 8
#error "libusb_control_setup has to be 8 bytes, check struct alignment"
#endif
```
and keep the `LIBUSB_PACKED_PUSH`/`LIBUSB_PACKED_POP` silent/best effort, since it absolutely doesn't matter for as long as it works in runtime.

We can even name it `LIBUSB_PRIVATE_PACKED_PUSH/LIBUSB_PRIVATE_PACKED_POP` and explicitly discaurage its use, even though it is in a public header.

libusb/libusbi.h
seanm: Yeah, I too prefer the gcc/clang way of annotating a struct with `__attribute__ ((packed))`. But MSVC does not have such a thing.

The whole idea of this PR was that it's better to at least do things the same way with all compilers, which all support this push/pop idea (though not with the exact same syntax).

Sad that even in C23 no such thing is standardized. Sigh.

libusb/libusb.h
seanm: See new commit.

libusb/libusb.h
seanm: See new commit.

libusb/libusb.h
```diff
@@ -1187,10 +1201,9 @@ struct libusb_control_setup {
 
 	/** Number of bytes to transfer */
 	uint16_t wLength;
-} LIBUSB_PACKED;
-#if defined(_MSC_VER) || defined(__WATCOMC__)
-#pragma pack(pop)
-#endif
+};
+LIBUSB_PACKED_POP
+static_assert(sizeof(struct libusb_control_setup) == 8, "struct packing failed");
```
Youw, line 1206: need explicit `#include <assert.h>`

libusb/libusb.h
```diff
@@ -83,12 +83,28 @@ typedef SSIZE_T ssize_t;
...
+#define LIBUSB_PACKED_PUSH _Pragma("pack(push, 1)")
+#define LIBUSB_PACKED_POP _Pragma("pack(pop)")
+#elif defined(_MSC_VER)
+#define LIBUSB_PACKED_PUSH __pragma(pack(push, 1))
+#define LIBUSB_PACKED_POP __pragma(pack(pop))
+#elif defined(__WATCOMC__)
+#define LIBUSB_PACKED_PUSH pragma pack(__push, 1)
+#define LIBUSB_PACKED_POP pragma pack(__pop, 1)
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)
+#warning Compiler not recognised; therefore not possible to pack structs as required.
```
Youw, line 103: now that we have static asserts - I see no reasons why fail build with warning-as-errors when alignment happens to be right even w/o pragmas

Youw (commented):
Looks good with regard to my last comments

libusb/os/windows_winusb.c
```diff
@@ -2470,19 +2470,14 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
 	UCHAR bDescriptorType;
 	uint8_t bString[252];  // UTF-16LE, must be even length
-} LIBUSB_PACKED;
-
-#ifdef _MSC_VER
-#pragma pack(pop)
-#endif
+};
+LIBUSB_PACKED_POP
+static_assert(sizeof(struct usbi_string_descriptor) == 254, "struct packing failed");
```
Youw, line 2480:
```suggestion
static_assert(sizeof(struct string_descriptor_s ) == 254, "struct packing failed");
```

libusb/libusb.h
```diff
@@ -83,12 +83,28 @@ typedef SSIZE_T ssize_t;
...
+#if defined(__GNUC__)
+#define LIBUSB_PACKED_PUSH _Pragma("pack(push, 1)")
+#define LIBUSB_PACKED_POP _Pragma("pack(pop)")
+#elif defined(_MSC_VER)
+#define LIBUSB_PACKED_PUSH __pragma(pack(push, 1))
+#define LIBUSB_PACKED_POP __pragma(pack(pop))
+#elif defined(__WATCOMC__)
+#define LIBUSB_PACKED_PUSH pragma pack(__push, 1)
+#define LIBUSB_PACKED_POP pragma pack(__pop, 1)
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)
```
Youw, line 102: wait, this condition looks weird - it says "otherwise we do not support C23 compilers"?

Issue #1455, opened 2024-02-02
Proper hotplug test utility


mcuee:
From here by @tormodvolden 

> I intend to work on a hotplug testing utility after the release, to be used for development (if no one beats me to it).

mcuee:
Some improbement ideas.
1) To be able to specify number of devicdes for monitoring.
2) Adopt similar idea to #1425

mcuee:
Some comments from #1706.

@sonatique has some good suggestions.
* https://github.com/libusb/libusb/issues/1706#issuecomment-3998864933

> So: in my opinion: this hotplugtest application is far from being a real "test" program. It's more like a demo application for illustrating how to use the API for seting-up callbacks. It supposes that things are executed in a specific order.
>
> If we want a more robust / flexible demo or a real test program for stress testing hotplug scenarios, then we shall either improve this one or write a new one. A first improvement to the demo would be for instance to:
> 
> 1) not open devices (it serves no purpose but for demoing the API), no need for the last check / warning
> 2) loop for ever (until ctrl-c or "Q" key press for instance)
>
> I think this way the application would behave more like you expected it.

mcuee:
@tormodvolden has some good comments in #1706.

> The hotplugtest kind of expects you would start the program, then plug in devices, and then optionally repeat variations of (plug some out again, in again) then when it believes all devices are unplugged again it will quit.
> The program is not a a stress-test or validation tool, I guess it was originally meant to be a simple example showing how to use the hotplug API.

sonatique:
@mcuee : this is my proposal:  https://github.com/libusb/libusb/pull/1776

As I said elsewhere hotplugtest was never meant to be a "test" program (despite its name), and this is actually obvious once you realize it is in the "examples" folder, not in the "tests" folder ;-)

Therefore I didn't add real test, but rathe I tried to make it more user-friendly and exampling more features and things.

PR #1550, opened 2024-08-12
examples/hotplugtest: Allow multiple devices attaching

tormodvolden:
If no device ID is specified, and for instance a hub is plugged, many devices will be attached (and later detached) at the same time. The old code counting only 2 events would then exit prematurely.

By counting attach and detach events separately, we preserve the old behaviour in the single-device case, but also allow more complex sequences to be handled in a way that mostly appears intuitive.

This is not fool-proof, and we can still end up leaving after a surprise detachment of a "pre-existing" device while we have one of the "new" devices open. In this case print a warning.

References #1455

### DIFF
--- a/examples/hotplugtest.c
+++ b/examples/hotplugtest.c
@@ -30,6 +30,7 @@ libusb_device_handle *handle = NULL;
 static int LIBUSB_CALL hotplug_callback(libusb_context *ctx, libusb_device *dev, libusb_hotplug_event event, void *user_data)
 {
 	struct libusb_device_descriptor desc;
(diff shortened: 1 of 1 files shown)


mcuee:
Good change. This should be safe to be merged.

mcuee:
@tormodvolden 
Now that libusb 1.0.28 has been released. Maybe this can be merged.

seanm (approved):

sonatique (approved):

tormodvolden:
Thanks for your reviews!

PR #1555, opened 2024-08-26
Less harsh handling of extra endpoint descriptors

PatriciaJonsson:
I have a touchscreen that has three interfaces; touchscreen, mouse/keyboard and stylus/pen. I can use the pen interface when I connect the device to Windows, Android and Linux, but when I use libusb I can't access the third interface with the pen. I've found that the device's interface descriptor claims to have one endpoint descriptor when there's in fact two. 
This means that libusb will stop parsing any interfaces left since it's pointing at an endpoint descriptor, but expects an interface descriptor. 
In my case, that meant that the second interface that claims to have only one endpoint but has two, will be the last interface parsed and the third will not be parsed.

Kernel seems to be more forgiving and uses the descriptor size to parse the descriptors and gives a notice when there are more endpoint descriptors than described. I've tried to follow this as well and use the descriptor size when parsing the endpoints instead of the reported bNumEndpoints and have a warning when this occurs.


tormodvolden:
That sounds like it makes sense. Please post lsusb -v output for the device in question.

PatriciaJonsson:
[lsusb.txt](https://github.com/user-attachments/files/16757181/lsusb.txt)

tormodvolden:
Please be clear in the commit message that this is a workaround for a broken device, and make reference to the given product. Otherwise it is unclear for someone who follows the USB standard why this is needed. For any workarounds and quirks we add we should keep a record of known devices it is needed for. Please also check spelling.

seanm:
> Please be clear in the commit message that this is a workaround for a broken device

Or better still in comments in code, no?

PatriciaJonsson:
> Does lsusb show the second endpoint in the second interface if your patch is applied?

[lsusb_with_fix.txt](https://github.com/user-attachments/files/16776764/lsusb_with_fix.txt)
No the second endpoint does not get listed. Don't know if I think it should. Doing a test with hidapi/hidraw all endpoints gets listed though. It's one or the other, either just list the ones accurate to bNumEndpoints or list everything that's there. 

> Please be clear in the commit message that this is a workaround for a broken device, and make reference to the given product. Otherwise it is unclear for someone who follows the USB standard why this is needed. For any workarounds and quirks we add we should keep a record of known devices it is needed for. Please also check spelling.

I'll look over both the comments and commit message

libusb/descriptor.c
```diff
@@ -274,27 +275,47 @@ static int parse_interface(libusb_context *ctx,
 
 		if (ifp->bNumEndpoints > 0) {
 			struct libusb_endpoint_descriptor *endpoint;
-			uint8_t i;
+
+			/* Endpoint desciptors reported */
+			if_desc_eps = ifp->bNumEndpoints;
 
 			endpoint = calloc(ifp->bNumEndpoints, sizeof(*endpoint));
```
tormodvolden, line 282: Aren't we gonna overflow this allocation if we end up with more endpoints than ifp->bNumEndpoints?

PatriciaJonsson: Ah good catch. For my product it shouldn't be a problem, but for a product with + 2 more endpoints it could. I'll look over the whole change and update

tormodvolden: You probably have your own ideas about how to deal with it, anyway I would have looked at making parse_endpoint() conditionally a read-only function (e.g. when passed a NULL buffer) and call it once to count endpoints, allocate, and call it again.

PatriciaJonsson: Yes, I am working on a different solution. I've found that my suggestion has more problems than the one you pointed out. The new idea is to keep the for loop from before and only parse the amount of endpoint provided by bNumEndpoints and then after, check if we're still pointing at an endpoint descriptor we're not expecting. If so we skip it and point to the next "proper" descriptor. 
I need to do some more testing before I push this suggestion

PR #1556, opened 2024-08-28
introduce new transfer state LIBUSB_TRANSFER_DISCONNECT

maybetaken:
While pollfd get a POLLERR from a fd, libusb tries to disconnect handle from libusb. The state reports LIBUSB_TRANSFER_NO_DEVICE, which is correct but not accurate, it is hard to distinguish from the user side. In most cases, we get LIBUSB_TRANSFER_NO_DEVICE through getting return value of ioctl, getting a POLLERR and disconnect handle with a new transfer state LIBUSB_TRANSFER_DISCONNECT is more accurate for this scenario.


fabiensanglard:
I don't think we need a `LIBUSB_TRANSFER_DISCONNECT` state.

Wouldn't the user code handle `LIBUSB_TRANSFER_DISCONNECT` exactly as `LIBUSB_TRANSFER_NO_DEVICE`?

mcuee:
Close this PR in this case.

PR #1557, opened 2024-08-28
fix fetech data after low level bus error reap completed

maybetaken:
Low level bus error reap completed is treated as LIBUSB_TRANSFER_ERROR Then it continues to fetch data, in some cases usb device is plugged out then ioctl will hang servel ms for waiting device data. This commit treats low level error and transfer completed as not required for another reap thusing fixing waiting ioctl return.


mcuee:
No so sure if this PR is correct or not. I tend to think not.

mcuee:
Close old PRs which does not seem to be correct and without updates for more than six months. Please re-open even if you think it is closed wrongly or you have new updates.

PR #1558, opened 2024-08-28
descriptor: Warn on invalid langid descriptor

### LINKED ISSUE #1548: nut_libusb_open get `some values` failed, retrying...
Not sure if related to libusb , but maybe any help from here would help to recognize some values for libusb NUT for Eaton 9E UPS
https://github.com/networkupstools/nut/pull/2571
https://github.com/networkupstools/nut/issues/1925

tormodvolden:
Closes #1548

Since we were just silently erring out on a failing langid retrieval, it was not easy to see if the problem is with the langid descriptor or with the wanted string descriptor.


mcuee:
This seems to be a safe to be merged.

seanm (approved):
This is such an incredibly conservative change, and it's been a year, why not just merge it?

Issue #1571, opened 2024-10-09
[linux] Missing netlink message  lead to hotplug not triggering


fabiensanglard:
I am encountering an issue with hotplug on linux.

Context: We are not using `udev` because we don't know if our users will have it installed. Instead we compile libusb to use `netlink` for hotplug detection. The machine I am using for the repro steps is a ThinkStation P620 (documentation [here](https://download.lenovo.com/pccbbs/thinkcentre_pdf/p620_ug_en.pdf)).

The end-user problem is that if they unplug a device and replug it in within 10s, the host can no longer communicate with the device (transfer complete with `LIBUSB_TRANSFER_NO_DEVICE` and hotplug `LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED` is never called).  

As long as users use the front USB ports, they encounter no issue. Hotplug is working as expected. However if users use the back USB 3.X ports, there is a problem with hotplugging (weirdly, the USB 2.0 on the back are fine). Note that several users have reported the same issue when using these ports.

I have tracked down the problem to netlink events coming late (or not at all). When a user unplug a device from the problematic port, here is the netlink messages libusb gets:

```
[2763.143728] [002e0835] libusb: debug [dump_message] unbind@/devices/pci0000:20/0000:20:08.1/0000:23:00.3/usb2/2-1/2-1:1.0
[2763.143735] [002e0835] libusb: debug [dump_message] ACTION=unbind
[2763.143741] [002e0835] libusb: debug [dump_message] DEVPATH=/devices/pci0000:20/0000:20:08.1/0000:23:00.3/usb2/2-1/2-1:1.0
[2763.143747] [002e0835] libusb: debug [dump_message] SUBSYSTEM=usb
[2763.143753] [002e0835] libusb: debug [dump_message] DEVTYPE=usb_interface
[2763.143757] [002e0835] libusb: debug [dump_message] PRODUCT=18d1/4ee7/510
[2763.143762] [002e0835] libusb: debug [dump_message] TYPE=0/0/0
[2763.143769] [002e0835] libusb: debug [dump_message] INTERFACE=255/66/1
[2763.143774] [002e0835] libusb: debug [dump_message] SEQNUM=8528
.
. 9 seconds later
.
[2772.636012] [002e0835] libusb: debug [dump_message] remove@/devices/pci0000:20/0000:20:08.1/0000:23:00.3/usb2/2-1/2-1:1.0
[...]

PR #1577, opened 2024-10-18
Enable linux netlink event monitoring for Android OS platform services

### LINKED ISSUE #1578: Enabling hotplug for Android OS platform services & Solution Proposal
As described in the comments of the commit 2f3bc98b0d and issue: #1154 the services running on Android on OS-level (in contrast to apps) still have permissions to access netlink and listen to hotplug events. 

The proposed solution (PR #1577) involves including an additional macro: ANDROID_OS that can be used to allow this category of services to use libusb with netlink event monitoring (and consequently hotplugging) enabled.

The services should use the libusb target that sets the cflag "-DANDROID_OS" (to be added to libusb's [Android.bp](https://android.googlesource.com/platform/external/libusb/+/81e4c314ab828810a0ca5af3926b1d2a6c616e78/Android.bp) - WIP).

Radu-Vele:
Introducing the ANDROID_OS macro that is supposed to be set by clients
of libusb that are running on Android at the OS level.

If Android_OS is set (and HAVE_LIBUDEV is not) then linux netlink event
monitoring (and consequently hotplug functionality) is enabled. This
works for Android services or simple binaries running at the OS level as
they have the necessary permissions, in contrast to usual Android apps
(see commit https://github.com/libusb/libusb/commit/2f3bc98b0d0f4766496df53c855685a5f0e5e7cf).

Note: the \_\_ANDROID\_\_ macro is set by GCC for all targets running on
Android (both OS-level services and apps).

---

I noticed there have been previous discussions about this topic in the 2f3bc98b0d commit comments and  #1168.  

---

Closes #1578

### DIFF
--- a/libusb/os/linux_usbfs.h
+++ b/libusb/os/linux_usbfs.h
@@ -174,7 +174,17 @@ static inline int linux_start_event_monitor(void)
 {
 #if defined(HAVE_LIBUDEV)
 	return linux_udev_start_event_monitor();
-#elif !defined(__ANDROID__)
+/*
+* __ANDROID__: preprocessor macro defined automatically by GCC for all Android
+*              targets (i.e. both Android native applications, and Android OS-level
+*              services)
+*
+* ANDROID_OS: compilation flag that should be set for using libusb from programs
+*             running on Android at OS level (e.g. Android platform services).
+*             The programs using libusb built with the ANDROID_OS flag must have
+*             permission to access netlink sockets.
+*/
+#elif !defined(__ANDROID__) || defined(ANDROID_OS)
 	return linux_netlink_start_event_monitor();
 #else
 	return LIBUSB_SUCCESS;
@@ -185,7 +195,7 @@ static inline void linux_stop_event_monitor(void)
 {
 #if defined(HAVE_LIBUDEV)
 	linux_udev_stop_event_monitor();
-#elif !defined(__ANDROID__)
+#elif !defined(__ANDROID__) || defined(ANDROID_OS)
 	linux_netlink_stop_event_monitor();
 #endif
 }
@@ -194,7 +204,7 @@ static inline void linux_hotplug_poll(void)
 {
 #if defined(HAVE_LIBUDEV)
 	linux_udev_hotplug_poll();
-#elif !defined(__ANDROID__)
+#elif !defined(__ANDROID__) || defined(ANDROID_OS)
 	linux_netlink_hotplug_poll();
 #endif
 }


fabiensanglard:
Is seems the problem you are trying to solve is that there are two users of `libusb` on Android.

- App: which don't have netlink access and for which registering for hotplug events will fails since they can't access netlink.
- Non-App (a.k.a Android OS platform service such as `adb` or any service running as a non-app uid) which can access netlink and should be able to register for hotplug events.

1. It seems another way to fix would be to introduce a second macro `ANDROID_OS` (and clarify that `ANDROID` actually means apps) and change hotplug system behave accordingly.

2. You shold create an issue where you describe the problem and how you want to solve it so  libusb reviewers can be more efficient at discussing the approach and reviewing this PR.

Radu-Vele:
Hi @tormodvolden,
Could you please take a look at our proposal when you get the time?

tormodvolden:
Is "ANDROID_OS" a commonly used macro for this purpose? Otherwise this needs documentation, we shouldn't just introduce a macro in the code without explanation.

Radu-Vele:
The  "ANDROID_OS" macro is introduced now to cover the use case of libusb at the OS level, so it's not commonly used yet.

Do you think it's suitable to add some explanatory comments in the linux_usbfs.h file where the ANDROID and ANDROID_OS macros are used or did you have a different place in mind?

Also, we intend to document the purpose of the macro in the libusb build file (Android.bp) from Android Open Source Platform code. 

Thank you!

fabiensanglard:
>  Do you think it's suitable to add some explanatory comments in the linux_usbfs.h

Since it is used there exclusively, it could be documented there. I could also be an opportunity to start documenting compilation flag. e.g: I could not really figure out the advantage of `HAVE_LIBUDEV` from the documentation (I had to ask around: https://github.com/libusb/libusb/issues/1571#issuecomment-2427589954).

> we intend to document the purpose of the macro in the libusb build file (Android.bp) from Android Open Source Platform code.

Just to clarify, the source of truth will be`libusb` and the macro will be extensively documented in `libusb`. The AOSP documentation will be a "plus" for aosp developers to understand the context right :) ?

Radu-Vele:
> Just to clarify, the source of truth will be`libusb` and the macro will be extensively documented in `libusb`. The AOSP documentation will be an "extra" for aosp developers to understand the build system context right :) ?

Yes indeed the source of truth will be `libusb`, thank you for clarifying. 

I added the explanatory comments in the linux_usbfs.h file at the moment. Please let me know what you think.

I am also happy to help with the creation of a dedicated page that explains the meaning of the compilation flags if you think that is more suitable, @tormodvolden.

hjelmn:
LGTM. Will merge.

Issue #1581, opened 2024-11-08
LIBUSB_TRANSFER_ADD_ZERO_PACKET doesn't work on Darwin


fabiensanglard:
The [LIBUSB_TRANSFER_ADD_ZERO_PACKET](https://libusb.sourceforge.io/api-1.0/group__libusb__asyncio.html#gga1fb47dd0f7c209b60a3609ff0c03d56da26b66334b6ec0537c49841ca623d901f) flag does not work as expected on Darwin. On my USB 3 device, with maxPacketSize = 1024, I see `libusb` sending ZLP for packets multiple of 5120 instead of 1024. Notice that 5120 =  1024 * 5.

This results in bulk transfers hanging on the device receiving the transfer because ZLP are not sent when they should.

I tracked the reason to an incorrect retrieval of maxPacketSize in `darwin_usb.c`. We use [GetPipePropertiesV3](https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/os/darwin_usb.c#L266) but this function retrieves a maxPacketSize premultiplied by maxBurst+1 ([documentation](https://lab.qaq.wiki/Lakr233/IOKit-deploy/-/blob/master/IOKit/usb/IOUSBLib.h#L3991)).

```
FULL maxPacketSize for this pipe, which includes both the mult and the burst
```

I see three ways to fix this issue.

1. Call [GetEndpointPropertiesV3](https://lab.qaq.wiki/Lakr233/IOKit-deploy/-/blob/master/IOKit/usb/IOUSBLib.h#L4004) in [darwin_get_pipe_properties](https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/os/darwin_usb.c#L260) 
 just after [GetPipePropertiesV3](https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/os/darwin_usb.c#L266) with the same `pipe_properties` parameter. I have verified that `wMaxPacketSize` receives the proper value (`GetEndpointPropertiesV3` documentation explicitly state that it is what we need:
 
```
The wMaxPacketSize is the BASE maxPacketSize as found in the endpoint descriptor, and has not been multiplied to take into account burst and mult.
```

[...]

PR #1582, opened 2024-11-10
Fix Zero-Length Packet on Macos

fabiensanglard:
The ZLPs are not being sent when they should because we retrieve
    and incorrect value for wMaxPacketSize via GetPipePropertiesV3.
    GetEndpointPropertiesV3 is what we need.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -262,8 +262,16 @@ static IOReturn darwin_get_pipe_properties(struct darwin_interface *cInterface, 
 
 #if (MAX_INTERFACE_VERSION >= 550)
   if (get_interface_interface_version() >= 550) {
+    // GetPipePropertiesV3 returns a "cooked" wMaxPacketSize (premultiplied by burst and mul). This not what we want.
+    // We call GetPipePropertiesV3 to fill the fields needed to call GetEndpointPropertiesV3.
     IOUSBEndpointProperties pipe_properties = {.bVersion = kUSBEndpointPropertiesVersion3};
     kresult = (*IOINTERFACE_V(cInterface, 550))->GetPipePropertiesV3 (IOINTERFACE(cInterface), pipe, &pipe_properties);
+    if (kIOReturnSuccess != kresult) {
+        return kresult;
+    }
+    
+    // GetEndpointPropertiesV3 returns the wMaxPacketSize without burst and mul multipliers.
+    kresult = (*IOINTERFACE_V(cInterface, 550))->GetEndpointPropertiesV3 (IOINTERFACE(cInterface), &pipe_properties);
     if (kIOReturnSuccess == kresult) {
       out->number = pipe_properties.bEndpointNumber;
       out->direction = pipe_properties.bDirection;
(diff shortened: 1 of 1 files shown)


seanm:
Look reasonable, though I'm no expert in this area.

Small typo in commit message: "and incorrect value" -> "an incorrect value"

Personally, I would also put more of the info from https://github.com/libusb/libusb/issues/1581 into the commit message itself. (git commit history will outlive github I'd wager.)

fabiensanglard:
- Fixed typo in commit message.
- Elaborated on the commit message to describe error better.

tormodvolden:
I hope there is no other code relying on darwin_get_pipe_properties() returning the "cooked" values (multiplied by burst and mul)?

fabiensanglard:
I checked that before uploading this PR. The method `darwin_get_pipe_properties` is called in three locations.

1. `submit_bulk_transfer`. `wMaxPacketSize` is used there and we are fixing it.
2. `submit_iso_transfer`. Only uses field `interval`.
3. `get_endpoints`. Only uses `direction` and `number`.

The cooked value was only used `submit_bulk_transfer` so I don't think this is breaking anything else.

tormodvolden:
Thanks a lot for checking and confirming this.

tormodvolden:
Do you know what the pre-550 interface version would return in this case? (If it supports USB3, I dunno). Just so that darwin_get_pipe_properties() would be consistent over all versions.

tormodvolden:
I just wonder if it is by convention that /pipe/ properties include the multipliers whereas /endpoint/ properties don't. And the name of our (internal) function darwin_get_pipe_properties() could be misleading in this case. But you have added comments about this so it should be OK.

fabiensanglard:
> Do you know what the pre-550 interface version would return in this case? (If it supports USB3, I dunno). Just so that darwin_get_pipe_properties() would be consistent over all versions.

Good call. According to its doc, function `GetPipeProperties` [returns](https://lab.qaq.wiki/Lakr233/IOKit-deploy/-/blob/master/IOKit/usb/IOUSBLib.h#L2412) the FULL maxPacket size. So it does support USB3. I can add a call to `GetEndpointProperties` to fix it there too.

That being said  [iousbdeviceinterface500](https://developer.apple.com/documentation/iokit/iousbdeviceinterface500), was included with `macOS 10.7.3`, which was [release in 2012](https://apple.fandom.com/wiki/Mac_OS_X_10.7.3#:~:text=3%20is%20an%20update%20to,Apple%20on%20February%201%2C%202012.). 

1. Do we want to support a 13 years old OS?
2. Regardless if we want to support it or not, the behavior should be consistent. So either we delete the fallback call to `GetPipeProperties` or we fix it with a subsequent call to `GetEndpointProperties`.

tormodvolden:
1. If the effort is minimal, I'd say yes. If we have larger problems we can ditch it. We are not Apple :) The percentage may be much higher in labs etc where libusb is in use, compared to the general consumer market.
2. Maybe leaving darwin_get_pipe_properties() to return the full "pipe" property is cleaner, and rather do the division in the caller, as was also suggested in the bug report?

tormodvolden:
Sounds good. If in the future some function will need the cooked value, we can e.g. add a full_max_packet_size to struct darwin_pipe_properties.

PR #1584, opened 2024-11-18
core: add new option to disable reap after usb is disconnected

BluerabbitY:
linux:

On some Linux systems (such as Linux 6.5 and Linux 6.8), after unplugging some USB devices, the execution time of the usb_disconnect() function in the kernel is too long, causing competition between the thread where handle_event() is located and the thread in the kernel, resulting in handle_event() thread blocking.

This option will disable Reap operation after disconnection, even if the system has the capability. This method is used to avoid the problem of competition between the kernel thread and the thread where the handle_event() related function is located after the device is unplugged, resulting in the loss of USB data that is still online in some millisecond level communication situations.

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -119,6 +119,8 @@ struct config_descriptor {
 struct linux_context_priv {
 	/* no enumeration or hot-plug detection */
 	int no_device_discovery;
+	/* no reap data after disconnection */
+	int no_reap_after_disconnect;
 };
 
 struct linux_device_priv {
@@ -441,11 +443,15 @@ static int op_set_option(struct libusb_context *ctx, enum libusb_option option, 
 {
 	UNUSED(ap);
 
-	if (option == LIBUSB_OPTION_NO_DEVICE_DISCOVERY) {
-		struct linux_context_priv *cpriv = usbi_get_context_priv(ctx);
+	struct linux_context_priv *cpriv = usbi_get_context_priv(ctx);
 
+	if (option == LIBUSB_OPTION_NO_DEVICE_DISCOVERY) {
 		usbi_dbg(ctx, "no device discovery will be performed");
 		cpriv->no_device_discovery = 1;
+		return LIBUSB_SUCCESS;
+	} else if (option == LIBUSB_OPTION_NO_REAP_AFTER_DISCONNECT) {
+		usbi_dbg(ctx, "no reap after disconnection will be performed");
+		cpriv->no_reap_after_disconnect = 1;
 		return LIBUSB_SUCCESS;
 	}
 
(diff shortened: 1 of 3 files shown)


libusb/os/linux_usbfs.c
```diff
@@ -441,12 +443,16 @@ static int op_set_option(struct libusb_context *ctx, enum libusb_option option,
...
 
-	if (option == LIBUSB_OPTION_NO_DEVICE_DISCOVERY) {
-		struct linux_context_priv *cpriv = usbi_get_context_priv(ctx);
+	struct linux_context_priv *cpriv = usbi_get_context_priv(ctx);
 
+	if (option == LIBUSB_OPTION_NO_DEVICE_DISCOVERY) {
 		usbi_dbg(ctx, "no device discovery will be performed");
 		cpriv->no_device_discovery = 1;
 		return LIBUSB_SUCCESS;
+	} else {
```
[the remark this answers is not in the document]
BluerabbitY: This is my oversight. Thanks for pointing it out. I have corrected it.

libusb/os/linux_usbfs.c
```diff
@@ -119,6 +119,8 @@ struct config_descriptor {
 struct linux_context_priv {
 	/* no enumeration or hot-plug detection */
 	int no_device_discovery;
+	/* no reap data after disconnection */
+	int no_reap_after_disconnect;
```
seanm, line 123: Aren't these boolean?  (I realize the other was `int`, so you just probably copied it...)

BluerabbitY: Yes, these are boolean, I don't know if I could include the "stdbool.h", so I follow the previous code.

tormodvolden:
I don't think we want to introduce a new option to workaround an issue in some Linux kernel versions. But if this issue indicates that we have a race condition (or deadlock?) that can show up in various kernels (with USBFS_CAP_REAP_AFTER_DISCONNECT?) or certain situations we should absolutely try to fix it. Do you see a way that this can be sorted out?

BluerabbitY:
> I don't think we want to introduce a new option to workaround an issue in some Linux kernel versions. But if this issue indicates that we have a race condition (or deadlock?) that can show up in various kernels (with USBFS_CAP_REAP_AFTER_DISCONNECT?) or certain situations we should absolutely try to fix it. Do you see a way that this can be sorted out?

During subsequent investigations, it was discovered that there is no way to completely solve this problem in libusb (user space). Whenever a USB-related system call function is invoked and usb_lock_device() is called within it, blocking due to waiting for the lock may occur.

The currently known system call functions that may be blocked are `open()/close()/ioctl()`.

The main reason for blocking is that when the USB is unplugged, the Linux system interrupt will occupy the lock, causing the user space system call to block and not return.

Currently, I have a preliminary solution: The locking mechanism of the ioctl function in the kernel has been modified to ensure that ioctl is non-blocking. Based on this change, added O_NONBLOCK to the open flag in libusb.

``` C
// in file drivers/usb/core/devio.c usbdev_do_ioctl()
if (file->f_flags & O_NONBLOCK)
{
	if (!usb_trylock_device(dev))
		return (connected(ps) ? -EAGAIN : -ENODEV);
}
else
{
	usb_lock_device(dev);
}

```

Additionally, `libusb_open()/libusb_close()` and `libusb_handle_events_timeout_completed()` need to run in separate threads. Testing is still ongoing, so I'm not yet certain if this will resolve the issue.

This change alters the kernel's behavior, so initially, I didn't want to apply it to all USB devices, but only to the necessary ones.

We are currently evaluating this change. If the kernel changes are accepted upstream, I will submit the corresponding changes to libusb.

b4549 (approved):

mcuee:
In this case, I think we should close this PR as it is not a libusb issue.

PR #1598, opened 2025-01-14
Handle size is 0 in usbi_reallocf() in libusbi.h

zkh8227:
Double Free Risk:
The original usbi_reallocf function does not handle the case where size is 0 correctly. When realloc is called with a size of 0, it is equivalent to calling free(ptr), and realloc returns NULL. If the function then checks !ret and calls free(ptr) again, it results in a double free error, which can lead to undefined behavior and potential crashes.

[Tormod added]
Note this can't happen with the current code, because usbi_reallocf() is only used in one place (in Linux-only code), and _size_ will never be zero here.

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -263,7 +263,7 @@ static inline void *usbi_reallocf(void *ptr, size_t size)
 {
 	void *ret = realloc(ptr, size);
 
-	if (!ret)
+	if (!ret && size != 0)
 		free(ptr);
 	return ret;
 }


Youw:
> When realloc is called with a size of 0, it is equivalent to calling free(ptr), and realloc returns NULL.

That is not true: https://en.cppreference.com/w/c/memory/realloc

> until C23:
> if `new_size` is zero, the behavior is *implementation defined* (null pointer may be returned (in which case ***the old memory block may or may not be freed***), or some non-null pointer may be returned that may not be used to access storage).

> since C23:
> if `new_size` is zero, *the behavior is undefined*.

---
So depending on the compiler the "before" or "after" behavior might be correct.
But you've found a good case.

Taking into account that:
> If there is not enough memory, the old memory block is not freed and null pointer is returned.

I believe the correct way of fixing this would be:
```c
void *ret = NULL;

if (size != 0) {
    ret = realloc(ptr, size);
    if (ret != NULL)
        free(ptr);
}
else {
    free(ptr);
}

return ret;
```

seanm:
`realloc` is part of C of course, but `reallocf` is from the BSDs. It might be sensible for libusb to copy its implementation from say FreeBSD: https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdlib/reallocf.c  We could also look up the other implementations in OpenBSD, etc.

seanm:
So @zkh8227's change basically makes libusb match FreeBSD, which is probably a good idea IMHO.  But: since C23 makes realloc(0-size) undefined, I would add an `assert(size != 0)` at the beginning. So in debug we'll catch (incorrectly) doing this, but in release we'll behave as FreeBSD.

I think what I'd suggest is

Youw:
> makes libusb match FreeBSD, which is probably a good idea

It is not.

FreeBSD implementation of `reallocf` (e.g. [here](https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdlib/reallocf.c)) most likely depends on the behavior of FreeBSD's implementation of `realloc` which is only one of the possible cases as per C standard. And libusb (at least that part of the code) is being compiled on a veriety of systems using a veriety of compilers.

Youw:
One more thing: `usbi_reallocf` is used only in like two places, and I don't see how `0` could even be a possible argument for the `size`. Where did you find a code path where you'd hit the issue you've described?

seanm:
> FreeBSD implementation of `reallocf` (e.g. [here](https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdlib/reallocf.c)) most likely depends on the behavior of FreeBSD's implementation of `realloc` which is only one of the possible cases as per C standard

reallocf() came from FreeBSD originally, so I'd wager everyone copied their implementation in order for the behaivour to be identical.  The linux libbsd version is the same for example:

https://cgit.freedesktop.org/libbsd/tree/src/reallocf.c?h=main

The macOS version is also the same:

https://github.com/apple-open-source-mirror/Libc/blob/5e566be7a7047360adfb35ffc44c6a019a854bea/stdlib/FreeBSD/reallocf.c#L33

So making the libusb version the same also still seems quite advisable to me...

Youw:
> So making the libusb version the same also still seems quite advisable to me...

I don't see anything advisable here:
1) Not unless someone checks the behavior with GCC, Clang, MinGW and MSVC.
2) https://github.com/libusb/libusb/pull/1598#issuecomment-2590894908

seanm:
>One more thing: usbi_reallocf is used only in like two places

In fact it looks like just 1 place, in linux_usbfs.c.

>and I don't see how 0 could even be a possible argument for the size

Agreed. Unless that loop overflows and wraps around, and you'd hit `LIBUSB_ERROR_NO_MEM` before that.

>I don't see anything advisable here:

The FreeBSD, OpenBSD, linux, and macOS implementations of `reallocf` are all identical, and you think libusb's `usbi_reallocf` should be different? Can you explain why?

Since there's only 1 use of this private function, we can reason as to the correctness of this change, and it seems to me that the change would essentially change nothing.

But better that `usbi_reallocf` behave ***identically*** to the function it's trying to mimic.

seanm:
From its name, `usbi_reallocf` is clearly trying to re-implement `reallocf`, agreed?  Assuming so, how shall we do so?

1. copy exactly what FreeBSD, OpenBSD, linux, and macOS do (which are all the identical)
2. do something *subtly* different from 1

Yes, Window's realloc() is maybe slightly different.  But how does that make choice 2 better than choice 1?!

Youw:
> is clearly trying to re-implement `reallocf`, agreed?

For starters - why does `reallocf` even exists? Because the behavior of  `realloc` is not well defined, so someone at some point come up with `reallocf` - as a *safer* and more clear version of `realloc`.

But unfortunately `reallocf` is not available on all of the platforms, so YES, `usbi_reallocf` suggested as an implementation of `reallocf` which is available to libusb on all platforms.

But it has to re-implement a safer **_behavior_**, not the implementation. And there is no point to copy a well known implementation that works on *some* platforms, if it is _not going to work_ on **all platforms**.

seanm:
> For starters - why does `reallocf` even exists?

It exists because common usage patterns of realloc can lead to memory leaks, specifically with `p = realloc(p, ...)`. clang-tidy can warn about it: 

https://clang.llvm.org/extra/clang-tidy/checks/bugprone/suspicious-realloc-usage.html

Notably, this is the exact pattern in linux_usbfs.c

> Because the behavior of `realloc` is not well defined

Sure it is. It's defined by the ISO C spec. Microsoft also defines their behaviour with zero size, but zero size has always been implemention-defined (or now undefined) behaviour. So MS can do whatever they want when size is zero, and so can anyone else. It's never been portable to pass zero size.

> But unfortunately `reallocf` is not available on all of the platforms, so YES, `usbi_reallocf` suggested as an implementation of `reallocf` which is available to libusb on all platforms.

Exactly.

> But it has to re-implement a safer **_behavior_**, not the implementation. 

What do you mean ***a*** safer behaviour?  reallocf isn't part of an ISO standard, but it does have a documented behaviour. You can't just implement a `reallocf()` that does random nice things. It's documented as "The reallocf() function is identical to the realloc() function, except that it will free the passed pointer when the requested memory cannot be allocated."  Indeed the `f` at the end means "***f***rees upon failure".

>And there is no point to copy a well known implementation that works on _some_ platforms, if it is _not going to work_ on **all platforms**.

Why wouldn't the canonical implementation work on Windows?

In fact, looking at Microsoft realloc docs, they say "If size is zero, then the block pointed to by memblock is freed; the return value is NULL, and memblock is left pointing at a freed block."

So today's code:

```c
static inline void *usbi_reallocf(void *ptr, size_t size)
{
	void *ret = realloc(ptr, size);

	if (!ret)
		free(ptr);
	return ret;
}
```

Means `usbi_reallocf(p, 0)` can perform a double-free on Windows. Which is presumable exactly why the canonical implementation also has the `&& size != 0` part that this PR adds.

tormodvolden:
Great discussion guys! So to summarize, is this correct? The code never calls usbi_reallocf() with 0, but if it would, this could cause a double-free on Windows?

seanm:
>So to summarize, is this correct?

I think so.

>The code never calls usbi_reallocf() with 0, but if it would, this could cause a double-free on Windows?

Correct.

On *nix, it makes the libusb implementation match the native implementations.

Youw:
Right. @tormodvolden I vote to remove `usbi_reallocf` from libusb altogether and stop trying to fix something that is not needed.

seanm:
It's not ***needed***, but it's beneficial.

If we remove it, this code:

```c
priv->descriptors = usbi_reallocf(priv->descriptors, alloc_len);
if (!priv->descriptors) {
	if (fd != wrapped_fd)
		close(fd);
	return LIBUSB_ERROR_NO_MEM;
}
```

would become harder to read and more prone to error:

```c
void *temp_descriptors = realloc(priv->descriptors, alloc_len);
if (!temp_descriptors) {
	free(priv->descriptors); // This necessary free() is the "f" in reallocf
	if (fd != wrapped_fd)
		close(fd);
	return LIBUSB_ERROR_NO_MEM;
}
else {
	priv->descriptors = temp_descriptors;
}
```

seanm (approved):

tormodvolden:
I am fine with the change, but can somebody please fix the commit message, for instance include the note I added in the PR description?

mcuee:
@zkh8227 

Would you please change the commit message as per the comment by @tormodvolden? Thanks.

seanm:
@sonatique this small safe change could also be merged ASAP.

mcuee:
@sonatique 

It will be good to merge this PR.

Right now we have too many PRs pending. 

After 1.0.30 release, I'd really like to see the number of pending PR drops to 50 first and then to something more manageable like 25. Same for open issues, I'd really like to see the number of open issues to drop to 100 first and later to something like 50.

seanm:
> It will be good to merge this PR.

Yes, this is extremely low risk and could be merged immediately.

tormodvolden:
@zkh8227 Do you have a name we can attribute this contribution to?

tormodvolden:
I fixed up the commit message and rebased to latest master.

tormodvolden:
Rereading the discussion, this change would actually only affect behavior on Windows (because other platforms are already safe against the hypothetical double-free)? If so the commit message is still not good.

seanm:
>@zkh8227 Do you have a name we can attribute this contribution to?

I'd surmise that after 1.5 years of beating this one-line dead horse to death, he no longer cares.

>If so the commit message is still not good. 

Why not?  The commit message I think is quite good. It even talks about "potential" issues in "future code", which is quite right. i.e. it's a defensive change to improve correctness.

tormodvolden:
Thanks, I think it got a bit better at least :) I just wondered if the whole "If realloc is called with a size of 0, it is equivalent to calling free(ptr), and realloc returns NULL." is only true for Windows? And the Linux comment that I had added, it is not only because the Linux code always call it with non-zero size, but also because the original code wouldn't cause double-free with zero size on Linux anyway.

PR #1606, opened 2025-02-18
core: Add block device, and COM port matching on Linux

### LINKED ISSUE #1575: Block device, and COM port matching.
I want to add support for finding the block device path or serial port path from a `libusb_device` struct.
I'm wondering the logistics of adding these functions. I want to add support for Linux, Mac OSX, and Windows, but I don't think i can add support for other operating systems. I also will add Linux first as I am am familiar with it. 

If I was going to add this feature do I need to add it to all platforms.
And I'm wondering thoughts on if this is a good feature.

Another library https://github.com/pololu/libusbp has a similar feature.

Mr-Bossman:
Adds libusb_get_blockdev_path and libusb_get_chardev_path. These functions allow the user to get the block device and COM port paths for a USB device. This is useful for identifying the device in a system.

Closes #1575


libusb/os/linux_usbfs.c
```diff
@@ -1407,6 +1408,129 @@ static int op_wrap_sys_device(struct libusb_context *ctx,
...
+	int ret;
+
+	ret = asprintf(&path, "%s/subsystem", sys_path);
+	if (ret < 0)
+		return LIBUSB_ERROR_NO_MEM;
+
+	subsystem_path = realpath(path, NULL);
+	free(path);
+	if (!subsystem_path && errno == ENOMEM)
+		return LIBUSB_ERROR_NO_MEM;
```
seanm, line 1423: If `subsystem_path` is NULL but errno is something else, shouldn't you be returning a libusb error here?

libusb/os/linux_usbfs.c
```diff
@@ -1407,6 +1408,129 @@ static int op_wrap_sys_device(struct libusb_context *ctx,
...
+static int get_subsytem(struct libusb_context *ctx, char *buf, size_t len,
+	const char *dir, const char* subsystem, int depth)
+{
+	DIR *dp;
+	struct dirent *entry;
+	struct stat statbuf;
+	char *path;
+	int ret;
+
+	if (depth >= 20)
```
seanm, line 1442: What is this magic 20? A comment would be good.

libusb/os/linux_usbfs.c
```diff
@@ -1407,6 +1408,129 @@ static int op_wrap_sys_device(struct libusb_context *ctx,
...
+		return LIBUSB_ERROR_NOT_FOUND;
+
+	ret = check_subsystem(dir, subsystem);
+	if (ret < 0)
+		return ret;
+
+	if (!ret) {
+		path = strrchr(dir, '/');
+		if(path) {
+			snprintf(buf, len, "/dev/%s", path + 1);
```
seanm, line 1452: Maybe I need more coffee, but if `/` is the very last character in the string, `path + 1` is pointing to the null terminator, right?

seanm (changes requested):

libusb/os/linux_usbfs.c
Mr-Bossman: `strrchr(dir, '/')` will be `/sde` as in it incldes the `/` if its not found it will be `NULL`.

Oh im dumb i can just do
`snprintf(buf, len, "/dev%s", path);` anyway. i was returning `sde` but changed it oops

seanm: That formulation is indeed better.

But what I was getting at was: what if (somehow) `/` is the *very last* character in `dir`?  Then `buf` will become `/dev/` and you'll return that. But that's a bogus result. Better to fail.  I think you need to also check the `strrchr` result (`path`) not just for NULL but `strlen(path) > 1`.

libusb/os/linux_usbfs.c
seanm: I don't think you addressed this one...

libusb/core.c
```diff
@@ -905,6 +907,76 @@ void API_EXPORTED libusb_free_device_list(libusb_device **list,
...
+/** \ingroup libusb_misc
+ * Get the block device path of USB resource.
+ * A string that contains a block device that is associated
+ * with the USB resource using the active config descriptor.
+ * An example path on *nix is `/dev/sdaX`
+ *
+ * \param dev a device
+ * \param iface_idx the <tt>bInterfaceNumber</tt> of the interface you wish to probe
+ * \param path output buffer for the USB block device path
+ * \param path_length size of path buffer
```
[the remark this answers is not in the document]
Mr-Bossman: I can make it allocate memory if you think that is better. Also, should I add the function to `/libusb/libusb-1.0.def` as I will be adding window support soon?

libusb/os/linux_usbfs.c
Mr-Bossman: It should be `errno != ENOENT` as we check all paths even if there is no `./subsystem` directory which would mean it doesn't have a subsystem.

libusb/os/linux_usbfs.c
Mr-Bossman: `dir` can not end in a `/` as it comes from `1469: ret = asprintf(&path, "%s/%s", dir, entry->d_name);`. `path` in this case will always be the caller's `/` + `entry->d_name`.

seanm: That's where it comes from today. But the correctness of the `check_subsystem` function should stand on its own. Imagine some future refactoring from someone that doesn't have the flow fresh in his mind and calls `check_subsystem` from some place new (hence the "(somehow)" in my previous comment).  Or imagine libusb someday having great testing where a fuzzer is generating crazy inputs.

libusb/os/linux_usbfs.c
seanm: If (somehow!) `realpath` returns NULL *and* sets `errno=ENOENT`, you will return the result of `asprintf()` (the number of characters printed). Is that what we want?

libusb/core.c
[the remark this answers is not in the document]
seanm: I do think it's better, but maybe that's just me. What do you think?  As a user of this API, how would you decide how big a buffer to make?  If you decide too small, you have to allocate bigger and then try again?  If you decide too big, you've wasted memory. Seems suboptimal to me.  I guess the only downside is that the caller will have to remember to call free(), vs maybe being able to get away with a small buffer on the stack.

But I don't have merge permission in this project, so I don't get to decide either. :)

I don't know much about Windows. I don't know what libusb-1.0.def is for example.

libusb/os/linux_usbfs.c
Mr-Bossman: no, but happened to work as a non-match would be a positive number.

libusb/core.c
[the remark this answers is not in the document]
Mr-Bossman: ok I changed it to allocate memory.

tormodvolden:
As posted in the corresponding issue, I am not sure we want to add this to libusb. I guess we should continue that discussion there. But thanks for submitting the code (and for reviewing it Sean) in any case, even if we don't merge it can still be useful for someone else.

PR #1607, opened 2025-02-18
CI: Remove build job for MSYS clang32

Mr-Bossman:
Upsream dropped support for clang32.
Remove the build job for it.
https://github.com/msys2/setup-msys2/pull/458


Mr-Bossman:
@mcuee any idea why macos build is failing now?

seanm:
IIRC it fails intermittently

mcuee:
> IIRC it fails intermittently

Indeed this happens from time to time. macOS test will fail from time to time. Strange.

tormodvolden:
Thanks for letting us know. But if they just dropped it in their dev branch, we are maybe not in a rush to delete it here? I suppose our CI builder is not using a MSYS dev snapshot? I would just leave it in until it doesn't work or causes other problems. Who knows, maybe it catches some errors that our other builders don't.

mcuee:
@tormodvolden 

FYI, msys2 has already removed 32bit build of libusb quite some time ago (early 2024).
* https://github.com/avrdudes/avrdude/pull/1687

I know this is kind of a different thing than this PR. But just want to give a data point.

mcuee:
@tormodvolden 

We should merge now since github action already failed.
https://github.com/libusb/libusb/actions/runs/13498133171/job/37709914785?pr=1610

>Run msys2/setup-msys2@v2
> Error: 'msystem' needs to be one of MSYS, MINGW32, MINGW64, UCRT64, CLANG64, CLANGARM64, got clang32

tormodvolden:
Thanks! It seems like we are actually pulling in latest setup-msys2 main branch on our CI.

PR #1625, opened 2025-03-20
Do not enumerate devices not initialized by udev

benzea:
Doing that can cause weird issues with permissions. One example is fprintd trying to open the fingerprint reader but selinux has not yet been tagged and it is not allowed to do so. Other scenarios can also be possible as it is not uncommon to change the user/group of devices using udev.

This should fix it, unfortunately the umockdev testing is not really optimal …

EDIT: Downstream bug report is https://bugzilla.redhat.com/show_bug.cgi?id=2346771

### DIFF
--- a/tests/umockdev.c
+++ b/tests/umockdev.c
@@ -409,6 +409,8 @@ test_fixture_add_canon(UMockdevTestbedFixture * fixture)
 		"E: DRIVER=usb\n"
 		"E: BUSNUM=001\n"
 		"E: DEVNUM=001\n"
+		"E: MAJOR=189\n"
+		"E: MINOR=128\n"
 		"E: DEVNAME=/dev/bus/usb/001/001\n"
 		"E: DEVTYPE=usb_device\n"
 		"A: bConfigurationValue=1\\n\n"
@@ -429,8 +431,7 @@ static void
 test_fixture_setup_libusb(UMockdevTestbedFixture * fixture, int devcount)
 {
 	libusb_device **devs = NULL;
-
-	libusb_init_context(/*ctx=*/&fixture->ctx, /*options=*/NULL, /*num_options=*/0);
+	struct timeval zero_tv = { 0 };
 
 	/* Suppress global log messages completely
 	 * (though, in some tests it might be interesting to check there are no real ones).
@@ -440,6 +441,11 @@ test_fixture_setup_libusb(UMockdevTestbedFixture * fixture, int devcount)
 	g_assert_cmpint(libusb_get_device_list(fixture->ctx, &devs), ==, devcount);
 	libusb_free_device_list(devs, TRUE);
 	libusb_set_log_cb (fixture->ctx, log_handler, LIBUSB_LOG_CB_CONTEXT);
+
+	/* Get the udev "add" event out of the way, part of the workaround that
+	 * umockdev does no create udev DB files.
+	 */
+	libusb_handle_events_timeout(fixture->ctx, &zero_tv);
 }
 
 static void
@@ -464,6 +470,8 @@ test_fixture_setup_empty(UMockdevTestbedFixture * fixture, UNUSED_DATA)
 {
 	test_fixture_setup_common(fixture);
 
+	libusb_init_context(/*ctx=*/&fixture->ctx, /*options=*/NULL, /*num_options=*/0);
+
 	test_fixture_setup_libusb(fixture, 0);
 }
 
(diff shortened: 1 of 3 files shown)


tests/umockdev.c
```diff
@@ -1110,6 +1135,37 @@ test_hotplug_add_remove(UMockdevTestbedFixture * fixture, UNUSED_DATA)
 	libusb_hotplug_deregister_callback(fixture->ctx, handle_remove);
 }
 
+static void
+test_hotplug_udev_not_ready(UMockdevTestbedFixture * fixture, UNUSED_DATA)
```
benzea, line 1139: Dunno, a test is nice, but then maybe having one that potentially breaks if umockdev is fixed isn't …

I opened https://github.com/martinpitt/umockdev/issues/267

3v1n0: > Dunno, a test is nice, but then maybe having one that potentially breaks if umockdev is fixed isn't …
> 
> I opened https://github.com/martinpitt/umockdev/issues/267

Yeah... Probably it makes sense to keep it in another PR that lands when umockdev is fixed?

tests/umockdev.c
```diff
@@ -409,6 +409,8 @@ test_fixture_add_canon(UMockdevTestbedFixture * fixture)
 		"E: DRIVER=usb\n"
 		"E: BUSNUM=001\n"
 		"E: DEVNUM=001\n"
+		"E: MAJOR=189\n"
```
benzea, line 412: Just FYI, this is the correct thing here and it is needed because the `is_initialized` check in the enumerator is special and considers any device without a device number always as initialized (`udev_device_get_is_initialized` behaves differently there).

mcuee:
Looks like the intention of this PR is good. But I am not able to review the implementation properly.

benzea:
> Looks like the intention of this PR is good. But I am not able to review the implementation properly.

I don't think it is that complicated. Most of the PR is adjusting the tests, but maybe you can ignore that? Either way, please ping me if you need someone else to look over it. I think that it would be nice if this makes it into the next release.

tormodvolden:
This sounds like something we want, and I think I have heard about issues caused by such "half-ready" devices. It seems from the downstream report this have had minimal to none testing though. Have you patched libusb in RedHat with this already?

BTW, you're saying the device will now be detected later on the udev "add" event. So the enumerator will (originally) also return devices for which there has been no udev add event signaled yet? And the add event coincides with "initialized" becoming true?

And I am fine with bumping umockdev requirements to the 3 year old 17.7 (available in the oldest supported Ubuntu LTS release as one data point).

benzea:
> This sounds like something we want, and I think I have heard about issues caused by such "half-ready" devices. It seems from the downstream report this have had minimal to none testing though. Have you patched libusb in RedHat with this already?

I am not with Red Hat anymore. I suppose I could submit a PR for the Fedora libusb package and try getting it in there.

> BTW, you're saying the device will now be detected later on the udev "add" event. So the enumerator will (originally) also return devices for which there has been no udev add event signaled yet? And the add event coincides with "initialized" becoming true?

Yup, exactly. The `enumerator` looks at sysfs, so it sees all devices known to the kernel. However, once udev is done, it writes a file into `/run/udev/data/` and fires the `"add"` event to notify all processes.

What the change effectively does is to check whether the file in `/run/udev/data/` has already been written. If it does not exist, then we know that an `"add"` signal will arrive in the future.

benzea:
So, the bugfix seems to work well. However, we have a bug in libmtp which runs from inside a udev rule and then fails to open the device and detect it as an MTP device.

A workaround might be to inspect the environment variables (udev exports all device properties there, see PROGRAM section in man page). Then we can explicitly include this device in the enumeration list (and log a warning) even though udev is not yet done with it.

The proper fix is for all programs run from udev rules to use libusb_wrap_sys_device.

sonatique:
@benzea : I am not familiar with udev, but I understand that this PR by itself is not sufficient to cover all cases? Are you going to complement it with what you described in your last message?

benzea:
> @benzea : I am not familiar with udev, but I understand that this PR by itself is not sufficient to cover all cases? Are you going to complement it with what you described in your last message?

**The PR itself is correct**. The problem is that there is buggy software out there that relies on incorrect behaviour.

EDIT: And yes, in general I am planning to add a workaround. Not sure when I'll test that one out though.

mcuee:
Does this PR affects those who build libusb without udev support (eg: relying on netlink)?

benzea:
> Does this PR affects those who build libusb without udev support (eg: relying on netlink)?

No, the code is not used in that case.

benzea:
My attempt at a workaround for the libmtp mtp-probe binary behaviour. Haven't really tested it yet, but one can see the general approach. I added the checks for the other environment variables just to be a bit more sure that we are in a udev environment, but that is probably overkill.

PR #1636, opened 2025-05-01
xusb: Fix error reporting on interface claim

tormodvolden:
There was a mixup of "ret" and "r", so use "r" consistently as in the rest of the function.


tormodvolden:
This is a regression so it would be nice to squeeze it into 1.0.29.

mcuee (approved):
Looks good to me.

Issue #1637, opened 2025-05-01
[macOS] Changes to remove compiler warnings breaks macOS re-enumeration timeout.


pjpetree:
libusb-1.0.28 introduced some changes to fix some compiler warnings. With these changes, part of the calculation of the elapsed µs time (the nsec portion) was cast as an `unsigned long long`. This cast incorrectly promotes small negative values to huge positive values cause the enumeration timeout to expire too soon. This, ultimately, causes some slower devices to fail to enumeration consistently.

For example, a delta between 1.9 s and 2.1 s means the `delta_sec` value will be 1 s and the `delta_nsec` value should be -0.8 s. Adding these two values would (correctly) make the delta 0.2 s. 

The code, unfortunately, casts the `delta_nsec` to an unsigned value which makes the value very large. This means adding the two values creates an invalidly large elapsed time, which is larger than the timeout, and the device is not re-enumerated.

seanm:
I'll take a look...

tormodvolden:
Thanks for reporting and the analysis. Can we say something about how likely this bug will kick in? Will most devices be so fast that it won't go wrong? Then some safe and unsafe intervals for the slower ones.

pjpetree:
Seems like it would be proportional to how long the device takes to re-enumerate.

E.g., if the device takes 250 ms to re-enumerate, the issue will only happen if the starting `nsec` value is ≥ 750 ms.  Which, if I'm not mistaken, would mean a probability to hit the issue of 0.25.  A device that takes 650 ms would have a probability of 0.65.

tormodvolden:
Just wondering if this warrants an accelerated new release. The issue was sitting in git master for 10 months before getting detected and reported (in PR #1610)  though.

sonatique:
I vote yes for the accelerated release.

PR #1610, opened 2025-02-24
darwin: Fix time calculation to handle negative nanosecond deltas

HotIce0:
Fix time calculation to handle negative nanosecond deltas for device re-enumerate

In the original code, the calculation of elapsed time could produce incorrect results when the nanosecond difference (delta_nsec) was negative. This situation occurs when the current time's nanoseconds are less than the start time's nanoseconds. 

The fix introduces a check for negative delta_nsec. If delta_nsec is negative, we adjust delta_sec by decrementing it by 1 and adding NSEC_PER_SEC to delta_nsec to ensure the elapsed time is calculated correctly.

This change ensures accurate time measurement in all scenarios.

PR #1638, opened 2025-05-01
Fixed issue #1637: incorrect subtraction of `struct timespec` structs

seanm:


tormodvolden:
Great. This looks like a safer pattern.

tormodvolden:
Oops, I noticed only now that this change is part of #1610. I didn't realize at the time that there was a regression discovered in that submission. @seanm does it seem fine to you to apply that one instead?

seanm:
Ugh, I remembered that one, but thought this was another example of the same pattern, not the exact same one.

Yes, let's merge the other one.

(I do wish we would merge PRs faster in libusb... it didn't occur to me that something fixed months ago wasn't in master...)

tormodvolden:
That fix unfortunately got a bit lost in the hurry to make a release. If I had realized its impact and that it was fixing a regression, it would of course have been included.

Issue #1649, opened 2025-06-04
libusb failed to build under VS2022 17.14.x version


mcuee:
I got a new laptop (Windows 11 Enterpris 23H2) and installed VS2022 Community Edition and add latest Windows SDK and Windows WDK, somehow libusb failed to build.

VS2022 shows that it is fully up to date at version 17.14.4.

Installation process:
https://learn.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk

```
PS C:\work\libusb\libusb> msbuild -m -v:m -p:PlatformToolset=v143,Platform=x64,Configuration=Release .\msvc\libusb.sln
MSBuild version 17.14.10+8b8e13593 for .NET Framework

  core.c
  descriptor.c
  events_windows.c
  hotplug.c
  io.c
  strerror.c
  sync.c
  threads_windows.c
  windows_common.c
  windows_usbdk.c
C:\work\libusb\libusb\libusb\os\windows_winusb.c(2131,62): error C2220: the following warning is treated as an error [C:\work\libusb\libusb\msvc\libusb_sta
tic.vcxproj]
  windows_winusb.c
        r = libusb_control_transfer(dev_handle, LIBUSB_ENDPOINT_OUT |
                                                                    ^

C:\work\libusb\libusb\libusb\os\windows_winusb.c(2131,62): warning C5287: operands are different enum types 'libusb_endpoint_direction' and 'libusb_request
_type'; use an explicit cast to silence this warning [C:\work\libusb\libusb\msvc\libusb_static.vcxproj]
        r = libusb_control_transfer(dev_handle, LIBUSB_ENDPOINT_OUT |
                                                                    ^
      C:\work\libusb\libusb\libusb\os\windows_winusb.c(2131,62):
      to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
        r = libusb_control_transfer(dev_handle, LIBUSB_ENDPOINT_OUT |
                                                                    ^

  getopt.c
  core.c
  descriptor.c
  events_windows.c
  hotplug.c
  io.c
  strerror.c
  sync.c
  threads_windows.c
  windows_common.c
  windows_usbdk.c
  getopt1.c
  windows_winusb.c
  getopt.vcxproj -> C:\work\libusb\libusb\build\v143\x64\Release\getopt.lib
[...]

PR #1650, opened 2025-06-04
windows: address warning C5287 treated as an error by Visual Studio 17.14.x+ compiler

sonatique:
Relates to https://github.com/libusb/libusb/issues/1649#issuecomment-2940138443


tormodvolden:
It really hurts to uglify the code with all those casts. I guess in the future we should use plain macros instead of enums for these, but it might be an API break.

Can't we just add  /Wv:18 to the msvc project settings?

seanm:
That is pretty ugly.

Why is this treated as an error?  It's valid C the way it is now (AFAIK). Are you compiling as C++ maybe?

Also, why do you cast to `uint8_t`?  enums are by default `int`-sized.  Aside: this is why I'd really like to (conditionally) adopt C23 enum sizes: https://github.com/libusb/libusb/pull/1417  It helps make clear what size enums are expected to be.

Anyway, it ***is*** kinda weird how libusb is mixing different enums, so the C5287 warning is interesting.  Maybe some of those should not be enums at all, but just constants.

sonatique:
@seanm : I am not using anything else than compiling from Visual Studio, and I guess it is using plain C. 
The project is configured with "Warning Level 4" (/W4) and "Treat Warnings as Errors" /WX .

Warning C5287 does make sense to me. I would be unhappy to silence it.

I am casting to uint8_t because it is the type the function that use the result of the OR is expecting, and I see no point in first casting to int. 

As you can see in xusb.c lines 751 and 764 there was already a cast to uint8_t. . 

@tormodvolden : using /Wv18 seems a bit too radical to me, especially considering that this warning bring an interesting problem to the table. The way I handled it might not be the prettiest, but it is easily understandable and is ready "as is" without needing some more complicated refactoring or complex decision making.

Also please note that there is only 1 change in libusb file themselves (namely windows_winusb.c), all other changes are in examples. This seems acceptable to me.

seanm:
>The project is configured with ... "Treat Warnings as Errors" /WX .

Oh, well, that explains it. If I build libusb with clang's `-Werror -Weverything` I get errors too. :)  Doesn't mean they should all be fixed.

>I would be unhappy to silence it.

But it doesn't necessary follow that a bunch of casts is the right fix.

I'd argue we should look deeper at why different enums are being mixed.

Does that compiler support C23? If you size the enums like in https://github.com/libusb/libusb/pull/1417 does that fix the warnings?

>...especially considering that this warning bring an interesting problem to the table.

What problem? I don't think any bug has been fixed here. I doubt codegen would even change.

seanm:
Perhaps the `libusb_request_type` enum should get new members: to indicate in/out direction. i.e. add `LIBUSB_REQUEST_TYPE_DIRECTION_IN`.

Then, for example, `LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_ENDPOINT_IN` would become `LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_REQUEST_TYPE_DIRECTION_IN` and we'd be ORing two things from the ***same enum***.

mcuee:
> The project is configured with "Warning Level 4" (/W4) and "Treat Warnings as Errors" /WX .

One option is to remove `/WX` and let the maintainers decide whether to deal with the warnings or not. We anyway have some warnings in other build.

sonatique:
@seanm : As far as I can tell and try, VS doesn't support C23. 
I think that even if it would support it, it wouldn't fix the issue since the problem is not the underlying integer type vs another integer type, but the mismatch of different enum types. We are mixing apple and pear, it doesn't matter whether they have the same size ;-)

> What problem?

Well the problem is that nobody with a decently recent Visual Studio is able to compile libusb or the example anymore. 
I think that this is a pretty serious one . The "cast" solution might not be the one we want to keep for the long run, but it is safe and convenient and could be merged quickly. 

It doesn't prevent any subsequent efforts to improve things.

@mcuee I would vote against removing /WX. In my experience this opens the door to have thousand of warnings that nobody understand / take care of, hiding bugs, as seen in other projects.

mcuee:
> @mcuee I would vote against removing /WX. In my experience this opens the door to have thousand of warnings that nobody understand / take care of, hiding bugs, as seen in other projects.

Noted. Thanks for the explanation.

> Well the problem is that nobody with a decently recent Visual Studio is able to compile libusb or the example anymore.
I think that this is a pretty serious one . 

I agree.

seanm:
> We are mixing apple and pear, it doesn't matter whether they have the same size ;-)

Yes, I guess it wouldn't fix this particular warning. But adding casts means we are still mixing apple and pear, just forcing them to look alike. 

>Well the problem is that nobody with a decently recent Visual Studio is able to compile libusb or the example anymore.
I think that this is a pretty serious one

Wait, you mean /WX is turned on in the project that libusb provides?  If so, that should be turned off IMNSHO. It leads to exactly these kinds of problems. You can never know in the *present* what new warnings might exist in the *future*.

>In my experience this opens the door to have thousand of warnings that nobody understand / take care of, hiding bugs, as seen in other projects.

This does not have to be the case. Warnings can be shown in CI, and we can fix them.

As for fixing these warnings, what do you think of the idea to expand the enums as I described?

sonatique:
@seanm : yes /WX is turned on by project configuration for libusb_dll , libusb_static and all examples in current master.

Regarding your proposed solution: yes that is a way of doing it, but keep in mind that we'll have to incorporate "libusb_request_recipient" as well.

maybe the new enum should be name "libusb_request" and have members such as
LIBUSB_REQUEST_TYPE_XY
LIBUSB_ENDPOINT_DIRECTION_XY
LIBUSB_RECIPTION_XY

It this way really is the preferred way, then I can do it. I am just wondering why we do have those 3 enums todays, there must be a good reason for it...

tormodvolden:
This topic has a long-term and a short-term dimension. If there will be a trend for compilers (and C standards) to barf at mixed enums, we should think about changing the API to use a common enum, or use macro constants instead. For long an often-heard advice has been to use enumerations for numbers instead of numeral macros, but this obviously clashes here. The point of these definitions (and maybe libusb at large) is to make it easy to write USB code that is simple and readable, and e.g. compose the flags for e.g. libusb_control_transfer() in a straight-forward way, so it seems very little elegant to need to throw in casts everywhere. I see little issue in changing winusb.c or internal files, but it would be with sadness I see these changes to the examples, suggesting that this is the way libusb should be used. I hope we can find a back-compatible solution that doesn't require downstream users to change all their code. More on the technical side, if we make common enums we certainly break ABI but that is tolerable, and we can maybe alias the old enum names so that old code still works? Anyway I'd expect very few applications to actually reference the enums? We may want to make breaking changes to the API one day, but I don't think this is what is gonna force us over to libusb 1.1 or 2.0.

At short term, I totally agree to make this warning non-breaking in our default configuration, and on our CI. I am surprised about /WX being enabled, since we have since long had warnings in our Windows builds. But maybe we had lots of warnings on the MinGW builds but none in the msvc builds? Newer compilers bring up more warnings than we can handle, and I don't think we have to care about all of them, there are more and more false positives. We already enable some warnings and other not in configure.ac (some with -Werror) and we should do the same for msvc. Downstream projects are free to do what they want. We cannot commit to making libusb build cleanly on any set of compilers and compiler warning configurations.

seanm:
> @seanm : yes /WX is turned on by project configuration for libusb_dll , libusb_static and all examples in current master.

Ah, sorry, I had not understood that originally.

I definitely vote to turn that off.  Again, some future version of VS will have some new warning, and break things again.  I'm a big fan of fixing warnings, but treating them as errors I don't think is helpful.

> Regarding your proposed solution: yes that is a way of doing it, but keep in mind that we'll have to incorporate "libusb_request_recipient" as well.
> 
> maybe the new enum should be name "libusb_request" and have members such as LIBUSB_REQUEST_TYPE_XY LIBUSB_ENDPOINT_DIRECTION_XY LIBUSB_RECIPTION_XY
> 
> It this way really is the preferred way, then I can do it.

I wasn't proposing adding a new `enum`, but rather adding new enumerations to existing `enums`.

If you have time to make a draft PR for us to discuss, that would be great.  If you don't have time, I could do it too.

>I am just wondering why we do have those 3 enums todays, there must be a good reason for it...

Maybe not. :) Sometimes things are unplanned and just end up a certain way over time.

>More on the technical side, if we make common enums we certainly break ABI but that is tolerable

I don't see any ABI breakage. I'm just suggesting adding to enumerations to existing `enums`...

sonatique:
> we should think about changing the API to use a common enum

There won't be a change int the API actually, given that the relevant parameter of libusb_control_transfer, i.e.  btRequestType is simply uint8_t 

I'll draft something.

sonatique:
Proposal of merging enums: here: https://github.com/libusb/libusb/pull/1655

sonatique:
This PR is no longer relevant

PR #1652, opened 2025-06-07
ss_usb_cap: rename wU2devExitLat from bU2devExitLat

aokblast:
According to the naming convention of libusb, a 2 byte member shoud be "w" prefix instead of "b"

### DIFF
--- a/libusb/hotplug.c
+++ b/libusb/hotplug.c
@@ -19,7 +19,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
-#include "libusbi.h"
+// #include "libusbi.h"
 
 /**
  * @defgroup libusb_hotplug Device hotplug event notification
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -984,7 +984,7 @@ struct libusb_ss_usb_device_capability_descriptor {
 	uint8_t  bU1DevExitLat;
 
 	/** U2 Device Exit Latency. */
-	uint16_t bU2DevExitLat;
+	uint16_t wU2DevExitLat;
 };
 
 /** \ingroup libusb_desc
--- a/android/examples/unrooted_android.c
+++ b/android/examples/unrooted_android.c
@@ -128,7 +128,7 @@ static void print_ss_usb_cap(struct libusb_ss_usb_device_capability_descriptor *
     LOGD("      wSpeedSupported:       %u\n", ss_usb_cap->wSpeedSupported);
     LOGD("      bFunctionalitySupport: %u\n", ss_usb_cap->bFunctionalitySupport);
     LOGD("      bU1devExitLat:         %u\n", ss_usb_cap->bU1DevExitLat);
-    LOGD("      bU2devExitLat:         %u\n", ss_usb_cap->bU2DevExitLat);
+    LOGD("      wU2devExitLat:         %u\n", ss_usb_cap->wU2DevExitLat);
 }
 
 static void print_bos(libusb_device_handle *handle)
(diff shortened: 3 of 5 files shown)


libusb/libusb.h
```diff
@@ -984,7 +984,7 @@ struct libusb_ss_usb_device_capability_descriptor {
 	uint8_t  bU1DevExitLat;
 
 	/** U2 Device Exit Latency. */
-	uint16_t bU2DevExitLat;
```
tormodvolden, line 987: This would change the exposed API. Maybe we will do that at some point, but not in a heartbeat.

libusb/hotplug.c
```diff
@@ -19,7 +19,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
-#include "libusbi.h"
```
tormodvolden: What does this do here?

aokblast: Sorry, fix it.

libusb/libusb.h
aokblast: The FreeBSD have a totally rewrite libusb but provides the API compatibility with the libusb/libusb. In FreeBSD's libusb, they use wU2DevExitLat. It will be great if we can fix it at some point. Do you think it is better to leave original name then use:

```c
#define wU2DevExitLat bU2DevExitLat
```

So that we can provide the backward compatibility?

tormodvolden: Of course, that is a very good idea.

seanm: >The FreeBSD have a totally rewrite libusb...

They do?! Know why? Have they submitted patches to be upstreamed?

tormodvolden:
We'd need that backward compatibility define with a note about keeping it until next major API revision.

libusb/libusb.h
aokblast: I am on the way to BSDCan. I will update the patch when I have time after reaching Ottawa.

I am a little bit uncertain if we should call the FreeBSD version "rewrite". The first commit of libusb in FreeBSD base is around 2000 and the commit message says "import from NetBSD". I think libusb/libusb does not exist when 2000 as the first commit of libusb/libusb is about 2007. After that, I see several commits (around 2010?) in FreeBSD's  libusb enhance the compatibility to libusb 1.0. I think the enhancement is for libusb/libusb. So I don't know if and how should we upstream FreeBSD's libusb.

seanm: So where is this code in FreeBSD base?  I tried searching, is it this: https://github.com/freebsd/freebsd-src/tree/main/lib/libusb

it looks familiar, but I don't see any of the expected file names...

aokblast: Yes, this is the correct directory. You could look at libusb.h and will see something you may be familiar with. We separate libusb with libusb, libusb10, libusb20 but I also cannot find any different from .pc file.

mcuee: @aokblast and @seanm 

If you want to know the history of libusb-1.0 backend for FreeBSD, you can look at the following thread from HPS back in July 2008.
https://sourceforge.net/p/libusb/mailman/libusb-devel/thread/200807041802.35524.hselasky@c2i.net/

HPS (Hans Petter Selasky) was developing the "new" USB stack for FreeBSD back in 2007/2008/2009. The new USB stack was part of FreeBSD 8.0 which was officially released on 25 November 2009.

FreeBSD's libusb is part of the system (BSD license), therefore FreeBSD cannot adopt libusb project (LGPL license) as is due to licensing differences. Rather FreeBSD's libusb implementation tries to be compatible with libusb-1.0 API as much as possible. HPS used to be actively participated in the discussions in libusb-devel mailing list. Unfortunately he passed away in June 2023.

I was one of the earlier testers for the FreeBSD USB stack. Back then I was still willing to rebuild FreeBSD from source codes. Now I am lazy and only use the FreeBSD release branch for most of the testing. Occassionally I will use the FreeBSD Current branch using the snapshot. :-)

mcuee: Reference: you can see I was actively testing libusb under FreeBSD at that time. Time flies.
https://sourceforge.net/p/libusb/mailman/libusb-devel/thread/200901241702.01466.hselasky%40c2i.net/

seanm: Ah, interesting thanks. That explains why the license [here](https://github.com/freebsd/freebsd-src/blob/main/lib/libusb/libusb.h) says BSD instead of LGPL.

A pity that this divergence exists, as libusb is under-resourced. Oh well.

aokblast:
I see libusb does not have NOTE file. So I leave some comments on the libusb header. Do you have any other recommendation for maintainers to notice this when next API release?

PR #1655, opened 2025-06-08
Group libusb_endpoint_direction, libusb_request_type, libusb_request_recipient  nums into a single libusb_request enum.

sonatique:
This avoid having to combine different enum type values when constructing uint8_t bmRequestType parameter value for libusb_control_transfer.

Note: this prevent compilation warnings C5287 with VS 2022 v14+

Fixes issue: https://github.com/libusb/libusb/issues/1649. Supersedes https://github.com/libusb/libusb/pull/1650

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -356,18 +356,6 @@ enum libusb_descriptor_type {
 #define LIBUSB_ENDPOINT_ADDRESS_MASK		0x0f	/* in bEndpointAddress */
 #define LIBUSB_ENDPOINT_DIR_MASK		0x80
 
-/** \ingroup libusb_desc
- * Endpoint direction. Values for bit 7 of the
- * \ref libusb_endpoint_descriptor::bEndpointAddress "endpoint address" scheme.
- */
-enum libusb_endpoint_direction {
-	/** Out: host-to-device */
-	LIBUSB_ENDPOINT_OUT = 0x00,
-
-	/** In: device-to-host */
-	LIBUSB_ENDPOINT_IN = 0x80
-};
-
 #define LIBUSB_TRANSFER_TYPE_MASK		0x03	/* in bmAttributes */
 
 /** \ingroup libusb_desc
@@ -437,10 +425,25 @@ enum libusb_standard_request {
 };
 
 /** \ingroup libusb_misc
- * Request type bits of the
+ * Bits for constructing values of the
  * \ref libusb_control_setup::bmRequestType "bmRequestType" field in control
  * transfers. */
-enum libusb_request_type {
+enum libusb_request {
+
+	/** Endpoint direction. Values for bit 7 of the
+	* \ref libusb_endpoint_descriptor::bEndpointAddress "endpoint address" scheme.
+	*/
+
+	/** Out: host-to-device */
+	LIBUSB_ENDPOINT_OUT = 0x00,
+
+	/** In: device-to-host */
+	LIBUSB_ENDPOINT_IN = 0x80,
+
+	/** Request type bits of the
+	* \ref libusb_control_setup::bmRequestType "bmRequestType" field in control
+	* transfers. */
+
 	/** Standard */
 	LIBUSB_REQUEST_TYPE_STANDARD = (0x00 << 5),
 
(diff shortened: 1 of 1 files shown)


seanm:
I initially thought this would cause a lot of source incompatibility, even within libusb itself, but nothing in libusb actually uses these enum types by name, everything uses raw uint8_t instead.

I was expecting code like:

```
enum libusb_request_recipient variable = ...
```

but everything is more like:

```
uint8_t variable = ...
```

Still, those enums are public API, so users may have code like the first example. We can always keep the old enums too and just mark them deprecated.

We can thus preserve source compatibility (with deprecations). And there's no ABI change.

libusb/libusb.h
```diff
@@ -437,10 +425,25 @@ enum libusb_standard_request {
 };
 
 /** \ingroup libusb_misc
- * Request type bits of the
+ * Bits for constructing values of the
  * \ref libusb_control_setup::bmRequestType "bmRequestType" field in control
  * transfers. */
-enum libusb_request_type {
+enum libusb_request {
```
Youw, line 431: I'd leave it a `request_type` taking into account that it is intended as a second parameter of [`libusb_control_transfer`](https://github.com/libusb/libusb/blob/e793d4658b2361b7eecf5a336046c1ced2df9b71/libusb/sync.c#L82) and it is named a `bmRequestType` and there is also a different parameter `bRequest`

Youw:
Should we document [here](https://github.com/libusb/libusb/blob/e793d4658b2361b7eecf5a336046c1ced2df9b71/libusb/sync.c#L82) that the value supposed to be a combination of bitfields of `enum libusb_request_type`?

seanm:
> Should we document [here](https://github.com/libusb/libusb/blob/e793d4658b2361b7eecf5a336046c1ced2df9b71/libusb/sync.c#L82) that the value supposed to be a combination of bitfields of `enum libusb_request_type`?

This is just a proof of concept for now.

@sonatique does this PR fix all those warnings already, or would we need additional similar changes?

libusb/libusb.h
sonatique: This "new" enum merge types that were named endpoint_direction, request_type and request_recipient and has 4 value specifically name "request_type" (_standard, _class, _vendor, _reserved) , but also values named _recipient_ and _direction_. So I thought it would be fair to rename it to stress that its scope changed.

But honestly I don't want to spend to much time arguing on it ;-) it seems quite minor to me given there is no clear convention enforced in the file, so I'll be happy to name it the way you prefer.

sonatique:
@seanm : if we want to 
a) keep all the old enums 
and 
b) introduce a new , 
then we need the new one to have a name that doesn't exist yet. This is in disagreement with @Youw comment. We need to resolve this.

@Youw
> Should we document [here](https://github.com/libusb/libusb/blob/e793d4658b2361b7eecf5a336046c1ced2df9b71/libusb/sync.c#L82) that the value supposed to be a combination of bitfields of enum libusb_request_type?

Well, I think we could, but technically the API doesn't enforce the use of these enum, the documentation doesn't even mention them. Users are free to use whatever method they want to create the correct uint8_t value. I see them more like helper values, to avoid having to look at the USB spec. In my opinion a cleaner way would be to have the function take 3 parameters, one of each enum type, but this would be a breaking change of course. 

Another way of doing it could be to add an helper function taking 3 different enum parameters and internally calling the current function (using "ugly" uint8_t case unfortunately).

@seanm : yes the change is this PR is sufficient to make all C5287 warnings disappear.

libusb/libusb.h
Youw: My argument/motivation stands.
And I agree, this is not a major issue.

jpue:
Hello everyone,

Is there any chance that these changes are going to merged soon?

I have been working on “Q Light Controller Plus” (QLC+), an open-source stage lighting software, and we recently noticed that builds using GCC 16 start failing in C++ code that interfaces with your library (see linked issue above for a detailed problem analysis).

To summarise:
GCC 16 switched the default C++ version from C++17 to C++20. Since C++20 (this issue also occurs on older `gcc` versions when compiling with an explicitly specified C++ version >= C++20), code such as the following `libusb` library call
```
r = libusb_control_transfer(m_handle, LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_RECIPIENT_INTERFACE | LIBUSB_ENDPOINT_OUT, …)
```
(https://github.com/mcallegari/qlcplus/blob/master/plugins/peperoni/unix/peperonidevice.cpp#L261)
generates a warning and because QLC+ uses `-Werror` by default, this becomes an error.

The problem is that this code snippet combines flags of the `enum libusb_request_type` with ones from `enum libusb_request_recipient`, which is a deprecated C++ feature (“Deprecate usual arithmetic conversions between two distinct enumeration types”, P1120R0).

Do you have any suggestion/plans to solve this problem? As more and more systems will start using GCC 16 in the coming months, other software written in C++ that use this library may encounter similar issues.

I look forward to hearing from you.

PR #1659, opened 2025-06-10
WIP: experimenting with Clang -fbounds-safety

seanm:
I've been playing with [-fbounds-safety](https://clang.llvm.org/docs/BoundsSafety.html), and libusb is a nice small codebase to try it with.

Here are some WIP patches.  A couple of them could be merged soon, the others are just to show and share. There's a lot left to do.

I've already found one place that will require an API change ideally, which is why I created this PR now, so that can be discussed...

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1977,10 +1977,18 @@ static inline void libusb_fill_control_transfer(
 	transfer->endpoint = 0;
 	transfer->type = LIBUSB_TRANSFER_TYPE_CONTROL;
 	transfer->timeout = timeout;
-	transfer->buffer = buffer;
 	if (setup)
-		transfer->length = (int) (LIBUSB_CONTROL_SETUP_SIZE
+	{
+		int length = (int) (LIBUSB_CONTROL_SETUP_SIZE
 			+ libusb_le16_to_cpu(setup->wLength));
+		transfer->buffer = buffer;
+		transfer->length = length;
+	}
+	else
+	{
+		transfer->buffer = NULL;
+		transfer->length = 0;
+	}
 	transfer->user_data = user_data;
 	transfer->callback = callback;
 }
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1,3 +1,5 @@
+#include <ptrcheck.h>
+
 /*
  * Public libusb header file
  * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
@@ -24,6 +26,8 @@
 
 #ifndef LIBUSB_H
 #define LIBUSB_H
+
+__ptrcheck_abi_assume_single()
 
 #if defined(_MSC_VER)
 #pragma warning(push)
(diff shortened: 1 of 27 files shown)


seanm:
@tormodvolden @sonatique care to take a look?

libusb/os/events_posix.c
```diff
@@ -224,16 +224,15 @@ int usbi_alloc_event_data(struct libusb_context *ctx)
...
+	ctx->event_data = NULL;
 	ctx->event_data_cnt = 0;
+
+	unsigned int cnt = 0;
 	for_each_event_source(ctx, ievent_source)
-		ctx->event_data_cnt++;
+		cnt++;
 
-	fds = calloc(ctx->event_data_cnt, sizeof(*fds));
+	fds = calloc(cnt, sizeof(*fds));
```
sonatique: @seanm That is interesting and I don't see a problem with this change, but I have difficulties understanding what the problem was for -fbounds_safety and how it's now gone.
From `fds `pointer "point of view", `ctx->event_data_cnt`  is just a variable holding the size of the allocated memory zone (well, the multipler of element size), as is `cnt`, at the instant where `calloc `is called
Why is -fbounds-safety complaining? Why does it care how the value of `ctx->event_data_cnt` is assigned to it? Because it indicates that the value may change later after `calloc`? But so can `cnt`. And in the end we'll use `ctx->event_data_cnt` for knowing how many elements there, which could be assigned whatever value... 

So, I am not complaining about the change, but just wondering: what is the rational behing -fbounds-safety complain and how the changes improve something for real, not just to please it?
Thanks!

seanm: Thanks for reviewing.

>Why is -fbounds-safety complaining?

First, I'll just say: there was no bug here, the code was fine with normal C.

-fbounds-safety is more like a different dialect of C, so it has different rules. It was complaining because one of those rules is that when a pointer tagged with `__sized_by` or `__counted_by` is changed, the count variable tied to it has to change too, and vice versa, and the changes have to be back-to-back. The document I linked explains in detail, but it is of course pretty long.

>how the changes improve something for real, not just to please it?

"just pleasing it" allows the code to be built with -fbound-safety, and then you get memory safety: buffer overruns will always crash, instead of silently overwriting memory. That improves something for real.

libusb/libusb.h
```diff
@@ -1927,10 +1927,18 @@ static inline void libusb_fill_control_transfer(
 	transfer->endpoint = 0;
 	transfer->type = LIBUSB_TRANSFER_TYPE_CONTROL;
 	transfer->timeout = timeout;
-	transfer->buffer = buffer;
 	if (setup)
-		transfer->length = (int) (LIBUSB_CONTROL_SETUP_SIZE
+	{
+		int length = (int) (LIBUSB_CONTROL_SETUP_SIZE
```
tormodvolden: Does it also require the helper variable?

seanm: IIRC, yes in this case. Because the compiler needs to see both the pointer field (buffer) and the size field (length) updated back to back with no intervening stuff.

https://clang.llvm.org/docs/BoundsSafety.html#maintaining-correctness-of-bounds-annotations

"Additionally, an explicit bounds annotation such as `int *__counted_by(count) buf` defines a relationship between two variables, buf and count: namely, that buf has count number of elements available. This relationship must hold even after any of these related variables are updated. To this end, the model requires that assignments to buf and count must be side by side, with no side effects between them. This prevents buf and count from temporarily falling out of sync due to updates happening at a distance."

PR #1678, opened 2025-08-24
CI: Add clang-tidy to linux workflow 

Mr-Bossman:
Attempt to add clang-tidy to linux workflow.

The "example bad commit" will be removed, it is just a demo to show it works.

I also made these three pulls as a demo
https://github.com/Mr-Bossman/libusb/pull/3
https://github.com/Mr-Bossman/libusb/pull/2
https://github.com/Mr-Bossman/libusb/pull/1

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,12 +6,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job
+      # can access it
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: setup prerequisites
         run: |
           sudo apt update
-          sudo apt install -y autoconf automake libtool libudev-dev m4 pkg-config clang llvm
+          sudo apt install -y autoconf automake libtool libudev-dev m4 pkg-config clang llvm clang-tidy wget
 
       - name: bootstrap
         run: ./bootstrap.sh
@@ -80,3 +84,11 @@ jobs:
           ./fuzz_bos_descriptor tests/fuzz/corpus/bos \
             -seed=1 -runs=200 -max_total_time=15 -timeout=5 -rss_limit_mb=1536 -print_final_stats=1
       # --- end sanitizers + fuzzer smoke ---
+
+      - name: clang-tidy
+        continue-on-error: true
+        env:
+          current_sha: ${{ github.sha }}
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          .private/ci-build.sh --clang-tidy --build-dir build-tidy --no-test -- --disable-udev
--- a/.private/ci-build.sh
+++ b/.private/ci-build.sh
@@ -7,6 +7,7 @@ scriptdir=$(dirname $(readlink -f "$0"))
 install=no
 test=yes
 asan=yes
+tidy=no
 
 while [ $# -gt 0 ]; do
 	case "$1" in
@@ -34,6 +35,10 @@ while [ $# -gt 0 ]; do
 		cflags+=" -Werror"
 		shift
 		;;
+	--clang-tidy)
+		tidy=yes
+		shift
+		;;
 	--)
 		shift
 		break;
@@ -53,6 +58,8 @@ if [ -e "${builddir}" ]; then
 	echo "ERROR: directory entry named '${builddir}' already exists" >&2
 	exit 1
 fi
+
+gitdir=$(git rev-parse --show-toplevel)
 
 mkdir "${builddir}"
 cd "${builddir}"
@@ -74,10 +81,51 @@ fi
 
 echo ""
 echo "Configuring ..."
-CFLAGS="${cflags}" CXXFLAGS="${cflags}" ../configure --enable-examples-build --enable-tests-build "$@"
+if [ "${tidy}" = "yes" ]; then
+	# Clang-Tidy needs to be run with Clang compiler
+	CFLAGS="${cflags}" CXXFLAGS="${cflags}" CC="clang" CXX="clang++" ../configure \
+		--enable-examples-build --enable-tests-build "$@"
+else
+	CFLAGS="${cflags}" CXXFLAGS="${cflags}" ../configure --enable-examples-build \
+		--enable-tests-build "$@"
+fi
 
 echo ""
 echo "Building ..."
+
+if [ "${tidy}" = "yes" ]; then
+	# $(@D) and $(<F) are GNU Make automatic variables.
+	# $(@D): '$(@D)' is equivalent to '$(dirname $@)'.
+	# $(<F): '$(<F)' is equivalent to '$(notdir $<)'.
+	# example: 'src/foo.c' -> 'builddir/src/foo.c.compdb.json'
+	# More info: https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html
+	# add CFLAGS here as automake escapes them
+	cflags+=" -MJ \$(@D)/\$(<F).compdb.json"
+	make -j4 -k CFLAGS="${cflags}" CXXFLAGS="${cflags}"
+
+	# Create compile_commands.json from all the .compdb.json files.
+	echo "[" > compile_commands.json
+	find . -name "*compdb.json" -exec cat {} \; >> compile_commands.json
+	echo "]" >> compile_commands.json
+
+	built_files=$(find . -name "*.compdb.json" -printf "${gitdir}/%P " | \
+		sed 's/.compdb.json//g')
+	reldr=$(realpath --relative-to="$(pwd)" "${gitdir}")
+
+	# Get llvmorg-21.1.0-rc3 clang-tidy-diff.py script from LLVM project.
+	wget https://raw.githubusercontent.com/llvm/llvm-project/refs/tags/llvmorg-21.1.0-rc3/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py
+
+	# Use clang-tidy-diff to check only the files that were built.
+	# clang-tidy-diff expects the files to be relative to the git root.
+	# shellcheck disable=SC2086, SC2154
+	git diff -U0 "${current_sha}^..${current_sha}" --dst-prefix="b/${reldr}/" \
+		--src-prefix="a/${reldr}/" -- $built_files | \
+		python3 clang-tidy-diff.py -p1 -warnings-as-errors="*" \
+		-config-file="${gitdir}/.clang-tidy"
+
+	exit $?
+fi
+
 make -j4 -k
 
 if [ "${test}" = "yes" ]; then
--- a/.private/ci-build.sh
+++ b/.private/ci-build.sh
@@ -8,6 +8,7 @@ install=no
 test=yes
 asan=yes
 docs=no
+tidy=no
 
 while [ $# -gt 0 ]; do
 	case "$1" in
@@ -37,6 +38,10 @@ while [ $# -gt 0 ]; do
 		;;
 	--build-docs)
 		docs=yes
+		shift
+		;;
+	--clang-tidy)
+		tidy=yes
 		shift
 		;;
 	--)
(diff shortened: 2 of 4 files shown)


.private/ci-build.sh
```diff
@@ -70,10 +77,48 @@ fi
...
+	find . -name "*compdb.json" -exec cat {} \; >> compile_commands.json
+	echo "]" >> compile_commands.json
+
+	built_files=$(find . -name "*.compdb.json" -printf "${gitdir}/%P " | \
+		sed 's/.compdb.json//g')
+	reldr=$(realpath --relative-to="$(pwd)" "${gitdir}")
+
+
+	# Get latest clang-tidy-diff.py script from LLVM project.
+	wget https://raw.githubusercontent.com/llvm/llvm-project/refs/heads/main/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py
```
Youw: using the very latest version from git is not always the best idea
```suggestion
	wget https://raw.githubusercontent.com/llvm/llvm-project/refs/tags/llvmorg-20.1.8/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py
```
fixing it to latest release would make it more predictable and stable

.github/workflows/linux.yml
```diff
@@ -48,3 +50,10 @@ jobs:
 
       - name: umockdev test
         run: .private/ci-container-build.sh docker.io/amd64/ubuntu:rolling
+
+      - name: clang-tidy
+        env:
+          before_sha: ${{ github.event.before }}
+          current_sha: ${{ github.sha }}
```
Youw: maybe it makes sense to run it this way only for PRs (and skip push events?)

and, maybe, use `github.event.pull_request.base.sha` for the `before_sha`, so that no issue is missed in case if PR consists of multiple commits

.private/ci-build.sh
Mr-Bossman: Sure. Turns out the `-warnings-as-errors` flag is only in [llvmorg-21.1.0-rc1](https://github.com/llvm/llvm-project/commit/ccfabe8380572d99a6833bfff09a62018aca1a15) and higher.

.github/workflows/linux.yml
Mr-Bossman: the `before_sha` and `current_sha` allows for both PRs and pushes to be checked, but I can change it to be just PRs.

`current_sha` is actually a merge hash so it has all the changes in it see https://github.com/Mr-Bossman/libusb/pull/1 for multi commits working.
iirc for pull events `github.event.pull_request.sha` is the same as `github.sha`.

seanm:
Amazing!  I don't know anything about this github integration stuff, but after spending some time triaging clang-tidy issues, it will be great to have PRs checked automatically before merge!  Thanks for doing this!

.github/workflows/linux.yml
Youw: > allows for both PRs and pushes to be checked

Right, but is there a point to check a single push? Everything that gets merged into master branch should pass the PR anyway, I don't see a real benefit running it per-commit/change there. Maybe it would make a better sense to have a separate job to run against the entire build to find all issues in the code in one row, if any.

As for PR - I still think it is much better to check agains the entire PR on each push, not against only the latest commit in the PR.

Mr-Bossman: > > allows for both PRs and pushes to be checked
> 
> Right, but is there a point to check a single push? Everything that gets merged into master branch should pass the PR anyway, I don't see a real benefit running it per-commit/change there. 

No there isn't really any benefit.

> Maybe it would make a better sense to have a separate job to run against the entire build to find all issues in the code in one row, if any.

I'm still not entirely sure what you mean?
> As for PR - I still think it is much better to check agains the entire PR on each push, not against only the latest commit in the PR.

Oh so you mean run the check per-commit not the whole PR squashed I'll add that.

Youw: > Oh so you mean run the check per-commit not the whole PR squashed I'll add that.

Hm, I either missread or missunderstood. I mean run check as if whole PR squashed. My understanding that right now it would only run check for a single (latest) commit in the PR.

> I'm still not entirely sure what you mean?

I mean to have a separate job to check the entire repository with clang-tidy. As a separate independent (re-)check.

Mr-Bossman: > > Oh so you mean run the check per-commit not the whole PR squashed I'll add that.
> 
> Hm, I either missread or missunderstood. I mean run check as if whole PR squashed. My understanding that right now it would only run check for a single (latest) commit in the PR.

There is an example PR which shows that errors in 2 different commits being found here https://github.com/Mr-Bossman/libusb/pull/1. though squashing does allow the first commit to fail and the second on fix it, thus making clang-tidy non-bisectable .

> 
> > I'm still not entirely sure what you mean?
> 
> I mean to have a separate job to check the entire repository with clang-tidy. As a separate independent (re-)check.

The repository still has quite a few errors so as of now i don't think thats possible, but ill look into it

Youw: > There is an example PR which shows that errors in 2 different commits being found here https://github.com/Mr-Bossman/libusb/pull/1. though squashing does allow the first commit to fail and the second on fix it, thus making clang-tidy non-bisectable .

In that case I don't think we need to change anything.

> The repository still has quite a few errors so as of now i don't think thats possible, but ill look into it

If we are to fix them in any foreseeable future, having a job to trigger/verify errors would be of help.
But that is something totally separate, independent from this PR.

seanm: > The repository still has quite a few errors so as of now i don't think thats possible, but ill look into it

It does?  I thought I had fixed all clang-tidy warnings (at least those enabled by our .clang-tidy file)... I'll recheck.

seanm: @Youw could you sync https://github.com/libusb/libusb-cmake with libusb master?  It's so much easier to use clang-tidy when I can get cmake to generate a compile_commands.json.

seanm: @Mr-Bossman here are some more clang-tidy fixes https://github.com/libusb/libusb/pull/1685

Mr-Bossman: @seanm this also generates the `compile_commands.json` i don't think it works for windows though

seanm:
BTW would it be possible to have clang-tidy on macOS too?  (could be in a followup PR.)

Mr-Bossman:
will do.

.github/workflows/linux.yml
Youw: > @Youw could you sync https://github.com/libusb/libusb-cmake with libusb master? It's so much easier to use clang-tidy when I can get cmake to generate a compile_commands.json.

done

seanm (approved):

.github/workflows/linux.yml
```diff
@@ -16,11 +16,13 @@ jobs:
...
       # can access it
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: setup prerequisites
         run: |
           sudo apt update
-          sudo apt install autoconf automake libtool libudev-dev m4
+          sudo apt install autoconf automake libtool libudev-dev m4 clang-tidy clang wget
```
tormodvolden, line 70: If clang is added here just because it is a dependency of clang-tidy, it is not needed here. apt takes care of dependencies.

.private/ci-build.sh
```diff
@@ -70,10 +77,48 @@ fi
 
 echo ""
 echo "Configuring ..."
+if [ "${tidy}" = "yes" ]; then
+# Clang-Tidy needs to be run with Clang compiler
```
tormodvolden: Typically you'd indent if-clauses.

.private/ci-build.sh
```diff
@@ -70,10 +77,48 @@ fi
...
 
 echo ""
 echo "Building ..."
+
+if [ "${tidy}" = "yes" ]; then
+	# $(@D): '$(@D)' is equivalent to '$(dirname $@)'.
+	# $(<F): '$(<F)' is equivalent to '$(notdir $<)'.
+	# example: 'src/foo.c' -> 'builddir/src/foo.c.compdb.json'
+	# add CFLAGS here as configure escapes them
+	cflags+=" -MJ \$(@D)/\$(<F).compdb.json"
```
tormodvolden, line 111: I think I never saw these. What kind of substitutions are these - what exactly performs them?

.private/ci-build.sh
```diff
@@ -70,10 +77,48 @@ fi
...
+	echo "[" > compile_commands.json
+	find . -name "*compdb.json" -exec cat {} \; >> compile_commands.json
+	echo "]" >> compile_commands.json
+
+	built_files=$(find . -name "*.compdb.json" -printf "${gitdir}/%P " | \
+		sed 's/.compdb.json//g')
+	reldr=$(realpath --relative-to="$(pwd)" "${gitdir}")
+
+
+	# Get latest clang-tidy-diff.py script from LLVM project.
```
tormodvolden: Well it is not the "latest" but a specific version, which I think is a wise choice.

And an empty line too much above.

tormodvolden:
I don't know the clang-tidy specifics much but the CI script looks sane. And this PR passes its new check at least. However, the "bad" commit PRs that you made didn't fail as expected, why is that?

.private/ci-build.sh
Mr-Bossman: make performs them. automake escapes them so make doesn't do substitutions. 
https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html

I'll update the comment to clarify.

.github/workflows/linux.yml
```diff
@@ -48,3 +50,11 @@ jobs:
 
       - name: umockdev test
         run: .private/ci-container-build.sh docker.io/amd64/ubuntu:rolling
+
+      - name: clang-tidy
+        continue-on-error: true
```
Mr-Bossman: @tormodvolden I added `continue-on-error: true` here so that CI wont actually fail. the result looks like this
image

.private/ci-build.sh
tormodvolden: Cool, I didn't know about the D and F ones.

.github/workflows/linux.yml
tormodvolden: Ah OK, that's nice.

tormodvolden: But will people see this on the PR front page and not just on the CI jobs page?

Mr-Bossman: Only on the CI jobs page. I can change it to fail the whole CI if needed. Unfortunately GitHub doesn't have a warn flag only pass fail.

Mr-Bossman: @tormodvolden Should I change it to fail the whole CI?

mcuee:
@Mr-Bossman 

Would you please help to bring this PR up-to-date to resolve the conflicts? Thanks.

mcuee:
Somehow Windows build failed.

mcuee:
@Mr-Bossman 

Just wondering if you can resolve the conflicts and then see if CI build will be okay or not now. Thanks.

Mr-Bossman:
@tormodvolden @mcuee i changed the way configure gets called so it uses the `env` command. a re-review would be good

.github/workflows/linux.yml
Youw: Let it pass the CI but in case of failure - lets add an explicit PR comment with failure details.
That was we both: technically do not fail the CI, so we can safely merge if needed, but still make the issue explicitly visible.

.private/ci-build.sh
```diff
@@ -80,13 +89,60 @@ fi
...
+	echo "[" > compile_commands.json
+	find . -name "*compdb.json" -exec cat {} \; >> compile_commands.json
+	echo "]" >> compile_commands.json
+
+	built_files=$(find . -name "*.compdb.json" -printf "${gitdir}/%P " | \
+		sed 's/.compdb.json//g')
+	reldr=$(realpath --relative-to="$(pwd)" "${gitdir}")
+
+	# Get llvmorg-21.1.0-rc3 clang-tidy-diff.py script from LLVM project.
+	wget https://raw.githubusercontent.com/llvm/llvm-project/refs/tags/llvmorg-21.1.0-rc3/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py
```
Youw: `-rc3`? maybe use latest stable version?
should we match the version of this script with llvm version that is currently in runtime?

Youw:
I did a quick research, why don't we use: https://github.com/marketplace/actions/clang-tidy-review ? It has nice integration with github, would add inline annotations right into the diff.

Mr-Bossman:
> I did a quick research, why don't we use: https://github.com/marketplace/actions/clang-tidy-review ? It has nice integration with github, would add inline annotations right into the diff.

somewhere there is a thread where we decided not to use it for some reason. I unfortunately cant find it now, though.

Youw:
> somewhere there is a thread where we decided not to use it for some reason. I unfortunately cant find it now, though.

I've practiced an alternative implementation with codex using that github action: #1951
a few PRs with examples how the integration would look like are on the way

Mr-Bossman:
@Youw @seanm @mcuee Sorry for the immense amount of spam, but this now also works.

PR #1681, opened 2025-08-25
Windows Github actions: specify supported architectures explicitly.

sonatique:
Windows Github actions use Windows SDK version 10.0.26100.0+ that does
not support 32-bit ARM development any longer. When letting build action
iterates over all possible platforms (i.e. architectures) included in
libusb.sln, builds fail when trying "ARM" (i.e. ARM 32-bit). Moreover
libusb.sln solution does not define "x86" but uses "Win32" instead.

This commit resolves these issues by explicitly iterating over:
Win32, x64, ARM64.

Note: ARM 32-bit is therefore not tested by Windows Gith actions, though
still supported by libusb solution, enabling ARM 32-bit build on machine
with earlier Windows SDK installed.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,8 +13,9 @@ jobs:
           - 'Debug'
           - 'Release'
         architecture:
-          - 'x86'
+          - 'Win32'
           - 'x64'
+          - 'ARM64'
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


sonatique:
I merged this PR without waiting for approval, given my previous action (merging https://github.com/libusb/libusb/pull/1353) broke master and given this does not touch the code, only the (new) Github Windows actions.

I hope this is OK for everyone.

Github Windows action details can still be improved anyway.

mcuee:
No issues. It is okay to merge such PR directly.

PR #1202, opened 2022-10-04
windows_winusb libusb_get_device_list: bus numbers are defined and assigned to root hubs directly during the hub pass, by enforcing the defintion of root hub as being a hub without parent.

sonatique:
windows: define bus numbers without using HCD

Bus numbers are defined and assigned to root hubs directly
during the hub pass of winusb libusb_get_device_list, using the
definition of a root hub as being a hub without parent at port 0.

Therefore the HCD pass is no longer used for defining root hubs.

Note: assigned bus number values for a given setup are not guaranteed
to remain constant when migrating from previous implementation to this
one.

The rational behind this change is that:

- Implementation using CM_Get_Child from HCD to get root hub devInst is
not working when using "FabulaTech Virtual USB Host Controller"
(USB-Ethernet bridge) under Hyper-V virtual machine.

- This new way of enumerating is inspired by "USB Device Viewer" Microsoft
driver sample code which is capable of correctly enumerating hubs and
devices in the above mentioned configuration as well as in regular
configurations.

PR #1682, opened 2025-08-26
Reorder some struct fields so mutex name appears before the variable it guards

seanm:
Initially, Clang Thread Safety annotation parsing required them ordered this way, or it won't compile. This has since been improved in newer Clang, but has not reached even the newest Xcode (16.4) yet, so this reording allow supporting more compilers.

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -385,14 +385,14 @@ struct libusb_context {
 	/* A flag to indicate that the context is ready for hotplug notifications */
 	usbi_atomic_t hotplug_ready;
 
+	/* Note paths taking both this and usbi_transfer->lock must always
+	 * take this lock first */
+	usbi_mutex_t flying_transfers_lock;
 	/* this is a list of in-flight transfer handles, sorted by timeout
 	 * expiration. URBs to timeout the soonest are placed at the beginning of
 	 * the list, URBs that will time out later are placed after, and urbs with
 	 * infinite timeout are always placed at the very end. */
 	struct list_head flying_transfers;
-	/* Note paths taking both this and usbi_transfer->lock must always
-	 * take this lock first */
-	usbi_mutex_t flying_transfers_lock; /* for flying_transfers and timeout_flags */
 
 #if !defined(PLATFORM_WINDOWS)
 	/* user callbacks for pollfd changes */
@@ -451,8 +451,8 @@ struct libusb_context {
 extern struct libusb_context *usbi_default_context;
 extern struct libusb_context *usbi_fallback_context;
 
-extern struct list_head active_contexts_list;
 extern usbi_mutex_static_t active_contexts_lock;
+extern struct list_head active_contexts_list;
 
 static inline struct libusb_context *usbi_get_context(struct libusb_context *ctx)
 {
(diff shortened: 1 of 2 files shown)


tormodvolden (approved):

tormodvolden:
Would be good if the commit message prepends "libusb.h:" and use consistently imperative form (even saves some chararacters in an already long summary :) ).

anotheruserofgithub:
I have two naive questions:
1. Support for `guarded_by` was initially done by https://github.com/llvm/llvm-project/pull/94216 but soon after replaced by https://github.com/llvm/llvm-project/pull/95455, which apparently also supports late parsing, so is it still required to reorder the mutex? The OP of the latter PR shows this example:
    ```c
    struct Example {
      int a_value_defined_before __attribute__ ((guarded_by(a_mutex)));
      struct Mutex *a_mutex;
    };
    ```
2. As far as I understand, in https://github.com/libusb/libusb/pull/1419 you only added a few annotations for demonstration, but in the end you want to add more, right? In this case, I suppose there would be the need for reordering basically all mutexes, e.g. `usb_devs_lock` in libusbi.h:373 (unless point 1 makes it useless)?

seanm:
1. It's great that Clang has those improvements now, but they were not there when I started this work in late 2023.  Also, just tried with Xcode 16.4 (current newest) and it still doesn't have that support yet in C mode, but does in C++ mode:

```
% xcrun clang++ -fsyntax-only -w ~/Desktop/test.c
% xcrun clang -fsyntax-only -w ~/Desktop/test.c
/Users/sean/Desktop/test.c:2:57: error: use of undeclared identifier 'a_mutex'
    2 |   int a_value_defined_before __attribute__ ((guarded_by(a_mutex)));
      |                                                         ^
1 error generated.
```

So to support as many compilers as we can, moving the mutex earlier seems better. It seems like a low-risk change to me.

Changing libusb to *also* compile as valid C++ (mostly needs additional casts) I also did in https://github.com/libusb/libusb/pull/1419/commits/f0328eae0117576a7f9b562f86d79c2b2708f1cb but my impression is the maintainers don't want it.

2. Yes, I would add more. But it's been almost 2 years and if I can't get the prerequisites merged, there's no point working on it further. I'm trying to give it another push now...

seanm:
> Would be good if the commit message prepends "libusb.h:" and use consistently imperative form (even saves some chararacters in an already long summary :) ).

Why should the summary mention libusb.h as opposed to darwin_usb.c?

tormodvolden:
> > Would be good if the commit message prepends "libusb.h:" and use consistently imperative form (even saves some chararacters in an already long summary :) ).
> 
> Why should the summary mention libusb.h as opposed to darwin_usb.c?

I originally meant libusbi.h :) Primarily because most of the change was in libusbi.h with just a small one in darwin_usb.c. But IMO the real reason for this prefix convention is not to be complete or accurate, but to make it easier for someone to peruse the git log and quickly see which commits are of interest, and maybe more important, which are not of interest. So someone working on Windows and would like to what's new relevant to them, can skip everything labeled "darwin" and "linux" and "CI" etc.

seanm:
OK, if that'll let it be merged, I won't argue. Done.

sonatique (approved):
I think the commit message doesn't follow the rules stated in the HACKING file

Commit messages should be formatted to 72 chars width and have a
free-standing summary line. See for instance "Commit Guidelines" on
https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project
or https://cbea.ms/git-commit/ about how to make well-formed commit
messages.

I would recommend adapting it, thanks.

seanm:
> Commit messages should be formatted to 72 chars

OK done. But if this kind of pedantry is important to you, you should add a git pre-commit hook.

sonatique:
> > Commit messages should be formatted to 72 chars
> 
> OK done. But if this kind of pedantry is important to you, you should add a git pre-commit hook.

;-)

I guess I am just inflicting to you what other inflicted to me before :
https://github.com/libusb/libusb/pull/1202#issuecomment-1384078860

Anyway, I know think that it makes scrolling the commit message easier.

By the way, having some mandatory commit hook for such kind of things seems a good idea. I don't know how to enforce them.

seanm:
@sonatique can you merge?  I would feel weird merging my own PR...

PR #1683, opened 2025-08-26
darwin: use threads_posix.h wrapper functions

seanm:
Instead of using pthread API directly, use the threads_posix.h wrapper functions.

This is handy because those can be tagged with Clang Thread Safety annotations (we can't change system header files).

Introduce usbi_cond_signal as a wrapper for pthread_cond_signal.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -57,15 +57,15 @@ static const mach_port_t darwin_default_master_port = 0;
 /* async event thread */
 /* if both this mutex and darwin_cached_devices_mutex are to be acquired then
    darwin_cached_devices_mutex must be acquired first. */
-static pthread_mutex_t libusb_darwin_at_mutex = PTHREAD_MUTEX_INITIALIZER;
-static pthread_cond_t  libusb_darwin_at_cond = PTHREAD_COND_INITIALIZER;
+static usbi_mutex_static_t libusb_darwin_at_mutex = USBI_MUTEX_INITIALIZER;
+static usbi_cond_t libusb_darwin_at_cond = USBI_COND_INITIALIZER;
 
 #define LIBUSB_DARWIN_STARTUP_FAILURE ((CFRunLoopRef) -1)
 
 static CFRunLoopRef libusb_darwin_acfl = NULL; /* event cf loop */
 static CFRunLoopSourceRef libusb_darwin_acfls = NULL; /* shutdown signal for event cf loop */
 
-static usbi_mutex_t darwin_cached_devices_mutex = PTHREAD_MUTEX_INITIALIZER;
+static usbi_mutex_t darwin_cached_devices_mutex = USBI_MUTEX_INITIALIZER;
 static struct list_head darwin_cached_devices;
 static const char *darwin_device_class = "IOUSBDevice";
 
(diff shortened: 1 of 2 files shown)


seanm:
> It is worth pointing this out in the commit message, that the wrappers were already (partially) used.

Done.

libusb/os/darwin_usb.c
```diff
@@ -57,8 +57,8 @@ static const mach_port_t darwin_default_master_port = 0;
 /* async event thread */
 /* if both this mutex and darwin_cached_devices_mutex are to be acquired then
    darwin_cached_devices_mutex must be acquired first. */
-static pthread_mutex_t libusb_darwin_at_mutex = PTHREAD_MUTEX_INITIALIZER;
-static pthread_cond_t  libusb_darwin_at_cond = PTHREAD_COND_INITIALIZER;
+static usbi_mutex_static_t libusb_darwin_at_mutex = USBI_MUTEX_INITIALIZER;
+static usbi_cond_t libusb_darwin_at_cond = PTHREAD_COND_INITIALIZER;
```
anotheruserofgithub: Why not aliasing this one too?
```suggestion
static usbi_cond_t libusb_darwin_at_cond = USBI_COND_INITIALIZER;
```

And L68:
```diff
-static usbi_mutex_t darwin_cached_devices_mutex = PTHREAD_MUTEX_INITIALIZER;
+static usbi_mutex_t darwin_cached_devices_mutex = USBI_MUTEX_INITIALIZER;
```

seanm: Great suggestions, thanks!  Done.

libusb/os/threads_posix.h
```diff
@@ -26,6 +26,7 @@
 #define PTHREAD_CHECK(expression)	ASSERT_EQ(expression, 0)
 
 #define USBI_MUTEX_INITIALIZER	PTHREAD_MUTEX_INITIALIZER
+#define USBI_COND_INITIALIZER	PTHREAD_COND_INITIALIZER
```
anotheruserofgithub: Maybe `#define USBI_COND_INITIALIZER` should be moved right above `typedef pthread_cond_t usbi_cond_t` L62, since this header seems to be organized by "blocks" mutex/cond/tls?

libusb/os/threads_posix.h
```diff
@@ -74,6 +75,10 @@ static inline void usbi_cond_destroy(usbi_cond_t *cond)
 {
 	PTHREAD_CHECK(pthread_cond_destroy(cond));
 }
+static inline void usbi_cond_signal(usbi_cond_t *cond)
```
anotheruserofgithub: Maybe `usbi_cond_signal()` should be moved right above `usbi_cond_broadcast()` L70, since both are about signaling?

libusb/os/threads_posix.h
anotheruserofgithub: Actually I think the best would be to keep both defines here and to add a blank line below them as a separation (I didn't see that the first "block" is for static mutex and then there is a second "block" for normal mutex).

anotheruserofgithub: And maybe for consistency it would be good to add an empty line below `#define USBI_MUTEX_INITIALIZER` in threads_windows.h:26 too?

libusb/os/threads_posix.h
seanm: Agreed. Done.

libusb/os/threads_posix.h
seanm: Actually I think I like your 1st idea better.

seanm:
@anotheruserofgithub thanks for your detailed review. Think this is ready now?

anotheruserofgithub:
> @anotheruserofgithub thanks for your detailed review. Think this is ready now?

Yes, this is great, sorry for delaying your work.

seanm:
> Yes, this is great, sorry for delaying your work.

Nonsense, your review was most appreciated.

seanm:
@sonatique can you merge this one too?

PR #1684, opened 2025-08-26
Clarify return value of usbi_mutex_trylock with named temporary variable

seanm:
There mutex wrapping functions are clearly thin wrappers over pthreads, yet the return value of usbi_mutex_trylock() is opposite that of the pthread_mutex_trylock() it wraps. Make the return value meaning clearer by introducing a named variable.

### DIFF
--- a/libusb/os/threads_posix.h
+++ b/libusb/os/threads_posix.h
@@ -51,7 +51,8 @@ static inline void usbi_mutex_unlock(usbi_mutex_t *mutex)
 }
 static inline int usbi_mutex_trylock(usbi_mutex_t *mutex)
 {
-	return pthread_mutex_trylock(mutex) == 0;
+	int mutexIsLocked = pthread_mutex_trylock(mutex) == 0;
+	return mutexIsLocked;
 }
 static inline void usbi_mutex_destroy(usbi_mutex_t *mutex)
 {
--- a/libusb/os/threads_windows.h
+++ b/libusb/os/threads_windows.h
@@ -50,7 +50,8 @@ static inline void usbi_mutex_unlock(usbi_mutex_t *mutex)
 }
 static inline int usbi_mutex_trylock(usbi_mutex_t *mutex)
 {
-	return TryEnterCriticalSection(mutex) != 0;
+    int mutexIsLocked = TryEnterCriticalSection(mutex) != 0;
+    return mutexIsLocked;
 }
 static inline void usbi_mutex_destroy(usbi_mutex_t *mutex)
 {


libusb/os/threads_posix.h
```diff
@@ -51,7 +51,8 @@ static inline void usbi_mutex_unlock(usbi_mutex_t *mutex)
 }
 static inline int usbi_mutex_trylock(usbi_mutex_t *mutex)
 {
-	return pthread_mutex_trylock(mutex) == 0;
+	int mutexIsNowLocked = pthread_mutex_trylock(mutex) == 0;
```
sonatique: I am fine with this way of doing, though I am not used to and I think a comment could have the same informative effect. 

I think the risk with this is that later at some point someone will just think the temporary variable is not necessary (and this is indeed the case, technically) and remove it for "cleaning" the code. 

I would prefer a comment but I can live with this.

sonatique (approved):

libusb/os/threads_posix.h
anotheruserofgithub: I agree with you, a comment would be more appropriate. Also I was confused because I first read it "mutex is no**t** locked" instead of "no**w**".

anotheruserofgithub: But if these files are supposed to be just thin wrappers, why not change the condition on the call side instead? As per GH search results there is only a single caller in io.c:1814.

seanm: The maintainers are usually extremely conservative about changing behaviour, so I figured reversing the behaviour would not be accepted. Also, it would then be backwards from the Windows perspective. It's not exactly public API, but it's also possible someone out there is using this function, and inverting its behaviour would break them. Personally, I wouldn't mind reversing it, but there's not a lot of upside really.

As for comments, in my view, the variable name *is* the comment. Comments are very prone to falling out of sync with code, and using a variable name that the compiler can check is often nicer. 2¢.

anotheruserofgithub: The return value was changed in 11cc9952823fbb47b7596aec30a205f68f3a8062, apparently such as to match `PTHREAD_CHECK()` and `WINAPI_CHECK()`. But to my opinion there is a tiny little inconsistency because the commit message says that `usbi_cond_timedwait()` was changed to return an error code corresponding to `enum libusb_error` instead of errno (thus `LIBUSB_SUCCESS = 0` on success), however `usbi_mutex_trylock()` returns a boolean, thus perhaps it should better `return pthread_mutex_trylock(mutex) == 0 ? 0 : LIBUSB_ERROR_OTHER;` and `libusb_try_lock_events()` should test `if (r < 0)`? https://github.com/libusb/libusb/blob/11cc9952823fbb47b7596aec30a205f68f3a8062/libusb/io.c#L1795-L1797 Also, currently `libusb_try_lock_events()` returns the inverse of `usbi_mutex_trylock()`: https://github.com/libusb/libusb/blob/11cc9952823fbb47b7596aec30a205f68f3a8062/libusb/io.c#L1774-L1778 Which is even weirder, kind of an anti-boolean... But this is public API I guess, so I understand you won't change that.

seanm:
@sonatique can you merge?  I would feel weird merging my own PR...

sonatique:
> @sonatique can you merge? I would feel weird merging my own PR...

OK, no problem, but this will be in maybe 2 hours from now.

PR #1685, opened 2025-08-26
More clang-tidy fixes

seanm:


examples/xusb.c
```diff
@@ -1193,8 +1193,6 @@ int main(int argc, char** argv)
 					test_mode = USE_XBOX;
 					break;
 				case 'h':
-					display_help(argv[0]);
-					return EXIT_SUCCESS;
 				default:
 					display_help(argv[0]);
 					return EXIT_FAILURE;
```
anotheruserofgithub: Hi! Small remark: `case 'h'` returned `EXIT_SUCCESS` while `default` returns `EXIT_FAILURE`. Do they expand to the same value somehow?

seanm: Oh crap, you're quite right!  I trusted the tool too blindly, this seems to be a false positive from clang-tidy.  Will revert this part...

Mr-Bossman: huh interesting.

Mr-Bossman:
LGTM!

examples/xusb.c
seanm: I think it might be this: https://github.com/llvm/llvm-project/issues/67662

tormodvolden (approved):
I find it a bit silly to have to write "a + (b * c)" but if this helps prevent real bugs, I can live with that.

seanm:
Do we think this is mergable now?

Would be nice to have since it is blocking https://github.com/libusb/libusb/pull/1678

PR #1686, opened 2025-08-26
Some code transformations required to allow -fbounds-safety to compile

seanm:
Should be no change in behaviour.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1977,10 +1977,18 @@ static inline void libusb_fill_control_transfer(
 	transfer->endpoint = 0;
 	transfer->type = LIBUSB_TRANSFER_TYPE_CONTROL;
 	transfer->timeout = timeout;
-	transfer->buffer = buffer;
 	if (setup)
-		transfer->length = (int) (LIBUSB_CONTROL_SETUP_SIZE
+	{
+		int length = (int) (LIBUSB_CONTROL_SETUP_SIZE
 			+ libusb_le16_to_cpu(setup->wLength));
+		transfer->buffer = buffer;
+		transfer->length = length;
+	}
+	else
+	{
+		transfer->buffer = NULL;
+		transfer->length = 0;
+	}
 	transfer->user_data = user_data;
 	transfer->callback = callback;
 }
--- a/libusb/os/events_posix.c
+++ b/libusb/os/events_posix.c
@@ -220,23 +220,24 @@ int usbi_disarm_timer(usbi_timer_t *timer)
 
 int usbi_alloc_event_data(struct libusb_context *ctx)
 {
-	struct usbi_event_source *ievent_source;
-	struct pollfd *fds;
-	size_t i = 0;
+	free(ctx->event_data);
+	ctx->event_data = NULL;
+	ctx->event_data_cnt = 0;
 
-	if (ctx->event_data) {
-		free(ctx->event_data);
-		ctx->event_data = NULL;
+	unsigned int cnt = 0;
+	struct usbi_event_source *ievent_source;
+	for_each_event_source(ctx, ievent_source)
+		cnt++;
+
+	if (cnt == 0) {
+		return 0;
 	}
 
-	ctx->event_data_cnt = 0;
-	for_each_event_source(ctx, ievent_source)
-		ctx->event_data_cnt++;
-
-	fds = calloc(ctx->event_data_cnt, sizeof(*fds));
+	struct pollfd *fds = calloc(cnt, sizeof(*fds));
 	if (!fds)
 		return LIBUSB_ERROR_NO_MEM;
 
+	size_t i = 0;
 	for_each_event_source(ctx, ievent_source) {
 		fds[i].fd = ievent_source->data.os_handle;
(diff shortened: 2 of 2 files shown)


libusb/os/events_posix.c
```diff
@@ -224,16 +224,15 @@ int usbi_alloc_event_data(struct libusb_context *ctx)
...
-	if (ctx->event_data) {
-		free(ctx->event_data);
-		ctx->event_data = NULL;
-	}
-
+	free(ctx->event_data);
+	ctx->event_data = NULL;
 	ctx->event_data_cnt = 0;
+
+	unsigned int cnt = 0;
```
tormodvolden, line 227: This should be at the top.

seanm: Do we still need to support C89?

tormodvolden: It is about consistency with the surrounding code. I am personally a fan of limiting scope for variables and declaring them when needed, but that is not the question here. It just lookw weird with some variables declared at the top and this single one in the middle.

seanm: I am also a fan of limiting scope and locality of reference.  How about I just rehabilitate the function then?  See new commit.

Also, clang static analyzer was warning in this function (both before and after my changes) `Use of memory allocated with size zero`. Not sure this is a true positive or false positive (the `for_each_event_source` macro is too gnarly for me the try and understand right now). I added an early return if `cnt` is zero. That seem correct?

seanm:
@sonatique what do you think?

mcuee:
@tormodvolden 

Are you fine with this PR now? Thanks.

libusb/libusb.h
```diff
@@ -1924,10 +1924,18 @@ static inline void libusb_fill_control_transfer(
 	transfer->endpoint = 0;
 	transfer->type = LIBUSB_TRANSFER_TYPE_CONTROL;
 	transfer->timeout = timeout;
-	transfer->buffer = buffer;
 	if (setup)
-		transfer->length = (int) (LIBUSB_CONTROL_SETUP_SIZE
+	{
+		int length = (int) (LIBUSB_CONTROL_SETUP_SIZE
```
sonatique, line 1982: Is the temporary `length` required by -fbouds-safety ?

sonatique (approved):

libusb/libusb.h
seanm: Yes: https://github.com/libusb/libusb/pull/1659#discussion_r2248354454

libusb/os/events_posix.c
tormodvolden: I'd prefer to keep this PR simple, changing as little as possible to make -fbounds-safety happy. Rehabilitating the function and all other changes make it much harder to review. We have made real regressions before while doing innocently looking rework for pleasing checking tools. I am not saying we shouldn't do those changes but they must be reviewed more deeply than I have time for now.

mcuee:
Pending review from @tormodvolden 

* https://github.com/libusb/libusb/pull/1686/files#r2426853868
> I'd prefer to keep this PR simple, changing as little as possible to make -fbounds-safety happy. Rehabilitating the function and all other changes make it much harder to review. We have made real regressions before while doing innocently looking rework for pleasing checking tools. I am not saying we shouldn't do those changes but they must be reviewed more deeply than I have time for now.

libusb/os/events_posix.c
seanm: IMHO this PR is very simple already. It's a small change to a 25 line private function with literally one caller. It's not hugely hard to reason about.

But I sympathize about not having time.

@sonatique has already approved, but another reviewer would always be nice. @Mr-Bossman perhaps you could take a look?

libusb/os/events_posix.c
```diff
@@ -220,30 +220,33 @@ int usbi_disarm_timer(usbi_timer_t *timer)
 
 int usbi_alloc_event_data(struct libusb_context *ctx)
 {
+	free(ctx->event_data);
```
Mr-Bossman, line 223: On Mac OSX freeing a NULL pointer is a runtime warning. I'm not sure why this happens as this is something Posix permits, but can you guard the free. Also @seanm while your at it I agree with tom on the C89-style variables. 

Side note I'm a bit busy with university sorry I haven't been as active.

seanm: >On Mac OSX freeing a NULL pointer is a runtime warning.

I think you're thinking of something else. There's no such runtime warning on macOS.

Mr-Bossman: > > On Mac OSX freeing a NULL pointer is a runtime warning.
> 
> I think you're thinking of something else. There's no such runtime warning on macOS.

Ah yes I was. That only happens when `-fsanitize=address` is enabled ignore my comment then.

seanm: >That only happens when -fsanitize=address

I just tried, and even then there is no such warning.  free(NULL) is well-defined and gives no warnings.

Mr-Bossman: I looked into this a bit more and it seems like the libasan version I'm using is broken, so LGTM.

PR #1521, opened 2024-06-24
windows: Fix bus number assignment corner case

rgantois:
Hi everyone,

Sorry about the duplicate pull request, I messed up my commit signature on the first one.

WinUSB assignes bus numbers sequentially to USB root hubs at enumeration time. In some corner cases, one root hub may be skipped by the enumeration loop due to it having been handled by a previous enumeration. The next enumerated root hub will then be assigned the skipped bus number, which will result in two different root hubs sharing the same bus number.

When skipping a root hub during enumeration, make sure to avoid reusing its bus number.

I discovered this issue while interacting with multiple different USB devices through pyusb. One device was on bus 3, port 3 and the other was on bus 4, port 3. Sometimes, in the same Python interpreter run, usb.core.find(bus=3, port_numbers=(3,), find_all=True) would return the device on bus 3 on the first call, and add the device on bus 4 to the returned results on the second call, claiming that both devices were on bus 3. This behavior only appeared if a reference to the first usb.core.find() call was kept in the Python script.

Issue #1165, opened 2022-07-14
USB hub reference is not decremented when a handle to a connected device is closed. 


Oldanko:
I have written a program to output the number of references to connected USB devices to see how the number changes after different USB events. As a result, I've observed that when I `libusb_open` a device that is connected to the USB hub, the number of references to that device increases by one, as well as the number of references to the USB hub. However, when I `libusb_close` the device, the number of references to the device decreases by one, but the number of references to the USB hub stays the same.

In order to write this program, as I couldn't find a function to get the number of references in libusb, I've added a function with the following definition:
`
void LIBUSB_CALL libusb_refcnt_device(libusb_device *dev, long* refcnt)
{
    *refcnt = usbi_atomic_load(&dev->refcnt);
}
`

Following is the C++ code of the program:
`
#include <iostream>
#include <vector>
#include "libusb/libusb.h"

void print_ref_cnts(libusb_context *context) {
    libusb_device **devs;
    auto devs_size = libusb_get_device_list(context, &devs);

    for(ssize_t i = 0; i < devs_size; i++) {
        auto p_device = devs[i];
        libusb_device_descriptor descriptor{};
        libusb_get_device_descriptor(p_device, &descriptor);

        long refs;
        libusb_refcnt_device(p_device, &refs);

        std::cout << std::hex
                  << " 0x" << descriptor.idVendor
                  << " 0x" << descriptor.idProduct
                  << std::dec
                  << " " << refs << std::endl;
    }

    libusb_free_device_list(devs, true);
}

auto open_devices(libusb_context *context, unsigned short vendor_id, unsigned short product_id) {
    libusb_device **devs;
    auto devs_size = libusb_get_device_list(context, &devs);

    std::vector<libusb_device_handle*> handles;

    for(ssize_t i = 0; i < devs_size; i++) {
        libusb_device_descriptor descriptor{};
        libusb_get_device_descriptor(devs[i], &descriptor);
        if(descriptor.idVendor == vendor_id && descriptor.idProduct == product_id) {
            libusb_device_handle* dh;
            libusb_open(devs[i], &dh);
            handles.push_back(dh);
        }
    }

    libusb_free_device_list(devs, true);
    return handles;
}

int main() {
    libusb_context *context;
    libusb_init(&context);

    std::string s;
    while(std::getline(std::cin, s)) {
        std::cout << "Before opening\n";
        print_ref_cnts(context);
        auto handles = open_devices(context, 0x9da, 0x9090);
[...]

PR #1687, opened 2025-08-27
user does not change refcnt, instead it indicates he wants to use the device

minicarpet:
This PR should resolve [issue #1165](https://github.com/libusb/libusb/issues/1165)

See comments into issue for more details.

**NOTE**: tested:

- using windows without hotplug
- using windows hotplug
- using private company program

Not tested for now:
- using linux

This PR change the way how user use devices:

- as explain into the issue: users should not change refcnt and access to this
- I moved ref inc and ref decrementation into internal (through `usbi_ref_device` & `usbi unref_device`)
- I kept `libusb_ref_device` & `libusb_unref_device` for compatibility
- User uses these functions (`libusb_ref_device` & `libusb_unref_device`) to tells libusb to keep the handle such "I want to work with this handle"
- User **MUST** call `libusb_unref_device` per `libusb_ref_device`

I have seen a problem (at least on windows) about bus number: the same number could be used multiple times (for different root hub). -> [another PR that is usefull to merge](https://github.com/libusb/libusb/pull/1521)

An other solution would be to manage HUB reference by not incrementing them when already discovered. But I had some issues doing this...

A sample for this would be:
```
/*
* Test suite program based of libusb-0.1-compat testlibusb
* Copyright (c) 2013 Nathan Hjelm <hjelmn@mac.ccom>
*
* This library is free software; you can redistribute it and/or
* modify it under the terms of the GNU Lesser General Public
* License as published by the Free Software Foundation; either
* version 2.1 of the License, or (at your option) any later version.
*
* This library is distributed in the hope that it will be useful,
* but WITHOUT ANY WARRANTY; without even the implied warranty of
* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
* Lesser General Public License for more details.
*
* You should have received a copy of the GNU Lesser General Public
* License along with this library; if not, write to the Free Software
* Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
*/

#include <stdio.h>
#include <string.h>
#include "libusb.h"
#include "libusbi.h"

int verbose = 0;

#ifdef __linux__
#include <errno.h>
#include <fcntl.h>
#include <unistd.h>

static int test_wrapped_device(const char *device_name)
{
	libusb_device_handle *handle;
	int r, fd;

	fd = open(device_name, O_RDWR);
	if (fd < 0) {
		printf("Error could not open %s: %s\n", device_name, strerror(errno));
		return 1;
	}
	r = libusb_wrap_sys_device(NULL, fd, &handle);
	if (r) {
		printf("Error wrapping device: %s: %s\n", device_name, libusb_strerror(r));
		close(fd);
		return 1;
	}
	print_device(libusb_get_device(handle), handle);
	close(fd);
	return 0;
}
#else
static int test_wrapped_device(const char *device_name)
{
	(void)device_name;
	printf("Testing wrapped devices is not supported on your platform\n");
	return 1;
}
#endif

void my_lib_usb_log_cb(libusb_context* ctx, enum libusb_log_level level, const char* str)
{
	UNUSED(ctx);
	UNUSED(level);
	printf("%s", str);
}

int main(int argc, char *argv[])
{
	const char *device_name = NULL;
	libusb_device **devs;
	libusb_device* stored_device[10] = { NULL };
	libusb_device_handle* handle_device[10] = { NULL };
	size_t index = 0;
	ssize_t cnt;
	int r, i;

	for (i = 1; i < argc; i++) {
		if (!strcmp(argv[i], "-v")) {
			verbose = 1;
		} else if (!strcmp(argv[i], "-d") && (i + 1) < argc) {
			i++;
			device_name = argv[i];
		} else {
			printf("Usage %s [-v] [-d </dev/bus/usb/...>]\n", argv[0]);
			printf("Note use -d to test libusb_wrap_sys_device()\n");
			return 1;
		}
	}

	r = libusb_init_context(/*ctx=*/NULL, /*options=*/NULL, /*num_options=*/0);
	if (r < 0)
		return r;

	libusb_set_log_cb(NULL, my_lib_usb_log_cb, LIBUSB_LOG_CB_GLOBAL);
	libusb_set_option(NULL, LIBUSB_OPTION_LOG_LEVEL, LIBUSB_LOG_LEVEL_DEBUG);

	if (device_name) {
		r = test_wrapped_device(device_name);
	} else {
		for (size_t j = 0; j < 2; j++)
		{
			printf("get device list\n");
			cnt = libusb_get_device_list(NULL, &devs);
			if (cnt < 0) {
				libusb_exit(NULL);
				return 1;
			}

			for (i = 0; devs[i]; i++)
			{
				if (j == 0 && devs[i]->device_descriptor.idVendor == 0x13B9)
				{
					stored_device[index++] = devs[i];
					printf("ref device\n");
					libusb_ref_device(devs[i]);
				}
			}

			printf("free device list\n");
			libusb_free_device_list(devs, 1);
		}
	}

	printf("open hexagon devices\n");
	for (size_t t = 0; t < index; t++)
	{
		r = libusb_open(stored_device[t], &handle_device[t]);
		if (r < 0)
		{
			printf("Failed to open device !");
			libusb_exit(NULL);
			return r;
		}
	}

	printf("unref devices\n");
	for (size_t t = 0 ; t<index ; t++)
	{
		libusb_close(handle_device[t]);
		libusb_unref_device(stored_device[t]);
	}

	libusb_exit(NULL);
	return r;
}
```

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -449,6 +449,7 @@ if (cfg != desired)
   * - libusb_init_context()
   * - libusb_interrupt_event_handler()
   * - libusb_interrupt_transfer()
+  *	- libusb_keep_device()
   * - libusb_kernel_driver_active()
   * - libusb_lock_events()
   * - libusb_lock_event_waiters()
@@ -472,6 +473,7 @@ if (cfg != desired)
   * - libusb_transfer_get_stream_id()
   * - libusb_transfer_set_stream_id()
   * - libusb_try_lock_events()
+  *	- libusb_unkeep_device()
   * - libusb_unlock_events()
   * - libusb_unlock_event_waiters()
   * - libusb_unref_device()
@@ -1335,16 +1337,59 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
 
 		libusb_unref_device(dev->parent_dev);
 
-		if (usbi_backend.destroy_device)
-			usbi_backend.destroy_device(dev);
+		if (usbi_atomic_load(&dev->keep_device) == 0) {
+			if (usbi_backend.destroy_device)
+				usbi_backend.destroy_device(dev);
 
-		if (!libusb_has_capability(LIBUSB_CAP_HAS_HOTPLUG)) {
-			/* backend does not support hotplug */
-			usbi_disconnect_device(dev);
+			if (!libusb_has_capability(LIBUSB_CAP_HAS_HOTPLUG)) {
+				/* backend does not support hotplug */
+				usbi_disconnect_device(dev);
+			}
+
+			for (int idx = 0; idx < LIBUSB_DEVICE_STRING_COUNT; ++idx) {
+				free(dev->device_strings_utf8[idx]);
+			}
+
+			free(dev);
 		}
+	}
+}
 
-		for (int idx = 0; idx < LIBUSB_DEVICE_STRING_COUNT; ++idx) {
-			free(dev->device_strings_utf8[idx]);
+/** \ingroup libusb_dev
+ * Keep the reference of a device.
+ * \param dev the device to reference
+ * \returns the same device
+ */
+libusb_device* LIBUSB_CALL libusb_keep_device(libusb_device* dev)
+{
+	usbi_atomic_store(&dev->keep_device, 1);
+
+	if (dev->parent_dev) {
+		libusb_keep_device(dev->parent_dev);
+	}
+
+	return dev;
+}
+
+/** \ingroup libusb_dev
+ * Stop keeping the reference of a device. If the reference count
+ * is zero, the device shall be destroyed.
+ * \param dev the device to unkeep
+ */
+void API_EXPORTED libusb_unkeep_device(libusb_device* dev)
+{
+	long refcnt;
+
+	usbi_atomic_store(&dev->keep_device, 0);
+
+	refcnt = usbi_atomic_load(&dev->refcnt);
+	assert(refcnt >= 0);
+
+	if (refcnt == 0) {
+		usbi_dbg(DEVICE_CTX(dev), "destroy device %d.%d", dev->bus_number, dev->device_address);
+
+		if (dev->parent_dev) {
+			libusb_unkeep_device(dev->parent_dev);
 		}
 
 		free(dev);
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1714,8 +1714,10 @@ ssize_t LIBUSB_CALL libusb_get_device_list(libusb_context *ctx,
 	libusb_device ***list);
 void LIBUSB_CALL libusb_free_device_list(libusb_device **list,
 	int unref_devices);
-libusb_device * LIBUSB_CALL libusb_ref_device(libusb_device *dev);
-void LIBUSB_CALL libusb_unref_device(libusb_device *dev);
+libusb_device* LIBUSB_CALL libusb_ref_device(libusb_device* dev);
+void LIBUSB_CALL libusb_unref_device(libusb_device* dev);
+libusb_device* LIBUSB_CALL libusb_keep_device(libusb_device* dev);
+void LIBUSB_CALL libusb_unkeep_device(libusb_device* dev);
 
 int LIBUSB_CALL libusb_get_device_string(libusb_device *dev,
 	enum libusb_device_string_type string_type, char *data, int length);
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -508,6 +508,7 @@ static inline void usbi_end_event_handling(struct libusb_context *ctx)
 }
 
 struct libusb_device {
+	usbi_atomic_t keep_device;
 	usbi_atomic_t refcnt;
 
 	struct libusb_context *ctx;


minicarpet:
I am not sure how to build for Linux-arm64 -> I thought clang64 aarch64 were building for Linux, but it seems not the case, I can't find out the procedure.

libusb/core.c
```diff
@@ -767,6 +767,55 @@ void usbi_disconnect_device(struct libusb_device *dev)
...
+
+	return dev;
+}
+
+void usbi_unref_device(libusb_device* dev)
+{
+	long refcnt = -1;
+
+	if (!dev)
+		return;
```
seanm: usbi_unref_device(NULL) does nothing, but usbi_ref_device(NULL) will derefernce NULL and crash. Is there a reason for this asymmetry?  (Oh, I just noticed after writing this it was already like that...)

libusb/core.c
```diff
@@ -767,6 +767,55 @@ void usbi_disconnect_device(struct libusb_device *dev)
...
+
+	if (usbi_atomic_load(&dev->refcnt) > 0) { /* decrement only if positive value */
+		refcnt = usbi_atomic_dec(&dev->refcnt);
```
seanm: This looks like a race condition to me... some other thread could change `dev->refcnt` between the load and decrement. Given the use of atomics, I suppose these functions are meant to be thread-safe?

libusb/core.c
```diff
@@ -767,6 +767,55 @@ void usbi_disconnect_device(struct libusb_device *dev)
...
+	if (refcnt == 0) {
+		usbi_unref_device(dev->parent_dev);
+		usbi_destroy_device(dev);
+	}
+}
+
+
+void usbi_destroy_device(libusb_device* dev)
+{
+	if (!dev->keep_device) {
```
seanm: All accesses to `keep_device` surely must use the atomic functions.

libusb/core.c
minicarpet: Yes, that's true. I did this because I got some refcnt value under 0 (because some parent refcnt are not incremented in case of alloc_transfer and maybe libusb_open.
The question is: should I really reworked the way refcnt is managed, and modify both fonctions? (Ie: increment each parent when incrementing device, but in such case I have to remove most of parent incrementing call in the code, especially in get_device_list) or do I wrap this into thread safe operation?

libusb/core.c
minicarpet: I tried to keep as today these fonctions

minicarpet:
@seanm I've made change regarding your comment.
Now few questions regarding this:
- regarding refcnt: sometimes it went to negative value (I have no reason for now) and this is why I Check if higher than 0. Should I spend more Time to improve this ?
I wanted to not change so much these functions...
- should I test hotplug Windows first (before merging) to verify if the issue is here as well ?

libusb/os/threads_posix.h
```diff
@@ -39,7 +39,11 @@ static inline void usbi_mutex_static_unlock(usbi_mutex_static_t *mutex)
 typedef pthread_mutex_t usbi_mutex_t;
 static inline void usbi_mutex_init(usbi_mutex_t *mutex)
 {
-	PTHREAD_CHECK(pthread_mutex_init(mutex, NULL));
+	pthread_mutexattr_t attr;
+
+	PTHREAD_CHECK(pthread_mutexattr_init(&attr));
+	PTHREAD_CHECK(pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE));
```
hjelmn: Generally speaking, the need for recursive mutexes shows that there are structural issues in the code. They should not be used unless we fully understand why they are needed for these changes.

hjelmn (commented):
One very high level comment before I dive into this (which I may not get to today). Generally speaking, if you are going to change this much code all at once please break it down. Once commit should introduce the new internal ref/unref methods (making them just call the API methods) and change all internal usage to the internal version. The next commit can then make the proposed behavioral changes. As is I need significant additional time to just break down the two different changes before I can comment on the merits of the proposed change.

Changes to the core of libusb and API behavior (which this appears to do) will require multiple maintainers, considerations to the library version number, etc. Once the above comment is addressed (I have no problem with internal APIs that do the same thing as libusb_ref/unref_device) it will be easier to see exactly what is being proposed.

hjelmn:
Also note. You absolutely MAY NOT do this:

I kept libusb_ref_device & libusb_unref_device for compatibility

These methods have well defined semantics and can not be changed. They can be deprecated in favor of something else if desired.

libusb/os/threads_posix.h
seanm: And changing this here would make many many mutexes recursive, which we likely don't want at all.

minicarpet:
To avoid recursive mutex, and make sure parents are really deleted, I need to change really how refcnt works. Is that ok to change it, ie:
- inc it when really necessary + inc all parents
- dec all parent each time a device is `unref`

I will need some times to handle that really, due to discover mode it is a little bit tricky.

I will spend some times next week to do that.

mcuee:
Thanks for working on Issue #1165. I think it is another long standing issues for Windows.
* #1165

PR #1406, opened 2023-12-28
windows: hotplug implementation

sonatique:
Hotplug implementation for Windows.

The idea is to use WM_DEVICECHANGE message to trigger a re-enumeration, then figure out which device(s) have been removed or added and trigger corresponding changes in ctx internal device list as well as relevant hotplug events.

I tried to minimize changes on existing code, basically reusing well tested main windows_winusb and windows_usbdk "get_device_list" implementation. I had to do small changes though, but without putting this code at risk, I think.

I tried to avoid re-enumerating on WM_DEVICECHANGE notifications, but didn't figure out how to get all required information about the triggering device. And anyway, all my attempts were actually doing a full enumeration, and I realized I was writing "get_device_list" again, but without taking all subtleties into account, so I reverted using existing code as much as possible.

As far as I can tell this approach works well, hotplugtest works as expected and I didn't see any issue so far.

Note that I used the approach of having a "hidden window" and a message pump in a dedicated thread. I tried to somehow mimic how it is done for linux, but dealing with Windows API. I understand having a "hidden Windows" might sound weird but in the end I didn't find any major drawback and we probably can consider this an implementation detail.

Now that I have a working solution as a reference, I will explore whether [CM_Register_Notification](https://learn.microsoft.com/en-us/windows/win32/api/cfgmgr32/nf-cfgmgr32-cm_register_notification) could be used to improve code and avoid the hidden window, as suggested here: https://github.com/libusb/libusb/issues/86#issuecomment-1724793991

I am totally OK to consider this PR only as a basis for discussion toward the optimal solution, but I thought it would worth publishing it since it could already be useful as is.

Beside providing hotplug events, this implementation avoid fully enumerating devices at system level every time user calls "libusb_get_device_list" since we're reusing ctx internal device list. I think that this alone should provide some desired improvements.

I had a hard time trying to comply with both coding style of Windows C and existing libusb-1.0 code for the Windows backend. I took some liberty in the new windows_hotplug.c file.

Let me know how all this works for you, I greatly welcome feedback regarding the code itself or any issue you may report when testing this code on your system. Thank you in advance!

PR #1688, opened 2025-08-28
fix: wrong PreprocessorDefinitions were set for hotplug config

minicarpet:
PreprocessorDefinitions used WINDOWS_HOTPLUG instead of LIBUSB_WINDOWS_HOTPLUG


sonatique:
@minicarpet , oups, thanks a lot, I missed that one when I renamed it, sorry.
I'll merge this asap as a follow up to #1406

sonatique (approved):

PR #1690, opened 2025-08-29
android: Fix intermittent open failure in get_usbfs_fd()

lukaszbg:
On Android get_usbfs_fd() fails intermittently because open retry delay set to 10ms is not always enough. This results in: libusb: debug [libusb_open] open 2.3
libusb: error [get_usbfs_fd] File doesn't exist, wait 10 ms and try again libusb: error [get_usbfs_fd] libusb couldn't open USB device /dev/bus/usb/002/003, errno=2 
libusb: debug [op_open] open failed with no device, but device still attached libusb: debug [libusb_open] open 2.3 returns -4

When experimenting with retry delay I was able to determine
that on Android it can be up to ~20ms.

This fix retries open 3 times instead of 1.

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -201,19 +201,25 @@ static int get_usbfs_fd(struct libusb_device *dev, int access_mode, int silent)
 	if (fd != -1)
 		return fd; /* Success */
 
+	/* TODO: fix race between netlink and usbfs https://github.com/libusb/libusb/issues/1691 */
 	if (errno == ENOENT) {
 		const long delay_ms = 10L;
 		const struct timespec delay_ts = { 0L, delay_ms * 1000L * 1000L };
+		uint8_t retry = 3;
 
-		if (!silent)
-			usbi_err(ctx, "File doesn't exist, wait %ld ms and try again", delay_ms);
+		while (retry-- > 0) {
+			if (!silent)
+				usbi_err(ctx, "File doesn't exist, wait %ld ms and try again", delay_ms);
 
-		/* Wait 10ms for USB device path creation.*/
-		nanosleep(&delay_ts, NULL);
+			/* Wait 10ms for USB device path creation.*/
+			nanosleep(&delay_ts, NULL);
 
-		fd = open(path, access_mode | O_CLOEXEC);
-		if (fd != -1)
-			return fd; /* Success */
+			fd = open(path, access_mode | O_CLOEXEC);
+			if (fd != -1)
+				return fd; /* Success */
+			if (errno != ENOENT)
+				break;
+		}
 	}
 
 	if (!silent) {


hjelmn:
Looks reasonable to me.

hjelmn (approved):
This looks fine to me. Retries the open if the dev file is not yet there.

hjelmn:
@tormodvolden Since this comes from my own organization please double check and let me know if you have any concerns.

tormodvolden:
This seems reasonable, just extending the existing retry that was probably experimentally determined.

The reason for the original retry was a race between netlink and the kernel, commit c59d574b. And that reminds me what @benzea proposes in https://github.com/libusb/libusb/pull/1067/commits/36a56585987f4036748232496d41b389e47f2255 PR #1067 (one of the commits that are not only static analyzer pleasing). But that's for udev which I believe is not used in Android.

tormodvolden:
It would make sense to mention "intermittent" in the commit summary as well.

lukaszbg:
> It would make sense to mention "intermittent" in the commit summary as well.

I updated the commit title. Thanks.

tormodvolden:
I don't see it changed in the commit though? Forgot to push?

benzea:
I really don't understand why a race would exist. I mean, why would the kernel send a netlink event before creating the device node?

lukaszbg:
> I don't see it changed in the commit though? Forgot to push?

I updated PR title instead of commit :/. Now the commit title is updated as well.

benzea:
> I really don't understand why a race would exist. I mean, why would the kernel send a netlink event before creating the device node?

Well, OK, not completely true. I see various possible explanations:
 1. We are listening for the wrong event. e.g. we need to wait for the usbdev driver to be bound or something.
 2. The event is correct, but Android has something similar to udev and device permissions are not yet updated. But that would again suggest that libusb is listening to the wrong event stream.
 3. There is a kernel bug.

Said differently, I consider a retry like this a really bad workaround. I am not surprised that it is feasible, but most likely there is a bug elsewhere that needs to be fixed. Someone should really look at how this is supposed to work and why this is going wrong sometimes. I listed some possibilities above, but there may be others.

This is not a full NACK on the patch, but I am saying that the goal must be to remove the loop/sleep entirely.

libusb/os/linux_usbfs.c
```diff
@@ -204,16 +204,19 @@ static int get_usbfs_fd(struct libusb_device *dev, int access_mode, int silent)
 	if (errno == ENOENT) {
 		const long delay_ms = 10L;
 		const struct timespec delay_ts = { 0L, delay_ms * 1000L * 1000L };
+		uint8_t retry = 3;
 
-		if (!silent)
-			usbi_err(ctx, "File doesn't exist, wait %ld ms and try again", delay_ms);
+		while (retry-- > 0) {
+			if (!silent)
+				usbi_err(ctx, "File doesn't exist, wait %ld ms and try again", delay_ms);
```
benzea, line 212: Could you please log errno (ideally store it on error and log it from that variable). "File doesn't exist" might be correct, but it could also be a permission error and I think it would be good if one is able to tell the difference.

benzea: Ah, I realise that this *is* ENOENT in the first loop iteration.

In that case, maybe the loop should break if errno is != ENOENT?

hjelmn:
> > I really don't understand why a race would exist. I mean, why would the kernel send a netlink event before creating the device node?
> 
> Well, OK, not completely true. I see various possible explanations:
> 
> 1. We are listening for the wrong event. e.g. we need to wait for the usbdev driver to be bound or something.
> 2. The event is correct, but Android has something similar to udev and device permissions are not yet updated. But that would again suggest that libusb is listening to the wrong event stream.
> 3. There is a kernel bug.
> 
> Said differently, I consider a retry like this a really bad workaround. I am not surprised that it is feasible, but most likely there is a bug elsewhere that needs to be fixed. Someone should really look at how this is supposed to work and why this is going wrong sometimes. I listed some possibilities above, but there may be others.
> 
> This is not a full NACK on the patch, but I am saying that the goal must be to remove the loop/sleep entirely.

I am not too familiar with Android but I remember there being some weirdness with USB devices due to custom software. It might be that there is a bug in either libusb's Android support or something in Android itself. I will let the original author speak to why this is needed. From my perspective it is harmless so not a big deal. If we suspect this is a bug in libusb we can open a bug to track fixing it and add a todo here referencing it.

Thanks for taking a look, the extra eyes help a lot.

libusb/os/linux_usbfs.c
lukaszbg: I added code to exit the while loop earlier if error is not ENOENT. Thanks

lukaszbg:
>>>I really don't understand why a race would exist. I mean, why would the kernel send a netlink event before creating the device node?
>>
>>Well, OK, not completely true. I see various possible explanations:
>>
>>We are listening for the wrong event. e.g. we need to wait for the usbdev driver to be bound or something.
The event is correct, but Android has something similar to udev and device permissions are not yet updated. But that would again suggest that libusb is listening to the wrong event stream.
There is a kernel bug.
Said differently, I consider a retry like this a really bad workaround. I am not surprised that it is feasible, but most likely there is a bug elsewhere that needs to be fixed. Someone should really look at how this is supposed to work and why this is going wrong sometimes. I listed some possibilities above, but there may be others.
>>
>>This is not a full NACK on the patch, but I am saying that the goal must be to remove the loop/sleep entirely.
>
>I am not too familiar with Android but I remember there being some weirdness with USB devices due to custom software. It might be that there is a bug in either libusb's Android support or something in Android itself. I will let the original author speak to why this is needed. From my perspective it is harmless so not a big deal. If we suspect this is a bug in libusb we can open a bug to track fixing it and add a todo here referencing it.
>
>Thanks for taking a look, the extra eyes help a lot.

This race does not seem to be android specific but a general linux trait. What I found is that on android it might take longer for the usbfs to be updated with device entry.  That said I can investigate the issue further however not now as I don't have bandwidth available. For now I submitted https://github.com/libusb/libusb/issues/1691 to track it.

PR #1692, opened 2025-09-04
Add GitHub Actions workflow for Android NDK build

alexbirkett:
Add a new GitHub Actions workflow that builds libusb for Android using the existing android/jni + ndk-build setup.

✅ What it does:

Builds libusb for the following Android ABIs:

arm64-v8a

armeabi-v7a

x86_64

Uses API level 21 (minimum 64-bit-compatible level).

Uploads the resulting .so file (and any built samples/tests) from libs/<ABI>/ as build artifacts per ABI.

NDK version is pinned to r27d, downloaded and cached automatically.

📦 Output:

Artifacts are named like:

libusb-android-arm64-v8a-api21/
└── libusb1.0.so

🔧 Implementation notes:

No changes to the source code or build system — this uses the existing Android.mk setup.

This is intended for CI validation and distributing prebuilt .so files for


.github/workflows/android.yml
```diff
@@ -0,0 +1,70 @@
...
+          set -euxo pipefail
+          mkdir -p .ndk
+          curl -L -o ndk.zip "https://dl.google.com/android/repository/android-ndk-${NDK_CHANNEL}-linux.zip"
+          unzip -q ndk.zip -d .ndk
+          rm ndk.zip
+
+      - name: Set NDK env + PATH
+        run: |
+          echo "NDK=${GITHUB_WORKSPACE}/.ndk/android-ndk-${NDK_CHANNEL}" >> $GITHUB_ENV
+          echo "${GITHUB_WORKSPACE}/.ndk/android-ndk-${NDK_CHANNEL}/toolchains/llvm/prebuilt/linux-x86_64/bin" >> $GITHUB_PATH
```
Youw: Do you think it worth bothering with all this "use external cached NDK" instead of using the one available on `ubuntu-latest` under `ANDROID_NDK_ROOT`?

Since you explicitly specify the `ANDROID_ABI`/`ANDROID_API` - that shoud be enough for as long as `ANDROID_NDK_ROOT` is up-to-date enough (and right now it is [`27.3.13750724`](https://github.com/libusb/libusb-cmake/actions/runs/17261381779/job/48983433050)).

I had no issues using it this was for a while: https://github.com/libusb/libusb-cmake/blob/000e06eb241927db0f34e6867637d39b849ab989/.github/workflows/builds.yml#L219

Youw (approved):
Other than my last comment - looks like a clear improvement, even if we don't immediately use the artifacts produced by these buids.

.github/workflows/android.yml
```diff
@@ -0,0 +1,70 @@
+name: Android
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build-android:
```
Youw: to be consistent with other builds:
```suggestion
on: [push, pull_request]

jobs:
  build:
```

.github/workflows/android.yml
alexbirkett: That's interesting. I did not know that the NDK was available under `ANDROID_NDK_ROOT` I will investigate

Youw: As of this moment: https://github.com/actions/runner-images/blob/57f0dc0e78cf05825b63f0a4b63932fa93b4ba3c/images/ubuntu/Ubuntu2404-Readme.md#android

alexbirkett:
I want the artifacts so that I don't have to pass around unattributable artifacts built on my local machine.

.github/workflows/android.yml
alexbirkett: `ANDROID_NDK_ROOT` seems to work @Youw  https://github.com/alexbirkett/libusb/actions/runs/17467661897

Youw (approved):
LGTM

Now we wait for other libusb admin(s) to merge this.

Mr-Bossman:
@alexbirkett Thanks so much for this! It would be nice if you squash the commits though. Otherwise, LGTM.

alexbirkett:
> @alexbirkett Thanks so much for this! It would be nice if you squash the commits though. Otherwise, LGTM.

great! squashed now

mcuee:
I like this kind of CI improvement.

And when it comes to such type of CI, I will tend to trust the review from @Youw.

mcuee (approved):

mcuee:
@sonatique @seanm and @tormodvolden 

I think this is safe to merge and include in libusb 1.0.30 release as it does not affect other things. Any objections?

tormodvolden:
Yeah, CI stuff doesn't affect the release files so this can go in anytime.

Thanks for your contribution @alexbirkett !

PR #1695, opened 2025-09-06
darwin: add missing mutex lock/unlock for claimed_interfaces

seanm:
Existing code comments say that
libusb_device_handle.claimed_interfaces is protected by libusb_device_handle.lock.  The function ep_to_pipeRef() accesses claimed_interfaces but does not hold the lock.

I examined all 9 callers:

- darwin_alloc_streams
- darwin_clear_halt
- darwin_free_streams
- darwin_abort_transfers
- submit_bulk_transfer
- submit_control_transfer
- submit_iso_transfer
- submit_stream_transfer
- darwin_async_io_callback

and none seem to hold the lock already.

Found with -Wthread-safety.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -477,6 +477,7 @@ static int ep_to_pipeRef(struct libusb_device_handle *dev_handle, uint8_t ep, ui
 
   usbi_dbg (ctx, "converting ep address 0x%02x to pipeRef and interface", ep);
 
+  usbi_mutex_lock(&dev_handle->lock);
   for (iface = 0 ; iface < USB_MAXINTERFACES ; iface++) {
     cInterface = &priv->interfaces[iface];
 
@@ -492,6 +493,7 @@ static int ep_to_pipeRef(struct libusb_device_handle *dev_handle, uint8_t ep, ui
             *interface_out = cInterface;
 
           usbi_dbg (ctx, "pipe %d on interface %d matches", *pipep, iface);
+          usbi_mutex_unlock(&dev_handle->lock);
           return LIBUSB_SUCCESS;
         }
       }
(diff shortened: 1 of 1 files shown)


seanm:
@hjelmn does this look right to you?

seanm:
@osy could you take a look at this?

osy:
This is a fix for a read race on claimed_interface (not write race) right? It looks fine to me but just a heads up, when I looked at the Darwin backend, I found a lot of read races and just reasoned that all of them would result in eventual errors that is similar to if the device was just hot unplugged.

seanm:
@osy Yes reading. This change is as a result of a -Wthread-safety compiler warning. According to existing comments:

```
	/* lock protects claimed_interfaces */
	usbi_mutex_t lock;
	unsigned long claimed_interfaces;
```

`claimed_interfaces` should be protected by `lock`.

hjelmn:
I wouldn't be surprised if there are races with hot-unplug. Thanks for starting to take a look at this. My time is often 110% booked so I haven't had the time needed to do some of the housekeeping the darwin backend needs.

Let me review this and make sure it isn't running afoul of something else. It is probably ok.

hjelmn:
This looks technically correct, the best kind of correct.

mcuee:
@sonatique and @tormodvolden 

This may be good to be included in libusb-1.0.30 release since @hjelmn has reviewed it.

sonatique:
LGTM. Worth noting that the returned *interface_out pointer is used by callers after the lock is released, so there's still a potential race if the interface is released concurrently. But that's a pre-existing issue, not something for this PR.
In my opinion this is a good incremental improvement, but we're not done with this sort of issue in Darwin...

mcuee:
@tormodvolden 

Please review as well. Thanks.

seanm:
@sonatique svp just merge this.

PR #1696, opened 2025-09-06
linux: change usbi_handle_disconnect signature to take ctx

seanm:
Change the signature of private function usbi_handle_disconnect() to take the context as a parameter so that it can eventually use a Clang Thread Safety REQUIRES(ctx->event_lock) decoration.


seanm:
@sonatique what do you think?

sonatique (approved):

PR #1708, opened 2025-10-01
darwin: handle error from GetConfigurationDescriptorPtr

osy:
When a device is disconnected or otherwise `dpriv->device` is invalid, `GetConfigurationDescriptorPtr` will return `kIOReturnNoDevice` and `IOUSBConfigurationDescriptorPtr` will not be set. This means that we have an uninitialized pointer that is read from which is... very bad.


libusb/os/darwin_usb.c
```diff
@@ -2151,7 +2151,10 @@ static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, b
   cached_configurations = alloca (sizeof (*cached_configurations) * descriptor.bNumConfigurations);
 
   for (i = 0 ; i < descriptor.bNumConfigurations ; ++i) {
-    (*dpriv->device)->GetConfigurationDescriptorPtr (dpriv->device, i, &cached_configuration);
+    kresult = (*dpriv->device)->GetConfigurationDescriptorPtr (dpriv->device, i, &cached_configuration);
+    if (kresult != kIOReturnSuccess) {
+      return LIBUSB_ERROR_NOT_FOUND;
+    }
```
seanm, line 2158: You probably need `dpriv->in_reenumerate = false;` before this early return.

osy: Good catch

libusb/os/darwin_usb.c
```diff
@@ -2213,8 +2216,8 @@ static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, b
   }
 
   for (i = 0 ; i < descriptor.bNumConfigurations ; ++i) {
-    (void) (*dpriv->device)->GetConfigurationDescriptorPtr (dpriv->device, i, &cached_configuration);
-    if (memcmp (cached_configuration, cached_configurations + i, sizeof (cached_configurations[i])) != 0) {
+    kresult = (*dpriv->device)->GetConfigurationDescriptorPtr (dpriv->device, i, &cached_configuration);
+    if (kresult != kIOReturnSuccess || memcmp (cached_configuration, cached_configurations + i, sizeof (cached_configurations[i])) != 0) {
       usbi_dbg (ctx, "darwin/reenumerate_device: configuration descriptor %d changed", i);
       return LIBUSB_ERROR_NOT_FOUND;
```
seanm, line 2223: Also need `dpriv->in_reenumerate = false;` before this early return?  In fact, there are other early returns in that function I'm not sure about too...

osy: After the point where it logs `darwin/reenumerate_device: waiting for re-enumeration to complete...`, `darwin_devices_attached` must be called which sets `in_reenumerate = false` or we hit the point where it logs `darwin/reenumerate_device: timeout waiting for reenumerate` and that also sets `in_reenumerate = false`.

seanm: Ah yes, makes sense now.

prayasnog-droid:
Samsung galaxy 5g

prayasnog-droid:
utmapp:fix/darwin-config

sonatique:
@mcuee user @prayasnog-droid doesn't seem to post very relevant comment or PR. Any way we can block it before we get flooded?

mcuee:
Blocked.

sonatique:
thanks @mcuee : how did you do it?

mcuee:
> thanks @mcuee : how did you do it?

You can try to follow the steps here next time.
https://docs.github.com/en/communities/maintaining-your-safety-on-github/blocking-a-user-from-your-organization

mcuee:
@tormodvolden 

This PR has been approved by @seanm and @sonatique, maybe it is worth merging for libusb-1.0.30 release.

seanm:
+1 to merging, this looks pretty conservative

tormodvolden:
Yes, thanks, this looks safe and very good.

PR #1713, opened 2025-10-20
Add null checks

wws13:
Fixing a crash if a sandboxed app doesn't have access to the USB drive

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -2143,7 +2143,7 @@ static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, b
 
   struct libusb_context *ctx = HANDLE_CTX (dev_handle);
 
-  if (!dpriv || !dpriv->device) {
+  if (!dpriv->device) {
     usbi_warn(ctx, "darwin_reenumerate_device: device interface is NULL");
     return LIBUSB_ERROR_NO_DEVICE;
   }


libusb/os/darwin_usb.c
```diff
@@ -2139,6 +2143,11 @@ static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, b
 
   struct libusb_context *ctx = HANDLE_CTX (dev_handle);
 
+  if (!dpriv || !dpriv->device) {
```
seanm: The check for `!dpriv` is not necessary, dpriv is already dereferenced 10 lines before, so would have crashed already at this point.

seanm:
lgtm

I think if you squash all the commits the warning about 'Commits must have verified signatures.' would go away. But if you don't know how to do that, no problem, I think we can squash on merge.

seanm:
@sonatique this should be safe to merge too I think.

tormodvolden:
Hi @wws13, would you like to share your full name, for inclusion in the AUTHORS file?

tormodvolden:
All, do you think we might have a problem elsewhere in the code, since the user ends up with a device handle for which there is no priv device? Should it have been filtered out earlier before coming that far? It is of course fine to have these null checks, but it could be indicative of something else being broken.

wws13:
> Hi @wws13, would you like to share your full name, for inclusion in the AUTHORS file?

Hi! Thanks for asking. Since it was just a minor fix, no need to add my full name to the AUTHORS file. Appreciate the offer though!

> All, do you think we might have a problem elsewhere in the code, since the user ends up with a device handle for which there is no priv device? Should it have been filtered out earlier before coming that far? It is of course fine to have these null checks, but it could be indicative of something else being broken.

Yes, this is a fix for the symptom, not the root cause. Unfortunately, I can't reproduce this crash. I fixed it based on user-reported crashes, and with this fix, there were fewer crashes. Since the crash only occurs on the macOS app in the sandbox, it's possible that the system is preventing some devices from opening, but this hasn't been verified.

Issue #926, opened 2021-06-11
Uwp (Windows.Devices.Usb) backend for libusb


ankurverma85:
The following documentation illustrates the apis available for interfacing with usb devices from uwp processes 

https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/talking-to-usb-devices-start-to-finish#step6

Its should theoretically be possible to use cppwinrt projections of these and bridge a c++ library as a libusb backend

This would not only modernize the windows support for libusb but also enable support for using the library on non desktop based windows hardware like Xbox and hololens etc.

mcuee:
Thanks. If you can implement a prototype and create a pull-request, that will be very welcome.

ankurverma85:
I'm a complete novice when it comes to the USB internals but I'd be happy to setup the scaffolding around Creating C++ static/library that uses cppwinrt projections to interact with the apis and serve as the initial seed.
We dont actually need any dot-net .
Should be simple enough to do a purely c++ <-> c libusb-backend interfacing. 

if someone can help out with the actual code to bridge the two together I can definitely setup a branch to get things started.

ankurverma85:
I did a basic cmake based scaffolding that tries to follow the examples and query all usb devices by implemeting the usbi backend call here:

https://github.com/ankurverma85/libusb/tree/uwp

To build:
If you have VS2019 you can open the folder cmake/ in there and run the test.exe to 
The CMakeLists.txt included is configured to only use the uwp backed (in uwp.cpp) on windows, so you wont get the other windows backends.

mcuee:
Thanks for the quick start. This will probably not be a simple effort.

I tried your branch, initially cmake-gui complains missing config.h.in, so I copy config.h from the msvc directory to the root directory and also the CMakeLists.txt file. Then cmake process is okay for VS2019.

However I can not build the solution.

```
Build started...
1>------ Build started: Project: usb-1.0, Configuration: Debug x64 ------
1>core.c
1>C:\work\libusb\libusb_uwp\libusb\libusbi.h(35,10): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory
1>descriptor.c
1>C:\work\libusb\libusb_uwp\libusb\libusbi.h(35,10): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory
1>hotplug.c
[...]

PR #1721, opened 2025-11-13
Windows WinRT/UWP Implementation

Tails86:
Relates to: https://github.com/libusb/libusb/issues/926

This branch accesses USB through WinRT which is compatible with Windows desktop as well as UWP. I tested this implementation mainly using WinUSB devices which it works best for. There are a few limitations to this implementation:
- WinRT requires C++17, so I figured that I may as well just use C++ standard library for threading and event handling to avoid Windows.h since much of the existing Windows platform code will not compile for UWP
- Devices with container ID of `{00000000-0000-0000-ffff-ffffffffffff}` (system device) or `{00000000-0000-0000-0000-000000000000}` (invalid) are ignored 
- The WinRT classes don't specifically call out isochronous pipes. I'm assuming that these pipes are accessed through the interrupt interfaces. I wasn't able to test this though.
- WinRT accesses USB devices through interfaces, and at least 1 interface must be available for connection in order to do the proper control transfers to list the device 
    - Side note: the device descriptor information provided by WinRT was inaccurate when I tried using them which is why a control transfer is required
- This should be able to access HID devices, but of course Windows denies access if they are already in use by a driver. I have not tested HID device access myself.
- Trying to set USB configuration will probably fail, but a best attempt is made to do so through a control transfer. There doesn't seem to be WinRT function calls for this.
- Hotplug is not implemented in this branch, but it would be feasible to do so in the future


sonatique (dismissed):

Tails86:
Apologies for the pipeline issues. I forgot to run through it all before making the pull request. It should all be set now.

Tails86:
I added signatures to all commits to remove the "Commits must have verified signatures." error

mcuee:
> I added signatures to all commits to remove the "Commits must have verified signatures." error

Thanks. This is not necessary though. You can just ignore the error.

Tails86:
I've discovered a couple issues after I've done some more extensive testing:
- I've run into an edge case where the function given to `asyncOp.Completed()` never gets executed, even when the `asyncOp` has been canceled. As far as I can tell from [the documentation](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncoperationwithprogress-2.completed), this isn't expected, and I can't think of any workarounds for this if I can't detect that it's happening.
    - The documentation recommends using `co_await` instead of directly setting the callback. I didn't want to use this if I didn't need to because it requires C++20 instead of the current C++17. I'm a bit doubtful, but maybe using this would help somehow?
    - This bug prevents the application from closing as it's waiting forever for this process to complete.
    - I was testing on Win32 app with external Win32 interactions, so maybe it would work better on a pure WINDOWS_STORE application?
- The code relies on `std::async` to synchronously run an async operation if it is run on a GUI thread. A new thread is not always used though, and I do run into assertion errors when trying this in Debug.

The code still generally works outside of these edge cases. Would you rather hold off until these issues are resolved or merge and open up bug issues?

mcuee:
@Tails86 

Just wondering if you have made any progress here. Thanks.

Tails86:
Sorry, this fell off my radar. Give me another week to try and address these issues, and I will report back.

Tails86:
I'm not so sure what happened, but I can't seem to get the existing code to work like it used to. I'll have to circle back to this later, so I'm closing this for now. Of course this code can be referenced if anyone else wants to take a stab at it.

mcuee:
> I'm not so sure what happened, but I can't seem to get the existing code to work like it used to. I'll have to circle back to this later, so I'm closing this for now. Of course this code can be referenced if anyone else wants to take a stab at it.

I see. No problem to close it for now. Take you time to revive it.

PR #1728, opened 2025-12-30
Replace remaining strcpy() with snprintf()

seanm:
strcpy() is dangerous because it's too easy to overrun a buffer.  snprintf is safer because the buffer size can be specified. Also, strcpy() is not allowed with -fbounds-safety.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -13,7 +13,6 @@ clang-analyzer-*,\
 -clang-analyzer-deadcode.DeadStores,\
 -clang-analyzer-optin.portability.UnixAPI,\
 -clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,\
--clang-analyzer-security.insecureAPI.strcpy,\
 -clang-analyzer-unix.Malloc,\
 misc-*,\
 -misc-no-recursion,\
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2917,7 +2917,7 @@ static void log_str(enum libusb_log_level level, const char *str)
 static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
 	const char *function, const char *format, va_list args)
 {
-	const char *prefix;
+	const char *prefix = NULL;
 	char buf[USBI_MAX_LOG_LEN];
 	int global_debug, header_len, text_len;
 	static int has_debug_header_been_displayed = 0;
@@ -2944,6 +2944,7 @@ static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
 
 	switch (level) {
 	case LIBUSB_LOG_LEVEL_NONE:	/* Impossible, but keeps compiler happy */
+		assert(0);
 		return;
 	case LIBUSB_LOG_LEVEL_ERROR:
 		prefix = "error";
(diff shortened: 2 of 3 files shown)


libusb/core.c
```diff
@@ -2959,24 +2959,38 @@ static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
...
-		/* Truncated log output. On some platforms a -1 return value means
-		 * that the output was truncated. */
-		text_len = (int)sizeof(buf) - header_len;
+	/* starting right after the end of the header, append the rest, but leave room for the terminator. */
+	int space = (int)sizeof(buf) - header_len - (int)strlen(USBI_LOG_LINE_END);
+	text_len = vsnprintf(buf + header_len, space, format, args);
+
+	if (text_len < 0) {
+		/* If somehow an error occurred, give up. */
+		return;
```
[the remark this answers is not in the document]
seanm: Returning -1 for truncation is pre-C99 behaviour, and seems maybe also with MSVC ≤ 2013.  What's libusb's minimum MSVC now?

libusb/core.c
```diff
@@ -2959,24 +2959,38 @@ static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
 			"libusb: %s [%s] ", prefix, function);
 	}
 
-	if (header_len < 0 || header_len >= (int)sizeof(buf)) {
-		/* Somehow snprintf() failed to write to the buffer,
+	if (header_len < 0) {
+		/* If somehow an error occurred, give up. */
+		return;
```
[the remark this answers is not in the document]
seanm: Done.

libusb/core.c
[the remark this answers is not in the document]
sonatique: Depends on: https://github.com/libusb/libusb/pull/1784 but I would say > 2015

libusb/core.c
```diff
@@ -2959,24 +2960,38 @@ static void log_v(struct libusb_context *ctx, enum libusb_log_level level,
 			"libusb: %s [%s] ", prefix, function);
 	}
 
-	if (header_len < 0 || header_len >= (int)sizeof(buf)) {
-		/* Somehow snprintf() failed to write to the buffer,
+	if (header_len < 0) {
```
sonatique: Could you remind me why you prefer to have the 2 ifs instead of 1 ? There was a reason linked to a compiler check but I don't remember.

libusb/core.c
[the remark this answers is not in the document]
mcuee: Yes, I agree we can drop VS2013/VS2015 support after 1.0.30 release.

seanm:
This is ready for re-review, I made a few tweaks.

.clang-tidy
```diff
@@ -13,7 +13,6 @@ clang-analyzer-*,\
 -clang-analyzer-deadcode.DeadStores,\
 -clang-analyzer-optin.portability.UnixAPI,\
 -clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,\
--clang-analyzer-security.insecureAPI.strcpy,\
```
sonatique, line 16: I am not familiar with how this work, but shouldn't we let it on to avoid someone introducing strcpy again later?

sonatique: ok got, it, it's the other way round, you enabled it

sonatique (approved):

libusb/core.c
seanm: There's no reason, it's an artifact from the first one just returning. I could merge them...

seanm: Done.

seanm:
@Youw do you want to give this one last look over before I merge?

Youw (approved):
LGTM

PR #1729, opened 2025-12-31
Convert various macros to functions

seanm:
Macros are well-known footguns. Functions provide type safety.

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -576,10 +576,10 @@ void usbi_get_real_time(struct timespec *tp);
  * 3. struct libusb_transfer (which includes iso packets) [variable size]
  *
  * You can convert between them with the functions:
- *  TRANSFER_PRIV_TO_USBI_TRANSFER
- *  USBI_TRANSFER_TO_TRANSFER_PRIV
- *  USBI_TRANSFER_TO_LIBUSB_TRANSFER
- *  LIBUSB_TRANSFER_TO_USBI_TRANSFER
+ *  usbi_transfer_priv_to_usbi_transfer
+ *  usbi_transfer_to_transfer_priv
+ *  usbi_transfer_to_libusb_transfer
+ *  usbi_libusb_transfer_to_usbi_transfer
  */
 
 struct usbi_transfer {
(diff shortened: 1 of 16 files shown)


libusb/libusbi.h
```diff
@@ -1549,6 +1460,66 @@ struct usbi_os_backend {
 
 extern const struct usbi_os_backend usbi_backend;
 
+static inline struct usbi_transfer * TRANSFER_PRIV_TO_USBI_TRANSFER(void * transfer_priv)
```
Youw: I cannot be sure what guidelines/best practices we're trying to follow for libusb, be by my own habits I always prohibit using `UPPER_SNAKE_CASE` in C/C++ for anything else except macros. Having it for a variable/function will get a very weird and hard to debug compilation issue in case if this name would leak from somewhere as a macro.

Youw (commented):
Can you update the PR description with the motivation for this change?

libusb/libusbi.h
seanm: In my own code I also use `UPPER_SNAKE_CASE` only for preprocessor stuff.

I'm not sure if libusb has any official standards for such.

I left the names the same only so that all the call sites didn't need to change too.  But I can do that if that's what we think is best.

seanm:
> Can you update the PR description with the motivation for this change?

The motivation is the usual one when it comes to C macros: no type safety, easy footguns, etc. etc. Any compiler today will certainly inline a 1 line macro, so its only advantage is gone.

libusb/libusbi.h
mcuee: I agree with @Youw here.

Youw:
> > Can you update the PR description with the motivation for this change?
> 
> The motivation is the usual one when it comes to C macros: no type safety, easy footguns, etc. etc. Any compiler today will certainly inline a 1 line macro, so its only advantage is gone.

I totally get that, but I generally encourage people to be disciplined-enough to write such things down (specially in the PR description) for other developers who may have different point of view, or may try to assume things like "maybe this is a preparation for something specific, and not just a refactoring/improvement".

libusb/libusbi.h
Youw: > I'm not sure if libusb has any official standards for such.

We're currently maitaining it, we better bring our best standards for the sake of the project long-term development.

---
It is hard to tell how big the diff will be, but I'd rather try and see.

seanm: OK I added another commit with just the renaming

Youw:
This highlights one more question from my side: should all these functins have a libusb-specific prefix, e.g.: `handle_ctx` -> `libusbi_handle_ctx`?
I get it gets very verbose but still, something worth discussing since we're considering renaming and already here.

seanm:
> This highlights one more question from my side: should all these functins have a libusb-specific prefix

Looking at core.c, many do, but not all.  I'd leave that for some other PR or discussion...

mcuee (approved):

mcuee:
Ah, one missing file to update.
https://github.com/libusb/libusb/actions/runs/26889358753/job/79310609437?pr=1729

```
/src/libusb_fuzzer.cc:44:37: error: use of undeclared identifier 'LIBUSB_TRANSFER_TO_USBI_TRANSFER'
   44 |   struct usbi_transfer *itransfer = LIBUSB_TRANSFER_TO_USBI_TRANSFER(transfer);
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

seanm:
> /src/libusb_fuzzer.cc:44:37: error: use of undeclared identifier 'LIBUSB_TRANSFER_TO_USBI_TRANSFER'
>    44 |   struct usbi_transfer *itransfer = LIBUSB_TRANSFER_TO_USBI_TRANSFER(transfer);

That file is not in the libusb repo, it seems to be from here: https://github.com/google/oss-fuzz/blob/c09755737ee2d135fb003b0fb02cc79e5c7c4fdb/projects/libusb/libusb_fuzzer.cc#L21

they include `libusbi.h` which isn't really a public header, but of course handy to be able to fuzz more functions.

Perhaps this is an argument for *not* renaming these macros/functions.

mcuee:
> > /src/libusb_fuzzer.cc:44:37: error: use of undeclared identifier 'LIBUSB_TRANSFER_TO_USBI_TRANSFER'
> > 44 |   struct usbi_transfer *itransfer = LIBUSB_TRANSFER_TO_USBI_TRANSFER(transfer);
> 
> That file is not in the libusb repo, it seems to be from here: https://github.com/google/oss-fuzz/blob/c09755737ee2d135fb003b0fb02cc79e5c7c4fdb/projects/libusb/libusb_fuzzer.cc#L21
> 
> they include `libusbi.h` which isn't really a public header, but of course handy to be able to fuzz more functions.

I see.

> Perhaps this is an argument for _not_ renaming these macros/functions.

One possibility is that we merge this PR first and then ask oss-fuzz to update.

But then there may be other such cases, so in the end we may have to keep the namaes of such macros/functions.

mcuee:
Let's hear from @sonatique and @tormodvolden as well on this topic.

Youw:
> But then there may be other such cases, so in the end we may have to keep the namaes of such macros/functions.

We could have functions with new names, and have macros as aliases for functions. This way we don't use UPPER_CAMEL_CASE for function names (directly) and still have backwarc compatibility.

sonatique:
Thanks @seanm : I went through the full diff and the conversion is faithful (the one semantic change, `transfer_priv_to_usbi_transfer` using `usbi_backend.transfer_priv_size` instead of `sizeof(*transfer_priv)`, is equivalent at its only call site and matches the allocation math in `io.c`). A few things before this can merge:

1. **Rebase needed** : the branch conflicts with master (`core.c`, `windows_winusb.c` moved).
2. **CIFuzz failure is caused by this PR** : the OSS-Fuzz harness (`libusb_fuzzer.cc` in google/oss-fuzz) still uses `LIBUSB_TRANSFER_TO_USBI_TRANSFER`. 
3. **macOS job needs a green rerun** (likely CI flakiness, not this PR, but we need a passing run on the current head).
4. Nit: new function signatures use inconsistent spacing (`is_ep_in (uint8_t ep)`, `type * name`) , please match the codebase's `type *name(args)` style.

seanm:
1. Fixed
2. I think we should just not rename these things, to keep source compatibility. i.e. drop the last commit.
3. Agreed
4. Let's decide about no. 2 first, because rebasing that will be a bit of a PITA. Also: expending brain cells on this when we have clang-format is a waste of time. We should try to push #1442 again some day soon...

sonatique:
@seanm : I concur with"not rename" for point 2.

seanm:
> @seanm : I concur with"not rename" for point 2.

Done.

And I think I've done 4 too.

Youw:
I think strictly correct and compatible way would be to do: https://github.com/seanm/libusb/pull/1

Youw (approved):

Youw:
I'm happy with current implementation.
@sonatique @tormodvolden if you don't have anything else to add here - we should be good to merge.

seanm:
@sonatique @tormodvolden did you want to review?  I'll merge at the end next week otherwise.

seanm:
@sonatique @Youw I've fixed the merge conflict, but there is `warning: macro '__EMSCRIPTEN_major__' has been marked as deprecated: Use __EMSCRIPTEN_MAJOR__ instead [-Wdeprecated-pragma]` from one bot.  I don't see how this PR causes that... anyone know what it is?

Youw:
@seanm that is a build warning unrelated to this change, ignore it.
As for failing built - that is a separate issue: #1934

PR #1730, opened 2025-12-31
Fix most remaining Clang -Wsign-conversion warnings

seanm:
- added cast
- changed some types
- added asserts

### DIFF
--- a/examples/fxload.c
+++ b/examples/fxload.c
@@ -94,7 +94,6 @@ int main(int argc, char*argv[])
 	const char *ext, *img_name[] = IMG_TYPE_NAMES;
 	int fx_type = FX_TYPE_UNDEFINED, img_type[ARRAYSIZE(path)];
 	int opt, status;
-	unsigned int i, j;
 	unsigned vid = 0, pid = 0;
 	unsigned busnum = 0, devaddr = 0, _busnum, _devaddr;
 	libusb_device *dev, **devs;
@@ -164,6 +163,7 @@ int main(int argc, char*argv[])
 
 	/* determine the target type */
 	if (type != NULL) {
+		int i;
 		for (i=0; i<FX_TYPE_MAX; i++) {
 			if (strcmp(type, fx_name[i]) == 0) {
 				fx_type = i;
@@ -190,7 +190,7 @@ int main(int argc, char*argv[])
 			logerror("libusb_get_device_list() failed: %s\n", libusb_error_name(status));
 			goto err;
 		}
-		for (i=0; (dev=devs[i]) != NULL; i++) {
+		for (unsigned int i=0; (dev=devs[i]) != NULL; i++) {
 			_busnum = libusb_get_bus_number(dev);
 			_devaddr = libusb_get_device_address(dev);
 			if ((type != NULL) && (device_path != NULL)) {
@@ -200,6 +200,7 @@ int main(int argc, char*argv[])
 			} else {
 				status = libusb_get_device_descriptor(dev, &desc);
 				if (status >= 0) {
+					unsigned int j;
 					if (verbose >= 3) {
 						logerror("examining %04x:%04x (%d,%d)\n",
 							desc.idVendor, desc.idProduct, _busnum, _devaddr);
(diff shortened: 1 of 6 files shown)


examples/fxload.c
```diff
@@ -91,7 +91,7 @@ int main(int argc, char*argv[])
 	const char *ext, *img_name[] = IMG_TYPE_NAMES;
 	int fx_type = FX_TYPE_UNDEFINED, img_type[ARRAYSIZE(path)];
 	int opt, status;
-	unsigned int i, j;
+	unsigned int j;
```
Youw: I believe moving `j` close to its usage would be very symetrical change to the one done for `i`, even if it isn't directly triggers any warnings right now

Youw (approved):
Overall this is good from "fix the warning" point of view.
But there is also anothre thing - this actually affects overall consistency and code style across implementation. We might want to consider what to do with it in the future.

mcuee (approved):

examples/sam3u_benchmark.c
```diff
@@ -157,15 +157,13 @@ static int benchmark_in(uint8_t ep)
...
-
 	get_timestamp(&tv_stop);
 
-	diff_msec = (tv_stop.tv_sec - tv_start.tv_sec) * 1000L;
+	long diff_msec = (tv_stop.tv_sec - tv_start.tv_sec) * 1000L;
 	diff_msec += (tv_stop.tv_usec - tv_start.tv_usec) / 1000L;
 
 	printf("%lu transfers (total %lu bytes) in %lu milliseconds => %lu bytes/sec\n",
-		num_xfer, num_bytes, diff_msec, (num_bytes * 1000L) / diff_msec);
+		num_xfer, num_bytes, diff_msec, (num_bytes * 1000L) / (unsigned long)diff_msec);
```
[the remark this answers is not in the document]
seanm: It's cast to unsigned long, so %lu is correct.  Anyway, this is just benchmark code...

examples/xusb.c
```diff
@@ -75,7 +75,7 @@ static void perr(char const *format, ...)
 #define CALL_CHECK(fcall) do { int _r=fcall; if (_r < 0) ERR_EXIT(_r); } while (0)
 #define CALL_CHECK_CLOSE(fcall, hdl) do { int _r=fcall; if (_r < 0) { libusb_close(hdl); ERR_EXIT(_r); } } while (0)
 #define B(x) (((x)!=0)?1:0)
-#define be_to_int32(buf) (((buf)[0]<<24)|((buf)[1]<<16)|((buf)[2]<<8)|(buf)[3])
+#define be_to_int32(buf) ((uint32_t)(((buf)[0]<<24)|((buf)[1]<<16)|((buf)[2]<<8)|(buf)[3]))
```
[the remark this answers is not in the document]
seanm: Done.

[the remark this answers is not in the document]
seanm: Done.

examples/fxload.c
seanm: Done.

seanm:
@sonatique @Youw ok I made some small changes today, this is ready for hopefully one last review.

libusb/os/events_posix.c
```diff
@@ -292,6 +293,7 @@ int usbi_wait_for_events(struct libusb_context *ctx,
 
 #ifdef HAVE_OS_TIMER
 	/* on timer configurations, fds[1] is the timer */
+	assert(nfds >= 2);
```
sonatique: I think that this assert can fire a false abort: 
HAVE_OS_TIMER is a compile-time condition; it doesn't guarantee the timer is in use at runtime. In io.c, if usbi_create_timer() fails, the timer fd is never added as an event source and usbi_using_timer(ctx) returns false. 

In that case nfds can be 1, so this unconditional assert would abort even though the if (usbi_using_timer(ctx) && fds[1].revents) below correctly guards the actual fds[1] access.

Suggest tying the assert to the same condition as the access:

assert(!usbi_using_timer(ctx) || nfds >= 2);

(or move it inside the if).

examples/sam3u_benchmark.c
```diff
@@ -160,15 +160,13 @@ static int benchmark_in(uint8_t ep)
 static void measure(void)
 {
 	struct timeval tv_stop;
-	unsigned long diff_msec;
-
 	get_timestamp(&tv_stop);
 
-	diff_msec = (tv_stop.tv_sec - tv_start.tv_sec) * 1000L;
+	long diff_msec = (tv_stop.tv_sec - tv_start.tv_sec) * 1000L;
```
sonatique, line 165: diff_msec changed from unsigned long to long: intentional? If yes, then why cast it back to (unsigned long) in the printf? We can keep it as long and change %lu to %l, I think

sonatique (changes requested):

libusb/os/events_posix.c
seanm: Whoops, I think I forgot about `&&` short circuiting!  Fixed.

examples/sam3u_benchmark.c
seanm: Done.

sonatique (approved):

PR #1731, opened 2026-01-02
Initial conservative attempt at starting to better specify and unify indentation

seanm:


Xcode/libusb.xcodeproj/project.pbxproj
Youw, line 1: was this automatically formatted or is it IDE artifact?

Youw (approved):
Surprisingly, there is very few actual code changes. The most significant I guess are changes to documentation in `core.c`.

Overall I like how this looks.

Xcode/libusb.xcodeproj/project.pbxproj
seanm: It was Xcode itself that moved things around in that file. Not sure why. Maybe because I'm now using a newer Xcode. But generally best to let it manage that file itself, hopefully subsequent changes will preserve this new order that it seems to like.

seanm: Oh wait, I somehow based this PR on an older version... I think it's fixable..

seanm: Fixed.

.git-blame-ignore-revs
Youw, line 1: this probably needs to be commited separately - once this PR is merged (probably squashed first), the has of squashed commit ID will be different...

seanm: Why would we squash this? I carefully divided it into separate commits.

sonatique (approved):

.git-blame-ignore-revs
sonatique: Do we really need this file? I feel it might be a bit of overkill refinement and things would be simpler without it, but I probably missed something. And I personally don't feel the need for squashing.

sonatique:
@seanm : would you like me to merge this PR asap?

.git-blame-ignore-revs
Youw: > Do we really need this file?

Yes, that's how we tell git history/blame to ignore certain commits (e.g. with white-space-only changes) when annotating when the last change was done to a line, etc.

> I feel it might be a bit of overkill refinement and things would be simpler without it

Completely independent from everything else, harmless for those who don't need it and very useful for those who do (me, for example).

---
Regarding squashing, I was under impression each PR to libusb is closed by a single (squashed) commit with VERSION_NANO incremented.
If we are to name this PR a non-functional change w/o VERSION_NANO update, I guess leaving this as is (and rebase-merge) should be fine.

seanm: >Regarding squashing, I was under impression each PR to libusb is closed by a single (squashed) commit with VERSION_NANO incremented.

Doesn't look like it. For example look at [this](https://github.com/libusb/libusb/pull/1685/commits) PR and actual history in master. Looks like 1 `VERSION_NANO` update per PR I think...

seanm:
> @seanm : would you like me to merge this PR asap?

It should be safe to do so I think.

The only slight risk is some file formats I don't know...  I know that for .c and .h trailing whitespace can be safely removed, and tab/space don't matter, but I'm only 95% sure for other formats, like those makefile files.  Do they care about tab vs space?

PS: but please don't squash.

.git-blame-ignore-revs
sonatique: @seanm @Youw : I am responsible for the way #1685 has been merged (i.e. one single VERSION_NANO increment on the last commit of the PR) but I am not sure it is correct, because the few commits before the last one are still on the previous value, which technically is not right. I did this to have the single "Closes" mention, obviously in the last commit in sync with the VERSION_NANO change. 
I read information here : https://github.com/libusb/libusb/wiki/Maintainers'-corner regarding the use of VERSION_NANO and how to merge PR, but I didn't really understand what to do with PR containing several commits: the last paragraph of "Merge of a commit that you didn't create ..." seems to indicate that VERSION_NANO shall be incremented for each commit, and this is consistent with the fact the increment is implemented using a commit hook. And by the way, when the hook is not run on every commit then the number jumps anyway by the number of commits since the last VERSION_NANO change.

I think @tormodvolden may have some hints about how to do it right, but the most consistent version seems to increment the VERSION_NANO with every commit, but to put a single "Closes" mention on the last commit.

sonatique:
@seanm : I am not sure about makefile and other neither, but shouldn't github checks catch issues with them?

seanm:
> @seanm : I am not sure about makefile and other neither, but shouldn't github checks catch issues with them?

I would hope, but I'm not very familiar with what the checks do. Checks never check everything. :)

seanm:
This is another example of where we should *not* squash and merge.

There are 2 approvals already, if no one objects soon I will 'rebase and merge' this later this week.

Issue #890, opened 2021-03-15
remove "num_iso_packets" from structure "struct usbi_transfer"


lerdmann:
"num_iso_packets" is defined in both, "struct usbi_transfer" and also in "struct libusb_transfer".

This can lead to a clash as function "libusb_alloc_transfer" will set the value in "struct usbi_transfer" whereas function "libusb_fill_iso_transfer" will set the value in "struct libusb_transfer" which can lead to a mismatch.

I suggest to change "libusb_alloc_transfer" to :

`DEFAULT_VISIBILITY
struct libusb_transfer * LIBUSB_CALL libusb_alloc_transfer(
	int iso_packets)
{
	size_t priv_size;
	size_t alloc_size;
	unsigned char *ptr;
	struct usbi_transfer *itransfer;
	struct libusb_transfer *transfer;

	assert(iso_packets >= 0);
	if (iso_packets < 0)
		return NULL;

	priv_size = PTR_ALIGN(usbi_backend.transfer_priv_size);
	alloc_size = priv_size
		+ sizeof(struct usbi_transfer)
		+ sizeof(struct libusb_transfer)
		+ (sizeof(struct libusb_iso_packet_descriptor) * (size_t)iso_packets);
	ptr = calloc(1, alloc_size);
	if (!ptr)
		return NULL;

	itransfer = (struct usbi_transfer *)(ptr + priv_size);
	itransfer->priv = ptr;
	usbi_mutex_init(&itransfer->lock);
	transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
	usbi_dbg("transfer %p", transfer);
        if (transfer) transfer->num_iso_packets = iso_packets;
	return transfer;
}`

and to either remove parameter "num_iso_packets" from structure "struct usbi_transfer" and also from function "libusb_fill_iso_transfer" or ignore it (leave as deprecated). That will ensure that there is only one value for "num_iso_packets" and that this value will always be an integral part of the transfer and not some internal control variable.

mcuee:
@lerdmann Just wondering if you can create a pull request for this. Thanks.

seanm:
While working on [-fbounds-safety](https://clang.llvm.org/docs/BoundsSafety.html) support (https://github.com/libusb/libusb/pull/1659) I've come across, I think, basically this same issue, or at least close enough that I'll comment here instead of creating a new Issue.

I think `libusb_fill_iso_transfer()` has a design flaw.

Its current signature and implementation is:

[...]

PR #1734, opened 2026-01-09
Issue #890, Add field to struct libusb_transfer to indicate actual buffer size

seanm:


libusb/libusb.h
```diff
@@ -1508,9 +1509,12 @@ struct libusb_transfer {
 	unsigned char *buffer;
 
 	/** Number of isochronous packets. Only used for I/O with isochronous
-	 * endpoints. Must be non-negative. */
+	 * endpoints. Must be non-negative and <= max_num_iso_packets. */
 	int num_iso_packets;
 
+	/** The count of the iso_packet_desc array below (its actual memory size). */
+	int max_num_iso_packets;
```
Youw, line 1558: this changes the struct layout of a public API and breaks binary compatibility, isn't it?

seanm: I guess so. I guess we'd need to increment LIBUSB_API_VERSION? or more?

Youw: Since the original libusb1.0 release - I don't believe that ever happened.
That would mean 1.1 release...

I believe all previous changes where backward (but not forward) compatible.

@mcuee can you confirm?

seanm: so the worry would be dynamic linking I guess.

I'll split this PR into 2, so the little commit can be merged sooner.

Youw: > so the worry would be dynamic linking I guess.

Sure it is. Most distros distribute binaries of dynamic libusb and applications that use it separately, and update those separately as well.

> I'll split this PR into 2

Thanks

seanm: Isn't this only a problem because `struct libusb_transfer` is exposed in the public header? (and thus other code may be assuming fields are at known offsets).

`struct libusb_context` on the other hand is not exposed in the public header.

Is there a reason for `struct libusb_transfer` to be public?

seanm: So I've wrapped the new field in `#ifdef __has_ptrcheck` so that it only affects `-fbounds-safety` builds.

Youw: > So I've wrapped the new field in `#ifdef __has_ptrcheck` so that it only affects `-fbounds-safety` builds.

That is even worse for binary compatibility. Not only the version of source/binary has to match - compilation options has to match as well, or the runtime will fail w/o a single warning...

mcuee: > Since the original libusb1.0 release - I don't believe that ever happened. That would mean 1.1 release...
> 
> I believe all previous changes where backward (but not forward) compatible.
> 
> @mcuee can you confirm?

Yes, I think that should be the case.

We have some 2.0 wishlist here but I do not think they will happen any time soon.
https://github.com/libusb/libusb/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A%222.0%20release%20(API%2FABI%20change)%22

seanm: > That is even worse for binary compatibility. 

There is no backwards compatibility with `-fbounds-safety` (aka `#ifdef __has_ptrcheck`) because libusb has never (and still does not) even compile with `-fbounds-safety`. So there's nothing to be backwards compatible with.

Youw: Fair point, but one still may end up compiling some project or a package with this option and libusb dynamically linked (e.g. using libusb1.0-dev package), and later use the system-installed dynamic libusb binary in rubtime (the same version) but totally incompatible...

seanm: Do you have an alternative proposal?

Youw: Technically there are two: 
1) Update minor version (libusb1.1) and mark it as potentially incompatible with libusb1.0
2) Intoduce a backward-compatible change, e.g. `struct libusb_transfer2`, and deprecate all functions which use `struct libusb_transfer`.
I imagine it will be quite a bit of a pain to introduce in a public API and support both new and deprecated ones (and there are some variations how to have that implemented).

If someone has alternative suggestions I'd gladly hear those.

libusb/io.c
```diff
@@ -1306,8 +1308,12 @@ struct libusb_transfer * LIBUSB_CALL libusb_alloc_transfer(
 	itransfer->num_iso_packets = iso_packets;
 	itransfer->priv = ptr;
 	usbi_mutex_init(&itransfer->lock);
-	struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
 
+	struct libusb_transfer *transfer = USBI_TRANSFER_TO_LIBUSB_TRANSFER(itransfer);
+#ifdef __has_ptrcheck
+	transfer->num_iso_packets = max_num_iso_packets;
```
anotheruserofgithub, line 1315:
```suggestion
	transfer->max_num_iso_packets = max_num_iso_packets;
```

libusb/libusb.h
```diff
@@ -41,6 +41,7 @@ typedef SSIZE_T ssize_t;
 #endif /* _SSIZE_T_DEFINED */
 #endif /* _MSC_VER */
 
+#include <assert.h>
```
anotheruserofgithub, line 48:
```suggestion
#ifdef __has_ptrcheck
#include <assert.h>
#endif
```

PR #1736, opened 2026-01-12

seanm:
If length is set to 0, also set data to NULL. When length is zero, usbi_utf8_copy doesn't use the pointer anyway, so there's no change in behaviour.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -1665,6 +1665,7 @@ int API_EXPORTED libusb_get_device_string(libusb_device *dev,
 	}
 	if (NULL == data) {
 		length = 0;
+		data = NULL;
 	} else if (length > 0) {
 		*data = 0;  // return an empty string on errors when possible
 	}


seanm:
@sonatique since it doesn't look like any release is coming soon, perhaps we could start merging some of my PRs?

sonatique:
> @sonatique since it doesn't look like any release is coming soon, perhaps we could start merging some of my PRs?

Yes sure, I'll go through them and merge.

As for having a release coming soon: as I understood, we both are maintainers, so there is basically nobody above us that will magically make it happen. It is in our hands, there are no "release guys" on duty for us ;-)

The legacy process that is currently in place is a bit complex, but I guess nothing that we couldn't manage with a reasonable amount of effort. If I understand, the current philosophy / guide-lines is to have people extensively test an RC before we can even think about releasing something, and I think that there is currently a lack of such available testers.

seanm:
> As for having a release coming soon: as I understood, we both are maintainers, so there is basically nobody above us that will magically make it happen. It is in our hands, there are no "release guys" on duty for us ;-)

For sure, I agree completely.

But to be totally honest: like everyone, I have limited time, and I have my priorities. I don't need a release; for me, building from master is just fine. Realistically, working on a release will never be at the top of my todo list.

tormodvolden:
No worries guys, I'll be release guy on duty :) As you might have guessed from recent PRs I've done baby steps towards the RC, I think it will happen this weekend. The RC will give us a lot more testers (and let us bring in more).

PR #1738, opened 2026-01-12

seanm:
Eliminated `remaining` variable in favour of just mutating the `size` parameter, which has to be kept in sync with the `buffer` parameter.

Should be no change in behaviour here.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -1408,12 +1408,11 @@ static int parse_iad_array(struct libusb_context *ctx,
 		iad_array->iad = iad;
 
 		/* Second pass: Iterate through desc list, fill IAD structures */
-		int remaining = size;
 		i = 0;
 		do {
 			header.bLength = buffer[0];
 			header.bDescriptorType = buffer[1];
-			if (header.bDescriptorType == LIBUSB_DT_INTERFACE_ASSOCIATION && (remaining >= LIBUSB_DT_INTERFACE_ASSOCIATION_SIZE)) {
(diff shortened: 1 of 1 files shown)


Youw (approved):
It may look like a logic change, but I believe all good, and now it is cleaner then it was

seanm:
> ...and now it is cleaner then it was

I'm finding a lot of the changes -fbounds-safety forces result in improvements. I guess making buffer lengths obvious to the compiler also makes them more obvious to us humans. :)

libusb/descriptor.c
```diff
@@ -1412,10 +1411,11 @@ static int parse_iad_array(struct libusb_context *ctx,
 				i++;
 			}
 
-			remaining -= header.bLength;
-			if (remaining < DESC_HEADER_LENGTH) {
+			if (size - header.bLength < DESC_HEADER_LENGTH) {
```
sonatique, line 1427: Why not first substract size -= header.bLength, then just compare size?

seanm: With -fbounds-safety, the size and buffer must be mutated back-to-back.

Details here https://clang.llvm.org/docs/BoundsSafety.html#maintaining-correctness-of-bounds-annotations :

"Additionally, an explicit bounds annotation such as `int *__counted_by(count) buf` defines a relationship between two variables, buf and count: namely, that buf has count number of elements available. This relationship must hold even after any of these related variables are updated. To this end, the model requires that assignments to buf and count must be side by side, with no side effects between them. This prevents buf and count from temporarily falling out of sync due to updates happening at a distance."

You can look at the example after that starting paragraph.

sonatique (approved):

mcuee (approved):

mcuee:
@sonatique 

I think this one can be merged now.

PR #1739, opened 2026-01-12
Use const in main() signature

seanm:
This prepares command-line entry points for `-fbounds-safety`, under which `const char *` is treated as a NUL-terminated string.

Most entry points now expose read-only argument strings. `tests/umockdev.c` keeps the standard mutable `argv` form because `g_test_init()` parses and rewrites the argument vector.

A Bounds Safety build was trialed in the existing macOS Xcode job. It was not retained because that job builds only the library target and currently fails in unrelated core code outside this preparation change.

Related to #1659.
Related to #1887.

Assisted-by: codex:gpt-5.6-sol

### DIFF
--- a/tests/umockdev.c
+++ b/tests/umockdev.c
@@ -1121,7 +1121,7 @@ test_hotplug_add_remove(UMockdevTestbedFixture * fixture, UNUSED_DATA)
 }
 
 int
-main(int argc, char **argv)
+main(int argc, char *argv[])
 {
 	g_test_init(&argc, &argv, NULL);
 
--- a/examples/hotplugtest.c
+++ b/examples/hotplugtest.c
@@ -99,7 +99,7 @@ static int LIBUSB_CALL hotplug_callback_detach(libusb_context *ctx, libusb_devic
 	return 0;
 }
 
-int main(int argc, char *argv[])
+int main(int argc, const char *argv[])
 {
 	libusb_hotplug_callback_handle hp[2];
 	int product_id, vendor_id, class_id;
--- a/examples/listdevs.c
+++ b/examples/listdevs.c
@@ -79,7 +79,7 @@ static int usage(void) {
 	return 1;
 }
 
-int main(int argc, char *argv[])
+int main(int argc, const char *argv[])
 {
 	int verbose = 0;
 	libusb_device **devs;
--- a/examples/testlibusb.c
+++ b/examples/testlibusb.c
@@ -271,7 +271,7 @@ static int test_wrapped_device(const char *device_name)
 }
 #endif
 
-int main(int argc, char *argv[])
+int main(int argc, const char *argv[])
 {
 	const char *device_name = NULL;
 	libusb_device **devs;
--- a/examples/xusb.c
+++ b/examples/xusb.c
@@ -1105,7 +1105,7 @@ static void display_help(const char *progname)
 	printf("If only the vid:pid is provided, xusb attempts to run the most appropriate test\n");
 }
 
-int main(int argc, char** argv)
+int main(int argc, const char* argv[])
 {
 	bool debug_mode = false;
 	const struct libusb_version* version;
@@ -1113,7 +1113,8 @@ int main(int argc, char** argv)
 	size_t i, arglen;
 	unsigned tmp_vid, tmp_pid;
 	uint16_t endian_test = 0xBE00;
-	char *error_lang = NULL, *old_dbg_str = NULL, str[256];
+	const char *error_lang = NULL;
+	char *old_dbg_str = NULL, str[256];
 
 	// Default to generic, expecting VID:PID
 	VID = 0;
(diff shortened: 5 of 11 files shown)


Youw (approved):

sonatique (approved):

seanm:
CI failures seem due to different `getopt` signature on BSD vs linux:

`int getopt(int argc, char * const argv[], const char *optstring);`

vs

`int getopt(int argc, char *argv[], const char *optstring);`

mcuee:
> CI failures seem due to different `getopt` signature on BSD vs linux:
> 
> `int getopt(int argc, char * const argv[], const char *optstring);`
> 
> vs
> 
> `int getopt(int argc, char *argv[], const char *optstring);`

Interesting, any fixes to this issue?

seanm:
@sonatique @Youw this is ready for final review now.

Youw:
Codex reviewed exact head `5e7bf31d5dab34c830ba2e8be5eb06c6466eea50`. Two items look worth addressing before merge:

1. **The `g_test_init()` cast should be avoided.** In [`tests/umockdev.c`](https://github.com/libusb/libusb/blob/5e7bf31d5dab34c830ba2e8be5eb06c6466eea50/tests/umockdev.c#L1124-L1127), `argv` has adjusted type `const char **`, so `&argv` is `const char ***`; casting that pointer object to `char ***` makes GLib access it through an incompatibly qualified type. `g_test_init()` is explicitly an in/out API and removes parsed entries from `argv` ([GLib documentation](https://docs.gtk.org/glib/func.test_init.html)). Keeping this particular entry point as `main(int argc, char **argv)` and retaining `g_test_init(&argc, &argv, NULL)` is the cleanest fix. A genuine mutable `char **` temporary would at least avoid casting the address of the const-qualified pointer object, but should be tested with the target compiler.

2. **The intended configuration is not validated.** The PR exists to prepare for `-fbounds-safety`, but none of the current CI jobs enables that flag. Normal green builds therefore do not demonstrate that the two new qualifier-removing casts—here and in `fxload.c`—work under the mode being targeted. Both casts are rejected by a `-Wcast-qual -Werror` probe. This is particularly relevant because Bounds Safety gives `const char *` implicit null-terminated semantics and applies additional cast rules ([Clang documentation](https://clang.llvm.org/docs/BoundsSafety.html)). An actual compile of these targets with the intended Apple Clang/`-fbounds-safety` configuration, preferably in CI, seems necessary.

The remaining declaration, definition, and call-site changes look internally consistent, and Codex did not find an obvious missed non-vendored `main(argc, argv)` entry point. At review time, all completed jobs passed; CIFuzz and AppVeyor were still pending.

Youw (approved):
This looks fine.
I asked codex to update CI for this as well, and check how much is missing yet.

examples/fxload.c
```diff
@@ -100,7 +100,7 @@ int main(int argc, char*argv[])
 	libusb_device_handle *device = NULL;
 	struct libusb_device_descriptor desc;
 
-	while ((opt = getopt(argc, argv, "qvV?hd:p:i:I:s:S:t:")) != EOF)
+	while ((opt = getopt(argc, (char*const*)argv, "qvV?hd:p:i:I:s:S:t:")) != EOF)
```
sonatique: Since getopt() requires mutable `char *` strings anyway, keeping `char *argv[]` in this file would avoid this cast.

tests/umockdev.c
```diff
@@ -1121,9 +1121,9 @@ test_hotplug_add_remove(UMockdevTestbedFixture * fixture, UNUSED_DATA)
 }
 
 int
-main(int argc, char **argv)
+main(int argc, const char *argv[])
 {
-	g_test_init(&argc, &argv, NULL);
+	g_test_init(&argc, (char***)&argv, NULL);
```
sonatique: g_test_init() rewrites the argv array, so this cast only hides the mutation from the type system. These tests are Linux-only and never built with -fbounds-safety, so keeping `char **argv` here would avoid the cast.

examples/fxload.c
seanm: Indeed I guess I don't care about the examples folder too much, at least not yet.

tests/umockdev.c
seanm: Indeed I guess I don't care about the tests folder too much, at least not yet.

-fbounds-safety is not-Apple only BTW.

sonatique:
@seanm : I am not against this PR since I understand the goals and benefit, but strictly speaking, I think `const char *argv[]` is technically non-standard, according to https://cigix.me/c17#5.1.2.2.1 .
"
or with two parameters (referred to here as argc and argv, though any names may be used, as they
are local to the function in which they are declared):

    int main(int argc, char *argv[]) { /* ... */ }
  
or equivalent;
"

So the question is whether we agree to use  non-standard main to please -fbouds-safety, or if we find another way.

Note: I was surprised myself when I check the C17 spec, because I have to say I think I used and saw use of const char *argv[] in serious programs.

sonatique (approved):

seanm:
I'm pretty sure the "or equivalent" means "plus or minus const/volatile keywords".

PR #1740, opened 2026-01-12

seanm:
Use a local variable to count so that setting the pointer and count in the structure can be delayed and done together.

Small behaviour change: move zero initialization to the beginning of the function, as such if early return occurs because of error, these fields will now be zeroed. Shouldn't change much in practice.


Youw (approved):
LGTM

mcuee (approved):

PR #1742, opened 2026-01-13
replace 0 size arrays with flexible

neheb:
The former is a GNU extension. The latter is standard C99.

### DIFF
--- a/libusb/os/linux_usbfs.h
+++ b/libusb/os/linux_usbfs.h
@@ -91,7 +91,7 @@ struct usbfs_urb {
 	int error_count;
 	unsigned int signr;
 	void *usercontext;
-	struct usbfs_iso_packet_desc iso_frame_desc[0];
+	struct usbfs_iso_packet_desc iso_frame_desc[];
 };
 
 struct usbfs_connectinfo {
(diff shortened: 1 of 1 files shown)


seanm:
Nice catch. There is also `LIBUSB_FLEXIBLE_ARRAY` for this purpose. Not sure anyone cares about C89 anymore (not me!) but you may as well use `LIBUSB_FLEXIBLE_ARRAY` since it still exists.

neheb:
That's in libusb.h. I'd rather not include it in these files.

Youw (approved):
LGTM
(just out of curiocity I wounder what copilot would say about this)

mcuee:
I think we can remove the support of VS2013 and VS2015. Both are no longer supported by Microsoft.

https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2013

https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2015

seanm:
> I think we can remove the support of VS2013 and VS2015. Both are no longer supported by Microsoft.

Just because MS doesn't support something doesn't mean other don't still want it. If we were going to do something big like adopt C23 then dropping old toolchains make sense, but dropping support for such a triviality is overkill.

Especially when we *already have* `LIBUSB_FLEXIBLE_ARRAY`, allowing both the GNU extension and new (C99) syntax.

neheb:
So this touches both Linux and windows files. How about changing the Linux ones only?

seanm:
BTW did you find this with some tool, or you just noticed it?

neheb:
Noticed. I’ve been getting rid of local zero length arrays in Linux kernel drivers in OpenWrt.

Youw:
I agree that using `[LIBUSB_FLEXIBLE_ARRAY]` instead of explicit `[]` would be a better and consistent across the project fix.

PR #1744, opened 2026-01-18
webusb: fix set_configuration binding

adriweb:
`setConfiguration` was only used in the very early days of WebUSB. It is nowadays `selectConfiguration`.
See https://developer.mozilla.org/en-US/docs/Web/API/USBDevice/selectConfiguration


mcuee:
Looks good to me.

PR #1746, opened 2026-01-21
windows: Support loading libusbK.dll from app directory

hartcw:
This allows app to ship their own version of the libusbK.dll in their package, and use that without having to install it into the Windows sytem directory.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2808,7 +2808,7 @@ cleanup_winusb:
 		usbi_info(ctx, "WinUSB DLL is not available");
 	}
 
-	hlibusbK = load_system_library(ctx, "libusbK");
+	hlibusbK = LoadLibraryA("libusbK.dll");
 	if (hlibusbK != NULL) {
 		LibK_GetVersion_t pLibK_GetVersion;
 		LibK_GetProcAddress_t pLibK_GetProcAddress;


libusb/os/windows_winusb.c
```diff
@@ -2808,7 +2808,7 @@ static bool winusbx_init(struct libusb_context *ctx)
 		usbi_info(ctx, "WinUSB DLL is not available");
 	}
 
-	hlibusbK = load_system_library(ctx, "libusbK");
+	hlibusbK = LoadLibraryA("libusbK.dll");
```
sonatique, line 2811: Thanks, nice idea (I have similar code on my private custom version). 
I would prefer to add something like this after your line, though:
`if(hlibusbK == NULL)
{
	hlibusbK = load_system_library(ctx, "libusbK");
}`

Unless LoadLibraryA also look into system paths (which I am not aware of but I may have miss something), I think your change breaks the default expected behavior.

hartcw: Yes I think LoadLibraryA() will check the standard system library locations as well.
So it first tries the directory where the app/executable was loaded, then the system dir (ie C:\Windows\System32), and then some other places.
(this is what chatgpt told me!)

sonatique: Hum, yes, I think you're right. In fact event libusb own documentation says so: the documentation for load_system_library says:
"Dynamically loads a DLL from the Windows system directory.  **Unlike the LoadLibraryA() function, this function will not search through any directories** to try and find the library."

So yes, if LoadLibraryA returns null, load_system_library won't find it neither.

However, I had a look at why load_system_library was introduced, and I found a very clear commit about it :
c3deb6dd from Chris Dickens in 2020, saying that:
"
Windows: Restrict path for dynamic library loading

The LoadLibraryA() function performs a search through various paths when
provided a library name that does not include a path element. All of the
libraries that libusb needs to dynamically load are installed in the
Windows system directory, thus it is not necessary to search any paths.

To harden the security of libusb and prevent loading libraries from
outside of the system directory should an attacker be able to alter the
environment or place an identically named library somewhere in the
search paths that occur before the Windows system directory, introduce a
function that calls LoadLibraryA() with a full path to the requested
library witin the Windows system directory. Note that we cannot call
SetDllDirectoryA() because as a library we should not alter the DLL
search path of the application. We also cannot use LoadLibraryExA() with
the LOAD_LIBRARY_SEARCH_* flags because those flags require a specific
security update to be installed on Vista and 7.
" 

So, instead of the modification I suggested, and in order to honor the above concerns, I guess we should do something like: 
if libusbK.dll exists in current dir then use it, else use load_system_library 

What do you think?

_Note: I also noted that the original code was only using the name "libusbK" without the ".dll" extension like you now do. So while your code might actually trigger a super-set of search as compared to load_system_library, I thought we should be careful with the addition of the .dll. But readony LoadLibraryA documentation told me that "If no file name extension is specified in the lpFileName parameter, the default library extension .dll" (see Remarks here : https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya)_

mcuee:
libusbK is mainly used for libusb0.sys and libusbk.sys driver. Typically the user will use a driver installer like Zadig or libusbK driver installer. Both apps will install libusbK.dll in the system directories.

sonatique:
> libusbK is mainly used for libusb0.sys and libusbk.sys driver. Typically the user will use a driver installer like Zadig or libusbK driver installer. Both apps will install libusbK.dll in the system directories.

I think the idea of the PR is to allow someone to only install the driver (.sys) using an .inf file or something and provide a (custom if needed) libusbK.dll side to the libusb dll to be used in priority.

This enable, for instance, to have a single driver but different applications that use different versions of libusbK.dll  bundled with them. This is also useful for testing libusbK or libusb without reinstalling everything with Zadig.

mcuee:
@hartcw 

Would you please modify the PR as per the last comment by @sonatique ? Thanks.

mcuee:
Reference from rufus.
https://github.com/pbatard/rufus/blob/073a100c8e0bfaf496f1374114cc39e15d28918f/src/rufus.c#L3271C1-L3281C2
```
// For delay-loaded DLLs, use LOAD_LIBRARY_SEARCH_SYSTEM32 to avoid DLL search order hijacking.
FARPROC WINAPI dllDelayLoadHook(unsigned dliNotify, PDelayLoadInfo pdli)
{
	if (dliNotify == dliNotePreLoadLibrary) {
		// Windows 7 without KB2533623 does not support the LOAD_LIBRARY_SEARCH_SYSTEM32 flag.
		// That is OK, because the delay load handler will interrupt the NULL return value
		// to mean that it should perform a normal LoadLibrary.
		return (FARPROC)LoadLibraryExA(pdli->szDll, NULL, LOAD_LIBRARY_SEARCH_SYSTEM32);
	}
	return NULL;
}
```

mcuee:
Reference from old version of rufus.
https://github.com/pbatard/rufus/blob/2d6a73e4f491230bed299e3dade051c5cb7b0099/src/rufus.c#L2980-L3004

```
	// Disable loading system DLLs from the current directory (sideloading mitigation)
	// PS: You know that official MSDN documentation for SetDllDirectory() that explicitly
	// indicates that "If the parameter is an empty string (""), the call removes the current
	// directory from the default DLL search order"? Yeah, that doesn't work. At all.
	// Still, we invoke it, for platforms where the following call might actually work...
	SetDllDirectoryA("");

	// Also, even if you use SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_SYSTEM32), you're
	// still going to be brought down if you link to wininet.lib or dwmapi.lib, as these two
	// perform their DLL invocations before you've had a chance to execute anything.
	// Of course, this is not something that security "researchers" will bother looking into
	// to try to help fellow developers, when they can get an ego fix by simply throwing
	// generic URLs around and deliberately refusing to practice *responsible disclosure*...
	// Finally, we need to perform the whole gymkhana below, where we can't call on
	// SetDefaultDllDirectories() directly, because Windows 7 doesn't have the API exposed.
	GetSystemDirectoryW(kernel32_path, ARRAYSIZE(kernel32_path));
	wcsncat(kernel32_path, L"\\kernel32.dll", ARRAYSIZE(kernel32_path) - wcslen(kernel32_path) - 1);
	// NB: Because kernel32 should already be loaded, what we do above to ensure that we
	// (re)pick the system one is mostly unnecessary. But since for a hammer everything is a
	// nail... Also, no, Coverity, we never need to care about freeing kernel32 as a library.
	// coverity[leaked_storage]
	pfSetDefaultDllDirectories = (SetDefaultDllDirectories_t)
		GetProcAddress(LoadLibraryW(kernel32_path), "SetDefaultDllDirectories");
	if (pfSetDefaultDllDirectories != NULL)
		pfSetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_SYSTEM32);
```

mcuee:
macOS CI build failure has nothing to do with this PR. Unfortunately it cannot be re-run now.

PR #1750, opened 2026-02-08
docs: Fix typo causing Doxygen error

tormodvolden:
Follow-up of commit 3dac3f2a0

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -2310,7 +2310,7 @@ int API_EXPORTED libusb_endpoint_supports_raw_io(libusb_device_handle* dev_handl
  * \returns another LIBUSB_ERROR code on other failure
  *
  * \see libusb_endpoint_supports_raw_io()
- * \seealso https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification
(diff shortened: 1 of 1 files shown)


sonatique (approved):

libusb/core.c
```diff
@@ -2310,7 +2310,7 @@ int API_EXPORTED libusb_endpoint_supports_raw_io(libusb_device_handle* dev_handl
  * \returns another LIBUSB_ERROR code on other failure
  *
  * \see libusb_endpoint_supports_raw_io()
- * \seealso https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification
+ * \see also https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification
```
seanm: Why even keep the word `also`?

tormodvolden: You are right, \see is special and currently this renders to:
```
See also
    libusb_endpoint_supports_raw_io()
    also https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/winusb-functions-for-pipe-policy-modification
```
which looks strange.

PR #1511, opened 2024-06-15
ci: Build with -Werror on Linux

tormodvolden:
This should fail on current master, but go green once #1508 is in place.

PR #1532, opened 2024-07-09
Added libusb_get_device_string() which works with a closed device

mliberty1:
Added libusb_get_device_string() which works with a closed device  #866

This new API call provides access to device strings, normally accessed using USB string descriptors, without opening the device.  The common use case is to filter on serial number for selecting one of multiple connected devices.

The commit adds the new API function, documentation, and support for the following platforms:

* Windows WinUSB
* Linux
* macOS

All platforms support manufacturer, product, and serial number strings.

The example/listdevs.c now has a "--verbose" command line argument that prints these strings.

Known issues:

* On one Ubuntu Linux machine, reading the /sys/bus/usb/ files to get the strings takes a variable amount of time on one device,  an xHCI root hub.  While the read can be fast, it can also take seconds.  Using "cat" at the command line has the exact same behavior.

References:

* https://github.com/libusb/libusb/issues/866
* https://github.com/libusb/libusb/pull/875
* https://github.com/libusb/libusb/pull/1258
* https://github.com/hjmallon/libusb/commit/d0f9fd4852dfceaa4f8f55700ed9b74c4e5406dd

Thank you to the prior work that made this much quicker:

* @mcuee
* @sonatique
* @hjmallon

I apologize anyone that I missed.  This feature has been in the works for a long time with multiple starts and stops.

And thank you to @Youw, @seanm and @Mr-Bossman for providing review feedback!

PR #1754, opened 2026-02-19
linux_usbfs.c: Avoid switch-enum warning in op_get_device_string()

tormodvolden:
(first pushed on top of #1511 to confirm it builds without warnings on CI)


tormodvolden:
(passed building with -Werror)

tormodvolden:
It wasn't limited to udev/netlink or not, but on all builds, introduced in commit e894c3e1 from PR #1532

PR #1755, opened 2026-02-22
Release 1.0.30 rc1

tormodvolden:
(exceptionally including nano version change, for build artifacts to have them correct)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ Aaron Luft
 Adam Korcz
 Addison Crump
 Adrian Bunk
+Adrien Bertrand
 Adrien Destugues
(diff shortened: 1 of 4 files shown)


seanm:
I'm not familiar with the .def format, but this generally lgtm

mcuee:
Great. I think we can release RC1 and then go through about two weeks of testing phase to see if there are major issues (newly introduced or regression). If no then we are good to go with 1.0.30 release.

tormodvolden:
The 1.0.30-rc1 has been released: https://sourceforge.net/p/libusb/mailman/message/59300066/
Please post test results to https://github.com/libusb/libusb/issues/1757

PR #1758, opened 2026-02-25
Fix reference counting for parent devices in device management

minicarpet:
This PR should resolve https://github.com/libusb/libusb/issues/1165 and replace https://github.com/libusb/libusb/pull/1687

since I've found some real issue regarding reference counting which were incrementing too much time, I unref hub when needed which lead to not keep Hub into referenced devices.

See comments into issue for more 
details.

**NOTE**:
tested using the [branch](https://github.com/minicarpet/libusb/tree/debug/test-debug):
- using windows without hotplug
- using windows hotplug
- using custom hub tree with private device
- using "normal device" such an usb headset

Tested with custom program:
- using private company program without hotplug (my program does not support hotplug yet): OK

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1988,19 +1988,14 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 				break;
 			case GEN_PASS:
 				port_nr = 0;
-#if defined(LIBUSB_WINDOWS_HOTPLUG)
 				if (priv->initialized) {
 					libusb_unref_device(parent_dev);
 					r = LIBUSB_SUCCESS;
-				}
-				else {
-#endif
+				} else {
 					if (!get_dev_port_number(*dev_info, &dev_info_data, &port_nr))
 						usbi_warn(ctx, "could not retrieve port number for device '%s': %s", dev_id, windows_error_str(0));
 					r = init_device(dev, parent_dev, (uint8_t)port_nr, dev_info_data.DevInst);
-#if defined(LIBUSB_WINDOWS_HOTPLUG)
 				}
-#endif
 				if (r == LIBUSB_SUCCESS) {
 					// Append device to the list of discovered devices
 					if (_discdevs) {


minicarpet:
macOS and linux labels can be removed

tormodvolden:
I like this much more than #1687 ...

If @sonatique can confirm that this is an understood problem with an obvious fix, maybe we can target this for 1.0.30 ? But otherwise no rush.

minicarpet:
> I like this much more than #1687 ...
> 
> If @sonatique can confirm that this is an understood problem with an obvious fix, maybe we can target this for 1.0.30 ? But otherwise no rush.

This is ready for review !

sonatique:
@tormodvolden @minicarpet if I remember correctly, I found this issue when working on hotplug and did this correction (I mean: the code within the "#if defined"s that are removed in this PR). 
Then, at the time of merging into master, I made sure that the code for the non-hotplug build would be 100% equivalent to the pre-existing version and I added some #if defined . I think that at that time I forgot about this particular fix (I should have dedicated a PR especially for it). 
So thanks @minicarpet for bringing it back.

tormodvolden:
When I looked at this yesterday it was a very simple fix (commit 8bb6391). Now the fix is in another place so it looks very different. Can you explain what the change is?

tormodvolden:
When two completely different solutions have the same commit message, maybe the commit message isn't explaining enough?

minicarpet:
I can: I thought first the issue was because hub were referenced too much, so I was doing first a untef to a hub when get_ancestror() were called.
But the real fix is about if the device is already init or not (init device leave immediately if device is init) but somewhere before the reference count were incremented, so we need to unref if device already init (in that case only). I just saw that it was correctly done for hotplug (see @sonatique comment). So in the end this is equivalent of what I've done.
This is hard to explain it really but Maily this is what happens.
About the commit message I can change it and add some details if you want ?

tormodvolden:
OK, so the hotplug path was doing the right thing, that also would be correct for the non-hotplug case, so you therefore simply removed the special-casing of hotplug?

Yes, thanks, IMO it would be nice with a little more explanation in the commit message, so it is easier to understand the change.

minicarpet:
Exact !
I will add details to the commit message this evening.

minicarpet:
@tormodvolden Done, I've rework commit message. Tells me if ok for you.

PR #1764, opened 2026-03-06
windows hotplug: hotplug events fire for hubs

sonatique:
The goal is to have hotplug events fired for hubs as well, since this
behavior is the de-facto standard under Linux.

Replace the per-device notification approach (RegisterDeviceNotification +
GUID_DEVINTERFACE_USB_DEVICE + DBT_DEVICEARRIVAL/DBT_DEVICEREMOVECOMPLETE)
with DBT_DEVNODES_CHANGED, which Windows broadcasts to all top-level
windows whenever the device tree changes — no registration required.

The reason for this change is that hubs only fire DBT_DEVNODES_CHANGED.
Since this event does not carry a device name, the code now relies on 2
boolean flags to determine device arrival and departure instead of
matching device names names.

Rapid events (e.g. a dock with many ports) are coalesced via a 10 ms
debounce timer with a 100 ms maximum delay ceiling, so we scan at most
once per burst while guaranteeing bounded latency during sustained event
storms.

Once the debounce timer fires, run a full re-enumeration via the Windows
backend, then diff the result using 2 new flags in winusb_device_priv:

seen_during_scan: set for every device the scan physically encounters.
Devices that remain unset after the scan have been physically removed.

seen_before_scan: explicitly set to true for all pre-existing devices
before the scan. Newly created devices start with false (usbi_alloc_device
uses calloc), so after the scan they can be identified as arrivals by the
combination seen_during_scan=true, seen_before_scan=false.

Note: made on top of #1761

### DIFF
--- a/libusb/os/windows_common.h
+++ b/libusb/os/windows_common.h
@@ -242,6 +242,10 @@ struct usbdk_device_priv {
 
 struct winusb_device_priv {
 	bool initialized;
+#if defined(LIBUSB_WINDOWS_HOTPLUG)
+	bool seen_during_scan; // set true for each device encountered during windows_get_device_list
+	bool seen_before_scan; // set true for each device encountered before windows_get_device_list
+#endif
 	bool root_hub;
 	uint8_t active_config;
 	uint8_t depth; // distance to HCD
@@ -275,7 +279,7 @@ struct usbdk_device_handle_priv {
 	// Not currently used
 	char dummy;
 };
- 
+
 enum WINUSB_ZLP {
 	WINUSB_ZLP_UNSET = 0,
 	WINUSB_ZLP_OFF = 1,
--- a/libusb/os/windows_hotplug.c
+++ b/libusb/os/windows_hotplug.c
@@ -25,16 +25,25 @@
 
 #include <stdio.h>
 #include <dbt.h>
-#include <usbiodef.h>
 
-/* The Windows Hotplug system is a three steps process.
- * 1. We create a monitor on GUID_DEVINTERFACE_USB_DEVICE via a hidden window.
- * 2. Upon notification of an event, we run the current windows backend to get
- *    the list of all devices.
- * 3. We use PDEV_BROADCAST_DEVICEINTERFACE->dbcc_name for finding device object
- *    in this list and generate LEFT or ARRIVED events libusb client via hotplug callbacks
+/* The Windows Hotplug system is a two steps process.
+ * 1. We create a hidden window and listen for DBT_DEVNODES_CHANGED, which Windows
+ *    broadcasts to all top-level windows whenever the device tree changes (no
+ *    registration required). Multiple rapid events (e.g. a hub with many children)
+ *    are coalesced via a short debounce timer so we only scan once the burst settles.
+ *    A maximum delay ceiling guarantees a scan fires within a bounded time even
+ *    during sustained bursts, preventing unbounded latency.
+ * 2. Upon timer expiry, we snapshot the current device list, run a full re-enumeration
+ *    via the Windows backend, then diff the result: newly found devices that have been
+ *    successfully initialized generate DEVICE_ARRIVED events; devices that were not
+ *    encountered by the re-enumeration (physically removed) generate DEVICE_LEFT events.
  */
 
+#define HOTPLUG_DEBOUNCE_TIMER_ID		1
+#define HOTPLUG_DEBOUNCE_MS				10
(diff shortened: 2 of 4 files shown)


seanm:
Nice work!  Great commit message, great code comments!

I don't know the Windows API though, so I'm not really the best reviewer, but I did read it all, and it looks reasonable.

mcuee:
Nice, now the hubs show up with this PR.

Acer Swift Go 14 Laptop 2024 model --> XikeStor Multifunction USB 3.2 Gen 2 Type C Hub --> Generic USB 3.0 Hub

Current git -- hubs do not show up as mentioned in Issue #1706.
```
PS C:\work\libusb\libusb> .\build\v145\x64\Release-Hotplug-MT\hotplugtest.exe
Device attached: 0bda:8156
No access to device: Operation not supported or unimplemented on this platform
Device attached: 05e3:0747
No access to device: Operation not supported or unimplemented on this platform
Device attached: 1ea7:0064
Device attached: 16c0:05dc
No access to device: Entity not found
Device attached: 046d:c548
No access to device: Entity not found
Device attached: 25a7:fa61
No access to device: Entity not found
Device detached: 0bda:8156
Device detached: 05e3:0747
Device detached: 1ea7:0064
Device detached: 16c0:05dc
Device detached: 046d:c548
Device detached: 25a7:fa61
```

This PR -- now USB hubs show up.
```
PS C:\work\libusb\libusb_pr1764> .\build\v145\x64\Release-Hotplug-MT\hotplugtest.exe
Device attached: 1ea7:0064
Device attached: 25a7:fa61
Device attached: 05e3:0747
No access to device: Operation not supported or unimplemented on this platform
Device attached: 16c0:05dc
Device attached: 046d:c548
Device attached: 0bda:8156
No access to device: Operation not supported or unimplemented on this platform
Device attached: 05e3:0625
No access to device: Operation not supported or unimplemented on this platform
Device attached: 05e3:0610
No access to device: Operation not supported or unimplemented on this platform
Device attached: 05e3:0610
No access to device: Operation not supported or unimplemented on this platform
Device attached: 05e3:0616
No access to device: Operation not supported or unimplemented on this platform
Device detached: 1ea7:0064
Device detached: 25a7:fa61
Device detached: 05e3:0747
Device detached: 16c0:05dc
Device detached: 046d:c548
Device detached: 0bda:8156
Device detached: 05e3:0625
Device detached: 05e3:0610
Device detached: 05e3:0610
Device detached: 05e3:0616
```
Screenshot 2026-03-06 210939

mcuee:
USB TreeViewer Gives more detailed info.
Screenshot 2026-03-06 212103
Screenshot 2026-03-06 212351

mcuee:
One more test for a "pure USB hub" which shows up as two hub devices.

Nothing will happen with current git.
```
PS C:\work\libusb\libusb> .\build\v145\x64\Release-Hotplug-MT\hotplugtest.exe
CTRL-C to exit
```

This PR is okay.
```
PS C:\work\libusb\libusb_pr1764> .\build\v145\x64\Release-Hotplug-MT\hotplugtest.exe
Device attached: 05e3:0612
No access to device: Operation not supported or unimplemented on this platform
Device attached: 05e3:0610
No access to device: Operation not supported or unimplemented on this platform
Device detached: 05e3:0610
Device detached: 05e3:0612
```

Screenshot 2026-03-06 212952

mcuee:
@sonatique 

BTW, you do not need to update version nano in the PR as it may cause conflict. I think the git commit step need to take care of that.

@tormodvolden can probably give you more clear instructions. I am not good at git.

sonatique:
> @sonatique
> 
> BTW, you do not need to update version nano in the PR as it may cause conflict. I think the git commit step need to take care of that.

Sorry the hook ran by mistake. I removed it and repushed.

mcuee:
@sonatique 
Please help to resolve the conflict and remove the changes to `version_nano.h`.

@tormodvolden 
Please help to review this one. I'd like this to be included in 1.0.30 release. It should be okay since I still do not enable hotplug by default for Windows.

tormodvolden:
Who merges this can fix up the version nano, but generally we should avoid changing it in PRs.

If this makes the hotplug experience closer to Linux and macOS I am all for it. I trust the reviews and testing, and yes, hotplug is not default yet.

tormodvolden:
The only thing I find negative here is the introduction of fixed (and arbitrary?) times, here for debounce. But sometimes they cannot be avoided.

PR #1765, opened 2026-03-06
ci: Retain artifacts from windows build

tormodvolden:
ci: Upload artifacts from the GitHub action windows builds

Retain the same files that we include in our release binary archive.


tormodvolden:
Thought I could spare appveyor from building this by removing appveyor.yml but it still builds.

tormodvolden:
Not sure why windows.yml is not building in this PR, maybe I broke it?

EDIT: I sure did.

mcuee:
@sonatique 

Please review as well. I think this is good to be in 1.0.30 release.

PR #1767, opened 2026-03-06
Specify SPDX-License-Identifier

seanm:
This is a standardized way to specify license and copyright in a machine-parseable way

See: https://spdx.dev/about/overview/

It's very helpful for supply chain and bill of materials tracking.

If we are agreed on this, I can do it in all files.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -5,6 +5,8 @@
  * Copyright © 2007-2008 Daniel Drake <dsd@gentoo.org>
  * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
  *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
--- a/Xcode/common.xcconfig
+++ b/Xcode/common.xcconfig
@@ -3,6 +3,8 @@
 // Copyright © 2012 Pete Batard <pete@akeo.ie>
 // For more information, please visit: <https://libusb.info>
 //
+// SPDX-License-Identifier: LGPL-2.1-or-later
+//
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
 // License as published by the Free Software Foundation; either
--- a/Xcode/debug.xcconfig
+++ b/Xcode/debug.xcconfig
@@ -3,6 +3,8 @@
 // Copyright © 2012 Pete Batard <pete@akeo.ie>
 // For more information, please visit: <https://libusb.info>
 //
+// SPDX-License-Identifier: LGPL-2.1-or-later
+//
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
 // License as published by the Free Software Foundation; either
--- a/Xcode/libusb.xcconfig
+++ b/Xcode/libusb.xcconfig
@@ -3,6 +3,8 @@
 // Copyright © 2012 Pete Batard <pete@akeo.ie>
 // For more information, please visit: <https://libusb.info>
 //
+// SPDX-License-Identifier: LGPL-2.1-or-later
+//
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
 // License as published by the Free Software Foundation; either
--- a/Xcode/libusb_debug.xcconfig
+++ b/Xcode/libusb_debug.xcconfig
@@ -3,6 +3,8 @@
 // Copyright © 2012 Pete Batard <pete@akeo.ie>
 // For more information, please visit: <https://libusb.info>
 //
+// SPDX-License-Identifier: LGPL-2.1-or-later
+//
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
 // License as published by the Free Software Foundation; either
--- a/Xcode/libusb_release.xcconfig
+++ b/Xcode/libusb_release.xcconfig
@@ -3,6 +3,8 @@
 // Copyright © 2012 Pete Batard <pete@akeo.ie>
 // For more information, please visit: <https://libusb.info>
 //
+// SPDX-License-Identifier: LGPL-2.1-or-later
+//
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
 // License as published by the Free Software Foundation; either
--- a/Xcode/release.xcconfig
+++ b/Xcode/release.xcconfig
@@ -3,6 +3,8 @@
 // Copyright © 2012 Pete Batard <pete@akeo.ie>
 // For more information, please visit: <https://libusb.info>
 //
+// SPDX-License-Identifier: LGPL-2.1-or-later
+//
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
 // License as published by the Free Software Foundation; either
--- a/android/config.h
+++ b/android/config.h
@@ -2,6 +2,8 @@
  * Android build config for libusb
  * Copyright © 2012-2013 RealVNC Ltd. <toby.gray@realvnc.com>
  *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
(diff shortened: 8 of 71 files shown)


mcuee:
I am okay with this proposal.

tormodvolden:
I have seen somewhere a recommendation (citation needed :-) ) to spell out "Copyright" and <year> <name> for legal reasons (e.g. just not the copyright symbol). I wonder is it safe to remove the old wording? You would think a standard like SPDX has this taken into consideration, but IANAL.

Also would be nice if e.g. @hjelmn and @dickens would chime in, since they have their copyrights many places.

mcuee:
We can probably follow Linux kernel.

Email address may be a bit tricky as it may change. But we can probably follow Linux to remove email address when we know the changes.

Example from Linux kernel.

Greg KH's email will not change for now. Alan Stern's email address will not change. David Brownell passed away in early 2011.

https://github.com/torvalds/linux/blob/master/drivers/usb/usb-skeleton.c
```
// SPDX-License-Identifier: GPL-2.0
/*
 * USB Skeleton driver - 2.2
 *
 * Copyright (C) 2001-2004 Greg Kroah-Hartman (greg@kroah.com)
 *
 * This driver is based on the 2.6.3 version of drivers/usb/usb-skeleton.c
 * but has been rewritten to be easier to read and use.
 */
````

https://github.com/torvalds/linux/blob/master/drivers/usb/host/ehci-hcd.c
```
// SPDX-License-Identifier: GPL-2.0+
/*
 * Enhanced Host Controller Interface (EHCI) driver for USB.
 *
 * Maintainer: Alan Stern <stern@rowland.harvard.edu>
 *
 * Copyright (c) 2000-2004 by David Brownell
 */
```

**Comment by mcuee** [comment]
Some Linux kernel source codes do not list email address.

https://github.com/torvalds/linux/blob/master/drivers/usb/core/hcd.c
```
// SPDX-License-Identifier: GPL-2.0+
/*
 * (C) Copyright Linus Torvalds 1999
 * (C) Copyright Johannes Erdfelt 1999-2001
 * (C) Copyright Andreas Gal 1999
 * (C) Copyright Gregory P. Smith 1999
 * (C) Copyright Deti Fliegl 1999
 * (C) Copyright Randy Dunlap 2000
 * (C) Copyright David Brownell 2000-2002
 */
```

**Comment by seanm** [comment]
> I have seen somewhere a recommendation (citation needed :-) ) to spell out "Copyright" and for legal reasons (e.g. just not the copyright symbol).

We are already using only the copyright symbol. (Also, I've never heard such a recommendation).

>I wonder is it safe to remove the old wording? You would think a standard like SPDX has this taken into consideration, but IANAL.

Removing the old text can be a future PR is anyone wants that.

> Also would be nice if e.g. @hjelmn and @dickens would chime in, since they have their copyrights many places.

It would, but I don't recall them participating for months/years, so we should be prepared to decide without them.

>We can probably follow Linux kernel.

I wonder why they've used SPDX-License-Identifier but not SPDX-FileCopyrightText.

**Comment by tormodvolden** [comment]
> We are already using only the copyright symbol.

No, we never use only the copyright symbol, there is always "Copyright" spelled out next to it.

> Removing the old text can be a future PR is anyone wants that.

Your example commit is removing the old text.

If your intention is just to _add_ the SPDX info, I have no concerns with that and it would be less important to collect other opinions.

**Comment by seanm** [comment]
> Your example commit is removing the old text.

Wait, which text are you talking about?  I didn't remove the whole license text.  The linux example https://github.com/torvalds/linux/blob/master/drivers/usb/core/hcd.c had no license text, only the string `SPDX-License-Identifier`.

The only thing I removed was the email addresses, but I could bring them back. I recall in another PR we talked about removing them.

**Comment by seanm** [comment]
> I have seen somewhere a recommendation (citation needed :-) ) to spell out "Copyright" and for legal reasons (e.g. just not the copyright symbol).

I've done some reading on this, and there's no difference.  Since © is shorter, and we sometimes have multiple people's names, I'll use just the symbol to keep the line shorter.

I've updated this PR with more (but not yet all) files changed.

File: libusb/core.c [hunk 1]
```diff
@@ -1,9 +1,9 @@
 /* -*- Mode: C; indent-tabs-mode:t ; c-basic-offset:8 -*- */
 /*
  * Core functions for libusb
- * Copyright © 2012-2023 Nathan Hjelm <hjelmn@cs.unm.edu>
- * Copyright © 2007-2008 Daniel Drake <dsd@gentoo.org>
- * Copyright © 2001 Johannes Erdfelt <johannes@erdfelt.com>
+ *
+ * SPDX-FileCopyrightText: © 2001 Johannes Erdfelt, 2007-2008 Daniel Drake, 2012-2023 Nathan Hjelm
```
mcuee: Multiple lines will be better.

seanm: I don't think SPDX parsing tools expect multiple lines, but I'll investigate...

seanm: From what I see, it has to be one line.

So for now, I'll concentrate on just adding the license...

mcuee: I see. Maybe it is worth tell the developer of the SPDX Parsing Tools.

mcuee (changes requested):
Unitl the SPDX parsing tool supports multiple lines, or we forget about the limited SPDX parsing tool, I will not approve this PR. In its current form, the chages to add  "SPDX-FileCopyrightText:" are NO-GO for me.

Edit -- it is good that this PR just concentrating on dding the license only.

mcuee (approved):

seanm:
So this PR now only adds `SPDX-License-Identifier`. Note that most code is LGPL, but a couple of files state GPL. Is this correct or accidental?

Youw (approved):
I never wanted to understand any of these licensing-related quirks and each time it makes me want to quit doing SW development at all.

But I don't see any obvious issues in bere.

Youw:
**Re: the dozens of repeated Copilot review comments** — they're all variations of one point, so to consolidate it in a single place: Copilot wants `LGPL-2.1+` written in SPDX's canonical form `LGPL-2.1-or-later`.

It's essentially right, with one nuance worth noting: `LGPL-2.1+` isn't strictly *invalid* — the trailing `+` ("or later") is valid SPDX license-**expression** syntax — but the predefined `LGPL-2.1+` **identifier** was deprecated when SPDX 2.0 introduced license expressions, in favour of the explicit `-only` / `-or-later` identifiers. Since this PR is specifically about machine-parseable license metadata, the non-deprecated `LGPL-2.1-or-later` is the form to prefer.

Practically it's a single `LGPL-2.1+` → `LGPL-2.1-or-later` find-and-replace across the touched files, which clears all of these comments at once.

Refs: canonical id → https://spdx.org/licenses/LGPL-2.1-or-later.html ; `LGPL-2.1+` is listed in the deprecated section of https://spdx.org/licenses/ . (The Linux kernel's `license-rules.rst` likewise standardizes on the `-or-later` / `-only` identifiers.)

*Assisted-by: claude-code:claude-opus-4-8*

Youw (approved):

Youw:
**CI root cause** — the failing checks (Fuzzing + the non-Windows `build` jobs) all abort at the same spot, the `bootstrap` / `autoreconf` step:

```
m4:configure.ac:2: Warning: excess arguments to builtin `m4_define' ignored
autom4te: error: /usr/bin/m4 failed with exit status: 1
autoreconf: error: aclocal failed with exit status: 1
```

`configure.ac:2` is:

```m4
m4_define([LU_VERSION_H], m4_include([libusb/version.h]))
```

so `libusb/version.h` is spliced into m4 as the second argument of `m4_define`. The license header this PR prepends to `version.h` contains top-level commas (e.g. `Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA`), which m4 reads as extra `m4_define` arguments — hence "excess arguments", and bootstrap aborts.

That also explains the green checks: the MSVC and Android/NDK jobs build from `.sln` / `ndk-build` and never run `autoreconf`, so they don't hit this.

The header was inserted directly above this existing line in the very same file:

```c
/* This file is parsed by m4 and windres and RC.EXE so please keep it simple. */
```

Fix: for `version.h` only, use the single-line tag `/* SPDX-License-Identifier: LGPL-2.1-or-later */` (no commas) instead of the full multi-line license block — i.e. honour that "keep it simple" note for this one m4-parsed file. The rest of the PR is unaffected.

*Assisted-by: claude-code:claude-opus-4-8*

seanm:
> contains top-level commas (e.g. `Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA`), which m4 reads as extra `m4_define` arguments —

OK, fixed by removing that stuff.

Youw (approved):
Now we should be good.

seanm:
I'll wait a few more days in case @tormodvolden or @sonatique want to comment further...

mcuee:
@Youw 
Nice analysis from AI.

PR #1769, opened 2026-03-07
windows: Remove ARM32 from MSVC, fix PlatformToolsetVersion

sonatique:
While working on the "getopt" PR, I realize building for ARM32 (ARM target) was broken because of undefined ref in props files. But after fixing them, I then realized support for building for ARM was removed from VS2026 (there are no more "Windows for ARM32" perspectives). I am therefore proposing to remove this altogether from MSCV builds.
 
Remove 32-bit ARM platform configurations from the MSVC solution and project files. ARM32 support was dropped by the v145 toolset (Visual Studio 2026+) and the configurations could not be built. ARM32 cross-compilation remains possible with other toolchains (e.g. MinGW).

Also fix two references to the undefined $(PlatformToolsetVersion) MSBuild property in Base.props and xusb.vcxproj, replacing them with $(LibusbPlatformToolsetVersion) which is properly defined in Configuration.Base.props and used consistently elsewhere.


mcuee:
@sonatique 

I agree to remove ARM32 from MSVC.

While you are on this one, I think it is also a good idea to update the following convinent PowerShell Script.
https://github.com/libusb/libusb/blob/master/msvc/build_all.ps1

1) Remove "ARM" which means ARM32
2) Add more Windows configurations: `Release-Hotplug`, `Release-Hotplug-MT`, `Debug-Hotplug`, `Debug-Hotplug-MT`.

I actually like to remove VS2013 and VS2015 support as well but I guess others may not like the idea. They do not support ARM64.

tormodvolden (approved):

sonatique:
> While you are on this one, I think it is also a good idea to update the following convinent PowerShell Script. https://github.com/libusb/libusb/blob/master/msvc/build_all.ps1
> 
> 1. Remove "ARM" which means ARM32
> 2. Add more Windows configurations: `Release-Hotplug`, `Release-Hotplug-MT`, `Debug-Hotplug`, `Debug-Hotplug-MT`.
> 
> I actually like to remove VS2013 and VS2015 support as well but I guess others may not like the idea. They do not support ARM64.

@mcuee : what about this: https://github.com/libusb/libusb/pull/1784

PR #1782, opened 2026-03-10
Add many explicit casts for C++ compatibility

seanm:
Building as C++ can be useful for various code analysis tools or compiler flags that only work in C++.  And so few changes are needed to achieve this, and C compatibility is of course not broken.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -782,7 +782,7 @@ void usbi_disconnect_device(struct libusb_device *dev)
 /* Perform some final sanity checks on a newly discovered device. If this
  * function fails (negative return code), the device should not be added
  * to the discovered device list. */
-int usbi_sanitize_device(struct libusb_device *dev)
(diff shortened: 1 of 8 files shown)


hjelmn:
Seems reasonable to me. Mostly just using proper casting where C lets us get away without.

seanm:
I *wouldn't* merge this for 1.0.30, but let's merge it right after.

Note that even with the very newest Xcode (26.3), `-Wthread-safety` still doesn't work properly in C, only works in C++.

mcuee (approved):

mcuee:
@seanm 

Please help to resolve the conflict.

Youw (approved):
I didn't find anything wrong with this.
I'm thinking maybe add a CI job to actually test the C++ compatibility if we're to keep fixes like this.
Other than that - looks good.

PR #1784, opened 2026-03-13
Update build_all.ps1 to match project configurations + remove VS2013 and VS2015 support

sonatique:


mcuee (approved):

mcuee:
@sonatique 

I like this changes.

But you may need to have another PR to remove VS2013 and VS2015 support for appveyor as that still has VS2013 and VS2015. While you do that, you may want to add VS2026 support as well.
https://github.com/libusb/libusb/blob/master/appveyor.yml

This is also why appveyor CI build fails.
https://ci.appveyor.com/project/LudovicRousseau/libusb/builds/53695597

Edit to add:
I think it is probably easier to combine the change to appveyor in this PR as well.

mcuee:
@tormodvolden 

Please review this change as well to see whether you like the idea or not. Thanks.

seanm:
There's some code cleanup that could be done if/when we drop these old MSVC. For example, search for `_snprintf`, hacks like this could be removed:

```
#if defined(_MSC_VER)
#define snprintf _snprintf
```

mcuee:
> There's some code cleanup that could be done if/when we drop these old MSVC. For example, search for `_snprintf`, hacks like this could be removed:
> 
> ```
> #if defined(_MSC_VER)
> #define snprintf _snprintf
> ```

Good point about cleanup. 

On the other hand, removing the old MSVC version itself only affects two places. Other things like the one above is not related to this PR.

libusbi.h
```
#if defined(_MSC_VER) && (_MSC_VER < 1900)
#include <stdio.h>
#define snprintf usbi_snprintf
#define vsnprintf usbi_vsnprintf
int usbi_snprintf(char *dst, size_t size, const char *format, ...);
int usbi_vsnprintf(char *dst, size_t size, const char *format, va_list args);
#define LIBUSB_PRINTF_WIN32
#endif /* defined(_MSC_VER) && (_MSC_VER < 1900) */
```

msvc/config.h --> already corrected by this PR.

Visual Studio Version [1, 2, 3, 4, 5, 6, 7, 8] | _MSC_VER Value
-- | --
Visual Studio 2026 (18.0) | 1950+
Visual Studio 2022 (17.0) | 1930+
Visual Studio 2019 (16.0) | 1920+
Visual Studio 2017 (15.0) | 1910+
Visual Studio 2015 (14.0) | 1900
Visual Studio 2013 (12.0) | 1800
Visual Studio 2010 (10.0) | 1600

mcuee:
@sonatique 

Please add the change to `libusbi.h` as well.

seanm (approved):
I'm not a Windows guys, but this looks reasonable.

mcuee:
@sonatique 

I am okay with either way.
1) merge this first and then have two others PR to cleanup for AppVeyor and left over codes mentioned by @sean.
2) or combine all the things here.

Maybe Option 1 is better.

PR #1785, opened 2026-03-13
core: fix crash on device destruction when already detached

### LINKED ISSUE #1599: Crash when re-enumerating a device with different class GUID under Windows
I am debugging a problem with libusb (current version, 1.0.27), which crashes in a corner case situation. To reproduce the problem a "cooperation from outside the app" is required, therefore I do not share directly a code snippet, just steps describing the situation. I wonder if somebody can comment - if more input is required, please let me know.

The problem is related to Windows platform and can be reproduced as follows:

1. The application (more precisely a library) enumerates the devices using libusb_get_device_list(), collects infos about the devices and (important) increases refcount (libusb_ref_device()) of the device(s) potentially interesting for the library users (more info can be queried about those devices later or the devices can be opened.
2. While that ref is taken, another driver was installed for given device (in the particular case where we observed the problem, WinUSB driver was installed for the device while the library was running.
3. Later a new device enumeration round (libusb_get_device_list()) was triggered. During this operation, libusb sees the device enumerated with a different class GUID. This happens inside winusb_get_device_list(), within following check
https://github.com/libusb/libusb/blob/de38189e8014fa393f4d8c1d9d3fdf5e2a95899d/libusb/os/windows_winusb.c#L1895-L1900
As a side note, this code was introduced in
https://github.com/libusb/libusb/commit/bb021598078eeea9efdbee131909f62c169a4d75
while fixing
https://github.com/libusb/libusb/issues/897
4. When seeing different GUID, libusb calls explicitly usbi_disconnect_device() on the device, removing it from the list and allocates the new one. It calls usbi_disconnect_device() explicitly, even if the call is immediately followed by libusb_unref_device() which would call usbi_disconnect_device() anyway if (important) there were no references to it.
5. Later, during cleanup, our library calls finally itself libusb_unref_device() on the "old" device which is no more enumerated. During this call the refcount goes to zero, libusb attempts to call usbi_disconnect_device() again, but because it was disconnected already in step 4 and the list pointers are NULL, the list_del call
https://github.com/libusb/libusb/blob/de38189e8014fa393f4d8c1d9d3fdf5e2a95899d/libusb/core.c#L751
crashes when dereferencing the NULL pointers.

I wonder whether in step 4 usbi_disconnect_device() should be called explicitly or if just the libusb_unref_device() should suffice, letting the disconnect happen only on last reference - I am not familiar enough with libusb internals to be confident about all consequences, thus I prefer to ask.  Or should the list_del() call be guarded to not happen if the list pointers are NULL already?  Or is the above use case somehow "illegal"? But the only thing our lib did in that case was that it took ref to the device which was later rediscovered with different class GUID.

Thanks for any hints. Jan

sonatique:
When a device's class GUID or instance ID changes during re-enumeration, the Windows backend calls usbi_detach_device() followed by libusb_unref_device(). If the application still holds a reference, the device is not freed immediately. Later, when the application drops its last reference, libusb_unref_device() calls usbi_disconnect_device() again (only for non-hotplug backends), causing a double list_del() on already-stale list pointers and crashing.

Guard the disconnect call with a check on dev->attached, which is already set to 0 by the earlier usbi_detach_device() call.

This fix is safe for all backends, not just Windows: if any backend detaches a device before its refcount drops to zero, the double- disconnect is now prevented.

Closes #1599

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1338,7 +1338,8 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
 		if (usbi_backend.destroy_device)
 			usbi_backend.destroy_device(dev);
 
-		if (!libusb_has_capability(LIBUSB_CAP_HAS_HOTPLUG)) {
+		if (!libusb_has_capability(LIBUSB_CAP_HAS_HOTPLUG) &&
+		    usbi_atomic_load(&dev->attached)) {
 			/* backend does not support hotplug */
 			usbi_disconnect_device(dev);
 		}
--- a/libusb/version_nano.h
+++ b/libusb/version_nano.h
@@ -1,1 +1,1 @@
-#define LIBUSB_NANO 12018
+#define LIBUSB_NANO 12020


mcuee:
Good analysis. I agree with the changes.

PR #1658, opened 2025-06-09
darwin I/O-less enumeration

tormodvolden:
This could potentially speed up things. Not tested.

Seems like there is a bcdUSB property from SDK 10.15 and on, defined as kUSBHostDevicePropertyStandardVersion in IOUSBHostFamilyDefinitions.h . Or as kUSBSpecReleaseNumber already from 10.10?

PR #1789, opened 2026-03-14
darwin: synthesize device descriptor from OS-cached values

### LINKED ISSUE #1564: Behringer UMC204HD "reboot" cycle
I have a Behringer UMC204HD USB DAC connected via a USB3 port on a Thunderbolt dock which gets into what looks like an endless reboot loop as long as I run an application that uses libusb. The example that interests me is the `usbmuxd` implementation from `libimobiledevice` but the same thing happens when I activate WebUSB in Chromium.

I presume that the underlying error is in the device firmware. If so it will probably not get fixed because Behringer cannot reproduce the issue.

Ignoring the issue is not an option as the power circuits in the DAC keep getting restarted leading to pops in attached speakers or phones (and lots of System.log pollution). For now my workaround is to disconnect the device when I need usbmuxd (to access data on an iPhone no longer supported on my Mac OS version) but that isn't satisfactory.

I see 2 better workarounds
- blacklist the device at the libusb level. This seems like a reasonable feature on platforms where libusb isn't more or less the system USB library and access to most devices will probably go through the system equivalent(?)
- blacklist or ignore it in usbmuxd

The latter will presumably be a lot easier for me to implement myself of libusb already has a way for client software to indicate that it's only interested in devices of a specific vendor, for instance? (This assumes devices don't get an init/reset command just to read their IDs...)

I presume libusb doesn't send initialisation commands to all attached devices continuously, so what sequence of events at what place should we be looking for? Is there a code path that keeps getting executed until a device responds in the desired way and that I could shortcut as an initial approach?

Pointers would be appreciated here given that I am familiar with neither libusb nor usbmuxd.

sonatique:
On macOS, darwin_cache_device_descriptor() sends a USB control transfer to every device during enumeration. This causes some devices (e.g. Behringer UMC204HD) to malfunction or reboot in a loop.

The macOS IOKit framework already caches all device descriptor fields in the IORegistry and IOUSBDeviceInterface. Read bcdUSB and bMaxPacketSize0 from the IORegistry, and the remaining fields via IOUSBDeviceInterface getter methods, to synthesize the device descriptor without any USB I/O.

If either IORegistry value is unavailable, fall back to the existing control transfer path unchanged.

Closes #1564

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1180,6 +1180,39 @@ static enum libusb_error darwin_cache_device_descriptor (struct libusb_context *
   (*device)->GetDeviceProduct (device, &idProduct);
   (*device)->GetDeviceVendor (device, &idVendor);
 
+  /* Try to synthesize device descriptor from OS-cached values (no USB I/O).
+   * If any value is unavailable, fall back to requesting descriptor from device. */
+  do {
+    IOUSBDeviceDescriptor *desc = &dev->dev_descriptor;
+    UInt16 bcdDevice, bcdUSB;
+
+    if (!get_ioregistry_value_number (dev->service, CFSTR("bcdUSB"), kCFNumberSInt16Type, &bcdUSB))
+      break;
+    if (!get_ioregistry_value_number (dev->service, CFSTR("bMaxPacketSize0"), kCFNumberSInt8Type, &desc->bMaxPacketSize0))
+      break;
+
+    desc->bcdUSB = libusb_cpu_to_le16 (bcdUSB);
+    desc->bDeviceClass = bDeviceClass;
+    (*device)->GetDeviceSubClass (device, &desc->bDeviceSubClass);
+    (*device)->GetDeviceProtocol (device, &desc->bDeviceProtocol);
+    desc->idVendor = libusb_cpu_to_le16 (idVendor);
+    desc->idProduct = libusb_cpu_to_le16 (idProduct);
+    (*device)->GetDeviceReleaseNumber (device, &bcdDevice);
+    desc->bcdDevice = libusb_cpu_to_le16 (bcdDevice);
+    (*device)->USBGetManufacturerStringIndex (device, &desc->iManufacturer);
+    (*device)->USBGetProductStringIndex (device, &desc->iProduct);
+    (*device)->USBGetSerialNumberStringIndex (device, &desc->iSerialNumber);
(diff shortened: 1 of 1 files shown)


tormodvolden:
This is weird. It looks almost identical to the code I posted in issue #1564. This is pretty straight-forward and I suppose there are not so many ways of doing it, but still, only some comment wording detail and whitespace is different.

seanm:
> This is weird. It looks almost identical to the code I posted in issue #1564.

@sonatique said "It is basically your idea / code".

sonatique:
Yes  @tormodvolden : sorry, I didn't want to steal you code. I just wanted things to move forward. I tried to find a solution and independently came to the same idea / implementation bits as you did last year. So instead of rewriting it from scratch I took your code and just completed it (for instance I get the bcdValue (0x200) from a call rather than hardcoding it (I implemented your FIXME at line 1143). I added a few more words to the comments and added the fallback to requesting the descriptor when nothing is cached. That's all more or less. That why I wrote:
"@tormodvolden please have a look at https://github.com/libusb/libusb/pull/1789 . It is basically your idea / code since I came to the almost exact solution, with a few clean-up. Should be good to go."

You can still claim ownership and open a PR with your WIP no problem for me, and thank you for the snipper !

tormodvolden:
I had already added the bcdValue readout and and hardware I/O fallback... https://github.com/libusb/libusb/issues/1564#issuecomment-2351174588 and https://github.com/tormodvolden/libusb/tree/darwin_ioless_enum :) So it was even more identical than what you may have thought - in fact seeing last week's comments I may now believe that you only looked up a single commit linked by mcuee last week instead of reading the thread and looking at the branch. Although that commit already had the I/O fallback that you "added" so I don't understand it completely, but nvm.

As seen in the comments, I was awaiting testing before making a PR of it. But I could of course have made a draft PR, I'll do that [EDIT: wait, I did that last year already #1658]. I think it is fair to be credited, not only because of the simple code, but also the idea and investigation together with the bug reporter.

sonatique:
Ok @tormodvolden, I closed mine. Sorry for the troubles.

seanm:
>As seen in the comments, I was awaiting testing before making a PR of it.

Testing is much easier when there's a PR for someone to fetch and build locally. The work you started has been untouched for 1.5 years, so I think it was great than @sonatique moved it forward.

How about we add `Co-authored-by:` in the commit message to give @tormodvolden credit and actually get this thing merged before another 1.5 years goes by.

PR #1790, opened 2026-03-16
Windows: Fix Error 31 in libusb_get_device_string by querying Language ID

Fishi96:
### Description of the Issue
Currently, on Windows, when `libusb_get_device_string()` is called, the underlying implementation constructs the control request for the string descriptor with the Language ID strictly hardcoded to `0`. 

While some devices tolerate a Language ID of `0`, on certain USB devices this results in a failure during the `DeviceIoControl` call, returning Windows System Error 31 (`ERROR_GEN_FAILURE` - A device attached to the system is not functioning).

### Root Cause
The USB specification dictates that string descriptors should be requested using a valid Language ID (LANGID) supported by the device. Requesting a specific string descriptor with `LANGID = 0` is out of spec, as index 0 is reserved for retrieving the array of supported LANGIDs itself.

### Proposed Fix
This patch modifies the behavior to correctly follow the USB specification:
1. It first issues a request for string descriptor index `0` to retrieve the array of supported Language IDs.
2. It extracts a valid Language ID (typically the first available one).
3. It then uses this correctly retrieved Language ID in the subsequent `DeviceIoControl` request to fetch the actual string descriptor.

This ensures proper compatibility with strictly compliant USB devices and resolves the Error 31 without breaking existing functional devices.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -2272,7 +2272,20 @@ static int winusb_get_device_string(libusb_device *dev,
 	BOOL result = DeviceIoControl(hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &sd, size,
 		&sd, size, &ret_size, NULL);
 
-	USHORT langId = result ? (USHORT)(sd.desc.bString[0] | (sd.desc.bString[1] << 8)) : 0;
+	if (!result) {
+		usbi_err(ctx, "could not access string descriptor 0 for '%s': %s", priv->dev_id, windows_error_str(0));
+		CloseHandle(hub_handle);
+		return LIBUSB_ERROR_ACCESS;
+	}
+
+	USHORT langId = 0;
+	if (sd.desc.bDescriptorType != LIBUSB_DT_STRING) {
+		usbi_warn(ctx, "descriptor 0 not a string descriptor for '%s'", priv->dev_id);
+	} else if (sd.desc.bLength < 4) {
+		usbi_warn(ctx, "string descriptor 0 too short for '%s'", priv->dev_id);
+	} else {
+		langId = (USHORT)(sd.desc.bString[0] | (sd.desc.bString[1] << 8));
+	}
 
 	size = sizeof(sd);
 	memset(&sd, 0, size);


mcuee:
Does this affect other OS?

Fishi96:
> Does this affect other OS?

I only tested on Windows and Linux. The patch works on Windows, and Linux is unaffected.

libusb/os/windows_winusb.c
```diff
@@ -2265,10 +2265,24 @@ static int winusb_get_device_string(libusb_device *dev,
 	sd.req.ConnectionIndex = (ULONG)dev->port_number;
 	sd.req.SetupPacket.bmRequest = LIBUSB_ENDPOINT_IN;
 	sd.req.SetupPacket.bRequest = LIBUSB_REQUEST_GET_DESCRIPTOR;
-	sd.req.SetupPacket.wValue = (LIBUSB_DT_STRING << 8) | string_descriptor_idx;
+	sd.req.SetupPacket.wValue = (LIBUSB_DT_STRING << 8) | 0;
 	sd.req.SetupPacket.wIndex = 0;
 	sd.req.SetupPacket.wLength = (USHORT)sizeof(sd.desc);
 
+	BOOL result = DeviceIoControl(hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &sd, size,
```
sonatique, line 2272: I think result shall be checked here: we shall not continue when this call fails (we'll get problem later for sure). Maybe add debug message and returns.

libusb/os/windows_winusb.c
```diff
@@ -2265,10 +2265,24 @@ static int winusb_get_device_string(libusb_device *dev,
...
+		&sd, size, &ret_size, NULL);
+
+	USHORT langId = result ? (USHORT)(sd.desc.bString[0] | (sd.desc.bString[1] << 8)) : 0;
```
sonatique: I think we shall validate the response (per USB 2.0 §9.6.7) before extracting langId, especially bLength to avoid overrun:
bLength must be >= 4 (2 bytes header + at least one 16-bit LANGID)
and
bDescriptorType must be USB_DT_STRING (0x03)

It think we shall not try to get langId when any of this two field is not as expected

libusb/os/windows_winusb.c
sonatique, line 2272: (Note: we now have 2 DeviceIoControl calls when we had 1, this may have a performance impact, but apart from caching I don't know how to avoid id.)

sonatique:
minor: typo in the comment: "read languege" shall be "read language"

libusb/os/windows_winusb.c
Fishi96: If conditions are not met, then we go with default zero for langId in main request, right?

Fishi96: I have pushed requested changes

libusb/os/windows_winusb.c
Fishi96: I have pushed requested changes

mcuee:
* https://github.com/libusb/libusb/pull/1796#issuecomment-4217724456

@Fishi96 
Thanks a lot for the contribution. We may use PR #1796 instead of this PR.

mcuee:
Closing this PR in favor of #1796.

@Fishi96 
Stilll thanks a lot for your contributions!

PR #1797, opened 2026-04-10
libusb.h: Fix typo in comment

smarvonohr:
The comment surely meant to say 3/2 instead of 2/3.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1684,7 +1684,7 @@ enum libusb_device_string_type {
  *  => 126 * 3 + 1
  *  => 382 bytes
  * 
- * Stay with 256 * 2/3 = 384 to be safe.
+ * Stay with 256 * 3/2 = 384 to be safe.
  */
 #define LIBUSB_DEVICE_STRING_BYTES_MAX  (384U)
  
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1681,7 +1681,7 @@ enum libusb_device_string_type {
  * 255 max descriptor length with 2 byte header 
  *  => 253 bytes UTF-16LE, no null termination (USB 2.0 9.6.7)
  *  => 126.5 codepoints
- *  => 126 * 3 + 1
+ *  => 127 * 3 + 1
  *  => 382 bytes
  * 
  * Stay with 256 * 3/2 = 384 to be safe.


mcuee:
Good catch. Thanks.

@sonatique 
This one can be safely merged.

mcuee (dismissed):

tormodvolden:
I agree to the correction. But what does 3/2 actually signify? Just that 384 is a "rounder" number than 382? Since we have a precise calculation, why don't we use the precise number 382? @mliberty1

mcuee:
Ah, looking at the comments, it has more typos. 

```
/** \ingroup libusb_desc
 * The maximum length for a device string descriptor in UTF-8.
 * 
 * 255 max descriptor length with 2 byte header 
 *  => 253 bytes UTF-16LE, no null termination (USB 2.0 9.6.7)
 *  => 126.5 codepoints
 *  => 126 * 3 + 1
 *  => 382 bytes
 * 
 * Stay with 256 * 2/3 = 384 to be safe.
 */
```

126 * 3 + 1 = 379

127 * 3 + 1 = 382

mcuee:
Examples from Device side USB stacks.

Raspberry Pi PICO SDK
https://forums.raspberrypi.com/viewtopic.php?t=349003
https://github.com/raspberrypi/pico-sdk/blob/2.2.0/src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c#L154-L162

```
const uint16_t *tud_descriptor_string_cb(uint8_t index, __unused uint16_t langid) {
#ifndef USBD_DESC_STR_MAX
#define USBD_DESC_STR_MAX (20)
#elif USBD_DESC_STR_MAX > 127
#error USBD_DESC_STR_MAX too high (max is 127).
#elif USBD_DESC_STR_MAX < 17
#error USBD_DESC_STR_MAX too low (min is 17).
#endif
    static uint16_t desc_str[USBD_DESC_STR_MAX];
```

mcuee (commented):
Hmm, we need to relook at the comments again.

mcuee:
It seems to me the comments kind overthink. We should be safe with 255.
bLength is a byte, max is 255.

https://www.usb.org/document-library/usb-20-specification

Screenshot 2026-04-11 211703
Screenshot 2026-04-11 211619

mliberty1:
You are right, it should be `3/2`, not` 2/3`.

The whole issue here is that the descriptor contains UTF-16LE, and we are storing UTF-8.  To be fully safe, the UTF-8 representation has to store all the UTF-16LE code points given the worst-case expansion as computed in this comment.

I erred on the side of allocating too many bytes just to be safe.  Since UTF-16LE is also null terminated, you could argue that we only need
(255 - 1 bLength - 1 bDescriptorType - 2 NULL) / 2 = 125 code points
=> 125 * 3 + 1 NULL = 376 bytes

However, if someone doesn't NULL terminate their string as an attack, libusb could have a problem.  

As the comment says, we reserve a few more bytes just to be completely safe.

mcuee:
In that case, two changes are needed.
```
/** \ingroup libusb_desc
 * The maximum length for a device string descriptor in UTF-8.
 * 
 * 255 max descriptor length with 2 byte header 
 *  => 253 bytes UTF-16LE, no null termination (USB 2.0 9.6.7)
 *  => 126.5 codepoints
 *  => 126 * 3 + 1 ==> 127 * 3 + 1
 *  => 382 bytes
 * 
 * Stay with 256 * 2/3 = 384 to be safe. ==> 256 * 3/2 = 384
 */
```

mcuee:
BUT I still have doubts that it is really meant to go above 255 Bytes...

mcuee:
@smarvonohr 

For the purpose of this PR, please add one more change I mentioned. Thanks.
```
126 * 3 + 1 ==> 127 * 3 + 1
```

smarvonohr:
> BUT I still have doubts that it is really meant to go above 255 Bytes...

It's not, at least on the USB transfer side. But it's converted from UTF-16 to UTF-8 inside libusb. This may or may not (depending on the actual contents) increase the required size in bytes.

> For the purpose of this PR, please add one more change I mentioned. Thanks.

Done.

mcuee (approved):

seanm (approved):

Issue #1727, opened 2025-12-30
RFC: abandon version_nano.h in favour of github "Merge pull request" feature


seanm:
1. version_nano.h is nice
2. github's "Merge pull request" feature is nice

But 1 prevents us from using 2, per https://github.com/libusb/libusb/wiki/Maintainers'-corner

Is this still a good trade-off?

I think no. Unless there's a way to have both, I find 2 more useful than 1. Thoughts?

mcuee:
I vote to give up the version_nano.h feature.

mcuee:
@tormodvolden and @sonatique 

Any objections?

sonatique:
No objection. I sometimes found it useful in my dev cycles but I favor being able to use standard GitHub merge.

Youw:
Is there a specific suggestion how to replace/abandone it?
There might be lots of code that relies on existance of `LIBUSB_NANO`...
A quick search revealed for example [this](https://github.com/Fraunhofer-IIS/flexiband/blob/4231f9aa01c5ac2f6cd5736cbf4f00893865c934/driver/unix/src/libusb_version_fixes.h#L3).

I don't like it either, and I tried to think of how to workaround it, and I don't remember having any good solutions at the time.

dickens:
One thing I’d like to point out is that the nano version provides a cross-reference to the specific commit (assuming the nano is properly kept in sync).  This allows anyone to know exactly which libusb source code is being used as this information appears in the debug log.  Unless the same sort of cross-reference is kept, I do not agree with removing the nano version.

dickens:
> There might be lots of code that relies on existance of `LIBUSB_NANO`... A quick search revealed for example [this](https://github.com/Fraunhofer-IIS/flexiband/blob/4231f9aa01c5ac2f6cd5736cbf4f00893865c934/driver/unix/src/libusb_version_fixes.h#L3).

I believe this is a non-issue.  The nano version is an internal value that is only accessible through the libusb version structure, thus no outside sources can directly use this.

mcuee:
> One thing I’d like to point out is that the nano version provides a cross-reference to the specific commit (assuming the nano is properly kept in sync).  This allows anyone to know exactly which libusb source code is being used as this information appears in the debug log.  Unless the same sort of cross-reference is kept, I do not agree with removing the nano version.

The replacement can be the git commit ID which many projects use to identify snapshot releases.

dickens:
[...]

PR #1799, opened 2026-04-14
CI: add automatic nano version increment workflow

sonatique:
Note: after merge one need to got to
GitHub Settings -> General -> Pull Requests and uncheck "Allow merge commits" and "Allow squash merging",
leaving only "Allow rebase merging"

Addresses #1727

### DIFF
--- a/.github/workflows/increment-nano.yml
+++ b/.github/workflows/increment-nano.yml
@@ -1,0 +1,181 @@
+# Auto-increment version_nano.h after each push to master.
+#
+# This replaces the local pre-commit / post-rewrite git hooks that
+# maintainers previously had to configure manually, and allows the
+# standard GitHub "Rebase and merge" button to be used on PRs.
+#
+# Behaviour:
+#
+#   PR merged via "Rebase and merge" (N commits):
+#     Bumps nano by N, adds ONE commit after the last rebased commit:
+#       "Bump nano version to 12349 (+3) for PR #1789 - Fix hotplug on macOS"
+#
+#   Direct push (N commits):
+#     Bumps nano by N, adds ONE commit after the pushed commits:
+#       "Bump nano version to 12349 (+3) for Fix hotplug on macOS"
+#
+# This preserves the original semantic: every commit that lands on
+# master "consumes" one nano number, even though the actual bump is
+# recorded in a single trailing commit.
+#
+# No force-push is involved.
+#
+# Loop prevention:
+#   Commits made with the default GITHUB_TOKEN do not trigger further
+#   workflow runs (GitHub policy).  As a belt-and-suspenders measure,
+#   the job also skips if the HEAD commit message starts with
+#   "Bump nano version".
+#
+# Concurrency:
+#   Only one instance runs at a time.  If two PRs are merged in quick
+#   succession, the second run queues and sees the already-bumped nano.
+
+name: Increment version nano
+
+on:
+  push:
+    branches:
+      - master
+
+# Serialize to avoid two runs reading the same nano value.
+concurrency:
+  group: version-nano-increment
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+
+jobs:
+  increment-nano:
+    runs-on: ubuntu-latest
+
+    # Do not run on our own nano-bump commits.
+    if: >-
+      !startsWith(github.event.head_commit.message, 'Bump nano version')
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v5
+        with:
+          # GITHUB_TOKEN: commits pushed with this token do NOT
+          # trigger further workflow runs — natural loop prevention.
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+
+      - name: Configure git identity
+        run: |
+          git config user.name  "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Count commits in this push
+        id: count
+        run: |
+          BEFORE="${{ github.event.before }}"
+          AFTER="${{ github.event.after }}"
+
+          # If 'before' is all zeros, this is a new branch (shouldn't
+          # happen on master, but handle it gracefully).
+          if echo "$BEFORE" | grep -qE '^0+$'; then
+            NUM_COMMITS=1
+          else
+            NUM_COMMITS=$(git rev-list --count "$BEFORE".."$AFTER")
+          fi
+
+          # Sanity: at least 1.
+          if [ "$NUM_COMMITS" -lt 1 ]; then
+            NUM_COMMITS=1
+          fi
+
+          echo "num_commits=$NUM_COMMITS" >> "$GITHUB_OUTPUT"
+          echo "Detected $NUM_COMMITS commit(s) in this push"
+
+      - name: Determine commit message context
+        id: msg
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Use the GitHub CLI to search for a merged PR that contains
+          # the commit SHA from this push event.  The CLI is more
+          # reliable than the REST endpoint for rebase-merged commits.
+          PR_INFO=$(gh pr list \
+            --repo "${{ github.repository }}" \
+            --search "${{ github.sha }}" \
+            --state merged \
+            --json number,title \
+            --jq '.[0] // empty' \
+          ) || true
+
+          if [ -n "$PR_INFO" ]; then
+            PR_NUMBER=$(echo "$PR_INFO" | jq -r '.number')
+            PR_TITLE=$(echo "$PR_INFO" | jq -r '.title')
+            echo "source=pr" >> "$GITHUB_OUTPUT"
+            echo "pr_number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+            # Sanitize title: remove characters that could break the
+            # commit message or git commands.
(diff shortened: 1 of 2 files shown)


seanm (approved):
Amazing! This is glorious!

sonatique:
Tested it on my fork, seems to work

mcuee (approved):

mcuee:
Nice one!

tormodvolden:
That's a very interesting take on it, outside the git hook box.

Your branch shows that it bumps the wrong way though:
[‎libusb/version_nano.h‎](https://github.com/sonatique/libusb/commit/b718aa0f763027a5f7d9a2b907f9dfbb7dde7bf4#diff-ea2cd4e7cb105b42808d104d44b1c5fad20981859094f60830bdba47467fc823)
```
@@ -1 +1 @@
-#define LIBUSB_NANO 12026
+#define LIBUSB_NANO 12025

```

sonatique:
> That's a very interesting take on it, outside the git hook box.
> 
> Your branch shows that it bumps the wrong way though: [‎libusb/version_nano.h‎](https://github.com/sonatique/libusb/commit/b718aa0f763027a5f7d9a2b907f9dfbb7dde7bf4#diff-ea2cd4e7cb105b42808d104d44b1c5fad20981859094f60830bdba47467fc823)
> 
> ```
> @@ -1 +1 @@
> -#define LIBUSB_NANO 12026
> +#define LIBUSB_NANO 12025
> ```

Hello @tormodvolden, 
The commit you're referring to was the merge of the CI into my branch. The nano version change was (wrongly) done locally. My local repos is / was a little bit of a mess because of the many origins I have, I am struggling with conflicting "master" branch... 

Anyway, the commit made by the CI is this one: https://github.com/sonatique/libusb/commit/3f441a192f6d2fd3a01457f7bc9bf10fb2f9a7dc 
it is the result of the CI action running after I merge a test PR, and I think it works:
Commit message is "Bump nano version to 12027 for PR https://github.com/sonatique/libusb/pull/3 - Test nano ci"
And the nano version increment by +1

I am still testing though. I will provide more clean evidences.

EDIT : 
here is the resulting CI Action commit for a commit that I directly pushed to my fork (without PR):
https://github.com/sonatique/libusb/commit/6580a2b10564d78a5f8749084ea00b698bfa6951
The increment is also correct.

EDIT:
This is the CI action commit after merging a PR with 2 commits:
https://github.com/sonatique/libusb/commit/60ef167047ba7546bcc1c230440a251b187cd203
I basically work but the increment is only 1, where it should be 2 if we want to keep the rule of the commit count.
I will fix this.
Also: I noticed a deprecation warning for "Node.js 20 actions are deprecated and this may not work as expected: actions/checkout@v4", but the fix is simple, I just need to use checkout@v5.

I'll be back soon with an fixed version.

Youw:
An oppinion - this is ugly. Yes, this PR would probablty work fine, but...
I've seen a project or two where similar aproach was used - each commit was followed by a version auto-increment commit, and git history for that looks awful.

Of course, force-pushes to master is our of the question, but I have at least 2 other alternatives:
1) A Github action that would be triggered on each commit/push to master, go to every active PR on github and push a commit with VERSION_NANO update as <version nano from master + 1>. That job will take like a few seconds to complete, and each and every PR will be ready to merge with VERSION_NANO already in place.
Of course this would require using only Squash+Merge, so these (potentially multiple) commits wouldn't end up in master git history.

2) We don't create a PR agains master, but rather against special staging branch. Once a PR is merged into that branch - a special job would force-push latest commit of that branch to update VERSION_NANO, and then rebase-merge that branch into master. This way:
- no manual handling of VERSION_NANO
- no force-pushes to master

---
Thoughts?

sonatique:
@mcuee @seanm @tormodvolden : I forced push an improved versions based on my tests. Fixes are:
- increment by N for PR with N commits (and still by 1 for single commit)
- correctly handle cases where a new PR / commit is merged before CI action for last PR / commit had opportunity to run (due to CI queue congestion). We end up with 2 actions running one after the other. This creates 2 sequential "nano bump" commits but the increment count is correct and message correctly points to the previous PR / commit.
- use non deprecated checkout action v5

As far as I can tell, there are no other issues.

Example of action commit for single commit: https://github.com/sonatique/libusb/commit/a4ce24f2fbb2f5aaaecb313855bdbc755fdb02a8

Example of action commit for a PR with 1 single commit, made after another 3-commit PR has been merged: https://github.com/sonatique/libusb/commit/17225db38ca6684213b2fb4671b739382bcd0c09

Example of action commit for a PR with 3 commits, made after previous action commit was made:
https://github.com/sonatique/libusb/commit/50490bd4748bacd8c5e4c914afe31e51b85e3fa6

In all cases the nano bump commit has the correct comment reference the PR or commit.

sonatique:
Hello @Youw 

Thank you for your comment. I see your point. 

We have one "Glorious" and one "Ugly" in the same PR, quite a stretch! ;-)

I won't say that the resulting git history would look beautiful, but I would tend to favor "just works" over beauty in this case.

Your solution certainly has benefits over mine, but I see one annoying point: it pushes some hassle on the contributor side in replacement for having the maintainers having to manage the nano version: contributors would have to deal with having their local commits no longer matching the remote ones. Might not be a big issue if the PR is merged as is, but this means at minimum doing a `pull --rebase` or something similar before updating the PR following comments. 
This literally means: every single contributors will potentially have to do this for all of his PR every time another PR is merged… 

I cannot easily estimate in practice how much additional work and annoyances this will cause, maybe only a little actually, but in theory this goes against the goal we're trying to reach here which is lowering the level of annoyance for contributors and maintainers...

Youw:
Then option 2 should handle all of it...

sonatique:
> Then option 2 should handle all of it...

Hum, maybe I didn't understand all of your point 2, but given the staging branch will always be rebased onto master after each PR merge, I think this "invalidate" all other PR branches because they will be on a "dangling branch", no?

I think I need to think more about it...

Are the example of other project that requires people to do their PR against a non-default branch?

mcuee:
BTW, I am still in favor of removing the `VERSION_NANO` completely, same as @seanm.

In any case, this PR can be in 1.0.31 and no need to be 1.0.30.

Youw:
> but given the staging branch will always be rebased onto master after each PR merge, I think this "invalidate" all other PR branches because they will be on a "dangling branch", no?

no more than pushing to master in a regular case

as for making people creating PR to staging branch - I personally haven't seen it, but it isn't a problem - even if PR is created against master - PR itself can be edited by the maintainer and targeting change can be changed with no issues (even right before the merge)

and both staging branch and master will be identical most of the time (except this short period after a PR is merged and before CI job force pushes to staging and then merging all of that to master)

---
But anyway, if we really all agree that removing VERSION_NANO is is a way to go - maybe it worth going that direction right away and not introduce any other workarounds in a first place.

Maybe right after 1.30 is released.

Maybe mark `VERSION_NANO` as deprecated using a preprocessor command in 1.30?

sonatique:
Hello @Youw I had time to think into more details about your proposal #2. Here is my opinion:

- It does not eliminate force-pushes, it just moves them:
The staging branch still gets force-pushed after each merge. From the perspective of any contributor with an open PR targeting staging, their branch is silently invalidated in exactly the same way as a force-push to master would be. The improvement is cosmetic.

- It actually creates a double rewrite per merged PR for pending contributors:
Consider what happens to a contributor "C" with an open PR against staging when another PR gets merged:
1. staging gets force-pushed to amend the tip with the nano bump. The tip SHA changes. C's branch is now based on a commit that no longer exists in staging's history.
2. staging then gets rebase-merged into master, producing brand new SHAs for every commit. staging is then reset to match master.

So C's branch goes through two base SHA rewrites per cycle, not one. Your argument that this is "no more than pushing to master in a regular case" holds for the rebase-merge step in isolation, but paired with the preceding force-push it becomes strictly worse than a normal master advance.

- There is a real race condition:
If two PRs are merged into staging in quick succession, two CI jobs run concurrently: both read the current nano value, both try to force-push staging, and at least one overwrites the other's work. My current PR handles this case gracefully: two sequential nano bump commits, both correct, no conflict.

- The contributor workflow change is non-trivial:
Asking contributors to target staging instead of master my work, but since we cannot enforcing it (no mechanism for that on GitHub PRs) maintainers will have to redirect every wrong PR before merging. That is a recurring manual step on (I guess) a lot of PR, which cuts against the goal of reducing friction.

- The staging branch infrastructure is heavyweight for what it buys:
To avoid nano bump commits in git history, option 2 needs: a staging branch, a force-push CI job, a rebase-merge CI job, and a soft contributor convention. That is a significant amount of moving parts for what is ultimately a history aesthetics concern. One extra commit per merged PR with a clear message is arguably informative rather than noisy, and can be filtered out of `git log` (with `--invert-grep --grep="^Bump nano"`) for anyone who prefers not to see them.

I agree the resulting history is not the prettiest, but the tradeoff seems clearly in favor of the simpler approach. If we end up removing nano version altogether after 1.0.30, this whole discussion is moot anyway.

So in my opinion: either remove nano altogether or user this PR, neither of this is perfect unfortunately.

Youw:
You're right about some of the points, but not about all of them (at least not in the way I see it).
But I don't think any of that matters - looks like we're going more towards abandoning the NANO version all along, and I'd rather suggest:
- abandone this change (just for the sake of keeping git history a bit cleaner if we're to abandone the VERSION_NANO)
- follow this plan: https://github.com/libusb/libusb/issues/1727#issuecomment-4308485528

sonatique:
Closed in favor of https://github.com/libusb/libusb/pull/1812

PR #1795, opened 2026-04-08
Address issue: race in libusb get device list and libusb free device list with win usb backend + claim_interface race

sonatique:
windows hotplug: fix device leak on detach of uninitialized devices:

When a physically removed device was never successfully initialized,
windows_refresh_device_list correctly calls usbi_detach_device (not
usbi_disconnect_device) since the device was never announced to users.

However, usbi_detach_device only removes the device from the list —
unlike usbi_disconnect_device, it does not post a DEVICE_LEFT hotplug
message. This means no message handler will ever drop the initial
allocation reference (refcnt=1 from usbi_alloc_device), so the device
struct is leaked.

Add the missing libusb_unref_device call after usbi_detach_device.
This is safe because:
 - The device was never initialized, so no user handle can exist.
 - The device has been removed from ctx->usb_devs, so it is no longer
   discoverable via libusb_get_device_list.
 - If a concurrent libusb_get_device_list snapshot holds a ref, unref
   drops refcnt to 1 (not 0), and the user's later
   libusb_free_device_list completes the cleanup.
   
windows: fix claim_interface race on shared device state:

libusb_claim_interface acquires dev_handle->lock before calling the
backend, but this lock is per-handle — created fresh during
libusb_open. When the same device is opened twice, each handle gets
its own independent mutex. Concurrent claim_interface calls on
different handles to the same device race on the shared
priv->usb_interface[] fields (endpoint, nb_endpoints), causing
double-free and use-after-free of the endpoint array.

The problem is worse for libusb_set_interface_alt_setting, which
releases dev_handle->lock *before* calling the backend (core.c:1921),
so winusb_set_interface_altsetting runs with no lock at all — even
a single-device single-handle scenario is unprotected against a
concurrent call from another thread using a different handle.

Add a per-device interface_lock mutex to struct winusb_device_priv
that serializes all modifications to usb_interface[] from
winusb_claim_interface and winusb_set_interface_altsetting, regardless
of which handle initiates the operation.

No lock is needed in the enumeration path (set_composite_interface,
set_hid_interface) because those run during winusb_get_device_list
under active_contexts_lock, before any handle can exist for the
device.

core: fix two race conditions in non-hotplug device list management

Fix two races between concurrent libusb_get_device_list and
libusb_free_device_list calls on non-hotplug backends (Windows
without hotplug, OpenBSD, NetBSD, Solaris, Emscripten, UsbDk).

Race 1 — half-initialized device visible in list:
[...]

PR #1809, opened 2026-04-22
tests: Add tests/testcore.c, a multi-threaded stress test that hammers

### LINKED ISSUE #1808: Include testcore in libusb tests
From @sonatique 
* https://github.com/libusb/libusb/pull/1795#issuecomment-4289750303

> I remove test program commit from this PR as per @mcuee opinion.
> It is still available here: https://github.com/libusb/libusb/commit/a1a5ca371be6acd3bf597c69b98c9bfb825ac183

Even though I recommend it to be left out of 1.0.30, it may be a good idea to have it in 1.0.31

sonatique:
libusb_get_device_list / libusb_free_device_list from N threads in parallel. It exists primarily to reproduce the concurrent-enumeration crashes reported in #1793 and the related set_composite_interface race surfaced during PR #1795 review.
The stressed code paths include:
 - usbi_get_device_by_session_id()
 - usbi_alloc_device()
 - usbi_connect_device()
 - libusb_unref_device() / ctx->usb_devs_lock
 - winusb_device_priv setup races in set_composite_interface and set_hid_interface (Windows, non-hotplug builds) The original C++20 reproducer was posted by smarvonohr and the C version with Win32 threads by mcuee in
https://github.com/libusb/libusb/pull/1795#issuecomment-4258288585 This version is adapted from mcuee's C version, with two changes that make it portable and easier to build:
 - Use the same PLATFORM_POSIX / PLATFORM_WINDOWS thread abstraction that stress_mt.c uses (pthread_create on POSIX, _beginthreadex on Windows, CreateThread on Cygwin), instead of bare pthread. The original would not build with MSVC.
 - Print the device list once before starting the worker threads, so the operator can confirm at a glance which devices are present. This is useful because the bug only manifests when enumeration touches certain device shapes (e.g. composite devices with HID children that exercise set_composite_interface). Add msvc/testcore.vcxproj following the same template as msvc/stress_mt.vcxproj so the test builds out of the box in Visual Studio against libusb_static.
Notes for runners:
 - The test only triggers the enumeration races in non-HOTPLUG builds. With Windows hotplug enabled, libusb_get_device_list is served from a cache and winusb_get_device_list never runs concurrently, so the bug does not manifest.
 - The default LOOPS = 100000 is intentionally large for the original bug-hunting use case but takes hours of real work even when the library is correct (each non-hotplug get_device_list does a full Windows USB enumeration, ~10-100 ms). Reduce LOOPS for a quick smoke test.
 - At least one composite USB device with HID-class child interfaces (USB game controller, headset with controls, Logitech Unifying receiver, etc.) must be present to exercise set_composite_interface. Pure UVC/UAC composite devices like webcams will NOT trigger that path because their child interfaces are not HID class and are not bound to any libusb-supported driver.
 
 Closes https://github.com/libusb/libusb/issues/1808

### DIFF
--- a/tests/testcore.c
+++ b/tests/testcore.c
@@ -1,0 +1,170 @@
+/*
+ * libusb concurrent get/free_device_list reproducer for issue #1793
+ *
+ * Repeatedly enumerates and frees the device list from multiple threads
+ * to stress concurrent access to:
+ *   - usbi_get_device_by_session_id()
+ *   - usbi_alloc_device()
+ *   - usbi_connect_device()
+ *   - libusb_unref_device()
+ *   - ctx->usb_devs_lock correctness
+ *   - winusb_device_priv setup race in set_composite_interface (Windows)
+ *
+ * Originally posted by mcuee in
+ * https://github.com/libusb/libusb/pull/1795#issuecomment-4258288585
+ * Adapted here to use the platform thread abstraction from stress_mt.c
+ * so it builds with MSVC.
+ */
+
+#include <config.h>
+
+#include <libusb.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#if defined(PLATFORM_POSIX)
+
+#include <pthread.h>
+#include <sched.h>
+typedef pthread_t thread_t;
+typedef void * thread_return_t;
+#define THREAD_RETURN_VALUE NULL
+#define THREAD_CALL_TYPE
+
+static inline int thread_create(thread_t *thread,
+	thread_return_t (*thread_entry)(void *arg), void *arg)
+{
+	return pthread_create(thread, NULL, thread_entry, arg) == 0 ? 0 : -1;
+}
+
+static inline void thread_join(thread_t thread)
+{
+	(void)pthread_join(thread, NULL);
+}
+
+static inline void thread_yield(void)
+{
+	sched_yield();
+}
+
+#elif defined(PLATFORM_WINDOWS)
+
+typedef HANDLE thread_t;
+#define THREAD_RETURN_VALUE 0
+#define THREAD_CALL_TYPE __stdcall
+
+#if defined(__CYGWIN__)
+typedef DWORD thread_return_t;
+#else
+#include <process.h>
+typedef unsigned thread_return_t;
+#endif
+
+static inline int thread_create(thread_t *thread,
+	thread_return_t (__stdcall *thread_entry)(void *arg), void *arg)
+{
+#if defined(__CYGWIN__)
+	*thread = CreateThread(NULL, 0, thread_entry, arg, 0, NULL);
+#else
+	*thread = (HANDLE)_beginthreadex(NULL, 0, thread_entry, arg, 0, NULL);
+#endif
+	return *thread != NULL ? 0 : -1;
+}
+
+static inline void thread_join(thread_t thread)
+{
+	(void)WaitForSingleObject(thread, INFINITE);
+	(void)CloseHandle(thread);
+}
+
+static inline void thread_yield(void)
+{
+	Sleep(0);
+}
+
+#endif /* PLATFORM_WINDOWS */
+
+#define THREADS 12
+#define LOOPS   100000
+
+struct worker_arg {
+	libusb_context *ctx;
+	int index;
+};
+
+static thread_return_t THREAD_CALL_TYPE worker(void *arg)
+{
+	struct worker_arg *wa = (struct worker_arg *)arg;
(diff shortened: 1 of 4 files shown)


mcuee:
Two quick comments.
1) Need to add the project to VS solution file. 
2) Need to add the test to auto-tools build as well.

mcuee:
For example, under macOS, without adding the new test to auto-tools build, it will not build testcore by default.

Manual build:
```
mcuee in ~/build/libusb/libusb_pr1809/tests on testcore-program-from-PR-1795 % clang -pthread -g -I /Users/mcuee/mybin/bingit_debug/include/libusb-1.0 -I.. testcore.c -L /Users/mcuee/mybin/bingit_debug/lib -lusb-1.0 -o testcore_debug

mcuee in ~/build/libusb/libusb_pr1809/tests on testcore-program-from-PR-1795 % ./testcore_debug
Device 0: 03eb:2106 class=0xff
Device 1: 04b4:8613 class=0xff
Device 2: 05e3:0747 class=0x00
Device 3: 0bda:8156 class=0x00
Device 4: 05e3:0610 class=0x09
Device 5: 05e3:0625 class=0x09
Device 6: 05ac:8015 class=0x09
Starting 12 threads, 100000 loops each...
Completed without crash.
```

tests/testcore.c
```diff
@@ -0,0 +1,170 @@
...
+	struct worker_arg args[THREADS];
+	int i;
+
+	if (libusb_init_context(&ctx, /*options=*/NULL, /*num_options=*/0) != 0) {
+		fprintf(stderr, "libusb_init_context failed\n");
+		return EXIT_FAILURE;
+	}
+
+	//libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, LIBUSB_LOG_LEVEL_DEBUG);
+
```
mcuee, line 199: Maybe can add the following option to test Linux non-hotplug situation. Just as a comment, no need to be too complicated.
    // This forces libusb to skip the automatic background discovery
    // and only scan when you explicitly call get_device_list.
   //  libusb_set_option(ctx, LIBUSB_OPTION_NO_DEVICE_DISCOVERY);

tests/testcore.c
```diff
@@ -0,0 +1,170 @@
...
+static inline void thread_yield(void)
+{
+	Sleep(0);
+}
+
+#endif /* PLATFORM_WINDOWS */
+
+#define THREADS 12
+#define LOOPS   100000
+
```
mcuee, line 92: Since this will be a generic test program, it is good to add these two as argument. It is also good to print the help for the usage.

The default value can be 4 and 100000. Or shall we use 4 and 10000?

mcuee:
> The default LOOPS = 100000 is intentionally large for the original bug-hunting use case but takes hours of real work even when the library is correct (each non-hotplug get_device_list does a full Windows USB enumeration, ~10-100 ms). Reduce LOOPS for a quick smoke test.

It would be good to make number of threads and number of loops to be arguments to the test program. If we choose the default loops to be 100,000, we may want to inform the user the above info.

tests/testcore.c
```diff
@@ -0,0 +1,170 @@
...
+static inline void thread_yield(void)
+{
+	Sleep(0);
+}
+
+#endif /* PLATFORM_WINDOWS */
+
+#define THREADS 12
+#define LOOPS   100000
```
seanm: I know this is just test code, but it would be nice to minimise use of #define for numerical constants as we write new code. Instead:

static const int THREADS = 12;

This allows better type checking.

mcuee:
Looks good to me with the latest changes.

sonatique:
> Looks good to me with the latest changes.

Would you be OK to approve the PR? I think merging it won't harm, no?

mcuee (approved):

mcuee:
Okay from my side. Approved.

Youw:
There is 1 downside to this development - it adds ~1.5-2 min to a single CI run. Maybe it is not much as an absolute value, but relatively that is 5m42s -> 9m23s (2 build runs) for one of the jobs, and 2m07 -> s3m01s for a different one, which is ~30% increase in CI run time.

Need to consider what to do with it. Either leave it as is or try to optimize something, e.g. do not run these tests(s) on each PR, but leave it for master pushes, etc.

@mcuee, @sonatique, @seanm

mcuee:
I think we can keep this only for git commit to the master.

seanm:
I'm accustomed to CI taking hours, so that increase doesn't bother me.

PR #1812, opened 2026-04-24
core: populate libusb_version::describe from git describe

### LINKED ISSUE #1727: RFC: abandon version_nano.h in favour of github "Merge pull request" feature

Youw:
Closes: #1727

Replaces the hardcoded `"https://libusb.info"` string in `libusb_version::describe` with a build-time identifier generated from `git describe --tags --always --dirty`.

This PR only **adds** the `describe` identifier — `LIBUSB_NANO` stays exactly as master maintains it. Freezing/retiring `LIBUSB_NANO` is a separate decision for later (see #1727); once developers agree to it, contributors will no longer need to touch any version file and the normal GitHub "Merge pull request" button becomes usable.

## Why `describe`?

`struct libusb_version` already has a `describe` field, currently hardcoded to `"https://libusb.info"` with a misleading "ABI compatibility only" comment. A GitHub code search shows it is in fact read by several downstream consumers — a 4D plugin, a Lua binding (`luausb`), `libheadunit`, `IDTechReader`, a bladeRF tutorial — always as an opaque display string piped into logs, tooltips or scripting-binding fields. Nobody parses it; nobody depends on its current value.

Repurposing it to hold the git-describe output is zero-ABI-break and strictly more useful: every existing call site starts showing a precise commit identifier (with a dirty-tree marker for free), which is exactly the "unique-per-commit" property `LIBUSB_NANO` provides today — but human-readable and recoverable with `git checkout`.

## Behavior API/ABI change

There is no change to API or ABI produced by this change. The only visible runtime changes:
- one libusb log line now includes the content of `libusb_version::describe`;
- the content of `libusb_version::describe` is no longer the hardcoded `"https://libusb.info"` string but the output of `git describe ...`;
- the version string embedded into the Windows RC / .dll attributes is updated to the same `git describe` identifier.

`LIBUSB_NANO`, the `nano` struct field, and the maintainer pre-commit hook that bumps it are all left untouched.

### Core changes

- **`build-aux/gen-describe.sh`** — new POSIX shell script that writes `libusb/version_describe.h` containing a `LIBUSB_DESCRIBE` macro. Resolution order:
  1. `.git` directory present → `git describe --tags --always --dirty` (with `safe.directory='*'` so containerised / cross-uid checkouts work). Hard-fails with an explicit error if `git` is missing.
  2. `.tarball-version` present (frozen by `make dist`) → use it verbatim.
  3. Neither → emit an `Unknown source (v1.0.31...)`-style string and warn. The `Unknown source` prefix is intentionally not a valid git revision so it can be told apart from a real identifier.

  The script is the single source of truth for the classification; it also has a `--classify` mode (`git`/`tarball`/`unknown`) that `configure.ac` calls to fail fast at configure time with a clear error message.

- **`build-aux/gen-describe.ps1`** — Windows companion: a PowerShell launcher that locates a matching `bash.exe` + `cygpath.exe` pair from the same Git for Windows / MSYS2 install (skipping the WSL `bash.exe` stub in `System32`), converts the paths to MSYS form via `cygpath`, and delegates to `gen-describe.sh`. No duplication of the classification logic. Wraps the bash invocation in a named system mutex (`Global\libusb-gen-describe`) so concurrent pre-build invocations from msbuild's `/m` parallel scheduling serialize on the file write instead of racing.

- **`libusb/core.c`** — includes `version_describe.h`, populates `libusb_version_internal.describe = LIBUSB_DESCRIBE`, and includes the describe string in the init-time debug log.

- **`libusb/libusb.h`** — re-words the `describe` field comment (previously "For ABI compatibility only.") with a documented contract: when the string does not start with `Unknown` and does not end with `-dirty`, it is a valid revision identifier accepted by `git rev-parse` / `git checkout` against the libusb repository, allowing the exact source tree to be recovered. The `nano` field comment is unchanged.

- **`libusb/libusb-1.0.rc`** — Windows `FileVersion` / `ProductVersion` resource strings now embed `LIBUSB_DESCRIBE` (the full git-describe identifier). The numeric `FILEVERSION` / `PRODUCTVERSION` 4-tuples still use `LIBUSB_NANO` since those slots demand integers.

`libusb/version_nano.h` and `HACKING` are intentionally **not** modified by this PR.

### Build-system wiring

`version_describe.h` is generated by every supported build system before `core.c` is compiled — there is no `__has_include` fallback:

- **Autotools**: `configure.ac` runs `gen-describe.sh --classify` early, `AC_MSG_ERROR`s if `.git` exists but `git` is missing, and `AC_MSG_WARN`s on the unknown path. `libusb/Makefile.am` regenerates the header on every build via a `.PHONY` rule (writing to the build tree, not the source tree) and adds it to `BUILT_SOURCES` / `CLEANFILES`. `dist-hook` writes `.tarball-version` into the tarball using `git describe`, so tarball users build without git.
- **MSVC**: `libusb_dll.vcxproj` and `libusb_static.vcxproj` carry a pre-build event invoking `gen-describe.ps1`. The launcher serializes concurrent invocations with the named mutex, so `msbuild /m` parallel builds work cleanly. Standalone use (opening `msvc/libusb.sln` in Visual Studio, or running `msbuild` directly) generates the header automatically — no separate setup step.
- **Xcode**: `libusb.xcodeproj` gets a shell-script build phase running `gen-describe.sh`, ordered before Headers/Sources.
- **Android**: `android/jni/libusb.mk` runs the script via `$(shell)` at parse time and `$(error)`s on failure.

The `.private/ci-container-build.sh` script (umockdev test container) installs `git` so the build inside the container can run `gen-describe.sh`.

### CI

Every workflow (`linux`, `macos`, `windows`, `android`, `msys2`, `msys2_clang64`, `msys2_clang64-aarch64`) checks out with `fetch-depth: 0` so `git describe --tags` produces the intended tag-prefixed `v1.0.30-NN-gabc1234` form rather than a bare hash.

## Test plan

Verified end-to-end on Linux/Autotools:

- [x] git checkout w/ git → describe baked into `.so` (`strings` confirms, `https://libusb.info` gone)
- [x] git checkout w/o git → configure aborts with an explicit "git is not on PATH" error
- [x] `make dist` writes `.tarball-version` into the tarball
- [x] Extracted tarball + no git → reads `.tarball-version`, builds cleanly, baked string in `.so`
- [x] No `.git` and no `.tarball-version` → configure warns + an `Unknown source (...)` string is baked into `.so`
- [x] Incremental rebuild does not rewrite `version_describe.h` when the describe string is unchanged
- [x] CI: macOS, Android, Linux (incl. umockdev container test), all MSYS2, full Windows MSVC matrix all green
- [x] Standalone MSVC build from a fresh clone (issue #1815) — bash/cygpath pairing fix verified


seanm:
This lgtm generally, though I don't know the language/syntax of various changed files.

I'll checkout and try the Xcode change in the coming days...

seanm:
> I'll checkout and try the Xcode change in the coming days...

I've tried it, and it indeed builds the file as expected:

```
% cat ./libusb/version_describe.h
/* Auto-generated by build-aux/gen-describe.sh. Do not edit. */
#ifndef LIBUSB_DESCRIBE
#define LIBUSB_DESCRIBE "v1.0.30-rc1-22-gde23a690"
#endif
```

seanm (approved):

tormodvolden:
I welcome this change. But I hope it is clear for everybody that this has to wait until after next release. So there is no rush.

mcuee:
Yes, I put the milestone as 1.0.31 and not 1.0.30.

Youw:
@tormodvolden do you think this is a big risk to include into 1.0.30? Or to avoid additional testing, follow planning, etc.?
What I'm trying to imply, that this is not yet replacing anything, only gives some functionality *in addition* to what we have, and we could still keep the existing VERSION_NANO (I'll remove the deprecation NOTE, etc.) as is, but have the additional `describe` populated and available.

Youw:
Unless, of course, 1.0.30 is ready to be released now (which I'm not aware of, simply because don't closely follow), in that case I agree that this is no good reason for a potential delay.

tormodvolden:
Hi Youw, I haven't looked too closely at the details here, so I am sorry if I missed something, but I saw "freeze LIBUSB_NANO" and that would not be possible before the release. An even if changing "describe" doesn't break the API, it is nice if we can wait and only warn about all these planned changes in the release announcement. And yes, 1.0.30 will be released soon. We will prepare an RC2 now and hopefully not need any changes before final release.

Youw:
> but I saw "freeze LIBUSB_NANO" and that would not be possible before the release

Yeah, that is purely a documentary change, can be delayed for as long as we're comfortable to keep it.

> hopefully not need any changes before final release

I totally agree with that.

---
I've also added explicit NOTE in the PR description about API/ABI change.

Youw:
@SteveRussell33 #1815 should now be fixed, please check it when you can

seanm:
> @SteveRussell33 #1815 should not be fixed, please check it when you can

Should "now" be fixed? :)

mcuee (approved):

sonatique (approved):

seanm:
> What I'm trying to imply, that this is not yet replacing anything, only gives some functionality in addition to what we have, and we could still keep the existing VERSION_NANO (I'll remove the deprecation NOTE, etc.) as is, but have the additional describe populated and available.

This would be fine with me.

But I'm also fine with waiting *if* an actual rc2 is released within say a week.

seanm:
@Youw can you fix the merge conflicts please?

We now have 3 approvals, and 1.0.30 is out. Would be nice to get this merged ASAP.

Youw:
Updated, no conflict.
Cleaned up wording, remove any mention regarding VERSION_NANO deprecation, or so. For now - that is purely an additional mechanism, so we can check/test it independently. Once we're ready - deprecation of VERSION_NANO, and full switch to `describe`, etc.

mcuee:
@tormodvolden @sonatique and @seanm 

I am in favour to merge this now. Any objections?

seanm:
@mcuee there are 3 approvals, so yes I think we should merge it ASAP.

seanm:
Amazing!  So now we can just do regular github PR merges?

mcuee:
> Amazing!  So now we can just do regular github PR merges?

I believe that is the case.

seanm:
> I believe that is the case.

I will try with https://github.com/libusb/libusb/pull/1738, which has 3 approvals.

Youw:
> So now we can just do regular github PR merges?

I believe not just yet. Latest agreement is to accept the new "git describe" mechanism into build and just have it along with VERSION_NANO. Next step is to explicitly deprecate VERSION_NANO (which hasn't been done by this PR as of yet), and only then we can safely abandone it and start using Github Merge PR button.

I'll open a second PR (UPD: #1839), lets continue the discussion there.

PR #1814, opened 2026-04-25
Fix/descriptor parser memory safety

### LINKED ISSUE #1813: Two memory-safety bugs in libusb/descriptor.c found by fuzzing (parse_interface NULL endpoint, parse_iad_array OOB read)
Hi, I ran a libFuzzer + ASAN/UBSan campaign against the descriptor parsers in `libusb/descriptor.c` (HEAD 77503a79) and found two issues. Both are reachable from a malformed config descriptor returned by an attached USB device, with no application-side input required.

Filing them together because they're in the same file and the same fuzz harness rediscovers both. Same general shape as #1039, but in the cross-platform parser rather than the Windows backend.

Tested on macOS arm64 with Homebrew clang 22.1.1; the parser code is platform-independent so this hits all backends.

---

## Bug A: `parse_interface` leaves `endpoint = NULL` with `bNumEndpoints > 0`

**File:** `libusb/descriptor.c`, around line 245.

In the inner "skip extra/class descriptors" loop:

```c
} else if (header->bLength > size) {
    usbi_warn(ctx, "short extra intf desc read %d/%u",
              size, header->bLength);
    return parsed;          // <-- bug
}
```

By the time this returns:

- `usb_interface->num_altsetting++` has already happened (line 221)
- `ifp->bNumEndpoints` was filled from `buffer[4]` (line 197)
- `ifp->endpoint` is still `NULL` (the `calloc` near line 284 never runs)

So a caller of `libusb_get_active_config_descriptor` / `libusb_get_config_descriptor` gets back a struct with `bNumEndpoints > 0` and `endpoint == NULL`. `libusb.h` documents `endpoint` as:

> Array of endpoint descriptors. This length of this array is determined by the bNumEndpoints field.

`examples/testlibusb.c` and `examples/xusb.c` follow this contract and would crash here.

### Reproducer (20 bytes of descriptor data)

```
09 02 19 00 01 01 00 a0 00     // config:    bNumInterfaces=1
09 04 00 00 01 ff 00 00 00     // interface: bNumEndpoints=1
07 05                          // truncated extra descriptor (claims bLength=7)
```

### Verifier (uses libusb's own descriptor.c source, no copy)

`verify.c` (also attached):

```c
#define ENABLE_LOGGING 0
#include "config.h"
#include "libusbi.h"
#include "descriptor.c"

#include <stdio.h>
#include <stdlib.h>

const struct usbi_os_backend usbi_backend = {0};
void usbi_log(struct libusb_context *c, enum libusb_log_level l,
              const char *f, const char *m, ...) { (void)c;(void)l;(void)f;(void)m; }
int libusb_control_transfer(libusb_device_handle *h, uint8_t a, uint8_t b,
                            uint16_t c, uint16_t d, unsigned char *e,
                            uint16_t f, unsigned int g) {
    (void)h;(void)a;(void)b;(void)c;(void)d;(void)e;(void)f;(void)g; return -99;
}

int main(void) {
    uint8_t in[20] = {
        0x09,0x02,0x19,0,0x01,0x01,0,0xa0,0,
        0x09,0x04,0,0,0x01,0xff,0,0,0,
        0x07,0x05
    };
    struct libusb_config_descriptor cfg = {0};
    parse_configuration(NULL, &cfg, in, sizeof(in));
    const struct libusb_interface_descriptor *alt =
        &cfg.interface[0].altsetting[0];
    printf("bNumEndpoints=%u  endpoint=%p\n",
           alt->bNumEndpoints, (void*)alt->endpoint);
    return 0;
}
```

Build and run from the repo root:

```sh
$ ./autogen.sh && make -C libusb
$ clang -O0 -g -fsanitize=address,undefined \
        -I . -I libusb -I libusb/os -DPLATFORM_POSIX=1 \
        -Wno-macro-redefined verify.c -o verify
$ ./verify
bNumEndpoints=1  endpoint=0x0
```

A real application's endpoint loop after this would SEGV.

### Suggested fix

```diff
 } else if (header->bLength > size) {
     usbi_warn(ctx, "short extra intf desc read %d/%u",
               size, header->bLength);
+    ifp->bNumEndpoints = 0;
     return parsed;
 }
```

Just keep the invariant: `bNumEndpoints > 0` implies `endpoint != NULL`.

---

## Bug B: `parse_iad_array` reads one byte past the buffer

**File:** `libusb/descriptor.c`, around line 1381.

```c
while (consumed < size) {
    header.bLength = buf[0];
    header.bDescriptorType = buf[1];
    if (header.bLength < DESC_HEADER_LENGTH) return LIBUSB_ERROR_IO;
    else if (header.bLength > size)               // <-- should be size - consumed
        return LIBUSB_ERROR_IO;
    ...
    buf += header.bLength;
    consumed += header.bLength;
}
```

The bound check uses the original `size` instead of remaining (`size - consumed`). With a single descriptor whose `bLength = size - 1`, `consumed` becomes `size - 1`; the loop re-enters and `buf[1]` is one byte past the malloc allocation.

`parse_configuration` and `parse_interface` in the same file already do this correctly (they decrement `size` and check `bLength > size`). Looks like `parse_iad_array` was written without the same convention.

### Reproducer (9 bytes)

```
08 00 00 00 00 00 00 00 ff
```

### ASAN output (driving `parse_iad_array` from the same verify harness)

```
==ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 1 at 0x... (0 bytes after 9-byte region)
    #0 parse_iad_array descriptor.c:1381
```

Reachable via `libusb_get_active_interface_association_descriptors` / `libusb_get_interface_association_descriptors`. 1 byte OOB read, so usually silent on stock allocators but a DoS under hardened ones (GuardMalloc, ASAN deployments, etc.).

### Suggested fix

```diff
-        else if (header.bLength > size) {
+        else if (header.bLength > size - consumed) {
             usbi_warn(ctx, "short config descriptor read %d/%u",
-                      size, header.bLength);
+                      size - consumed, header.bLength);
             return LIBUSB_ERROR_IO;
         }
```

---

Both fixes are 1 line each. I have a PR ready that includes both fixes plus a new `tests/fuzz/fuzz_ss_ep_companion.c` harness for OSS-Fuzz coverage (the same harness that surfaces both bugs). Happy to open it whenever you've had a chance to confirm.

----
The attachments:

[fuzz_ss_ep_companion.c](https://github.com/user-attachments/files/27081613/fuzz_ss_ep_companion.c)
[verify.c](https://github.com/user-attachments/files/27081614/verify.c)

1. **`verify.c`** ：the unity-include verifier referenced in the body. Reproduces both bugs against libusb's actual `descriptor.c` source.
2. **`fuzz_ss_ep_companion.c`**： the single libFuzzer harness that surfaces both bugs (Bug A in ~1 second; Bug B from a 9-byte minimized input on the same corpus).

MarkLee131:
Fix https://github.com/libusb/libusb/issues/1813


libusb/descriptor.c
```diff
@@ -246,6 +246,10 @@ static int parse_interface(libusb_context *ctx,
 				usbi_warn(ctx,
 					  "short extra intf desc read %d/%u",
 					  size, header->bLength);
+				/* Keep the invariant: bNumEndpoints > 0 implies
+				 * endpoint != NULL. The endpoint array isn't
+				 * allocated yet on this early return. */
+				ifp->bNumEndpoints = 0;
```
Youw, line 252: so this should be applied to other early returns we have above, isn't it?

tests/fuzz/fuzz_descriptor_parsers.c
```diff
@@ -0,0 +1,81 @@
...
+                            uint16_t wValue, uint16_t wIndex,
+                            unsigned char *data, uint16_t wLength,
+                            unsigned int timeout) {
+    (void)dev_handle; (void)bmRequestType; (void)bRequest; (void)wValue;
+    (void)wIndex; (void)data; (void)wLength; (void)timeout;
+    return LIBUSB_ERROR_NOT_SUPPORTED;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (size < LIBUSB_DT_CONFIG_SIZE || size > 8192)
```
seanm, line 48: What is 8192?

seanm (approved):

libusb/descriptor.c
MarkLee131: Good catch to ask. I checked the two earlier `return parsed;` sites in this function (the "unexpected descriptor" branch and the "short intf descriptor read" branch above), and they both return *before* `usb_interface->num_altsetting++` on line 221. At that point the new altsetting slot exists in the realloc'd array but isn't yet visible to callers iterating `0 .. num_altsetting`, so the `bNumEndpoints > 0 && endpoint == NULL` invariant break can't be observed there. 
Only this return (line 253) sits after the `num_altsetting++` commit and after `bNumEndpoints` has been copied from `buffer[4]`, which is why the reset is needed here specifically. 
Happy to expand the inline comment to spell that out if you'd like.

tests/fuzz/fuzz_descriptor_parsers.c
MarkLee131: It's an upper bound on the fuzzer input size. Real USB configuration descriptors are bounded by `wTotalLength` (a `uint16_t`, so ≤ 64 KB in theory, and ≤ ~4 KB in practice for any device I've seen), and the IAD array path is similarly small. 8192 is comfortably above both while keeping libFuzzer iterations fast and the corpus small. 
Happy to replace it with a named constant (e.g. `MAX_FUZZ_INPUT`) and a short comment if you'd prefer.

tormodvolden: Yes, please, it would be nice to have a comment noting this is a "comfortable" size and not a hard limit of any kind.

BTW, would it be possible to have human-readable corpus seed definitions instead of the (small) binary files, with some kind of preprocessing? That would allow comments and make it less "magic".

seanm: @MarkLee131 are you an LLM?

MarkLee131: Hi @seanm, no, I’m a human contributor... English is not my first language, so I use chatgpt to polish the wording of my replies......

seanm: Yes, it very much sounds like ChatGPT. :)

>BTW, would it be possible to have human-readable corpus seed definitions instead of the (small) binary files, with some kind of preprocessing? That would allow comments and make it less "magic".

@tormodvolden would that really add so much value? Fuzzing is of course about (pseudo-)random input, so knowing which random garbage corresponds to which struct field could be helpful in debugging I suppose, but in my experience, once a fuzzer finds a repro case, it's usually pretty easy to figure it out in a debugger.

I wouldn't hold up this great PR for that anyway.

tormodvolden: The inputs in question are not random so it would have an explanatory value. The author has tried to pack a description into the file names, I think we would agree that it is better to have it explained in text as well. Apart from avoiding blobs. But I agree this shouldn't hold up the real code fixes, it is also why it was a "BTW" question and not a request for changes. It can improved later. The code fixes look good.

tormodvolden (approved):

tests/fuzz/fuzz_descriptor_parsers.c
tormodvolden: I also added a short comment about 8192, taken from the author's explanation above.

PR #1823, opened 2026-05-03
Release 1.0.30-rc2

tormodvolden:
(exceptionally including nano version change, for build artifacts to have them correct)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,6 +25,7 @@ Alexander Mot
 Alexander Pyhalov
 Alexander Schlarb
 Alexander Stein
+Alex Birkett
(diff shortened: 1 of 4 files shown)


mcuee:
Good to go.

mcuee:
Interestingly macOS build hangs again. I think we should really relax the stress_mt test.

Of course that is nothing to do with this PR.

tormodvolden:
stress_mt just gets stuck on macOS from time to time, randomly. It is tracked in another ticket. Let's see if the continued work on macOS concurrency issues will help.

seanm (approved):

PR #1825, opened 2026-05-05
windows: Avoid warning when building without logging

tormodvolden:


mcuee (approved):

seanm (approved):

mcuee:
This should be merged for 1.0.30 release.

tormodvolden:
I am also fine with merging this before the release. The change doesn't affect the binaries for the normal builds (with logging). Not sure why it changes the --disable-log build since it looks like an unused assignment that would be optimized away anyway, but it is probably a secondary optimization artifact, and without risk.

PR #1827, opened 2026-05-07
darwin: null-terminate device string buffer

antoinemoreau:
The `get_device_string` documentation ([libusbi.h](https://github.com/libusb/libusb/blob/master/libusb/libusbi.h)) states:

> _"This function must not write more than length bytes into data, including the null terminator. Returns the actual length in bytes including the null terminator on success."_

The macOS implementation (`darwin_get_device_string` in `os/darwin_usb.c`) doesn't properly handle the null terminator.

`CFStringGetBytes` writes into the buffer without adding a null terminator. `libusb_get_device_string` then calls `usbi_utf8_copy` that reads past the valid string bytes into uninitialised heap memory, sometimes appending garbage to device strings.

I observed this symptom on one of my projects and found a fix to be consistent with the behavior we already have on windows and linux, and the documentation:

Pass `length - 1` to `CFStringGetBytes` to reserve one byte, then explicitly write `'\0'` at the end of the string. Return value updated to `cfUsedIndex + 1` to stay consistent with the documented `get_device_string` function (length including null terminator).

I couldn't find any issue referencing this bug, but I may have missed it!

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -1673,7 +1673,7 @@ int API_EXPORTED libusb_get_device_string(libusb_device *dev,
 	if ((string_type < 0) || (string_type >= LIBUSB_DEVICE_STRING_COUNT)) {
 		return LIBUSB_ERROR_INVALID_PARAM;
 	}
-	if (length < 0) {
+	if (length <= 0) {
 		return LIBUSB_ERROR_INVALID_PARAM;
 	}
 	if (NULL == data) {
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1009,13 +1009,14 @@ static int darwin_get_device_string(struct libusb_device *dev,
 
   long cfUsedIndex = 0;
   CFStringGetBytes(cf, CFRangeMake(0, CFStringGetLength(cf)), kCFStringEncodingUTF8, '?', false,
-    (uint8_t *) buffer, length, &cfUsedIndex);
+    (uint8_t *) buffer, length - 1, &cfUsedIndex);
   CFRelease(cf);
 
   if (cfUsedIndex <= 0)
     return LIBUSB_ERROR_NOT_FOUND;
 
-  return (int) cfUsedIndex;
+  buffer[cfUsedIndex] = '\0';
+  return (int) cfUsedIndex + 1;
 }
 
 static int get_configuration_index (struct libusb_device *dev, UInt8 config_value) {
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -972,6 +972,9 @@ static void darwin_exit (struct libusb_context *ctx) {
 static int darwin_get_device_string(struct libusb_device *dev, 
     enum libusb_device_string_type string_type, char *buffer, int length) {
   
+  if (length <= 0)
+    return LIBUSB_ERROR_INVALID_PARAM;
+
   struct darwin_cached_device *priv = DARWIN_CACHED_DEVICE(dev);
   io_iterator_t deviceIterator;
   io_service_t service;


libusb/os/darwin_usb.c
```diff
@@ -1009,13 +1009,14 @@ static int darwin_get_device_string(struct libusb_device *dev,
 
   long cfUsedIndex = 0;
   CFStringGetBytes(cf, CFRangeMake(0, CFStringGetLength(cf)), kCFStringEncodingUTF8, '?', false,
-    (uint8_t *) buffer, length, &cfUsedIndex);
+    (uint8_t *) buffer, length - 1, &cfUsedIndex);
```
seanm, line 1015: if length is 0, this will pass -1 to CFStringGetBytes().  Not sure what it does, but probably nothing good.

Perhaps the public libusb_get_device_string() should also change its check of:

```
	if (length < 0) {
	}
```

to `length <= 0`...

seanm (changes requested):

libusb/os/darwin_usb.c
antoinemoreau: Done, changed both length checks in libusb_get_device_string() and darwin_get_device_string().

seanm (approved):

seanm:
@sonatique @tormodvolden since this is a new API we should merge this fix for 1.0.30 too IMHO.

libusb/descriptor.c
tormodvolden, line 1682: Looks like the original code was explicitly allowing length 0. But I don't know why.

tormodvolden:
Agree, this looks like an important fix for the new get_device_string() API.

libusb/descriptor.c
tormodvolden: @mliberty1 Maybe you can clarify, if length 0 was meant to be allowed.

mliberty1: The USB spec technically allows zero-length string descriptors, which are "00 00" in UTF-16.  Since C string functions traditionally do not include the NULL terminator in length, this ends up being zero length.

However, I did not test how the platforms handled this.  If the platforms return a zero length string, I figured just go with it.

mcuee: In this case, the PR is not correct, right?

mcuee: What about Linux?

seanm: >zero-length string descriptors, which are "00 00" in UTF-16

"00 00" in UTF-16 is 2 bytes, is it not?

The function parms are:

```
 * \param data the data buffer for the UTF-8 encoded string.
 * \param length the size of the data buffer in bytes.  
 *      USB string descriptors cannot be longer than 
 *      LIBUSB_DEVICE_STRING_BYTES_MAX.
```

If the given buffer is zero length, you can't even write the NUL terminator into it.  If the caller gives no buffer, there's nothing to be done, and the caller surely made a mistake.

mcuee:
Linux code for reference:
https://github.com/libusb/libusb/blob/master/libusb/os/linux_usbfs.c

```
static int op_get_device_string(struct libusb_device *dev, 
		enum libusb_device_string_type string_type, char *buffer, int length)
{
	ssize_t r;
	int fd;
	struct linux_device_priv *priv = usbi_get_device_priv(dev);
	struct libusb_context* ctx = DEVICE_CTX(dev);
	const char * attr;

	switch (string_type) {
		case LIBUSB_DEVICE_STRING_MANUFACTURER: attr = "manufacturer"; break;
		case LIBUSB_DEVICE_STRING_PRODUCT: attr = "product"; break;
		case LIBUSB_DEVICE_STRING_SERIAL_NUMBER: attr = "serial"; break;
		case LIBUSB_DEVICE_STRING_COUNT:
			/* intentional fall-through, avoid -Wswitch-enum */
		default:
	}
	fd = open_sysfs_attr(ctx, priv->sysfs_dir, attr);
	if (fd < 0)
		return LIBUSB_ERROR_IO;

	r = read(fd, buffer, length - 1);  // leave space for null terminator
	if (r < 0) {
		r = errno;
		close(fd);
		if (r == ENODEV)
		usbi_err(ctx, "attribute %s read failed, errno=%zd", attr, r);
		return LIBUSB_ERROR_IO;
	}
	close(fd);
	buffer[r] = 0;  // add null terminator
	while (r && ((buffer[r] == 0) || (buffer[r] == '\n'))) {
		buffer[r--] = 0;
	}
	++r;
	return r;
}
```

libusb/descriptor.c
mcuee: > What about Linux?

From the Linux code, it seems to me this PR is correct after all.

mcuee (approved):

seanm:
> r = read(fd, buffer, length - 1);  // leave space for null terminator

So same bug. 0 - 1 = SIZE_T_MAX = boom.

mliberty1:
So, are we talking about two different things?  I think the original issue was with the length returned by `darwin_get_device_string`, right?

Validation of the input length parameter is different.  For the input parameters, if the user provides a valid pointer with zero length, that violates normal C semantics.  Checking both pointer!=NULL and length > 0 is always valid, and checking these arguments has no downside other than a small bit of overhead.  It's always safe to add these checks.  I have no opinion on adding this check or not.

seanm:
> So, are we talking about two different things?

I'm talking about the code change in this PR.

mcuee:
> > r = read(fd, buffer, length - 1);  // leave space for null terminator
> 
> So same bug. 0 - 1 = SIZE_T_MAX = boom.

That means the changes to descriptor.c is correct, right?

mcuee:
> Validation of the input length parameter is different. For the input parameters, if the user provides a valid pointer with zero length, that violates normal C semantics. Checking both pointer!=NULL and length > 0 is always valid, and checking these arguments has no downside other than a small bit of overhead. It's always safe to add these checks. I have no opinion on adding this check or not.

So you are okay with the change to descriptor.c, right?

> So, are we talking about two different things? I think the original issue was with the length returned by `darwin_get_device_string`, right?

So you are not okay with the change to `darwin_usb.c`, right?

mcuee:
From USB 2.0 Spec

Screenshot 2026-05-12 134905

antoinemoreau:
> I think the original issue was with the length returned by `darwin_get_device_string`, right?

It is not really about the returned value itself.

The main issue comes from `libusb_get_device_string` that copies the data into the output buffer without considering the value returned by the `get_device_string` implementation.
Since the darwin implementation doesn't ensure there is a null terminator, `usbi_utf8_copy` can read into uninitialized memory.

I see two different solutions:
- each implementation properly adds the null terminator (that's what the documentation says)
- `libusb_get_device_string` uses the `get_device_string` return value to copy the data, and ensures there is a null terminator

mliberty1:
PR looks good to me

PR #1828, opened 2026-05-11
linux: (perf) add urb preallocation for bulk and iso

jranson:
This change adds 2 new functions `libusb_prealloc_*_urbs` + respective changes in `submit_*_transfer` and `handle_*_completion` that allow linux/android callers to optionally preallocate and reuse URBs in both bulk and isochronous modes to eliminate all URB heap operations during transfers.

Note: this does not add the optimizations to Windows/Darwin. Let me know if I should do that in separate PRs or in this one. I'm happy to update all of them. Also this is my first time contributing to libusb, so apologies if I am missing anything.

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1934,6 +1934,8 @@ struct libusb_transfer * LIBUSB_CALL libusb_alloc_transfer(int iso_packets);
 int LIBUSB_CALL libusb_submit_transfer(struct libusb_transfer *transfer);
 int LIBUSB_CALL libusb_cancel_transfer(struct libusb_transfer *transfer);
 void LIBUSB_CALL libusb_free_transfer(struct libusb_transfer *transfer);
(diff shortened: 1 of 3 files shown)


mcuee:
This seems to be a good improvement. Please add Windows and Linux implementation in this PR.

mcuee:
It is very nice that you add the test for Linux as well. 

I am not sure if you can add a generic test application as well, say for the generic bulkloop/bulksrcsink/isoloop/isosrcsync example, or whatever device you have.

We do not have a good such example as libusbK kBench.
https://github.com/mcuee/libusbk/tree/master/BmFW/CYPRESS_FX3
https://github.com/mcuee/libusbk/tree/master/libusbK/src/kBench

I use Cypress examples under Linux and macOS for testing.
https://github.com/Ho-Ro/cyusb_linux/blob/main/src/09_cyusb_performance.cpp

jranson:
Sounds good. I am setting to Draft for now and will set back to Ready once the other platforms have been implemented.

PR #1830, opened 2026-05-12
windows: stop clearing endpoint data before a successful claim

asahtik:
On Windows, libusb clears endpoint data before attempting to claim the interface. This changes the state when attempting to claim an already claimed interface from a different device handle.

~This seems to be asymptomatic on master, but~ it causes issues on [this PR](https://github.com/libusb/libusb/pull/1181). Regardless, if I'm not missing something crucial, I believe this fix will reduce the risk of any future regressions.


mcuee:
> This seems to be asymptomatic on master, but it causes issues on https://github.com/libusb/libusb/pull/1181.

@asahtik 

Just wondering what kind of issues you have with PR #1181 on top of git.

sonatique (approved):
This change seems good overall and a prerequisite for #1181

sonatique:
@asahtik : could you tell us more about your scenario  (claim against an interface already held by another handle)? Thank you!

asahtik:
Sorry for the late response. The scenario that caused me issues is using the [XLink](https://github.com/luxonis/XLink) library. It exposes two important functions:

 - `XLinkFindAllSuitableDevices` which for each suitable device opens a handle, claims the interface, sends & reads some data and then frees the interface and closes the device handle
 - `XLinkBootMemory` which  opens a handle to the specified device, claims the interface and sends the firmware in batches
 
 The problem appeared when one thread was polling `XLinkFindAllSuitableDevices` while another was trying to boot a device - the batched transfer stopped working once the device's interface was claimed from the first thread.

I'm still not sure why exactly the mentioned PR started causing issues, but we haven't had any problems since applying this patch.

sonatique:
@mcuee I reviewed the PR and tested on hardware. I think it is it would be nice to have it merged soon because it would unblock #1181 and enable a nice 1.0.31 release.

@asahtik I would like to add one correction to the description: this is not asymptomatic on master. priv->usb_interface[iface].endpoint is device-scoped while claim and release are handle-scoped, so clearing it before the backend call means any failed attempt destroys the table an already-successful claim depends on. interface_by_endpoint() then returns -1 and all transfers on that interface fail with LIBUSB_ERROR_NOT_FOUND.

It reproduces on current master with a single handle, no composite device, no #1181:

```
claim(A, 0)                        = LIBUSB_SUCCESS
clear_halt(A, 0x01)                = LIBUSB_SUCCESS        <- baseline
set_interface_alt_setting(A, 0, 1) = LIBUSB_ERROR_IO       <- device has only altsetting 0
clear_halt(A, 0x01)                = LIBUSB_ERROR_NOT_FOUND <- interface now broken
```

So any application that probes an alternate setting and gets a rejection breaks its own interface until it re-claims.

The two-handle case reproduces too (libusbK device, with a79e8f14 from #1323 applied so a competing claim actually fails):

```
claim(B, 0)                        = LIBUSB_ERROR_BUSY
clear_halt(A, 0x01)                = LIBUSB_ERROR_NOT_FOUND <- handle A broken by handle B

```

That is your XLink symptom, and it explains why #1181 exposed it: on master the second handle fails earlier at libusb_open() (#1177), so you never reach the claim.

With this PR applied, both cases return LIBUSB_SUCCESS and the interface keeps working. Tested on x64 against a libusbK device. 

Could you expand the commit message so it records what this actually fixes? Something like:
"
windows: stop clearing endpoint data before a successful claim

Move the endpoint table reset out of winusb_claim_interface() and winusb_set_interface_altsetting() into windows_assign_endpoints(), which only runs once the backend operation has succeeded. 

priv->usb_interface[iface].endpoint is device-scoped while claim and release are handle-scoped, so clearing it up front let a failed attempt destroy the table an already-successful claim still depends on, leaving interface_by_endpoint() unable to match any endpoint of that interface.

Both callers already hold interface_lock, so the write stays inside the lock, and nb_endpoints is now always reset together with the pointer.
"

Thank you very much!

mcuee (approved):

sonatique:
@Youw @seanm : I wish to merge this PR soon, would you like more time to review it? Thanks!

libusb/os/windows_winusb.c
```diff
@@ -548,6 +548,9 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 	const struct libusb_interface_descriptor *if_desc;
 	int i, r;
 
+	safe_free(priv->usb_interface[iface].endpoint);
+	priv->usb_interface[iface].nb_endpoints = 0;
```
seanm: Setting it to 0 after seems to be exactly what `safe_free` is for, no?

seanm: Or sorry, need more coffee, didn't see the `nb_`

seanm (approved):
I don't know this code or OS, but looks reasonable.

mcuee:
@Youw 

Please review as well. Thanks.

libusb/os/windows_winusb.c
Youw: This closes the wipe-before-claim hazard, but one residual window remains: the table is cleared here, and `windows_assign_endpoints` can still fail afterwards — `libusb_get_active_config_descriptor` (`active_config == 0`, transient `NO_MEM`), `if_desc == NULL`, or the `malloc` below — in which case a sub-API op that *succeeded* still returns an error with the shared table wiped. That is the same end state master reached on those paths (incomplete hardening, not a regression), but the cross-handle scenario from the commit message can still fire through it: `hid_claim_interface` has no cross-handle exclusivity (per-handle `INTERFACE_CLAIMED` flag only), so handle B's claim of an interface handle A is using can succeed, clear the table here, then fail the descriptor fetch — and A is broken again.

To close it completely, build the new table into locals and install it only on success (the install has to stay before the `configure_endpoints` call at the bottom, which reads the table from `priv`), roughly:

```c
	uint8_t *endpoints = NULL;
	...
	if (if_desc->bNumEndpoints > 0) {
		endpoints = malloc(if_desc->bNumEndpoints);
		if (endpoints == NULL) {
			r = LIBUSB_ERROR_NO_MEM;
			goto end;
		}
		for (i = 0; i < if_desc->bNumEndpoints; i++) {
			endpoints[i] = if_desc->endpoint[i].bEndpointAddress;
			usbi_dbg(HANDLE_CTX(dev_handle), "(re)assigned endpoint %02X to interface %u", endpoints[i], iface);
		}
	} else {
		usbi_dbg(HANDLE_CTX(dev_handle), "no endpoints found for interface %u", iface);
	}

	safe_free(priv->usb_interface[iface].endpoint);
	priv->usb_interface[iface].endpoint = endpoints;
	priv->usb_interface[iface].nb_endpoints = if_desc->bNumEndpoints;
```

The minimal alternative is to move this clear down to just after the `if_desc == NULL` check (its pre-PR position, now also resetting `nb_endpoints`), which leaves only the malloc-failure window open. Either is fine as a follow-up too — the PR as-is is already a strict improvement.

Youw (commented):
Reviewed at the code level (no hardware on my side): traced all three sub-API claim/altsetting paths, every reader/writer of `usb_interface[].endpoint`/`nb_endpoints`, the locking, and the failure-state matrix vs master.

The move is correct and I found no regression:

- Both call sites hold `interface_lock` across the sub-API op plus `windows_assign_endpoints`, and (as of current master) every reader of the table takes the same lock, so the relocated clear adds no concurrency window.
- No sub-API `claim_interface`/`set_interface_altsetting` implementation reads or partially mutates the endpoint table, so clearing after the op instead of before is invisible to them.
- A table that now survives a failed claim stays dormant through the failing handle: every consumer first requires `HANDLE_VALID(...api_handle)`, which a failed claim leaves invalid.
- Keeping state on a failed altsetting matches the Linux/darwin backends' keep-state-on-failure semantics — the old wipe-first behavior was the outlier.

One inline note below about the residual clear-then-fail window inside `windows_assign_endpoints` itself — optional hardening, not a blocker.

While auditing the surrounding code I ran into a few pre-existing issues in the same area — none introduced or worsened by this PR. Listing them here so they don't get lost; happy to file separate issues:

1. **Out-of-bounds write in `winusbx_configure_endpoints`** (line 3580, since 0ce010c / v1.0.25): `zlp` is `uint8_t zlp[USB_MAXENDPOINTS]` (32 bytes) but is indexed with the raw `bEndpointAddress`, so every IN endpoint (0x81+ → index 129+) writes past the array into the neighboring `interface_handle[]` entries. Harmless on typical single-interface devices (it hits calloc-zeroed unused slots, and being intra-struct it never trips ASan), but on a composite device with several interfaces open it zeroes bytes of a live `dev_handle`/`api_handle`. The transfer-path use at line 4209 is OUT-only and therefore in-bounds for conformant devices. Fix would be a direction-aware index (e.g. `(ep & 0x0F) | ((ep & 0x80) >> 3)`) or skipping the `zlp` reset for IN endpoints.
2. **Claim-succeeded-but-`windows_assign_endpoints`-failed wedges** (the exact transition this PR touches; all reproducible on master too): (a) HID: `hid_claim_interface` has already set the per-handle `INTERFACE_CLAIMED` flag (line 5244) and core records no claim on error (core.c:1867-1869), so the same handle's retry returns `LIBUSB_ERROR_BUSY` forever and release returns `NOT_FOUND` (core.c:1905-1908) — wedged until `libusb_close`. (b) winusbx associated interfaces: the `GetAssociatedInterface` branch lacks the already-initialized short-circuit the `Initialize` branch has (line 3644), so a retry fails `ERROR_ALREADY_EXISTS` and the failure path (3731-3733) overwrites the still-valid `api_handle` with `INVALID_HANDLE_VALUE` — that WinUSB handle is then never freed at close, and the interface stays `BUSY`. (c) plain `Initialize` path: benign in comparison — a retry hits 3644 and fully recovers.
3. Nit: in the `configure_endpoints` dispatch (584-586) the per-interface branch drops the return value (`r =` missing, unlike the device-level branch above). Unreachable failure today (`CHECK_WINUSBX_AVAILABLE` would have failed the preceding claim first), but worth aligning whenever this code is touched next.

_Drafted with Claude Code._

mcuee:
@asahtik 

Please help to address the comments from @Youw when you got the time. Thanks.

PR #1832, opened 2026-05-18
windows hotplug: undo RegisterClass in stop_event_monitor

sl1200mk2:
`windows_stop_event_monitor()` destroys the hidden event window but never calls `UnregisterClass()` to undo the `RegisterClass()` done in `init_wnd_class()`. On the second `libusb_init()` of a process, `RegisterClass()` then fails because the class already exists for the HINSTANCE, `init_wnd_class()` returns false, and the event thread exits with -1.

After that, no `WM_DEVICECHANGE` is ever received, no hotplug callback ever fires. The failure is silent because `usbi_err()` is a no-op without `ENABLE_LOGGING`.

This patch:
- adds `UnregisterClass()` symmetric to the existing `RegisterClass()`;
- resets `windows_event_hwnd` and `windows_event_thread_handle` to NULL so a subsequent start is idempotent;
- moves `#define WND_CLASS_NAME` above `windows_stop_event_monitor()` so the symbol is visible there.

Reproducer: any app that does `libusb_init` / `libusb_exit` / `libusb_init` in the same process.

### DIFF
--- a/libusb/os/windows_hotplug.c
+++ b/libusb/os/windows_hotplug.c
@@ -43,6 +43,8 @@
 #define HOTPLUG_DEBOUNCE_MS             10
 #define HOTPLUG_DEBOUNCE_MAX_DELAY_MS   100
 
+#define WND_CLASS_NAME TEXT("libusb-1.0-windows-hotplug")
+
 static ULONGLONG first_debounce_tick;
 static HWND windows_event_hwnd;
 static HANDLE windows_event_thread_handle;
@@ -97,6 +99,13 @@ int windows_stop_event_monitor(void)
 		log_error("CloseHandle");
 		return LIBUSB_ERROR_OTHER;
 	}
+
+	/* Symmetric with RegisterClass in init_wnd_class. Without this, a
+    * subsequent libusb_init in the same process fails to recreate
+    * the class and the event thread dies silently. */
+    UnregisterClass(WND_CLASS_NAME, GetModuleHandle(NULL));
+    windows_event_hwnd = NULL;
+    windows_event_thread_handle = NULL;
 
 	return LIBUSB_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


sonatique (approved):
Nice fix and a clear write-up, thank you very much for the fix!
One nit: the four new lines (the comment continuation and the three statements) are space-indented but I think that the rest of the file uses tabs: could you switch them to tabs to match?

sl1200mk2:
github things are a little bit obscure to me... If you can do it it could save me a lot of time...

seanm:
Now that 1.0.30 is done, hopefully I can proceed with the auto-reformatting of the codebase (https://github.com/libusb/libusb/pull/1731) which will fix this automatically.

mcuee:
@sonatique 

I think you can still merge this one and we can fix the formating later.

sonatique:
Update to the matter: Thanks for this, @sl1200mk2 : the diagnosis is spot-on and the reproducer (init / exit / init in one process) is exactly the right way to trigger it. This is the same failure tracked in #1862.

Since you opened this, master landed a fix for the same bug in c66599c3 (windows: fix hotplug event monitor restart) and 5b870b63 (windows: register hotplug window class with this copy's module handle) from PR #1863. 
The thing is the issue was open and solved without realizing some work was done here already... This happens when there are more things in parallel than people to handle them. 

PR #1863 took a slightly different tack:

UnregisterClass is called at the event thread's exit rather than in windows_stop_event_monitor, so it also covers the CreateWindow-failure path and an abnormally terminated thread, not just the orderly stop.

init_wnd_class now tolerates ERROR_CLASS_ALREADY_EXISTS as a recovery, so a class leaked by a crashed thread doesn't permanently break hotplug.

The class is keyed to this libusb copy's own module (via GetModuleHandleEx from a local symbol) instead of GetModuleHandle(NULL), so two independent copies of libusb in one process (e.g. one static-linked into the app and one bundled as a DLL by a plugin) register distinct classes instead of colliding.

I resolved your branch against current master (while wanting to solve conflicts) and I double-checked, and the result is identical to what's already there, so the change is fully covered. Because the remaining delta would reintroduce GetModuleHandle(NULL), I'm going to close this as already-fixed rather than merge it.

Thanks again for your work.

sl1200mk2:
@sonatique it makes sense.

thank you for your work.

++

PR #1826, opened 2026-05-06
windows: Avoid -Wpragma-pack warnings on Clang builds

sl1200mk2:
The Microsoft headers <pshpack1.h> / <poppack.h> exist solely to issue
#pragma pack(push, 1) / #pragma pack(pop). Clang on MSYS2 (UCRT64)
emits -Wpragma-pack every time these headers are included, since the
alignment is modified inside an included file. The same warnings do
not appear under Apple Clang or OpenBSD clang, so this is specific to
MSYS2 clang.
Replace the includes by the equivalent #pragma pack directives, which
are already used elsewhere in the codebase (libusb.h, libusbi.h,
os/windows_winusb.c) and are portable across MSVC, GCC, and Clang.
The semantics are identical on all targets.
Tested on MSYS2 UCRT64 + Clang: build is now warning-free for the
Windows backend headers.
Closes: #1395
Assisted-by: claude-code:claude-opus-4-7
Signed-off-by: Nicolas Bats <sl1200mk2@gmail.com>

PR #1833, opened 2026-05-18
windows: Use #pragma pack consistently instead of includes (2)

tormodvolden:
A leftover commit from #1826 that I somehow missed out on when merging that PR (commit 47495346).

Below from Nicolas' original commit summary:

> The Microsoft headers <pshpack1.h> / <poppack.h> exist solely to issue emits -Wpragma-pack every time these headers are included, since the alignment is modified inside an included file. The same warnings do not appear under Apple Clang or OpenBSD clang, so this is specific to MSYS2 clang.
> Replace the includes by the equivalent #pragma pack directives, which are already used elsewhere in the codebase (libusb.h, libusbi.h, os/windows_winusb.c) and are portable across MSVC, GCC, and Clang. The semantics are identical on all targets.
> Tested on MSYS2 UCRT64 + Clang: build is now warning-free for the Windows backend headers.

### DIFF
--- a/libusb/os/windows_common.h
+++ b/libusb/os/windows_common.h
@@ -160,7 +160,7 @@ enum windows_version {
 
 extern enum windows_version windows_version;
 
-#include <pshpack1.h>
+#pragma pack(push, 1)
 
 typedef struct USB_DEVICE_DESCRIPTOR {
 	UCHAR  bLength;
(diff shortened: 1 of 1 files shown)


mcuee (approved):

Youw (approved):

seanm:
I can't help but note: if we were just using github's "merge" button this accidental omission would not have happened. In other words, our old/current workflow is more dangerous in this regard as a PR as-reviewed, is not actually what gets merged.

seanm (approved):

PR #1836, opened 2026-05-20
windows: configure endpoints for composite devices

### LINKED ISSUE #1835: potential bug / suggested patch: AUTO_CLEAR_STALL not enabled for composite devices on Windows
# AUTO_CLEAR_STALL not enabled for composite devices

| Field | Value |
|-------|-------|
| **Date** | 2026-05-18 |
| **Version** | libusb 1.0.29.11953 (source from libusb/libusb GitHub, v1.0.29 release) |
| **Platform** | Windows 10/11, WinUSB backend |
| **Compiler** | GCC 16.1.0 (MinGW-w64) |

---

## Summary

On Windows, when a USB device uses the composite driver (USBCCGP), the WinUSB pipe policy `AUTO_CLEAR_STALL` is never configured for its endpoints. This causes endpoint stalls to be unrecoverable in practice, even though `libusb_clear_halt()` is called by the application.

For non-composite devices using WinUSB directly, `AUTO_CLEAR_STALL` (along with other pipe policies) **is** correctly configured via `winusbx_configure_endpoints()`.

---

## Root Cause

In `windows_winusb.c`, the function `windows_assign_endpoints()` is called after an interface is successfully claimed. At the end of this function, it calls the backend's `configure_endpoints` function pointer:

```c
// Extra init may be required to configure endpoints
if (priv->apib->configure_endpoints)
    r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
```

The problem: for the `USB_API_COMPOSITE` backend, `configure_endpoints` is `NULL`:

```c
{
    USB_API_COMPOSITE,
    "Composite API",
    ...
    NULL,   /* configure_endpoints */    // <-- HERE
    composite_claim_interface,
    ...
},
```

While for `USB_API_WINUSBX` (non-composite), it is set:

```c
{
    USB_API_WINUSBX,
    "WinUSB-like APIs",
    ...
    winusbx_configure_endpoints,         // <-- Set correctly
    winusbx_claim_interface,
    ...
},
```

As a result, `winusbx_configure_endpoints()` — which sets `AUTO_CLEAR_STALL`, `ALLOW_PARTIAL_READS`, `SHORT_PACKET_TERMINATE`, `PIPE_TRANSFER_TIMEOUT`, etc. — is **never** called for interfaces on composite devices.

---

## How to Reproduce

1. Use a USB composite device (handled by Windows USBCCGP driver) where one or more interfaces use WinUSB. Example: a device with:
   - Interface 0: WinUSB vendor-specific (bulk endpoints)
   - Interface 1+2: CDC ACM (serial)

2. Open the device with libusb, claim the WinUSB interface, and perform bulk transfers.

3. Trigger any condition that causes an endpoint STALL (e.g., device NAKs, protocol error, buffer overrun on the device side).

4. Observe: the endpoint remains permanently halted. Even though the application calls `libusb_clear_halt()`, the recovery is unreliable:
   - Sometimes `clear_halt` appears to succeed but the next transfer stalls again
   - Timing-dependent: fast retry may work, slow retry may not
   - Multiple in-flight async transfers may all fail before `clear_halt` runs
   - The device effectively becomes unresponsive, requiring re-enumeration

5. For comparison: the same device configured as a non-composite single-interface WinUSB device does **not** exhibit this problem, because `AUTO_CLEAR_STALL` is set via `winusbx_configure_endpoints()`.

---

## Patch

**File:** `libusb/os/windows_winusb.c`
**Function:** `windows_assign_endpoints()` (around line 568 in v1.0.29)

### Original

```c
// Extra init may be required to configure endpoints
if (priv->apib->configure_endpoints)
    r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);

if (r == LIBUSB_SUCCESS)
    priv->usb_interface[iface].current_altsetting = altsetting;
```

### Patched

```c
// Extra init may be required to configure endpoints
if (priv->apib->configure_endpoints)
    r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
else if (priv->usb_interface[iface].sub_api != SUB_API_NOTSET)
    winusbx_configure_endpoints(priv->usb_interface[iface].sub_api, dev_handle, iface);

if (r == LIBUSB_SUCCESS)
    priv->usb_interface[iface].current_altsetting = altsetting;
```

---

## Patch Explanation

The added `else if` clause handles composite devices where `configure_endpoints` is `NULL`. It uses the per-interface `sub_api` (which is correctly set to the WinUSB sub-api for each interface during enumeration) to call `winusbx_configure_endpoints()` directly.

### Safety

- **Non-composite devices (`USB_API_WINUSBX`):** take the existing `if`-branch as before. No change in behavior.
- **Composite devices (`USB_API_COMPOSITE`):** fall through to the `else if`. The guard (`sub_api != SUB_API_NOTSET`) ensures we only call it for interfaces that actually use a WinUSB-like driver.
- **HID/unsupported backends:** `sub_api` is `SUB_API_NOTSET`, so the `else if` is skipped. No change in behavior.

The return value of `winusbx_configure_endpoints()` in the `else if` branch is intentionally **not** assigned to `r`, since endpoint configuration is best-effort and should not cause `claim_interface` to fail.

jerry9871:
For composite USB devices (USB_API_COMPOSITE), the configure_endpoints function pointer is NULL, causing winusbx_configure_endpoints() to never be called. This means pipe policies like AUTO_CLEAR_STALL are not set for composite device interfaces, unlike non-composite WinUSB devices.

Without AUTO_CLEAR_STALL, endpoint stalls on composite devices become unrecoverable - libusb_clear_halt() works unreliably due to timing issues with async transfers, often requiring device re-enumeration.

Add an else-if branch in windows_assign_endpoints() to call winusbx_configure_endpoints() directly for composite device interfaces that have a valid WinUSB sub_api, ensuring consistent pipe policy configuration regardless of device topology.

Closes #1835

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -575,6 +575,9 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 	// Extra init may be required to configure endpoints
 	if (priv->apib->configure_endpoints)
 		r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
+	else if (priv->usb_interface[iface].apib->configure_endpoints)
+		priv->usb_interface[iface].apib->configure_endpoints(
+			priv->usb_interface[iface].sub_api, dev_handle, iface);
 
 	if (r == LIBUSB_SUCCESS)
 		priv->usb_interface[iface].current_altsetting = altsetting;


libusb/os/windows_winusb.c
```diff
@@ -575,6 +575,8 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 	// Extra init may be required to configure endpoints
 	if (priv->apib->configure_endpoints)
 		r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
+	else if (priv->usb_interface[iface].sub_api != SUB_API_NOTSET)
```
sonatique: Thanks for the fix and the thorough issue write-up.
One concern with the guard, though. Issue #1835 says HID/unsupported backends are safe because their sub_api is SUB_API_NOTSET. However, I think that for HID interfaces sitting inside a composite device that isn't quite right: during enumeration, get_api_type() calls get_sub_api() to fill in (api, sub_api), then the loop passes that pair to set_composite_interface(), which stores sub_api on the interface: for USB_API_HID, get_sub_api returns the index into hid_driver_names[] = {"HIDUSB", "MOUHID", "KBDHID"}, so priv->usb_interface[iface].sub_api ends up as 0, 1, or 2 (set at line 1424), not SUB_API_NOTSET.

With the current patch that means a composite device with a HID interface will hit the new else if and call winusbx_configure_endpoints(0|1|2, ...) on a HID HANDLE:

CHECK_WINUSBX_AVAILABLE only checks that the DLL slot is loaded, it doesn't reject the index.
WinUSBX[*].SetPipePolicy() then gets called against a HID handle. The OS will reject those calls, but the function only emits usbi_dbg and still returns LIBUSB_SUCCESS.
So in practice it's probably harmless (silent debug-level failures, no functional regression), but it's semantically wrong and noisy in the logs, and it relies on the coincidence that the HID name indices overlap with the WinUSB sub-API indices (notably SUB_API_LIBUSB0 == 1 happens to skip most of the policy loop).

Suggest either tightening the guard:

```
else if (priv->usb_interface[iface].apib->id == USB_API_WINUSBX)
    winusbx_configure_endpoints(priv->usb_interface[iface].sub_api, dev_handle, iface);
```
or, matching the dispatch pattern used by composite_claim_interface and friends, going through the per-interface apib:

```
else if (priv->usb_interface[iface].apib->configure_endpoints)
    priv->usb_interface[iface].apib->configure_endpoints(
        priv->usb_interface[iface].sub_api, dev_handle, iface);
```
The latter auto-extends if any other sub-API ever grows a configure_endpoints hook.

sonatique (changes requested):

libusb/os/windows_winusb.c
jerry9871: updated in 1f12114f. Now dispatches through the per-interface apib->configure_endpoints pointer, which is NULL for HID interfaces (skipped) and points to winusbx_configure_endpoints for WinUSB interfaces (called). Matches the dispatch pattern used elsewhere in the composite driver.

libusb/os/windows_winusb.c
```diff
@@ -575,8 +575,9 @@ static int windows_assign_endpoints(struct libusb_device_handle *dev_handle, uin
 	// Extra init may be required to configure endpoints
 	if (priv->apib->configure_endpoints)
 		r = priv->apib->configure_endpoints(SUB_API_NOTSET, dev_handle, iface);
-	else if (priv->usb_interface[iface].sub_api != SUB_API_NOTSET)
-		winusbx_configure_endpoints(priv->usb_interface[iface].sub_api, dev_handle, iface);
+	else if (priv->usb_interface[iface].apib->configure_endpoints)
```
sonatique, line 578: Very good, thanks you. 
(just squash before merging obviously)

sonatique (approved):

libusb/os/windows_winusb.c
jerry9871: Ok, now in a single commit. I am glad I could contribute

mcuee:
Related issue here.
* https://github.com/libusb/libusb/issues/1841

mcuee (approved):

jerry9871:
A single check reports as 'cancelled' - doesn't seem related to the change itself, but rather looks like a runner failure. How to restart/fix this?

mcuee:
> A single check reports as 'cancelled' - doesn't seem related to the change itself, but rather looks like a runner failure. How to restart/fix this?

You can ignore that. There is a known issue for macOS runner to occassionally hang.

mcuee:
Anyway, I have triggered a re-run and now all checks are passed.

mcuee:
@sonatique 

I will let you merge this one.

PR #1844, opened 2026-06-03
Update altsetting and num_altsetting back-to-back as needed by -fboun…

seanm:
…ds-safety

Update altsetting and num_altsetting back-to-back as needed by -fbounds-safety which requires a pointer and the variable that counts its size be updated together with no intervening statements. This was not the case before, with ->altsetting updated first, then the fields would be partially populated, then the possibility of early return, then -> num_altsetting was updated.

Now extract fields from the buffer into temp variables and check those for the early return situation. Only if early return does not occur, then realloc the buffer and populate the new libusb_interface_descriptor.

### DIFF
--- a/libusb/descriptor.c
+++ b/libusb/descriptor.c
@@ -364,40 +364,39 @@ static int parse_configuration(struct libusb_context *ctx,
 		return LIBUSB_ERROR_IO;
 	}
 
-	config->bLength = buffer[0];
-	config->bDescriptorType = buffer[1];
-	config->wTotalLength = ReadLittleEndian16(&buffer[2]);
-	config->bNumInterfaces = buffer[4];
-	config->bConfigurationValue = buffer[5];
-	config->iConfiguration = buffer[6];
-	config->bmAttributes = buffer[7];
-	config->MaxPower = buffer[8];
-	if (config->bDescriptorType != LIBUSB_DT_CONFIG) {
+	uint8_t bLength = buffer[0];
+	uint8_t bDescriptorType = buffer[1];
+	uint16_t wTotalLength = ReadLittleEndian16(&buffer[2]);
+	uint8_t bNumInterfaces = buffer[4];
+	uint8_t bConfigurationValue = buffer[5];
+	uint8_t iConfiguration = buffer[6];
+	uint8_t bmAttributes = buffer[7];
+	uint8_t MaxPower = buffer[8];
+
+	if (bDescriptorType != LIBUSB_DT_CONFIG) {
 		usbi_err(ctx, "unexpected descriptor 0x%x (expected 0x%x)",
-			 config->bDescriptorType, LIBUSB_DT_CONFIG);
+			 bDescriptorType, LIBUSB_DT_CONFIG);
 		return LIBUSB_ERROR_IO;
-	} else if (config->bLength < LIBUSB_DT_CONFIG_SIZE) {
-		usbi_err(ctx, "invalid config bLength (%u)", config->bLength);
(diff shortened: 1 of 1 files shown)


Youw (approved):
This looks good.
Let me also check with an agent.

seanm:
@sonatique what do you think of this one?  (@Youw has approved already)

libusb/descriptor.c
```diff
@@ -425,31 +434,42 @@ static int parse_configuration(struct libusb_context *ctx,
...
+	config->bDescriptorType = bDescriptorType;
+	config->wTotalLength = wTotalLength;
+	config->bConfigurationValue = bConfigurationValue;
+	config->iConfiguration = iConfiguration;
+	config->bmAttributes = bmAttributes;
+	config->MaxPower = MaxPower;
+
 	return size;
 
 err:
```
sonatique, line 475: This introduces a memory leak on every error path: `config->interface` is now only assigned at `finish:`, so on `goto err`, `clear_configuration(config)` sees `config->interface == NULL` and only frees `config->extra`. The `usb_interface` array from `calloc()` , plus the `altsetting`/`endpoint`/`extra` allocations of every already-parsed interface,  leaks. (In master this was safe because `config->interface` and `config->bNumInterfaces` were set before the parse loop.)

All three `goto err` paths are affected, and two are reachable with malformed descriptor data alone: an extra descriptor with `bLength < 2`, or `parse_interface()` failing (its own `err:` only clears the interface it was parsing, not the earlier ones).

Committing the pointer/count pair at `err:` too restores master's cleanup behavior, and keeps the pair adjacent as `-fbounds-safety` wants:

```suggestion
err:
	config->interface = usb_interface;
	config->bNumInterfaces = bNumInterfaces;
```

This is safe because on every `goto err` path `bNumInterfaces` still holds the original count passed to `calloc()` (it is only reduced right before `goto finish`/`break`), and `clear_interface()` on the calloc-zeroed tail entries is a no-op,  so `clear_configuration()` then frees everything correctly.

libusb/descriptor.c
```diff
@@ -425,31 +434,42 @@ static int parse_configuration(struct libusb_context *ctx,
...
+			if (!newExtra) {
 				r = LIBUSB_ERROR_NO_MEM;
 				goto err;
 			}
 
-			memcpy(extra + config->extra_length, begin, (size_t)len);
-			config->extra = extra;
-			config->extra_length += (int)len;
+			memcpy(newExtra + config->extra_length, begin, (size_t)len);
+			config->extra = newExtra;
```
sonatique: nit: locals are snake_case in this codebase — new_extra_len / new_extra.

sonatique (changes requested):
Thanks @seanm : the direction seems right to me, and the parse_interface() half is strictly better than before (fields validated in temporaries before the realloc, pointer/count updated adjacently). One regression in parse_configuration() needs fixing before merge: the error path now leaks everything the function allocated  (see inline comment).

Unrelated to the code: the macOS CI job was cancelled (runtime bound), not failed. Worth a re-run to get a full green matrix.

libusb/descriptor.c
seanm: Fixed.

libusb/descriptor.c
seanm: Nice catch. Fixed.

sonatique (approved):

seanm:
>[sonatique](https://github.com/sonatique) approved these changes

Great, I will merge later this week unless anyone else has any comments...

PR #1847, opened 2026-06-06
Replace libusb_fill_control_transfer to make length of buffer explicit

seanm:
The function signature provided a buffer but no associated length.  This is not so safe, and won't compile with -fbounds-safety.

Other similar functions like libusb_fill_bulk_transfer, libusb_fill_bulk_stream_transfer, libusb_fill_interrupt_transfer, etc. all sensibly group both together.

Deprecate libusb_fill_control_transfer() and replace it with a new function basically the same but with length explicit.


libusb/libusb.h
```diff
@@ -1979,10 +1980,38 @@ uint32_t LIBUSB_CALL libusb_transfer_get_stream_id(
  * \param buffer data buffer. If provided, this function will interpret the
  * first 8 bytes as a setup packet and infer the transfer length from that.
  * This pointer must be aligned to at least 2 bytes boundary.
+ * \param length length of data buffer in bytes
  * \param callback callback function to be invoked on transfer completion
  * \param user_data user data to pass to callback function
  * \param timeout timeout for the transfer in milliseconds
  */
+static inline void libusb_fill_control_transfer_new(
```
Youw: I think "new" is uncommon suffix for this purpose, while index is often used in *nix systems, other libraries.
```suggestion
static inline void libusb_fill_control_transfer2(
```
This is a subject for discussion.

seanm: Yes, that name is a temporary placeholder.  I could not think of a good new name, at least not quickly.

libusb/libusb.h
```diff
@@ -1979,10 +1980,38 @@ uint32_t LIBUSB_CALL libusb_transfer_get_stream_id(
  * \param buffer data buffer. If provided, this function will interpret the
  * first 8 bytes as a setup packet and infer the transfer length from that.
  * This pointer must be aligned to at least 2 bytes boundary.
+ * \param length length of data buffer in bytes
  * \param callback callback function to be invoked on transfer completion
```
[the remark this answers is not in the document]
seanm: Bot analysis not right, but did make me realize I did not update the docs enough.

libusb/libusb.h
```diff
@@ -1979,10 +1980,38 @@ uint32_t LIBUSB_CALL libusb_transfer_get_stream_id(
...
+		assert (length >= (int) (LIBUSB_CONTROL_SETUP_SIZE
+			+ libusb_le16_to_cpu(setup->wLength)));
+		transfer->buffer = buffer;
+		transfer->length = length;
+	}
+	else
+	{
+		transfer->buffer = NULL;
+		transfer->length = 0;
+	}
```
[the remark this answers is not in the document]
seanm: Bot analysis not right, but did make me realize I did not update the docs enough.

libusb/libusb.h
```diff
@@ -1537,7 +1538,7 @@ struct libusb_transfer {
 	 * of the following functions:
 	 * - libusb_fill_bulk_transfer()
 	 * - libusb_fill_bulk_stream_transfer()
-	 * - libusb_fill_control_transfer()
```
sonatique, line 1543: This function implementation is later declared as obsolete, so I think we shall keep the declaration here, no?

libusb/libusb.h
```diff
@@ -1974,18 +1975,48 @@ uint32_t LIBUSB_CALL libusb_transfer_get_stream_id(
...
+{
+	transfer->dev_handle = dev_handle;
+	transfer->endpoint = 0;
+	transfer->type = LIBUSB_TRANSFER_TYPE_CONTROL;
+	transfer->timeout = timeout;
+	if (buffer)
+	{
+#ifndef NDEBUG
+		struct libusb_control_setup *setup = (struct libusb_control_setup *)(void *)buffer;
+		assert (length == (int) (LIBUSB_CONTROL_SETUP_SIZE
```
sonatique: So: we shall give a length value, but this value shall be exactly equal to another value. I understand the intent but:
- should the assert be changed to >= ?
- I understand that the check is not done on release, I think it defeat the purpose of the change. Should we always log an error (maybe return an error also, i.e. have the new function return an error value) and refuse to go on when length is too small?

sonatique (changes requested):

libusb/libusb.h
seanm: I actually had it as `>=` previously, but changed it after reviewing copilot's comments, because I convinced myself they would/should always be the same.

>I think it defeat the purpose of the change.

The purpose of the change can be summarized as: "we should always easily know how long a buffer is".  An analogous problem is `sprintf` vs `snprintf`. They both take a buffer, but the latter knows how long the buffer is, but the former does not. If your job is to write the code for these functions, you cannot implement `sprintf` safely because you don't know how long your buffer is, and you can't be sure you won't write too much.

`libusb_fill_control_transfer` has a similar problem. The 3rd parameter is a buffer, but we don't know how long it is. Well, we sort of do, because we assume the buffer is actually a `libusb_control_setup` that has the length embedded in there. But that's too clever for the compiler. The compiler (with -fbounds-safety) needs an annotated correlation between a buffer and its length, for example the (abbreviated) prototype of `snprintf` is `snprintf(char* __counted_by(size) string, size_t size, ...)`. Note how the `__counted_by` attribute allows the compiler to know the pairing to the length variable.

The assert is only there to sanity check that the old way gives the same number as the new way. Just in case.

I had another thought though: every call to `libusb_fill_control_transfer` is preceded by a call to `libusb_fill_control_setup`. Maybe I could solve this problem another way: by combining these two functions to make a new function that does both things...

sonatique:
@seanm : still working on that one?

libusb/libusb.h
seanm: I brought it back.

libusb/libusb.h
seanm: >I had another thought though: every call to libusb_fill_control_transfer is preceded by a call to libusb_fill_control_setup. Maybe I could solve this problem another way: by combining these two functions to make a new function that does both things...

@sonatique see `libusb_fill_control_transfer3` in my latest push. What do you think of this alternate solution?

sonatique: I like this direction better than transfer2. Deriving wLength from buffer_len - LIBUSB_CONTROL_SETUP_SIZE removes the second source of truth entirely, so there's no length to get wrong and nothing to assert: that's a real improvement over transfer2, where the assert is compiled out in release and so doesn't actually prevent the mismatch it's guarding against.

That said, I think there is a few things to consider before this can land:

1. It has to be static inline. As written it's a plain void ... definition in libusb.h, which will give duplicate-symbol link errors as soon as the header is included in more than one translation unit (same form as the other helpers.)

2. NULL-buffer path crashes. The setup->... writes happen before the if (buffer) check, but the assert allows !buffer && buffer_len == 0. In that case setup is NULL and we write through it (in release too, since the assert is gone). The setup-populating block needs to move inside if (buffer).

3. Docs: worth stating explicitly that the buffer must be sized exactly to setup + data (no oversized scratch buffer), since wLength is now inferred from the buffer length.

Also a general note: #include <assert.h> in the public header leaks assert/NDEBUG-dependent behavior into consumers, I'd prefer to avoid that in libusb.h if we can.

sonatique (changes requested):

libusb/libusb.h
seanm: > Deriving wLength from buffer_len - LIBUSB_CONTROL_SETUP_SIZE removes the second source of truth entirely

Yeah, that's my thinking too.

> That said, I think there is a few things to consider before this can land:

Yeah, this was just to demonstrate the idea, it's still WIP.

>     1. It has to be static inline.

Actually, I think it's wiser to move the implementation from the .h to the .c.

> Also a general note: #include <assert.h> in the public header leaks assert/NDEBUG-dependent behavior into consumers, I'd prefer to avoid that in libusb.h if we can.

Another reason to put the implementation into the .c.

OK, so I'll go with the transfer3 idea and update this PR...

seanm: OK, see new draft.

seanm: @sonatique did you have more thoughts on this one?

sonatique:
@seanm :

Concept-wise I'm sold: folding the setup packet into the fill call removes the second source of truth entirely, so there's no length left to get wrong and nothing to assert. 

I consider this is still a draft, meaning this won't get merge as-is, right?

**1. Name.**  I'd suggest **`libusb_fill_control_setup_transfer()`**. It follows the existing precedent in the family:` libusb_fill_bulk_stream_transfer()` is a variant of `libusb_fill_bulk_transfer()` taking extra parameters, named by inserting the qualifier before `_transfer`, which is exactly this case. It keeps `control_setup` intact as the established term it is, it names the actual difference from `libusb_fill_control_transfer()` (this one fills the setup for you), and it sorts next to its two siblings in the header and the docs.

**2. `void` or `int`?** My June question: all four `assert()`s vanish under `NDEBUG`, so a NULL transfer or a mis-sized buffer is silent UB in every release build. An error return costs nothing here. I'd prefer `int` with `LIBUSB_ERROR_INVALID_PARAM`. The symmetry argument for `void` applies to the other `libusb_fill_*` helpers, but those are inline and can't validate anything; this one is out-of-line precisely so it can. 

**3. Deprecation scope.** Please don't deprecate `libusb_fill_control_setup()`. It contradicts the new function's own docs, which say a NULL buffer is legal and that you must populate buffer and length later: that's precisely what `libusb_fill_control_setup()` is for, so we'd have no non-deprecated way to do something we document. It would also warn across a lot of correct third-party code for no safety gain. Keep the deprecation on `libusb_fill_control_transfer()` only.

Also: moving the implementation into `io.c` turns this into a real exported symbol, so it now needs `LIBUSB_CALL` on both the declaration and the definition (without it, it's cdecl while the rest of the API is stdcall on Windows) and an entry in `libusb/libusb-1.0.def`. Neither is caught by the Linux CI.

The rest is for when you take it out of draft: a couple of `size_t`/`int` conversions in `sync.c` and `io.c` that would reintroduce warnings we cleaned up in bc1ddf43, a ChangeLog entry under 1.0.31, and some remaining doc references to `libusb_fill_control_setup()`. 

What is your plan. do you want to turn this draft PR into an actual PR?

seanm:
> Concept-wise I'm sold: folding the setup packet into the fill call removes the second source of truth entirely, so there's no length left to get wrong and nothing to assert.
> 
> I consider this is still a draft, meaning this won't get merge as-is, right?

Correct.

> **1. Name.** I'd suggest **`libusb_fill_control_setup_transfer()`**. It follows the existing precedent in the family:` libusb_fill_bulk_stream_transfer()` is a variant of `libusb_fill_bulk_transfer()` taking extra parameters, named by inserting the qualifier before `_transfer`, which is exactly this case. It keeps `control_setup` intact as the established term it is, it names the actual difference from `libusb_fill_control_transfer()` (this one fills the setup for you), and it sorts next to its two siblings in the header and the docs.

Indeed `libusb_fill_control_transfer3` is just a placeholder name (part of why this is draft).

> **2. `void` or `int`?** My June question: all four `assert()`s vanish under `NDEBUG`, so a NULL transfer or a mis-sized buffer is silent UB in every release build. An error return costs nothing here. I'd prefer `int` with `LIBUSB_ERROR_INVALID_PARAM`. The symmetry argument for `void` applies to the other `libusb_fill_*` helpers, but those are inline and can't validate anything; this one is out-of-line precisely so it can.

It's a question of what the API contract is.  i.e. is checking for NULL the job or the *caller* or *callee*?  `strlen(NULL)` is UB because the contract requires the caller to check for NULL.  Same for these libusb API.  The existing `libusb_fill_control_setup()` also does not check for NULL, it's the caller's job.  The asserts are there to be helpful to the programmer that messed up. They serve as documentation of the API contract, and help static analyzers too.

> **3. Deprecation scope.** Please don't deprecate `libusb_fill_control_setup()`. It contradicts the new function's own docs, which say a NULL buffer is legal and that you must populate buffer and length later: that's precisely what `libusb_fill_control_setup()` is for, so we'd have no non-deprecated way to do something we document. It would also warn across a lot of correct third-party code for no safety gain. Keep the deprecation on `libusb_fill_control_transfer()` only.

The whole point is to deprecate the old function, because its signature is unsafe: it takes a buffer of unknown size.

> What is your plan. do you want to turn this draft PR into an actual PR?

Once we all agree on what to do, yes.

seanm:
Another thought occurred: 

`libusb_fill_control_setup` could be made safe by changing it's first parameter from `char*` to `struct libusb_control_setup *`.  Then the compiler knows that the buffer is at least the size of the struct, instead of an unknown size.

This too would be a breaking change, so we'd need a `libusb_fill_control_setup2` (well, a better name than that).

sonatique:
> The whole point is to deprecate the old function, because its signature is unsafe: it takes a buffer of unknown size.

Yes I got but my point is that the documentation of libusb_fill_control_transfer3 says: _"It is also legal to pass a NULL buffer (and zero length) to this function, but remember that **you must then populate the buffer and length fields later**."_

The setup filling code is inside `if(buffer)` so no setup packet is written at all. So the user has to do it afterward and the function meant for this is precisely `libusb_fill_control_setup` which will be now deprecated, so there is no longer helper to do this and the user have to hand-roll the struct with manual `libusb_cpu_to_le16` calls which is precisely the error-prone thing that these helpers exist to prevent.
Is your idea to force the manual way? Or what did I miss?

seanm:
> Yes I got but my point is that the documentation of libusb_fill_control_transfer3 says: _"It is also legal to pass a NULL buffer (and zero length) to this function, but remember that **you must then populate the buffer and length fields later**."_

Well that I just copied from the `libusb_fill_control_transfer` docs, but I would be ok with the new `libusb_fill_control_transfer3` *requiring* a non-null buffer.  That ok for you?

sonatique:
> Well that I just copied from the `libusb_fill_control_transfer` docs, but I would be ok with the new `libusb_fill_control_transfer3` _requiring_ a non-null buffer. That ok for you?

Honestly I think it would be better to have the opinions of more maintainers before deprecating a function that probably is used correctly by all libusb users given how long it has been around. I mean: the safety benefit vs annoyance is difficult to estimate.

seanm:
@sonatique I recently realized there is a much easier solution to this:

```diff
diff --git a/libusb/libusb.h b/libusb/libusb.h
index b9dbb7fa..155ce314 100644
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -1949,7 +1949,7 @@ static inline struct libusb_control_setup *libusb_control_transfer_get_setup(
  * \ref libusb_control_setup::wLength "wLength" field of
  * \ref libusb_control_setup
  */
-static inline void libusb_fill_control_setup(unsigned char *buffer,
+static inline void libusb_fill_control_setup(unsigned char * __sized_by(LIBUSB_CONTROL_SETUP_SIZE) buffer,
        uint8_t bmRequestType, uint8_t bRequest, uint16_t wValue, uint16_t wIndex,
        uint16_t wLength)
 {
@@ -2000,7 +2000,7 @@ uint32_t LIBUSB_CALL libusb_transfer_get_stream_id(
  */
 static inline void libusb_fill_control_transfer(
        struct libusb_transfer *transfer, libusb_device_handle *dev_handle,
-       unsigned char *buffer, libusb_transfer_cb_fn callback, void *user_data,
+       unsigned char * __sized_by_or_null(LIBUSB_CONTROL_SETUP_SIZE) buffer, libusb_transfer_cb_fn callback, void *user_data,
        unsigned int timeout)
 {
        struct libusb_control_setup *setup = (struct libusb_control_setup *)(void *)buffer;
```

It somehow never occurred to me (until recently) that a constant can be used with `__sized_by`, it doesn't have to be another parameter!

So I guess I'll just abandon this PR...

Youw:
I think we can create a separate issue to track backward incompatible changes like this, and maybe introduce them if a next major version release would ever happen.
Until then - an annotation is a fine compatible solution.

PR #1845, opened 2026-06-05
ci: test building the library as C++ on Linux, macOS and Windows

Youw:
## Summary

Adds a blocking `build-cxx` job to the Linux, macOS, and Windows workflows so C++ buildability is continuously verified instead of remaining an untested property.

This is a CI-only change: it does not alter the library or run the test suite under C++. Each job performs configure/build validation only and has a 60-minute timeout.

## Platform coverage

- **Linux / g++**: bootstraps and configures normally, then builds the sources as C++20 with `g++`, `-x c++`, and C++-appropriate `AM_CFLAGS`.
- **macOS / clang++**: follows the same autotools path with `clang++` and C++20.
- **Windows / MSVC**: generates `msvc/Directory.Build.props` with `CompileAs=CompileAsCpp`, then builds the Release x64 `libusb_dll` project. This project override is required because MSBuild otherwise emits `/TC` for `.c` files, overriding `CL=/TP`.

## Prerequisites and follow-ups

The prerequisite C++-compatibility work has landed:

- #1782 covers the common code and Darwin backend.
- #1848 covers the Linux and Windows backends.
- #1899 fixes the `events_posix.c` C++ error exposed after merging newer `master`.

The macOS job also exposes pre-existing `va_start` and Apple SDK anonymous-enum warnings. Their source fixes are isolated in #1904 so this PR remains CI-only. The unrelated Libtool configure probe warning that `-single_module` is obsolete is intentionally left alone.

## Verification and current CI status

- Parsed all GitHub Actions workflow YAML files successfully.
- Linux C++20 build passes locally with g++ 11 under WSL.
- Windows MSVC Release x64 build passes locally with 0 warnings and 0 errors.
- Linux, macOS, and Windows `build-cxx` jobs all pass on the current head.
- `git diff --check` passes.

Related to #1730
Related to #1782
Related to #1848
Related to #1899
Related to #1904

Assisted-by: claude-code:claude-opus-4-8
Assisted-by: codex:GPT-5.6-Sol

PR #1848, opened 2026-06-06
Make the Linux and Windows backends compile as C++

Youw:
## Summary

Building libusb as C++ (useful for static-analysis tools and compiler flags that only run in C++ mode) needs a set of explicit casts plus a couple of C-vs-C++ keyword differences. #1782 makes the **common code + the Darwin backend** build as C++; this PR is the **orthogonal companion** covering the **Linux and Windows** pieces #1782 doesn't touch. It's independent of #1782 (no overlapping changes) and can land before or after it.

## What it fixes

- **`libusbi.h` (atomics)** — g++ can't include `<stdatomic.h>` in C++ mode before C++23 (the `_Atomic` keyword); this was by far the biggest blocker on Linux (>400 cascading errors). Use the GCC `__atomic` builtins for any GCC C++ build (previously only the Haiku path did). C and clang builds are unchanged.
- **`os/threads_posix.c`** — map C11 `_Thread_local` to `thread_local` under `__cplusplus` (g++ rejects `_Thread_local` in C++).
- **`os/linux_usbfs.c`** — explicit casts on `void*` allocations and `usbi_get_*_priv()` accessors; cast an `int` transfer status to its enum; reorder the `usbi_os_backend` initializer to the struct's declaration order (C++ requires designated initializers in order).
- **Windows backend** (`windows_common.{c,h}`, `windows_winusb.{c,h}`, `windows_usbdk.c`, `events_windows.c`) — explicit casts on `void*` allocations/accessors; compare GUIDs with `memcmp` (`IsEqualGUID` takes references in C++); one enum→`uint8_t` narrowing cast for `/WX`.

All casts are no-ops in C and preserve behavior.

## Verification (reproduced locally on Windows + WSL)

- **Linux / g++** (`-std=c++20`): the `.so` builds cleanly.
- **Windows / MSVC** (`/TP`, `CompileAs=CompileAsCpp`): the DLL builds cleanly (0 warnings under `/WX`).
- With #1782 applied, the full C++ build is **0 errors** on both compilers. Independently, on latest master + this PR (without #1782) the only remaining C++ errors are in #1782's files — confirming these fixes are complete and orthogonal.

The C++-build CI proposed in #1845 exercises exactly this.

Related to #1782
Related to #1845
Assisted-by: claude-code:claude-opus-4-8


libusb/os/windows_winusb.c
```diff
@@ -1950,7 +1950,7 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 						libusb_unref_device(dev);
 						goto alloc_device;
 					}
-					if (!IsEqualGUID(&priv->class_guid, &dev_info_data.ClassGuid)) {
+					if (memcmp(&priv->class_guid, &dev_info_data.ClassGuid, sizeof(GUID)) != 0) {
```
Youw: Why `IsEqualGUID` was replaced by `memcmp`?

libusb/os/threads_posix.c
```diff
@@ -81,9 +81,17 @@ int usbi_cond_timedwait(usbi_cond_t *cond,
 		return LIBUSB_ERROR_OTHER;
 }
 
+/* C++ spells C11's _Thread_local as the keyword thread_local; map to whichever
+ * the current compiler accepts so this file builds as both C and C++. */
```
Youw: double check if this wording is accudate - does libusb a 100% require C11? maybe it simply works on posix systems too? maybe simplify wording and only mention C++ vs current platform. (is this comment even needed?)

libusb/libusbi.h
```diff
@@ -104,11 +104,11 @@ typedef volatile LONG usbi_atomic_t;
...
-#if defined(__HAIKU__) && defined(__GNUC__) && !defined(__clang__)
-/* The Haiku port of libusb has some C++ files and GCC does not define
- * anything in stdatomic.h when compiled in C++11 (only in C++23).
- * This appears to be a bug in gcc's stdatomic.h, and should be fixed either
- * in gcc or in Haiku. Until then, use the gcc builtins. */
+#if defined(__GNUC__) && !defined(__clang__) && (defined(__cplusplus) || defined(__HAIKU__))
+/* GCC's <stdatomic.h> cannot be used when compiling as C++ before C++23:
+ * it relies on the C-only _Atomic keyword, so the header fails to parse.
+ * Fall back to the GCC __atomic builtins, which work in both C and C++.
+ * This also covers the Haiku port, which builds some files as C++ with GCC. */
```
Youw: - this looses the information about originla Haiku issue - need to preserver it
- the comment mentiones "before C++23" but doesn't really handle it; should we have a fallback but only for pre-C++23? need to check

libusb/os/windows_winusb.c
Youw: `IsEqualGUID` differs between C and C++ here. In C (`guiddef.h`) it's a macro that expands to `!memcmp(a, b, sizeof(GUID))` and takes pointers; in C++ the SDK provides `IsEqualGUID(const GUID&, const GUID&)`. So the existing `IsEqualGUID(&priv->class_guid, &dev_info_data.ClassGuid)` (passing `GUID*`) doesn't compile as C++. The `memcmp(...)` here is exactly what the C macro expands to — behavior-identical and valid in both C and C++.

If you'd rather keep `IsEqualGUID`, the alternative is an `#ifdef __cplusplus` that drops the `&` for the C++ path — happy to switch, but `memcmp` seemed cleaner.

libusb/os/threads_posix.c
Youw: Simplified in 2dcc3871 to `/* C uses _Thread_local; C++ uses the thread_local keyword. */`, dropping the C11 framing. (libusb does require C11 — `configure.ac` errors out otherwise — but the comment doesn't need to state that.)

libusb/libusbi.h
Youw: Reworded in 2dcc3871:
- Kept the Haiku origin: "This was first hit by the Haiku port, which builds some libusb files as C++ with GCC."
- On C++23: the GCC `__atomic` builtins work in C and every C++ version, so a C++23 build would just use them too — there's no need for a separate pre-/post-C++23 path. The comment now says that rather than implying a missing C++23 fallback.

libusb/os/windows_winusb.c
Youw: still want to use `IsEqualGUID` explicitly (with a helper function that covers C/C++ differences

Youw: Switched to your suggested approach in c3708add: added a `usbi_guid_equal(const GUID *, const GUID *)` helper in `windows_common.h` that wraps `IsEqualGUID` with the `#ifdef __cplusplus` inside (pointers in C, `*deref` to bind the `const GUID&` refs in C++), and reverted the call site from `memcmp` back to it. Verified it builds clean as both C and C++.

Naming note: I used the `usbi_` internal prefix (the convention across the tree) rather than `libusbi_` — happy to rename if you prefer.

sonatique (approved):

libusb/os/linux_usbfs.c
```diff
@@ -2836,13 +2836,13 @@ const struct usbi_os_backend usbi_backend = {
...
-	.get_config_descriptor = op_get_config_descriptor,
-	.get_config_descriptor_by_value = op_get_config_descriptor_by_value,
-
 	.wrap_sys_device = op_wrap_sys_device,
 	.open = op_open,
 	.close = op_close,
+
+	.get_active_config_descriptor = op_get_active_config_descriptor,
+	.get_config_descriptor = op_get_config_descriptor,
+	.get_config_descriptor_by_value = op_get_config_descriptor_by_value,
```
seanm, line 2847: Why did these move?

Youw: Order of members initialisation needs to be same as order of declaration in C++, otherwise that is a warning

Youw:
I don't see any active change requests on this - merging.

PR #1853, opened 2026-06-16
CI: Create MSI libusb runtime installers for Windows

antoniovazquezblanco:
In linux, LibUSB is usually packaged by the distros but this is not true for Windows. Also, there is no official LibUSB runtime installer packages for windows.

This leads to a lack of common practices for apps that use LibUSB on Windows...
For example, pyocd has created https://pypi.org/project/libusb-package/ as a way to ship the library in python.
For the pyusb or libusbdotnet packages it is difficult to locate the libusb library because of a lack of standardized path and also being unable to point to an official libusb runtime installation...
Some people opt for bundling the library but it is not that easy because of the different supported architectures and not being among the best practices in some package environments.

I think it would be positive to have an official runtime installation package for Windows that could be referenced whenever an application does not find the LibUSB library in the path. That would simplify things for the dependent applications. Also people may opt for bundling the libusb installer in their installers.

I hope this is useful.
Thanks!

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -41,6 +41,10 @@ jobs:
         run: |
           msbuild msvc/libusb.sln /m -property:Configuration=${{ matrix.configuration }} -property:Platform=${{ matrix.architecture }}
 
+      - name: Build MSI installer
+        if: startsWith(matrix.configuration, 'Release')
+        run: msbuild msvc/installer/libusb_installer.wixproj -restore -p:Platform=${{ matrix.architecture }} -p:LibusbDllConfiguration=${{ matrix.configuration }} -p:BuildLibusbDll=false
+
       - name: Upload artifacts
         uses: actions/upload-artifact@v7
         with:
@@ -52,3 +56,4 @@ jobs:
             build/**/dll/libusb-1.0.exp
             build/**/dll/libusb-1.0.pdb
             build/**/lib/libusb-1.0.lib
+            build/installer/**/*.msi
--- a/msvc/installer/libusb.wxs
+++ b/msvc/installer/libusb.wxs
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  WiX source for the libusb runtime MSI installer.
+
+  It installs the DLL under "Program Files\libusb\bin"
+  (or "Program Files (x86)\libusb\bin" for x86) and prepends
+  that directory to the system PATH.
+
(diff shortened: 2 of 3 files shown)


Youw:
Maybe this is a good start, but at least this has to be well documented with suggested use cases, compatibility conserns, etc.

Maybe we could use gstreamer installer as a model to repeat (they use a global envoronment variable in addition to registry keys).

I see a registry key being written but there is no option for "user-scope" installation as far as I can tell (not sure if it is needed, but it poped my mind instantly as not uncommon option on Windows).

mcuee:
FYI, I am not a fan of pyocd's libusb-package for pyusb myself as it is outdated and with limitations.
https://github.com/pyocd/libusb-package/
* https://github.com/pyusb/pyusb/issues/549

And I have removed the old outdated bundled libusb from libusbdotnet project.
* https://github.com/LibUsbDotNet/LibUsbDotNet/issues/301

But I think this one is in the right direction as it will come from the official libusb project.

antoniovazquezblanco:
> But I think this one is in the right direction as it will come from the official libusb project.

Nice to see positive feedback towards the idea. The other non official solutions were also problematic to me.


> I see a registry key being written but there is no option for "user-scope" installation as far as I can tell (not sure if it is needed, but it poped my mind instantly as not uncommon option on Windows).

I can work on this if this is the consensus. If I am not mistaken, you are talking about having an user-scoped installation that does not require privilege elevation vs a system-wide one with privilege requirements.


> Maybe this is a good start, but at least this has to be well documented with suggested use cases, compatibility conserns, etc.

How may I document the use cases? Any pointers?
Regarding compatibility concerns, you will be able to have a global installation of libusb (as you were able in the past), but this time assisted by the installer.
For any software packaging their own versions, loading precedence will favour same-folder libraries, keeping compat.
Do you have any compat concerns?

Thanks!

Youw:
> you are talking about having an user-scoped installation that does not require privilege elevation vs a system-wide one with privilege requirements

right, maybe having an installer option to select the installation method. not sure if this can be easily achieved with wix

> How may I document the use cases? Any pointers?

I guess we need to come up with a few specific use-cases and document them, so that others would use it exactly how the whole thing is designed. This would make it easier for both users and us, maintainers, later.

> Do you have any compat concerns?

No. Many project most likely will continue to use their own shipped versiosn, even if there will be an option to use system/user-installed version of the library. That's absolutely fine for as long as we do not try to replace their shipped version with the installed on in their runtime.

PR #1856, opened 2026-06-19
libusb_get_device_list returned incorrect old cached information on Windows

udoudou:
On Windows, for a libusb_device* pointer a (device A) that has already been opened or refed by libusb_open or libusb_ref_device, if device A is unplugged and plugged in again, using libusb_get_device_list to retrieve the device list will still return the old cached libusb_device* pointer a, which will still record the previous device address.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1294,6 +1294,46 @@ static bool get_dev_port_number(HDEVINFO dev_info, SP_DEVINFO_DATA *dev_info_dat
 			NULL, (PBYTE)port_nr, sizeof(*port_nr), &size) && (size == sizeof(*port_nr));
 }
 
+static bool get_dev_addr(struct libusb_device *parent_dev, uint8_t port_number, uint8_t *dev_addr)
+{
+	struct libusb_context *ctx = NULL;
+	struct winusb_device_priv *parent_priv;
+	USB_NODE_CONNECTION_INFORMATION_EX conn_info;
+	HANDLE hub_handle;
+	DWORD size;
+	if (parent_dev == NULL) {	// Root hub has no device address
+		return false;
+	}
+	ctx = DEVICE_CTX(parent_dev);
+	parent_priv = usbi_get_device_priv(parent_dev);
+	if (parent_priv->apib->id != USB_API_HUB) {
+		usbi_warn(ctx, "device '%s' is not a hub", parent_priv->dev_id);
+		return false;
+	}
+	// Even when the hub device is unplugged, it can still be opened and a handle can be obtained
+	// normally (tested on Win11 25H2).
+	hub_handle = CreateFileA(parent_priv->path, GENERIC_WRITE, FILE_SHARE_WRITE,
+							NULL, OPEN_EXISTING, 0, NULL);
+	if (hub_handle == INVALID_HANDLE_VALUE) {
+		usbi_warn(ctx, "could not open hub %s: %s", parent_priv->dev_id,
+				windows_error_str(0));
+		return false;
+	}
+	conn_info.ConnectionIndex = (ULONG)port_number;
+	// If no device is connected to the corresponding port (or the hub device does not exist), it
+	// will still return success, and DeviceAddress will be 0 (tested on Win11 25H2).
+	if (!DeviceIoControl(hub_handle, IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, &conn_info, sizeof(conn_info),
+		&conn_info, sizeof(conn_info), &size, NULL)) {
+		usbi_warn(ctx, "could not get node connection information for hub '%s' port %lu: %s",
+			parent_priv->dev_id, ULONG_CAST(port_number), windows_error_str(0));
+		CloseHandle(hub_handle);
+		return false;
+	}
+	*dev_addr = (uint8_t)conn_info.DeviceAddress;
+	CloseHandle(hub_handle);
+	return true;
+}
+
 static int enumerate_hcd_root_hub(struct libusb_context *ctx, const char *dev_id,
 	DEVINST devinst)
 {
@@ -1652,7 +1692,7 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 	char *dev_interface_path = NULL;
 	unsigned long session_id;
 	DWORD size, port_nr, install_state;
-	uint8_t bus_number = 0;
+	uint8_t bus_number = 0, device_address;
 #if defined(ENABLE_LOGGING)
 	char guid_string[MAX_GUID_STRING_LENGTH];
 #endif
@@ -1948,15 +1988,24 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 
 					if (strcmp(priv->dev_id, dev_id) != 0) {
 						usbi_dbg(ctx, "device instance ID for session [%lX] changed", session_id);
-						usbi_detach_device(dev); // usbi_detach_device is equivalent do usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
+						usbi_detach_device(dev); // usbi_detach_device is equivalent to usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
 						libusb_unref_device(dev);
 						goto alloc_device;
 					}
 					if (!IsEqualGUID(&priv->class_guid, &dev_info_data.ClassGuid)) {
 						usbi_dbg(ctx, "device class GUID for session [%lX] changed", session_id);
-						usbi_detach_device(dev); // usbi_detach_device is equivalent do usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
+						usbi_detach_device(dev); // usbi_detach_device is equivalent to usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
 						libusb_unref_device(dev);
 						goto alloc_device;
+					}
+					if (priv->initialized && parent_dev) {
+						device_address = 0;
+						if (get_dev_addr(parent_dev, dev->port_number, &device_address) && device_address != dev->device_address) {
+							usbi_dbg(ctx, "device address for session [%lX] changed", session_id);
+							usbi_detach_device(dev); // usbi_detach_device is equivalent to usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
+							libusb_unref_device(dev);
+							goto alloc_device;
+						}
 					}
 #if defined(LIBUSB_WINDOWS_HOTPLUG)
 					priv->seen_during_scan = true;


libusb/os/windows_winusb.c
```diff
@@ -1958,6 +1958,20 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
 						libusb_unref_device(dev);
 						goto alloc_device;
 					}
+					if (priv->initialized) {
```
Youw: **Root hubs churn here.** Root hubs reach `priv->initialized = true` (`:1230`) but never get a `device_address` assigned (the `:1208` assignment is only on the `parent_dev != NULL` path), so it stays `0`. This block then compares `0 != SPDRP_ADDRESS(non-zero)` and detaches + re-allocates every root hub on every enumeration. Gating on a non-zero address avoids that:

```suggestion
					if (priv->initialized && dev->device_address != 0) {
```

(This only fixes the root-hub churn — the `SPDRP_ADDRESS`-vs-`device_address` semantics issue below still stands.)

*— suggestion by Claude*

Youw (commented):
> **🤖 Automated review by Claude** — this review and its inline `suggestion` blocks were generated by Claude (`claude-opus-4-8`, via Claude Code) at a maintainer's request. Findings are advisory; please verify before acting. The suggestions are mechanical and have **not** been tested on hardware.

## Summary

The intent is right and the eviction plumbing is correct, but as written the fix likely doesn't do what it claims — it compares two different quantities — and it adds root-hub churn.

### 🔴 Blocker — `SPDRP_ADDRESS` is a *port number* (per this file's own docs), not the USB device address

The new check evicts the cached device when `dev->device_address != (uint8_t)SPDRP_ADDRESS`, but those are different values:

- `dev->device_address` is set at `windows_winusb.c:1208` from `conn_info.DeviceAddress` (`IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX` on the parent hub) — the **USB bus address** (1–127), reassigned on every enumeration. That is what goes stale.
- `SPDRP_ADDRESS` is what this file's own `get_dev_port_number()` uses as a **port-number** fallback (`:1288`): *"the address may be the port number, which is true for the Microsoft driver."*

Consequences:

- On a **same-port replug** — the exact scenario (same `DevInst`/`session_id`, so the existing `dev_id` check at `:1947` doesn't fire) — `SPDRP_ADDRESS` (the port) is unchanged, so the new check **never fires** and the stale device isn't evicted. The reported bug isn't actually fixed.
- Where a device's address ≠ its port number, the check can fire on **every scan** with no replug, churning the device and emitting spurious `DEVICE_LEFT`/arrived hotplug events.
- A *different*-port replug is already handled by the existing `dev_id` instance-ID check, so the new check is redundant there.

**Recommendation:** to detect a changed USB address, re-derive it from the same source the cache uses — `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX.DeviceAddress` — not `SPDRP_ADDRESS`. (Detail in the inline comment.)

### 🟠 Root hubs churn every scan

Root hubs get `priv->initialized = true` (`:1230`) but never get a `device_address` (the `:1208` assignment is only on the `parent_dev != NULL` path), so it stays `0`. The new block then compares `0 != SPDRP_ADDRESS(non-zero)` → detach + unref + re-allocate every root hub on every enumeration. The inline suggestion mitigates this.

### ✅ What's correct

- The `usbi_detach_device` + `libusb_unref_device` + `goto alloc_device` mechanism is refcount-balanced and memory-safe — a faithful copy of the existing `dev_id`/`class_guid` eviction blocks (no double-free / leak / use-after-free).
- The SetupAPI call is well-formed (NULL reg-type, buffer/length, `size == sizeof` short-read guard, safe `DWORD`→`uint8_t`).

### Nits

Missing space in `)){`; the "equivalent **do**" typo (copied from the existing blocks); the failure `usbi_warn` says "address" but reads the port property.

Inline comments carry one-click `suggestion`s for the mechanical items (root-hub guard, brace, typo). The blocker is a design issue and needs a rethink rather than a one-line patch.

*— Review generated by Claude (`claude-opus-4-8`)*

libusb/os/windows_winusb.c
```diff
@@ -1958,6 +1958,20 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
+						DWORD address;
+						if (pSetupDiGetDeviceRegistryPropertyA(*dev_info, &dev_info_data, SPDRP_ADDRESS,
+								NULL, (PBYTE)&address, sizeof(address), &size) && (size == sizeof(address))){
```
Youw: Nit: missing space before the brace (the rest of the file uses `) {`).

```suggestion
								NULL, (PBYTE)&address, sizeof(address), &size) && (size == sizeof(address))) {
```

*— suggestion by Claude*

libusb/os/windows_winusb.c
```diff
@@ -1958,6 +1958,20 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
+							if (dev->device_address != (uint8_t)address) {
```
Youw: **🔴 This compares the wrong quantity.** `dev->device_address` (set at `:1208` from `conn_info.DeviceAddress`) is the USB bus address; `SPDRP_ADDRESS` is documented in this very file (`get_dev_port_number`, `:1288`) as *"the port number… true for the Microsoft driver."* So:

- **same-port replug** (the targeted case): the port is unchanged → this never fires → the stale device isn't evicted → the bug isn't fixed;
- elsewhere it can fire on **every scan** with no replug → spurious `DEVICE_LEFT` events and churn.

To compare the real USB address, re-derive it the way the cache does — `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX` on the parent hub (`.DeviceAddress`, as at `:1146`–`1208`) — rather than `SPDRP_ADDRESS`. That needs the hub handle, so it's a design change rather than a one-line suggestion — flagging instead of suggesting.

*— review by Claude*

libusb/os/windows_winusb.c
```diff
@@ -1958,6 +1958,20 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
+								usbi_dbg(ctx, "device address for session [%lX] changed", session_id);
+								usbi_detach_device(dev); // usbi_detach_device is equivalent do usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
```
Youw: Nit: `equivalent do` → `equivalent to` (typo carried over from the `dev_id`/`class_guid` blocks at `:1949`/`:1955`).

```suggestion
								usbi_detach_device(dev); // usbi_detach_device is equivalent to usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
```

*— suggestion by Claude*

libusb/os/windows_winusb.c
udoudou: Yes. I only verified the problem scenarios I found, so I did not find this error. I have corrected this; please check again.

Youw: NOTE: I'll ask Claude to review again, but I'm not able to actually check this myself, so we're doing best theoretical effort here.

libusb/os/windows_winusb.c
```diff
@@ -1948,16 +1984,25 @@ static int winusb_get_device_list(struct libusb_context *ctx, struct discovered_
...
 					if (!IsEqualGUID(&priv->class_guid, &dev_info_data.ClassGuid)) {
 						usbi_dbg(ctx, "device class GUID for session [%lX] changed", session_id);
-						usbi_detach_device(dev); // usbi_detach_device is equivalent do usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
+						usbi_detach_device(dev); // usbi_detach_device is equivalent to usbi_disconnect_device but for the firing of LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT
 						libusb_unref_device(dev);
 						goto alloc_device;
 					}
+					if (priv->initialized && parent_dev) {
+						device_address = 0;
+						if (!get_dev_addr(parent_dev, dev->port_number, &device_address) || device_address != dev->device_address) {
```
Youw: **Transient failure → spurious eviction.** This treats *any* `get_dev_addr()` failure (parent-hub `CreateFileA` fails, or the IOCTL fails) the same as a genuine address change: detach (fires `LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT`) + re-allocate. But a device that's still listed in this enumeration is still present — a truly removed device wouldn't be enumerated here at all (and the `seen_during_scan` sweep reaps the unseen ones). So a momentary hub-open/contention failure would tear down a working device and emit a phantom `DEVICE_LEFT` then `DEVICE_ARRIVED` that applications act on.

Safer to evict only on a *positive* mismatch — if the address can't be read, treat it as "unknown → keep" rather than "changed → evict". This doesn't weaken the fix: a real same-port replug still makes `get_dev_addr()` succeed and return the new address, so the mismatch (and eviction) still happens.

```suggestion
						if (get_dev_addr(parent_dev, dev->port_number, &device_address) && device_address != dev->device_address) {
```

*— review by Claude*

libusb/os/windows_winusb.c
```diff
@@ -1294,6 +1294,42 @@ static bool get_dev_port_number(HDEVINFO dev_info, SP_DEVINFO_DATA *dev_info_dat
...
 }
 
+static bool get_dev_addr(struct libusb_device *parent_dev, uint8_t port_number, uint8_t *dev_addr)
+{
+	struct libusb_context *ctx = NULL;
+	struct winusb_device_priv *parent_priv;
+	USB_NODE_CONNECTION_INFORMATION_EX conn_info;
+	HANDLE hub_handle;
+	DWORD size;
+	if (parent_dev == NULL) {	//Root hub has no device address
```
Youw: Trivial nit: missing space after `//` (the file convention is `// `).

```suggestion
	if (parent_dev == NULL) {	// Root hub has no device address
```

*— suggestion by Claude*

Youw (commented):
> **🤖 Re-review by Claude** — this follow-up review and its inline `suggestion` block were generated by Claude (`claude-opus-4-8`, via Claude Code) at a maintainer's request. Findings are advisory and have not been hardware-tested.

Thanks for the rework — the new `get_dev_addr()` helper is the right fix. Re-reviewed against `d94da8b1`:

**✅ Resolved**
- **🔴 Blocker (comparing the wrong quantity)** — fixed. `get_dev_addr()` opens the parent hub and queries `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX`, comparing the real `conn_info.DeviceAddress` against the cached `dev->device_address` — i.e. the actual USB bus address the cache stored at allocation, so a same-port replug is now detected. This is the approach suggested in the original review.
- **Root-hub churn** — fixed. The check is gated on `parent_dev`, and `get_dev_addr()` returns `false` for a NULL parent, so root hubs (which never get a non-zero address) no longer detach/re-allocate on every scan.
- **Typo `equivalent do` → `to`** — fixed in all three blocks.
- **Brace-spacing nit** — obsolete (the `SPDRP_ADDRESS` line is gone).

The helper correctly mirrors the existing allocation-path IOCTL usage (`:1137`–`1152`); `parent_priv->apib` is always non-NULL (defaults to `USB_API_UNSUPPORTED`), so the `!= USB_API_HUB` guard is safe; and the block only runs in the HUB/GEN passes where the parent is always a hub, so that guard won't misfire.

**🟡 One new consideration (not a blocker)** — inline on `:1999`: treating *any* `get_dev_addr()` failure as "address changed" can evict a still-present device on a transient hub-open failure, firing spurious `DEVICE_LEFT`/`DEVICE_ARRIVED`. Suggest evicting only on a positive mismatch. Plus one trivial style nit inline.

libusb/os/windows_winusb.c
Youw: ✅ Resolved by `d94da8b1`. The new `get_dev_addr()` helper re-derives the real USB address via `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX` on the parent hub and compares it against the cached `dev->device_address` — exactly the bus address the cache stored at allocation, so a same-port replug is now detected. This is the approach suggested above; the fix is correct, so I'm resolving this thread.

One unrelated robustness consideration on the new code (transient `get_dev_addr()` failure → spurious eviction) is posted as a fresh review comment on `:1999`.

*— Claude (claude-opus-4-8), via Claude Code*

libusb/os/windows_winusb.c
udoudou: My initial concern was that if the corresponding hub device was lost or the corresponding port had no device, the address information might never be successfully obtained. Therefore, I intentionally treated the failure as an address change. However, I just tested this on my local Windows 11 machine, and the Windows API implementation seems to be independent of the presence of the hub device or port device; the API still returns success normally, only the DeviceAddress will be 0. Therefore, my concern was unnecessary; the check should only be performed upon successful acquisition.

Youw (commented):
> **🤖 Final review round by Claude** — generated by Claude (`claude-opus-4-8`, via Claude Code) at a maintainer's request. Advisory and not hardware-tested on my end; @udoudou verified the Windows behavior on Win11 25H2.

Re-reviewed against `72449a57`. **Everything raised across the previous rounds is now addressed, and the implementation looks correct.**

**Resolved this round**
- **Transient-failure churn (`:1999`)** — now evicts only on a *positive* mismatch (`get_dev_addr(...) && device_address != dev->device_address`). @udoudou's Win11 25H2 testing shows this is also *more* correct than I'd assumed: a lost hub / empty port doesn't make the IOCTL fail — it returns success with `DeviceAddress == 0`. So the two cases separate cleanly:
  - device genuinely gone → `get_dev_addr()` succeeds with addr `0` → `0 != cached_addr` → evicted ✅ (cached non-root devices always have a non-zero address, so removal is still reliably detected);
  - transient hub-open/contention failure → `get_dev_addr()` returns `false` → device kept, no phantom `DEVICE_LEFT` ✅.

  The two `// ... (tested on Win11 25H2)` comments documenting this are a nice touch.
- **`//` comment spacing (`:1304`)** — fixed.

**From earlier rounds (still good)**
- 🔴 wrong-quantity blocker → fixed via the parent-hub `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX`;
- root-hub churn → gated on `parent_dev`;
- `equivalent do`→`to` typos / brace nit → fixed / obsolete.

**No new issues.** One inherent, accepted trade-off worth stating once: each cached non-root device now costs a hub-open + IOCTL per `libusb_get_device_list()` scan (previously just string compares). That's the necessary price of detecting a same-port replug and it mirrors the allocation path's existing per-device cost, so it's fine.

LGTM on the logic. Since I can't exercise the hardware, I'd defer the final functional sign-off to your Win11 testing — but from a code-correctness standpoint this is good to merge.

*— review by Claude (claude-opus-4-8), via Claude Code*

PR #1861, opened 2026-06-26
hotplug: fix callback use-after-free races

### LINKED ISSUE #1859: Two use-after-free races in `libusb/hotplug.c` on `ctx->hotplug_cbs` (unlocked free in `usbi_hotplug_exit`, unlocked access in the ENUMERATE loop)
While looking at the `hotplug_cbs` locking I came across two spots in `libusb/hotplug.c` where a `struct usbi_hotplug_callback` (which holds the function pointer `cb`) can be freed while another thread is still using it. Both are reachable from plain multi-threaded use of the exported API. Confirmed with ASan on real USB devices, tested against `19afc23a`.

They're in the same file, a few lines apart, and fall out of the same `hotplug_cbs` locking protocol, so reporting them together.

---

## Bug A: `usbi_hotplug_exit` frees `hotplug_cbs` without `hotplug_cbs_lock`

`libusb/hotplug.c`, `usbi_hotplug_exit`, lines 200-204:

```c
	/* free all registered hotplug callbacks */
	for_each_hotplug_cb_safe(ctx, hotplug_cb, next_cb) {
		list_del(&hotplug_cb->list);
		free(hotplug_cb);
	}
```

This is the only accessor of `ctx->hotplug_cbs` that does not take `ctx->hotplug_cbs_lock`. Every other one does: `usbi_hotplug_process` (301), `libusb_hotplug_register_callback` (388), `libusb_hotplug_deregister_callback` (443), `libusb_hotplug_get_user_data` (481). The sibling loop just below it (freeing discovered devices, line 219) got `usb_devs_lock` added in `6c0ae1ab` / #1452 — but that fix left the `hotplug_cbs` loop above it unlocked.

So a second thread in `libusb_exit` → `usbi_hotplug_exit` can `free(hotplug_cb)` at line 203 while another thread is mid-iteration over `hotplug_cbs` (the event thread in `usbi_hotplug_process`, which releases the lock across the user callback at 312-314, or — as the PoC drives it — the registration thread's ENUMERATE loop, see Bug B).

**Reproducer**

```c
#include <pthread.h>
#include <unistd.h>
#include "libusb.h"

static libusb_context *ctx;
static volatile int go;

static int LIBUSB_CALL cb(libusb_context *c, libusb_device *d,
                          libusb_hotplug_event e, void *u) {
	go = 1;                 /* inside the unlocked ENUMERATE loop */
	usleep(300000);
	return 0;
}

static void *exit_thread(void *_) {
	while (!go) usleep(1000);
	libusb_exit(ctx);       /* -> usbi_hotplug_exit -> hotplug.c:203 free, no lock */
	return _;
}

int main(void) {
	pthread_t t;
	libusb_hotplug_callback_handle h;
	libusb_init_context(&ctx, NULL, 0);
	pthread_create(&t, NULL, exit_thread, NULL);
	libusb_hotplug_register_callback(ctx, LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED,
		LIBUSB_HOTPLUG_ENUMERATE, LIBUSB_HOTPLUG_MATCH_ANY,
		LIBUSB_HOTPLUG_MATCH_ANY, LIBUSB_HOTPLUG_MATCH_ANY, cb, NULL, &h);
	pthread_join(t, NULL);
	return 0;
}
```

Build — first an ASan libusb, then the reproducer against it (from the repo root):

```sh
$ ./autogen.sh && make -C libusb CFLAGS="-g -O1 -fsanitize=address,undefined"
$ gcc -I libusb -include libusb/config.h -fsanitize=address,undefined -g -O1 -pthread \
      poc_hotplug_exit_uaf.c -o poc_hotplug_exit_uaf \
      -L libusb/.libs -Wl,-rpath,$(pwd)/libusb/.libs -lusb-1.0
$ ASAN_OPTIONS=halt_on_error=1:detect_leaks=0 ./poc_hotplug_exit_uaf
```

Needs at least one USB device present (the ENUMERATE loop must run at least one iteration); real or otherwise.

**ASan:**

```
==ERROR: AddressSanitizer: heap-use-after-free
READ of size 1 at ... thread T0
    #0 usbi_hotplug_match_cb            hotplug.c:241
    #1 libusb_hotplug_register_callback hotplug.c:415   // ENUMERATE loop, no lock
freed by thread T2 here:
    #1 usbi_hotplug_exit                hotplug.c:203   // unlocked free (this bug)
    #2 libusb_exit                      core.c:2779
    #1 libusb_hotplug_register_callback hotplug.c:368
```

**Fix:**

```diff
 	/* free all registered hotplug callbacks */
+	usbi_mutex_lock(&ctx->hotplug_cbs_lock);
 	for_each_hotplug_cb_safe(ctx, hotplug_cb, next_cb) {
 		list_del(&hotplug_cb->list);
 		free(hotplug_cb);
 	}
+	usbi_mutex_unlock(&ctx->hotplug_cbs_lock);
```

---

## Bug B: ENUMERATE for-loop accesses `hotplug_cb` unlocked and ignores `NEEDS_FREE`

`libusb/hotplug.c`, `libusb_hotplug_register_callback`, lines 414-418:

```c
		for (i = 0; i < len; i++) {
			usbi_hotplug_match_cb(devs[i],
					LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED,
					hotplug_cb);
		}
```

After the callback is published on `hotplug_cbs` under the lock (397) the lock is dropped (399). This loop then calls `usbi_hotplug_match_cb(..., hotplug_cb)` — reading `hotplug_cb->flags` and invoking the function pointer `hotplug_cb->cb(...)` (260) — **without the lock, without checking `USBI_HOTPLUG_NEEDS_FREE`, without a safe iterator.**

`libusb.h` (line 2407) documents that calling `libusb_hotplug_deregister_callback()` from within a callback is safe, and returning `1` deregisters. But the event thread's `usbi_hotplug_process` honors a deregister by freeing the cb through its **locked** cleanup (317-318 / 336-337) while the registration thread is still iterating it unlocked. The next iteration touches freed memory. The dispatch loop (309) and the cleanup loop (333) both already guard with the `NEEDS_FREE` check; the ENUMERATE loop is the only caller of `usbi_hotplug_match_cb` that doesn't.

**Reproducer**

```c
#include <pthread.h>
#include <unistd.h>
#include "libusb.h"

static libusb_context *ctx;
static volatile int done;

static void *event_thread(void *_) {
	struct timeval tv = {0, 0};
	while (!done)
		libusb_handle_events_timeout_completed(ctx, &tv, NULL);
	return _;
}

static int LIBUSB_CALL cb(libusb_context *c, libusb_device *d,
                          libusb_hotplug_event e, void *u) {
	libusb_hotplug_deregister_callback(c, 1);   /* set NEEDS_FREE, signal event */
	usleep(200000);                             /* let event thread free us */
	done = 1;
	return 0;
}

int main(void) {
	pthread_t t;
	libusb_hotplug_callback_handle h;
	libusb_init_context(&ctx, NULL, 0);
	pthread_create(&t, NULL, event_thread, NULL);
	usleep(50000);
	libusb_hotplug_register_callback(ctx, LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED,
		LIBUSB_HOTPLUG_ENUMERATE, LIBUSB_HOTPLUG_MATCH_ANY,
		LIBUSB_HOTPLUG_MATCH_ANY, LIBUSB_HOTPLUG_MATCH_ANY, cb, NULL, &h);
	pthread_join(t, NULL);
	libusb_exit(ctx);
	return 0;
}
```

Build as in Bug A (same ASan libusb + same `gcc` line, swapping the source name).

**ASan:**

```
==ERROR: AddressSanitizer: heap-use-after-free
READ of size 1 at ... thread T0
    #0 usbi_hotplug_match_cb            hotplug.c:241
    #1 libusb_hotplug_register_callback hotplug.c:415   // ENUMERATE loop
freed by thread T2 here:
    #1 usbi_hotplug_process              hotplug.c:337   // cleanup free (locked)
    #2 handle_event_trigger              io.c:2187
    #3 libusb_handle_events_timeout_completed io.c:2376
    #1 libusb_hotplug_register_callback hotplug.c:368
```

**Fix** (mirror the dispatch loop's existing check at 309):

```diff
 		for (i = 0; i < len; i++) {
+			usbi_mutex_lock(&ctx->hotplug_cbs_lock);
+			if (hotplug_cb->flags & USBI_HOTPLUG_NEEDS_FREE) {
+				usbi_mutex_unlock(&ctx->hotplug_cbs_lock);
+				break;
+			}
+			usbi_mutex_unlock(&ctx->hotplug_cbs_lock);
 			usbi_hotplug_match_cb(devs[i],
 					LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED,
 					hotplug_cb);
 		}
```

---

Both fixes are a few lines each. I can open a PR with both plus the two reproducers.

smaeljaish771:
usbi_hotplug_exit() freed the hotplug callback list without holding hotplug_cbs_lock, while every other path operating on that list already takes the lock.

The LIBUSB_HOTPLUG_ENUMERATE path also kept using the registered callback object after dropping the lock. If another thread deregistered and freed that callback, the enumerate loop could read freed memory. Keep the handle locally, look up the callback under the lock for each enumerated device, copy the callback data to the stack, and invoke the user callback from that copy after unlocking.

Reformulating the hotplug iteration macros for Clang thread-safety analysis is left out, as it is an independent change.

Fixes: #1859

### DIFF
--- a/libusb/hotplug.c
+++ b/libusb/hotplug.c
@@ -53,7 +53,8 @@
  * libusb_hotplug_register_callback() because of the \ref LIBUSB_HOTPLUG_ENUMERATE
  * flag, the callback return value is ignored. In other words, you cannot cause a
  * callback to be deregistered by returning 1 when it is called from
- * libusb_hotplug_register_callback().
+ * libusb_hotplug_register_callback(). Deregistering the callback from within
+ * such a callback stops further enumeration.
  *
  * Callbacks for a particular context are automatically deregistered by libusb_exit().
  *
@@ -417,13 +418,14 @@ int API_EXPORTED libusb_hotplug_register_callback(libusb_context *ctx,
 
 		for (i = 0; i < len; i++) {
 			struct usbi_hotplug_callback hotplug_cb_copy;
+			struct usbi_hotplug_callback *matching_cb;
 			int found = 0;
 
 			usbi_mutex_lock(&ctx->hotplug_cbs_lock);
-			for_each_hotplug_cb(ctx, hotplug_cb) {
-				if (handle == hotplug_cb->handle) {
-					if (!(hotplug_cb->flags & USBI_HOTPLUG_NEEDS_FREE)) {
-						hotplug_cb_copy = *hotplug_cb;
+			for_each_hotplug_cb(ctx, matching_cb) {
+				if (handle == matching_cb->handle) {
+					if (!(matching_cb->flags & USBI_HOTPLUG_NEEDS_FREE)) {
+						hotplug_cb_copy = *matching_cb;
 						found = 1;
 					}
 					break;
--- a/libusb/hotplug.c
+++ b/libusb/hotplug.c
@@ -198,10 +198,12 @@ void usbi_hotplug_exit(struct libusb_context *ctx)
 		return;
 
 	/* free all registered hotplug callbacks */
+	usbi_mutex_lock(&ctx->hotplug_cbs_lock);
 	for_each_hotplug_cb_safe(ctx, hotplug_cb, next_cb) {
 		list_del(&hotplug_cb->list);
 		free(hotplug_cb);
 	}
+	usbi_mutex_unlock(&ctx->hotplug_cbs_lock);
 
 	/* free all pending hotplug messages */
 	while (!list_empty(&ctx->hotplug_msgs)) {
@@ -348,6 +350,7 @@ int API_EXPORTED libusb_hotplug_register_callback(libusb_context *ctx,
 	libusb_hotplug_callback_handle *callback_handle)
 {
 	struct usbi_hotplug_callback *hotplug_cb;
+	libusb_hotplug_callback_handle handle;
 
 	/* check for sane values */
 	if (!events || (~VALID_HOTPLUG_EVENTS & events) ||
@@ -389,6 +392,7 @@ int API_EXPORTED libusb_hotplug_register_callback(libusb_context *ctx,
 
 	/* protect the handle by the context hotplug lock */
 	hotplug_cb->handle = ctx->next_hotplug_cb_handle++;
+	handle = hotplug_cb->handle;
 
 	/* handle the unlikely case of overflow */
 	if (ctx->next_hotplug_cb_handle < 0)
@@ -399,7 +403,7 @@ int API_EXPORTED libusb_hotplug_register_callback(libusb_context *ctx,
 	usbi_mutex_unlock(&ctx->hotplug_cbs_lock);
 
 	usbi_dbg(ctx, "new hotplug cb %p with handle %d",
-		 (void *) hotplug_cb, hotplug_cb->handle);
+		 (void *) hotplug_cb, handle);
 
 	if ((flags & LIBUSB_HOTPLUG_ENUMERATE) && (events & LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED)) {
 		ssize_t i, len;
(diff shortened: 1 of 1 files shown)


libusb/hotplug.c
```diff
@@ -399,29 +403,47 @@ int API_EXPORTED libusb_hotplug_register_callback(libusb_context *ctx,
...
+			libusb_hotplug_deregister_callback(ctx, handle);
 			return (int)len;
 		}
 
 		for (i = 0; i < len; i++) {
+			struct usbi_hotplug_callback hotplug_cb_copy;
+			int found = 0;
+
+			usbi_mutex_lock(&ctx->hotplug_cbs_lock);
+			for_each_hotplug_cb(ctx, hotplug_cb) {
```
sonatique: the lookup reuses the outer `hotplug_cb` as its iterator. It's safe  (unused afterward), but maybe not explicit enough when reading the code, and maybe not super-safe against refactoring.
What about using a dedicated value?

sonatique (approved):
Thanks for turning this around so quickly. It looks to me that both races are correctly addressed, and I'm happy with the approach. You even went further than the patch sketched for issue B, if I understand correctly.

I noticed that with your code, calling `libusb_hotplug_deregister_callback()` from *within* an ENUMERATE callback now stops  enumeration at the next device. whereas before it kept firing for the remaining devices. I think this is more  correct, but maybe it worth documenting it.

libusb/hotplug.c
smaeljaish771: Done. Documented the ENUMERATE deregistration behavior and used a dedicated iterator.

sonatique:
Thanks @smaeljaish771 , all good to me.
Would be nice to have a second approval before merging.

seanm:
> Thanks @smaeljaish771 , all good to me. Would be nice to have a second approval before merging.

I should have time this week to look this over too...

mcuee:
Ping @seanm

seanm:
@smaeljaish771 please rebase and resolve conflicts, and I'll then review.

sonatique:
@smaeljaish771 : would like someone else to do the rebase?

PR #1863, opened 2026-07-02
windows: fix hotplug monitor restart, and register the window class per libusb copy

### LINKED ISSUE #1862: windows_hotplug does not seem to work after deregistering and registering callback a second time
In RPCS3 on Windows, we have a class that handles usb passthrough (see [sys_usbd](https://github.com/RPCS3/rpcs3/blob/6a2ad0a0aa17c994ae95f492bfd27a2a2b3c2b0f/rpcs3/Emu/Cell/lv2/sys_usbd.cpp))
In this class, we call `libusb_hotplug_register_callback` and eventually `libusb_hotplug_deregister_callback` when closing a game.
When initiating the thing again (booting the same game again in the emulator), hotplug events are not received anymore (see "`hotplug_callback`").

We only added hotplug for windows recently, before that, we periodically enumerated devices.
To my knowledge the same code works on linux and is only broken on windows.

We use libusb v1.0.30, but the same happens when I compile with the latest master commit.

sonatique:
The first commit is the actual fix for #1862: it unregisters the window class when the event thread ends, makes windows_start_event_monitor wait until the notification window exists (so a bring-up failure now propagates out of libusb_init instead of being lost), and resets the monitor state on stop. This alone fully restores hotplug across restart.

Second commit is an independent robustness improvement, not required to close the issue. The class was registered against the executable's module handle, so multiple independent copies of libusb in one process (e.g. one linked statically and one bundled as a DLL) collide on the shared class name; keying it to each copy's own module handle lets them coexist. It also stops registering a class whose window procedure lives in the libusb DLL against the executable.

Note: I made a littel reproducer program that I didn't ship with this PR, tell me if I should.

Fixes: #1862

### DIFF
--- a/libusb/os/windows_hotplug.c
+++ b/libusb/os/windows_hotplug.c
@@ -58,50 +58,105 @@ static LRESULT CALLBACK windows_proc_callback(HWND hwnd, UINT message, WPARAM wP
 
 int windows_start_event_monitor(void)
 {
+	// Signaled by the event thread once its notification window exists, so
+	// that a failure to bring up the monitor is reported to the caller.
+	// Owned by this function; the thread only receives it to signal it,
+	// and must not use it after this function returns.
+	HANDLE window_ready = CreateEvent(NULL, TRUE, FALSE, NULL);
+	if (window_ready == NULL)
+	{
+		log_error("CreateEvent");
+		return LIBUSB_ERROR_OTHER;
+	}
+
+	DWORD thread_id;
+
 	windows_event_thread_handle = CreateThread(
 		NULL, // Default security descriptor
 		0, // Default stack size
 		windows_event_thread_main,
-		NULL, // No parameters to pass to the thread
+		window_ready, // The event the thread signals once its window exists
 		0, // Start immediately
-		NULL // No need to keep track of thread ID
+		&thread_id // Used to reap the thread if waiting for readiness fails
 	);
 
 	if (windows_event_thread_handle == NULL)
 	{
 		log_error("CreateThread");
+		CloseHandle(window_ready);
 		return LIBUSB_ERROR_OTHER;
 	}
 
-	return LIBUSB_SUCCESS;
+	// Wait until the notification window exists or the thread failed and exited
+	HANDLE handles[2] = { window_ready, windows_event_thread_handle };
+	int r = LIBUSB_ERROR_OTHER;
+
+	switch (WaitForMultipleObjects(2, handles, FALSE, INFINITE))
+	{
+	case WAIT_OBJECT_0:
+		r = LIBUSB_SUCCESS;
+		break;
+	case WAIT_OBJECT_0 + 1:
+		usbi_err(NULL, "event thread exited before its notification window was created");
+		break;
+	default:
+		log_error("WaitForMultipleObjects");
+		// Should not happen with two valid handles, but do not abandon a
+		// live thread: a queued WM_QUIT ends its message loop even if it
+		// is posted before the loop starts. PostThreadMessage fails until
+		// the thread has a message queue, so retry while it is running.
+		while (!PostThreadMessage(thread_id, WM_QUIT, 0, 0))
+		{
+			if (WaitForSingleObject(windows_event_thread_handle, 10) != WAIT_TIMEOUT)
+			{
+				break; // The thread exited on its own
+			}
+		}
+		WaitForSingleObject(windows_event_thread_handle, INFINITE);
+		break;
+	}
+
+	if (r != LIBUSB_SUCCESS)
+	{
+		CloseHandle(windows_event_thread_handle);
+		windows_event_thread_handle = NULL;
+		windows_event_hwnd = NULL; // May have been set before the thread was reaped
+	}
+
+	CloseHandle(window_ready);
+
+	return r;
 }
 
 int windows_stop_event_monitor(void)
 {
-	if (windows_event_hwnd == NULL)
+	if (windows_event_thread_handle == NULL)
 	{
+		// The event monitor is not running, there is nothing to stop
 		return LIBUSB_SUCCESS;
 	}
 
-	if (!SUCCEEDED(SendMessage(windows_event_hwnd, WM_CLOSE, 0, 0)))
-	{
-		log_error("SendMessage");
-		return LIBUSB_ERROR_OTHER;
-	}
+	// The WM_CLOSE handler destroys the notification window, which in turn
+	// posts WM_QUIT and thereby ends the event thread's message loop.
+	SendMessage(windows_event_hwnd, WM_CLOSE, 0, 0);
 
+	int r = LIBUSB_SUCCESS;
 	if (WaitForSingleObject(windows_event_thread_handle, INFINITE) != WAIT_OBJECT_0)
 	{
 		log_error("WaitForSingleObject");
-		return LIBUSB_ERROR_OTHER;
+		r = LIBUSB_ERROR_OTHER;
 	}
 
 	if (!CloseHandle(windows_event_thread_handle))
 	{
 		log_error("CloseHandle");
-		return LIBUSB_ERROR_OTHER;
+		r = LIBUSB_ERROR_OTHER;
 	}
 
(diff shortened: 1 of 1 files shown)


Megamouse:
It seems to work now.
The callback is being called when un/plugging my usb device even after booting a game multiple times in one session.

sonatique:
@Megamouse , great, thank you for the feedback!
@mcuee @seanm @Youw @tormodvolden : waiting for your review / approvals, thanks!

libusb/os/windows_hotplug.c
```diff
@@ -58,50 +58,90 @@ static LRESULT CALLBACK windows_proc_callback(HWND hwnd, UINT message, WPARAM wP
...
+	switch (WaitForMultipleObjects(2, handles, FALSE, INFINITE))
+	{
+	case WAIT_OBJECT_0:
+		r = LIBUSB_SUCCESS;
+		break;
+	case WAIT_OBJECT_0 + 1:
+		usbi_err(NULL, "event thread exited before its notification window was created");
+		break;
+	default:
+		log_error("WaitForMultipleObjects");
```
Youw, line 103: **Minor (unreachable in practice):** the `default:` case here is only `WAIT_FAILED`, which can't realistically happen with two valid handles — but if it ever did, this branch abandons a *live* event thread while NULLing `windows_event_thread_handle`:
- `windows_stop_event_monitor` would then no-op, leaking the running thread and its window;
- a subsequent `libusb_init` would start a second event thread whose `RegisterClass` hits the reuse path, giving two windows on two threads racing the shared `first_debounce_tick`;
- `CloseHandle(window_ready)` below can race the live thread's upcoming `SetEvent(window_ready)` — a signal on a closed (or recycled) handle.

If you want to harden it cheaply: capture the thread id from `CreateThread`, and in this branch `PostThreadMessage(tid, WM_QUIT, 0, 0)` + `WaitForSingleObject(thread, INFINITE)` before the cleanup below — that reaps the thread cleanly whether it is before or after `CreateWindow` (a queued `WM_QUIT` ends the NULL-filter loop even if posted before the loop starts). Also fine to just leave it as is given the branch is effectively dead — your call.

*— review by Claude*

libusb/os/windows_hotplug.c
```diff
@@ -225,35 +303,49 @@ static DWORD WINAPI windows_event_thread_main(LPVOID lpParam)
...
+
 	MSG msg;
 	BOOL ret_val;
 
-	while ((ret_val = GetMessage(&msg, windows_event_hwnd, 0, 0)) != 0)
+	// Note: no window handle filter here. GetMessage retrieves WM_QUIT
+	// regardless of any filter, but once the notification window has been
+	// destroyed a window handle filter would reference an invalid window,
+	// and clean termination would then rely on unspecified ordering inside
+	// GetMessage. With a NULL filter the loop simply ends on WM_QUIT.
```
Youw: Nit on the comment: the documented guarantee (“GetMessage always retrieves WM_QUIT messages no matter which values you specify”) is about `wMsgFilterMin`/`wMsgFilterMax` — the message-*range* filter — not the window-handle filter, so “regardless of any filter” slightly over-claims it. The conclusion (NULL filter) is exactly right; suggested tightening:

```suggestion
	// Note: no window handle filter here. The documentation only guarantees
	// that WM_QUIT is retrieved regardless of the message-range filter; with
	// a window handle filter, clean termination would instead depend on
	// GetMessage failing with -1 once the notification window (the filter)
	// has been destroyed. With a NULL filter the loop simply ends on WM_QUIT.
```

*— suggestion by Claude*

Youw (commented):
> **🤖 Automated review by Claude** — this review and its inline `suggestion` block were generated by Claude (model `claude-fable-5`, via Claude Code) at a maintainer's request. Findings are advisory and have not been hardware-tested by the reviewer.

## Summary
Both commits are correct and carefully reasoned — nice work, including the commit messages. The root-cause chain of #1862 (window class survives thread exit → `RegisterClass` fails `ERROR_CLASS_ALREADY_EXISTS` on restart → event thread dies silently because `windows_start_event_monitor` never waited on bring-up) is real, and the fix addresses every link: unregister on thread end (and on the `CreateWindow` failure path), propagate bring-up failure out of `libusb_init`, and reset the monitor state in stop.

## Verified
- **Lifecycle**: start/stop run only on first-init/last-exit (`windows_common.c` `init_count`, serialized by core), so the start→stop→start sequencing the fix relies on holds. If start fails, `windows_event_thread_handle` is NULLed, so a later `windows_stop_event_monitor` correctly no-ops.
- **Synchronization**: `windows_event_hwnd` is written before `SetEvent(window_ready)` and only read by other threads after the wait in start returns — proper happens-before. The `window_ready` ownership contract (thread must not touch it after signaling) is documented at both ends and upheld on the success and thread-exit paths.
- **`get_hotplug_instance()` caching**: all callers run on the event thread, and successive event threads are serialized by the join in stop, so the unsynchronized `static` is safe as the comment claims. `GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS` on a file-scope static is the standard way to key the class to this libusb copy, and `UNCHANGED_REFCOUNT` is right here.
- **`ERROR_CLASS_ALREADY_EXISTS` reuse**: sound *after* per-module keying — within one module the leftover class necessarily points at this copy's `windows_proc_callback`. It also doubles as recovery for the abnormal-exit case where `UnregisterClass` failed because the window still existed (GetMessage `-1` break), which is a nice property.
- **NULL `GetMessage` filter**: the right fix. With the old `windows_event_hwnd` filter the loop in practice only terminated via the `-1`/invalid-handle error path after the window was destroyed; NULL filter + `WM_QUIT` is the idiomatic clean exit, and no other windows exist on that thread. Dropping the `SUCCEEDED()` checks on `SendMessage`/`TranslateMessage`/`DispatchMessage` is also correct — those return `LRESULT`/`BOOL`, not `HRESULT`s, so the old checks were no-ops at best.

## Minor notes (inline)
1. The `WAIT_FAILED` branch in `windows_start_event_monitor` abandons a potentially live event thread — unreachable in practice, but see inline note for a cheap hardening option.
2. The new comment above the message loop slightly over-claims what the docs guarantee about `WM_QUIT` and filters — inline `suggestion` with tightened wording.

## Noticed in passing (pre-existing, out of scope)
In `windows_common.c`, when `windows_start_event_monitor()` fails, `windows_init`'s `init_exit` path cleans up backends/htab but not the just-created per-context IOCP port and `windows_iocp_thread` — the thread blocks forever in `GetQueuedCompletionStatus` and core frees the ctx it holds. Pre-existing (was reachable via `CreateThread` failure), but this PR makes start failure a realistic path, so it may deserve a follow-up.

## Tests
Re your reproducer question: with failure propagation in place, `tests/stress.c` (repeated `libusb_init_context`/`libusb_exit`) now catches the #1862 class-leak regression on any build with `LIBUSB_WINDOWS_HOTPLUG` defined (the MSVC props enable it unconditionally) — the second init would fail loudly. So I don't think the reproducer needs to ship in the tree; attaching it to #1862 for posterity would still be useful. If you want in-tree coverage of the original user-level symptom, a small register→deregister→register hotplug-callback cycle in `tests/` would be the closest match, but I'd call that optional.

Youw (approved):

libusb/os/windows_hotplug.c
sonatique: Hardened as suggested: the thread id is now captured and this branch posts WM_QUIT (retrying while the thread is still running, since PostThreadMessage fails until the thread has a message queue) and then joins before the cleanup. The failure cleanup also resets windows_event_hwnd, which may have been set by then. 
Note: if the window already exists when the reap fires, the thread's UnregisterClass fails (logged) and the hidden window leaks. This seemed proportionate for a branch requiring WaitForMultipleObjects to fail on two valid handles.

libusb/os/windows_hotplug.c
sonatique: Good catch! Applied your wording verbatim, and fixed the same over-claim in the commit message.

Youw (approved):
This now looks good

Issue #1869, opened 2026-07-11
darwin: stress_mt hangs macOS CI (~20% of runs) — event-thread startup/shutdown handshake can wait forever


Youw:
> **🤖 Analysis by Claude** — this investigation and write-up were generated by Claude (model `claude-fable-5`, via Claude Code) at a maintainer's request. Findings are from code analysis and CI-run forensics; no macOS hardware was available to reproduce locally.

> [!IMPORTANT]
> **Update (2026-07-12):** the "unlatched startup/shutdown handshake" hypothesis below, and the "CI runners are device-less" assumption it partly rests on, were **both wrong** — see the [follow-up comment](https://github.com/libusb/libusb/issues/1869#issuecomment-4950063680) for a real captured stack trace from a reproduced CI hang. The event thread turns out to be healthy; all 8 worker threads are wedged waiting on a lost transfer completion, matching @mcuee's TSan-confirmed `open_count`/`capture_count` race (#1798 item 1, fixed by #1864). Kept below for the record of how the investigation got there.

## Symptom

`stress_mt` hangs forever on the macOS GitHub CI, the job dies at the 6-hour GitHub timeout as `cancelled`, and all evidence is lost (automake buffers the test's output into `test-suite.log`, which is never shown).

Sampled forensics from [run 28896022560](https://github.com/libusb/libusb/actions/runs/28896022560) (branch `conversion-warnings`, 2026-07-08):

```
00:19:00  Making check in tests
00:19:12  PASS: stress
          (silence for 5h58m)
06:17:52  ##[error]The operation was canceled.
06:17:54  Terminate orphan process: pid (13341) (stress_mt)
```

**Frequency: of the last 100 macOS workflow runs, 21 were cancelled** — every one I sampled ran exactly ~6h before being killed, i.e. they are all this hang, including runs on `master` itself ([28894513259](https://github.com/libusb/libusb/actions/runs/28894513259), [28881034605](https://github.com/libusb/libusb/actions/runs/28881034605), [28819210881](https://github.com/libusb/libusb/actions/runs/28819210881)). Hangs were reported as far back as May 2024 in #1494, and the rate is essentially unchanged before/after #1780's merge (2026-07-02): cancelled runs appear continuously through April–July 2026 on both sides of that date.

## Constraint analysis — where the hang can and cannot be

[...]

PR #1864, opened 2026-07-03
darwin: serialize open/capture state (supersedes #1679)

sonatique:
This PR fixes items 1 and 2 of the concurrency audit in #1798, superseding #1679 per
[@osy's suggestion](https://github.com/libusb/libusb/pull/1679#issuecomment-4233506257) that a new PR take over the stalled one.

**Item 1: unsynchronized `open_count` / `capture_count`.** Both counters on `struct darwin_cached_device` were read-modify-written with no synchronization from `darwin_open`, `darwin_close`, `darwin_restore_state`, `darwin_reset_device`, `darwin_detach_kernel_driver`, `darwin_attach_kernel_driver` and `darwin_capture_release_interface`. Concurrent open/close/detach on different handles to the same physical device could double-`USBDeviceOpenSeize`, skip last-close cleanup, or corrupt capture state. This is the crash class behind #1635 and the `test4.cpp` (sudo) reproducer from the #1798 thread.

**Item 2: same-thread deadlock in the close path.** Any per-device locking scheme deadlocks if `darwin_close` holds the lock while releasing claimed interfaces, because `libusb_release_interface` chains into `darwin_capture_release_interface` -> `darwin_attach_kernel_driver` -> `darwin_reenumerate_device` -> `darwin_restore_state`, which need the same lock.

- One per-device mutex `dpriv->lock` guards **both** counters. A single lock (instead of   #1679's `open_mutex` + `capture_mutex`) removes the ABBA deadlock by construction: the   close path (open -> capture order) and the reset path (capture -> open order) can no longer nest two locks in opposite orders.

- Public entry points are thin wrappers taking the lock; internal callers that already   hold it use the `*_locked` variants (same structure as #1679; names avoid the leading  underscore since file-scope identifiers starting with `_` are reserved in C).

- `darwin_close` releases claimed interfaces **before** entering the critical section.   The release loop only touches per-handle state, so it doesn't need the device lock.  It now also runs when `dpriv->device` is NULL (disconnected device); the interface plug-ins are per-handle objects and every downstream path checks `dpriv->device`;  previously those plug-ins and their run-loop sources were leaked.  `darwin_restore_state`'s internal close/open cycle is unaffected (it temporarily zeroes `claimed_interfaces`, so it never released interfaces there anyway).

- Serializing first-open against last-close also closes a window where a concurrent  `USBDeviceOpenSeize` could race the closing thread's `USBDeviceClose`.
-
[...]

PR #1870, opened 2026-07-11
ci: bound job runtimes and preserve test logs on failure

Youw:
Two small, independent CI robustness improvements, applicable to all platforms:

## 1. `timeout-minutes: 60` on every workflow job

All jobs previously ran with GitHub's default 6-hour limit, so a single wedged job (a hung test, a stuck runner, an unresponsive external service) burned six hours of runner time before dying as `cancelled`. Job *execution* time (`started_at`→`completed_at` per job, which is what `timeout-minutes` actually measures — not run-level `createdAt`/`updatedAt`, which is skewed by queueing and unrelated metadata updates and can be off by orders of magnitude) over the last ~60 completed runs per workflow:

| Workflow | median | p90 | slowest observed |
|---|---|---|---|
| Android (per matrix job) | <1 min | <1 min | ~1 min |
| CIFuzz | ~12 min | ~12 min | ~12 min |
| linux | ~4.5 min | ~5 min | ~5 min |
| macOS | ~1.5 min | ~2 min | ~2.4 min |
| MSYS2 build | ~5.5 min | ~6 min | ~6.5 min |
| MSYS2 clang64 build | ~3.5 min | ~4 min | ~8.5 min |
| MSYS2 aarch64 clang64 | ~7 min | ~7.5 min | ~8 min |
| Windows (per matrix job) | <1 min | ~1.5 min | ~3.5 min |

A uniform 60-minute cap is 5–100× the slowest observed job of any workflow — no realistic false-positive risk — while cutting worst-case waste up to 6×.

CIFuzz is worth calling out specifically since fuzzing time isn't inherently bounded by build size the way a normal compile is: the tight, low-variance distribution above (~12 min every time, over 59 sampled runs including 8 failures) isn't luck, it's structural — the workflow already passes `fuzz-seconds: 600` to `run_fuzzers`, hard-capping the actual fuzzing at 10 minutes regardless of what changed; the rest is fixed build overhead. `timeout-minutes: 60` is ~5× that ceiling, not a guess.

(The `macOS` figures above exclude the pre-fix `stress_mt` hangs from #1869 — those runs legitimately ran until the old 6-hour default killed them, which is exactly the failure mode this PR's timeout is meant to bound. With #1864/#1865 merged, that class of run no longer occurs; every macOS run in the current history completes in ~1–2.5 min.)

## 2. Preserve test logs when a test-running job fails or is cancelled

The autotools test driver buffers each test's stdout/stderr into `tests/<test>.log`; the console only shows a PASS/FAIL summary. `ci-build.sh` already dumps `test-suite.log` inline when `make check` fails, but that dump is lost when the job is cancelled (e.g. hits the new timeout) and doesn't include `config.log`. The five workflows that run the test suite (linux, macOS, and the three MSYS2 flavors) now upload `build*/config.log` + `build*/tests/*.log` with `actions/upload-artifact` (already used elsewhere in these workflows) under `if: failure() || cancelled()`, so whatever evidence exists survives in downloadable form.

On the "is there a third-party action for this" question: for log preservation the official `actions/upload-artifact` *is* the idiomatic tool, so nothing exotic is needed. (No popular off-the-shelf action covers the other thing this branch briefly did — sampling a hung process's stacks before killing it — but that machinery is no longer needed here.)

---

History: this branch originally carried a temporary `stress_mt` hang watchdog that was used to capture the stack trace which root-caused #1869 (since fixed by #1864/#1865, verified over 71 consecutive clean CI runs). That diagnostic served its purpose and has been replaced by this general, issue-agnostic cleanup; the old machinery remains documented in #1869/#1872 if it's ever needed again.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -5,6 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -3,6 +3,7 @@ on: [pull_request]
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
     - name: Build Fuzzers
       id: build
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -5,6 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v6
         with:
@@ -84,3 +85,15 @@ jobs:
           ./fuzz_bos_descriptor tests/fuzz/corpus/bos \
             -seed=1 -runs=200 -max_total_time=15 -timeout=5 -rss_limit_mb=1536 -print_final_stats=1
       # --- end sanitizers + fuzzer smoke ---
+
+      # The autotools test driver buffers each test's output into its own
+      # tests/<test>.log; keep them (and config.log) when something goes wrong.
+      - name: Upload test logs
+        uses: actions/upload-artifact@v7
+        if: failure() || cancelled()
+        with:
+          name: test-logs-linux
+          path: |
+            build*/config.log
+            build*/tests/*.log
+          if-no-files-found: ignore
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -10,6 +10,7 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     runs-on: macos-latest
+    timeout-minutes: 60
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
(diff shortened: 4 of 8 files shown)


mcuee (approved):

mcuee:
Nice. Thanks for looking into this issue and come out with an automatic solution.

Youw:
> 🤖 Update by Claude (`claude-fable-5`, via Claude Code).

Confirmed working as designed: reran this branch's macOS job up to 20 times looking for the hang. It reproduced on attempt 9 — [run 29169932932](https://github.com/libusb/libusb/actions/runs/29169932932/job/86615466593), `compile` step ran 2408s (~40min), the watchdog fired, captured `sample` stacks of the hung `stress_mt`, and killed it — the job failed cleanly in ~41min instead of the previous silent 6h timeout.

The captured trace was directly useful: full write-up and root-cause correction on #1869. Short version: it shows all 8 worker threads wedged on a lost transfer completion (matches @mcuee's TSan finding in #1869, #1798 item 1), not the mechanism I'd originally guessed at in #1869 before this watchdog existed.

The current HEAD check on this PR is red from that same reproduction run (expected — that's the watchdog doing its job, not a bug in it). Happy to trigger a clean rerun if you'd like the check green before merging.

Youw:
> 🤖 Note by Claude (`claude-fable-5`, via Claude Code).

Reworked per maintainer request, now that the hang this branch was chasing is root-caused and fixed (#1869 → #1864/#1865): the temporary `stress_mt` watchdog + inline log decoration are gone, and the PR is now a standalone CI improvement — a 60-minute `timeout-minutes` on **all eight workflows** (not just macOS), plus `actions/upload-artifact` on `failure() || cancelled()` for the test/config logs of the five test-running workflows. Branch history rewritten to a single clean commit; title/description updated accordingly. The comments above this one refer to the old diagnostic incarnation.

Youw:
This is a generic standalone improvement now.

sonatique (approved):

PR #1871, opened 2026-07-12
windows: fix I/O completion port and thread leak on init failure

Youw:
`windows_init()` creates the per-context I/O completion port and its worker thread before starting the optional Windows hotplug event monitor. If a later initialization step failed, the previous error path left the worker blocked in `GetQueuedCompletionStatus()` while holding the context pointer that libusb core immediately frees; the thread and I/O completion port handles also leaked.

This path becomes reachable when `windows_start_event_monitor()` failures propagate from `libusb_init()`, as introduced by #1863.

This PR now:

- Adds shared `windows_iocp_teardown()` cleanup for both failed initialization and normal context exit.
- Keeps the completion port open until after the worker has exited, preventing the worker from observing a closed or recycled handle.
- Tracks the worker lifecycle atomically. During initialization, the worker uses a bounded wait and observes an exit request even if posting the NULL completion packet fails; after successful initialization it returns to the existing indefinite wait behavior.
- Decrements `init_count` after every failed context initialization, while retaining first-initialization global cleanup only for the first-context failure case.

Verified with MSVC Release x64 and Release-Hotplug x64 builds, with warnings treated as errors.

Related to #1863
Depends on #1875

Assisted-by: Claude:claude-fable-5


Youw:
This escalated quickly. Interesting how "corner case not handled" case found by AI turned into non-trivial refactoring of part of the code.

libusb/os/windows_common.h
```diff
@@ -366,10 +366,20 @@ struct windows_backend {
...
+	WINDOWS_IOCP_INITIALIZING = 0, // windows_init() in progress, the thread uses timed waits
+	WINDOWS_IOCP_RUNNING,          // windows_init() succeeded, the thread blocks indefinitely
+	WINDOWS_IOCP_EXIT_REQUESTED,   // teardown in progress, the thread must exit
+};
+
 struct windows_context_priv {
 	const struct windows_backend *backend;
 	HANDLE completion_port;
 	HANDLE completion_port_thread;
+	volatile LONG completion_port_state; // enum windows_iocp_state
```
sonatique: libusbi.h already provides a portable atomic abstraction (usbi_atomic_t with usbi_atomic_load/usbi_atomic_store, libusbi.h:91-129), used e.g. for ctx->hotplug_ready. Since this flag only needs atomicity/visibility, could we use it here instead of a hand-rolled volatile LONG + Interlocked calls? That keeps windows_common.c on the same abstraction as the rest of the tree (C11 atomic_long on MinGW rather than relying on its Interlocked implementations). If applied, the comment on lines 370-371 should say "with the usbi_atomic helpers" instead of "with the Interlocked functions".

libusb/os/windows_common.c
```diff
@@ -432,11 +432,23 @@ static enum windows_version get_windows_version(void)
...
 
+// While the context is still initializing, the I/O completion port thread
+// waits with this timeout so that it notices an exit request even if posting
+// the exit completion packet failed, see windows_iocp_teardown()
+#define IOCP_INIT_WAIT_TIMEOUT_MS 100
+
+static enum windows_iocp_state windows_iocp_state_get(struct windows_context_priv *priv)
+{
+	// Atomic read of the state, with a full memory barrier
+	return (enum windows_iocp_state)InterlockedCompareExchange(&priv->completion_port_state, 0, 0);
```
sonatique: With usbi_atomic_t this becomes a plain usbi_atomic_load, and the CAS-as-read idiom (and its barrier comment on line 442) can go away. The helper is thin enough that callers could also just call usbi_atomic_load directly, like the hotplug_ready readers in hotplug.c do.

libusb/os/windows_common.c
```diff
@@ -512,6 +537,36 @@ static unsigned __stdcall windows_iocp_thread(void *arg)
...
+static void windows_iocp_teardown(struct libusb_context *ctx)
+{
+	struct windows_context_priv *priv = usbi_get_context_priv(ctx);
+
+	if (priv->completion_port_thread != NULL) {
+		// Request the thread to exit. While the context is still
+		// initializing the thread waits with a finite timeout and checks
+		// this state, so it exits even if posting the exit completion
+		// packet below fails.
+		InterlockedExchange(&priv->completion_port_state, WINDOWS_IOCP_EXIT_REQUESTED);
```
sonatique: usbi_atomic_store(&priv->completion_port_state, WINDOWS_IOCP_EXIT_REQUESTED);

libusb/os/windows_common.c
```diff
@@ -564,11 +619,14 @@ static int windows_init(struct libusb_context *ctx)
 		goto init_exit;
 	}
 
-	// And a dedicated thread to wait for I/O completions
+	// And a dedicated thread to wait for I/O completions. The thread starts
+	// in the initializing state and uses timed waits until initialization
+	// has fully succeeded; thread creation orders this store before any
+	// read the thread performs.
+	priv->completion_port_state = WINDOWS_IOCP_INITIALIZING;
```
sonatique: This one is more than style: it's currently a plain assignment while every other access is Interlocked. It works (the field is still zero from the context calloc and thread creation orders the store), but if the field becomes usbi_atomic_t, the store macro keeps all accesses uniform:
Suggested:
usbi_atomic_store(&priv->completion_port_state, WINDOWS_IOCP_INITIALIZING);

libusb/os/windows_common.c
```diff
@@ -586,15 +644,25 @@ static int windows_init(struct libusb_context *ctx)
 	windows_initial_scan_devices(ctx);
 #endif
 
+	// Initialization has fully succeeded, switch the I/O completion port
+	// thread to indefinite blocking waits
+	InterlockedExchange(&priv->completion_port_state, WINDOWS_IOCP_RUNNING);
```
sonatique: usbi_atomic_store(&priv->completion_port_state, WINDOWS_IOCP_RUNNING);

sonatique (changes requested):

sonatique:
@Youw : interestingly, the usbi_atomic_load / usbi_atomic_store I recommended are currently not fully implemented for MSVC.
I made a PR for improving this: https://github.com/libusb/libusb/pull/1875

sonatique (approved):

Youw:
updated
I guess we will wait for https://github.com/libusb/libusb/pull/1875 first

sonatique:
@Youw #1875 is merged. Maybe you can rebase this one, and let the CI run, then we should be able to merge it I guess

PR #1873, opened 2026-07-13
darwin: hold the handle lock across interface state changes

sonatique:
# darwin: hold the handle lock across interface state changes

Fixes items 5 and 6 of #1798, plus the crash @mcuee reproduced with the point-6 test
(https://github.com/libusb/libusb/issues/1798#issuecomment-4882194652):
`libusb_set_interface_alt_setting` checks the claimed bit under `dev_handle->lock`,
releases it, then calls the backend — so a concurrent `libusb_release_interface` can
tear down the interface plug-in under `get_endpoints` (NULL SEGV, plain ASan on master).

Backend entry points that mutate per-handle interface state now take
`dev_handle->lock`, always before `dpriv->lock` (the order the core established):
`set_configuration`, `set_interface_altsetting` and `clear_halt` become
wrapper + `_locked` pairs (with `ep_to_pipeRef` split too, so the altsetting path
doesn't self-deadlock on the lock from #1695); `reset_device`/`detach`/`attach` take it
before `dpriv->lock`, so `darwin_restore_state` runs with both locks and its
`claimed_interfaces` writes are protected.

Consequence handled with care: reset can now hold `dev_handle->lock` across a
re-enumeration, so nothing on the event thread may take that lock.
`darwin_async_io_callback`'s zero-length-packet write was the one offender — the pipe
is now cached in the transfer private data at submission. This also fixes a
pre-existing bug where the callback ignored the pipe lookup result and could
dereference an uninitialized interface pointer.

Annotation note: the entry points previously carried an *aspirational*
`REQUIRES(dev_handle->lock)` (the core never held the lock across the backend
function-pointer call — invisible to the analysis). They now actually acquire it and
are annotated `EXCLUDES(dev_handle->lock)`, with `REQUIRES` moved to the `_locked`
variants — the whole chain is now compiler-checked. Zero new `-Wthread-safety`
warnings; @seanm, a run of your analysis would be a welcome confirmation.

## Testing

@mcuee: `test1798_point6.cpp` (plain ASan, no sudo) is the targeted reproducer —
expectation: no crash. The point-7 test (`active_config`) will still report under
TSan; that's #1798 item 7, queued separately.

Reference: #1798 (items 5 and 6)

Assisted-by: claude-code:claude-fable-5

PR #1874, opened 2026-07-13
darwin: make first_config and active_config atomic

sonatique:
# darwin: make first_config and active_config atomic

Fixes item 7 of #1798.

`active_config` is written by `darwin_set_configuration` and by
`darwin_check_configuration` (running on the hotplug thread during
enumeration and re-enumeration) and read by `darwin_get_configuration`,
`darwin_get_active_config_descriptor`, `darwin_reenumerate_device` and
`darwin_restore_state`, with no common lock. `first_config` has the same pattern
(written by `darwin_check_configuration`, read by `darwin_claim_interface`'s
set-configuration fallback).

Whether concurrent `libusb_get_configuration`/`libusb_set_configuration` should return
*ordered* results is an API-contract question (the Linux backend has the same pattern),
but at the C level the unsynchronized mixed-thread access is a data race and therefore
undefined behavior. Making both fields `_Atomic UInt8` removes the UB cheaply, same
size/alignment, always lock-free, no new ordering semantics claimed, and the fields are
never on transfer-hot paths.

One call needed rewriting: `GetConfiguration` took the field's address, which is not
possible with an atomic. It now reads into a local and stores only on success, the
previous code ignored the result, so the value was indeterminate when the call failed;
Keeping the last known value is well-defined.

@mcuee: the point-7 TSan test (`test1798_point7.cpp`) is the targeted check, expectation: the `active_config` data-race report disappears.
Note: to save you a build cycle: #1873  (items 5/6) and this (item 7) merge cleanly with each other, so one combined build tests both.

Reference: #1798 (item 7)

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1098,10 +1098,10 @@ static int darwin_get_active_config_descriptor(struct libusb_device *dev, void *
   struct darwin_cached_device *priv = DARWIN_CACHED_DEVICE(dev);
   int config_index;
 
-  if (0 == priv->active_config)
+  if (0 == atomic_load_explicit(&priv->active_config, memory_order_relaxed))
     return LIBUSB_ERROR_NOT_FOUND;
 
-  config_index = get_configuration_index (dev, priv->active_config);
+  config_index = get_configuration_index (dev, atomic_load_explicit(&priv->active_config, memory_order_relaxed));
   if (config_index < 0)
     return config_index;
 
@@ -1154,13 +1154,13 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
   if (0x05ac == libusb_le16_to_cpu (dev->dev_descriptor.idVendor) &&
       0x8005 == libusb_le16_to_cpu (dev->dev_descriptor.idProduct)) {
     usbi_dbg (ctx, "ignoring configuration on root hub simulation");
-    dev->active_config = 0;
+    atomic_store_explicit(&dev->active_config, 0, memory_order_relaxed);
     return LIBUSB_SUCCESS;
   }
 
   /* find the first configuration */
   kresult = (*darwin_device)->GetConfigurationDescriptorPtr (darwin_device, 0, &configDesc);
-  dev->first_config = (kIOReturnSuccess == kresult) ? configDesc->bConfigurationValue : 1;
+  atomic_store_explicit(&dev->first_config, (kIOReturnSuccess == kresult) ? configDesc->bConfigurationValue : 1, memory_order_relaxed);
 
   /* check if the device is already configured. there is probably a better way than iterating over the
      to accomplish this (the trick is we need to avoid a call to GetConfigurations since buggy devices
@@ -1188,21 +1188,22 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
     /* device is configured */
     if (dev->dev_descriptor.bNumConfigurations == 1) {
       /* to avoid problems with some devices get the configurations value from the configuration descriptor */
-      dev->active_config = dev->first_config;
+      atomic_store_explicit(&dev->active_config, atomic_load_explicit(&dev->first_config, memory_order_relaxed), memory_order_relaxed);
     } else {
       /* devices with more than one configuration should work with GetConfiguration.
          active_config is atomic, so read into a local and store only on
          success */
       UInt8 active_config;
       if (kIOReturnSuccess == (*darwin_device)->GetConfiguration (darwin_device, &active_config)) {
-        dev->active_config = active_config;
+        atomic_store_explicit(&dev->active_config, active_config, memory_order_relaxed);
       }
     }
   } else
     /* not configured */
-    dev->active_config = 0;
+    atomic_store_explicit(&dev->active_config, 0, memory_order_relaxed);
 
-  usbi_dbg (ctx, "active config: %u, first config: %u", dev->active_config, dev->first_config);
+  usbi_dbg (ctx, "active config: %u, first config: %u", atomic_load_explicit(&dev->active_config, memory_order_relaxed),
+            atomic_load_explicit(&dev->first_config, memory_order_relaxed));
 
   return LIBUSB_SUCCESS;
 }
@@ -1468,6 +1469,13 @@ static enum libusb_error darwin_get_cached_device(struct libusb_context *ctx, io
         ret = LIBUSB_ERROR_NO_MEM;
         break;
       }
+
+      /* calloc zero-fills the storage but does not give the atomic members a
+         specified initialized state in C11; initialize them explicitly before
+         list_add publishes the object */
+      atomic_init(&new_device->first_config, 0);
+      atomic_init(&new_device->active_config, 0);
+      atomic_init(&new_device->in_reenumerate, false);
 
       /* add this device to the cached device list */
       list_add(&new_device->list, &darwin_cached_devices);
@@ -1797,7 +1805,7 @@ static void darwin_close (struct libusb_device_handle *dev_handle) EXCLUDES(dev_
 static int darwin_get_configuration(struct libusb_device_handle *dev_handle, uint8_t *config) {
   struct darwin_cached_device *dpriv = DARWIN_CACHED_DEVICE(dev_handle->dev);
 
-  *config = dpriv->active_config;
+  *config = atomic_load_explicit(&dpriv->active_config, memory_order_relaxed);
 
   return LIBUSB_SUCCESS;
 }
@@ -1828,7 +1836,7 @@ static int darwin_set_configuration_locked(struct libusb_device_handle *dev_hand
     if (dev_handle->claimed_interfaces & (1U << i))
       darwin_claim_interface (dev_handle, i);
 
-  dpriv->active_config = (UInt8)config;
+  atomic_store_explicit(&dpriv->active_config, (UInt8)config, memory_order_relaxed);
 
   return LIBUSB_SUCCESS;
 }
@@ -1982,13 +1990,13 @@ static int darwin_claim_interface(struct libusb_device_handle *dev_handle, uint8
     return darwin_to_libusb (kresult);
 
   /* make sure we have an interface */
-  if (!usbInterface && dpriv->first_config != 0) {
-    usbi_info (ctx, "no interface found; setting configuration: %d", dpriv->first_config);
+  if (!usbInterface && 0 != atomic_load_explicit(&dpriv->first_config, memory_order_relaxed)) {
+    usbi_info (ctx, "no interface found; setting configuration: %d", atomic_load_explicit(&dpriv->first_config, memory_order_relaxed));
 
     /* set the configuration. the caller already holds dev_handle->lock, so
        call the locked variant (the darwin_set_configuration entry point
        would deadlock here). */
-    ret = darwin_set_configuration_locked (dev_handle, (int) dpriv->first_config);
+    ret = darwin_set_configuration_locked (dev_handle, (int) atomic_load_explicit(&dpriv->first_config, memory_order_relaxed));
     if (ret != LIBUSB_SUCCESS) {
       usbi_err (ctx, "could not set configuration");
       return ret;
@@ -2271,7 +2279,7 @@ static enum libusb_error darwin_restore_state (struct libusb_device_handle *dev_
     return LIBUSB_ERROR_NOT_FOUND;
   }
 
-  if (dpriv->active_config != active_config) {
+  if (atomic_load_explicit(&dpriv->active_config, memory_order_relaxed) != active_config) {
     usbi_dbg (ctx, "darwin/restore_state: restoring configuration %d...", active_config);
 
     ret = darwin_set_configuration_locked (dev_handle, active_config);
@@ -2344,7 +2352,7 @@ static void darwin_adopt_pending_device (struct darwin_cached_device *dpriv) EXC
 static int darwin_reenumerate_device (struct libusb_device_handle *dev_handle, bool capture) REQUIRES(dev_handle->lock) {
   struct darwin_cached_device *dpriv = DARWIN_CACHED_DEVICE(dev_handle->dev);
   unsigned long claimed_interfaces = dev_handle->claimed_interfaces;
-  uint8_t active_config = dpriv->active_config;
+  uint8_t active_config = atomic_load_explicit(&dpriv->active_config, memory_order_relaxed);
   UInt32 options = 0;
   IOUSBDeviceDescriptor descriptor;
   IOUSBConfigurationDescriptorPtr cached_configuration;
@@ -2479,7 +2487,7 @@ static int darwin_reset_device_locked (struct libusb_device_handle *dev_handle) 
   ret = darwin_reenumerate_device (dev_handle, false);
   if ((ret == LIBUSB_SUCCESS || ret == LIBUSB_ERROR_NOT_FOUND) && dpriv->capture_count > 0) {
     int capture_count;
-    uint8_t active_config = dpriv->active_config;
+    uint8_t active_config = atomic_load_explicit(&dpriv->active_config, memory_order_relaxed);
     unsigned long claimed_interfaces = dev_handle->claimed_interfaces;
 
     /* save old capture_count */
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1186,12 +1186,18 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
     IOObjectRelease (firstInterface);
 
     /* device is configured */
-    if (dev->dev_descriptor.bNumConfigurations == 1)
+    if (dev->dev_descriptor.bNumConfigurations == 1) {
       /* to avoid problems with some devices get the configurations value from the configuration descriptor */
       dev->active_config = dev->first_config;
-    else
-      /* devices with more than one configuration should work with GetConfiguration */
-      (*darwin_device)->GetConfiguration (darwin_device, &dev->active_config);
+    } else {
+      /* devices with more than one configuration should work with GetConfiguration.
+         active_config is atomic, so read into a local and store only on
+         success */
+      UInt8 active_config;
+      if (kIOReturnSuccess == (*darwin_device)->GetConfiguration (darwin_device, &active_config)) {
+        dev->active_config = active_config;
+      }
+    }
   } else
     /* not configured */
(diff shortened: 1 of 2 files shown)


libusb/os/darwin_usb.h
```diff
@@ -126,7 +126,12 @@ struct darwin_cached_device {
   usbi_mutex_t          lock;          /* protects open_count and capture_count */
   int                   open_count;    /* GUARDED_BY(lock) */
   int                   capture_count; /* GUARDED_BY(lock) */
-  UInt8                 first_config, active_config, port;
+  UInt8                 port;
+  /* first_config and active_config are written by darwin_check_configuration
+     on the enumeration/hotplug paths and read from user threads with no
+     common lock. Using atomic remove the data race. */
+  _Atomic UInt8         first_config;
+  _Atomic UInt8         active_config;
```
Youw, line 136: **[P2] Initialize both allocated atomics before publication**

`darwin_cached_device` is created by `calloc()` at `darwin_usb.c:1454`, but zeroing allocated storage does not give its atomic subobjects a specified initialized state; [WG14 recommends `atomic_init()` immediately after allocation](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2886.htm). This has actual read-before-store paths: a failed `GetConfiguration()` for a newly discovered multi-configuration device skips the store and the log at line 1193 immediately reads `active_config`, while the root-hub early return at lines 1142-1146 leaves `first_config` unset for later readers. Please `atomic_init(..., 0)` both fields before `list_add()` publishes the object.

---

libusb/os/darwin_usb.c
```diff
@@ -1174,12 +1174,18 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
...
-    else
-      /* devices with more than one configuration should work with GetConfiguration */
-      (*darwin_device)->GetConfiguration (darwin_device, &dev->active_config);
+    } else {
+      /* devices with more than one configuration should work with GetConfiguration.
+         active_config is atomic, so read into a local and store only on
+         success */
+      UInt8 active_config;
+      if (kIOReturnSuccess == (*darwin_device)->GetConfiguration (darwin_device, &active_config)) {
+        dev->active_config = active_config;
```
Youw: **[P2] Spell all atomic accesses explicitly**

This assignment and every existing plain access to these members become implicit sequentially consistent operations. The PR's own [macOS Xcode log](https://github.com/libusb/libusb/actions/runs/29245761883/job/86802153874) records 32 warnings (16 sites for each of x86_64 and arm64) because the project enables `-Watomic-implicit-seq-cst`; downstream warnings-as-errors builds will fail. [Commit `8ed83638`](https://github.com/libusb/libusb/commit/8ed83638ce1929a63d52782c610953174e4d305f) fixed the same warning for `in_reenumerate`. Please convert every access to explicit atomic loads/stores, preferably with `memory_order_relaxed` because these cached bytes do not publish other state and the PR does not require ordering semantics.

---

Youw (commented):
I found two issues: both newly allocated atomics can be read before initialization on existing error/early-return paths, and the type change makes the Xcode build emit 32 `-Watomic-implicit-seq-cst` warnings. Details inline.

---

libusb/os/darwin_usb.h
sonatique: Fixed new commit: both fields are atomic_init-ed right after the calloc, before list_add publishes the object, and the pre-existing in_reenumerate gets the same treatment since it had the identical gap. Both read-before-store paths you identified now read a specified value.

libusb/os/darwin_usb.c
sonatique: Fixed in new commit: all 16 sites converted to explicit atomic_load_explicit/atomic_store_explicit with memory_order_relaxed, per your suggestion (these are independent cached bytes; no ordering claimed, matching the commit's stated semantics). in_reenumerate intentionally keeps its seq-cst operations, which its completion handshake relies on. The macOS CI job's warning count is the direct check, it should now read zero,

Youw (approved):

mcuee:
Great. I should be able to test this over the weekend.

mcuee:
Great. The following test results show this PR is good whereas libusb git has issues.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point7 % cat test1798_point7.cpp   
#include <iostream>
#include <libusb.h>
#include <pthread.h>
#include <atomic>
#include <unistd.h>
#include <vector>

// Control flags
std::atomic<bool> keep_running{true};
std::atomic<int>  target_config{0};
std::atomic<long> race_detections{0};

struct TestContext {
    libusb_device_handle* handle;
    int cfg_off; // 0
    int cfg_on;  // 1
};

// Thread A: The Writer
// Triggers darwin_set_configuration -> writes dpriv->active_config WITHOUT lock.
void* writer_func(void* arg) {
    TestContext* ctx = (TestContext*)arg;
    while (keep_running) {
        // Set to Config 1
        target_config.store(ctx->cfg_on);
        if (libusb_set_configuration(ctx->handle, ctx->cfg_on) == 0) {
            // Configuration successfully changed hardware-side
        }

        // Set to Config 0 (Unconfigured)
        target_config.store(ctx->cfg_off);
        if (libusb_set_configuration(ctx->handle, ctx->cfg_off) == 0) {
            // Configuration successfully changed hardware-side
        }
    }
    return nullptr;
}

// Thread B: The Reader
// Triggers darwin_get_configuration -> reads dpriv->active_config WITHOUT lock.
void* reader_func(void* arg) {
    TestContext* ctx = (TestContext*)arg;
    int current_cfg = -1;
    
    while (keep_running) {
        // We read what the writer just successfully told the hardware to do
        int expected = target_config.load();
        
        // Point 7: This core call returns dpriv->active_config without a lock.
        int r = libusb_get_configuration(ctx->handle, &current_cfg);
        
        if (r == 0) {
            // If the write to active_config hasn't propagated or was lost in a race,
            // current_cfg will not match the expected state.
            if (current_cfg != expected) {
                race_detections++;
            }
        }
    }
    return nullptr;
}

int main(int argc, char** argv) {
    libusb_context* usb_ctx = nullptr;
    if (libusb_init(&usb_ctx) < 0) return 1;

    // Use a standard USB device. Most devices support Config 1 and Config 0.
    // Replace with your specific VID/PID.
    uint16_t vid = 0x04b4; 
    uint16_t pid = 0x8613; 

    libusb_device_handle* handle = libusb_open_device_with_vid_pid(usb_ctx, vid, pid);
    if (!handle) {
        printf("Error: Could not open device %04x:%04x. Check permissions/Sudo.\n", vid, pid);
        libusb_exit(usb_ctx);
        return 1;
    }

    TestContext t_ctx = { handle, 0, 1 }; 
    pthread_t writer, reader;

    printf("Starting Point 7 Race Test (C++11)...\n");
    printf("Targeting unprotected dpriv->active_config access.\n");
    printf("Toggling between Config 0 and Config 1.\n");

    pthread_create(&writer, nullptr, writer_func, &t_ctx);
    pthread_create(&reader, nullptr, reader_func, &t_ctx);

    // Run the stress test
    for(int i = 0; i < 20; ++i) {
        sleep(1);
        if (race_detections > 0) {
            printf("Detected Inconsistencies: %ld\n", race_detections.load());
        }
    }

    keep_running = false;
    pthread_join(writer, nullptr);
    pthread_join(reader, nullptr);

    printf("Final Inconsistency Count: %ld\n", race_detections.load());
    if (race_detections > 0) {
        printf("RESULT: Point 7 verified. active_config access is NOT synchronized.\n");
    } else {
        printf("RESULT: No race detected in this window. High system load may be required.\n");
    }
    
    libusb_close(handle);
    libusb_exit(usb_ctx);
    return 0;
}

mcuee in ~/build/libusb/libusb_test/issue1798_point7 % cat build_tsan_debug.sh   
clang++ -std=c++11 -g -fsanitize=thread -fno-omit-frame-pointer -pthread -I /Users/mcuee/mybin/bin_git_tsan_debug/include/libusb-1.0 test1798_point7.cpp -L /Users/mcuee/mybin/bin_git_tsan_debug/lib -lusb-1.0 -o ./test1798_point7_git_tsan

clang++ -std=c++11 -g -fsanitize=thread -fno-omit-frame-pointer -pthread -I /Users/mcuee/mybin/bin_pr1873_tsan_debug/include/libusb-1.0 test1798_point7.cpp -L /Users/mcuee/mybin/bin_pr1874_tsan_debug/lib -lusb-1.0 -o ./test1798_point7_pr1874_tsan

mcuee in ~/build/libusb/libusb_test/issue1798_point7 % sh build_tsan_debug.sh    

mcuee in ~/build/libusb/libusb_test/issue1798_point7 % ./test1798_point7_git_tsan
Starting Point 7 Race Test (C++11)...
Targeting unprotected dpriv->active_config access.
Toggling between Config 0 and Config 1.
==================
WARNING: ThreadSanitizer: data race (pid=93951)
  Write of size 1 at 0x00010cb02d69 by thread T2:
    #0 darwin_set_configuration <null> (libusb-1.0.0.dylib:arm64+0x177a4)
    #1 libusb_set_configuration <null> (libusb-1.0.0.dylib:arm64+0x3dc8)
    #2 writer_func(void*) test1798_point7.cpp:26 (test1798_point7_git_tsan:arm64+0x10000082c)

  Previous read of size 1 at 0x00010cb02d69 by thread T3:
    #0 darwin_get_configuration <null> (libusb-1.0.0.dylib:arm64+0x17570)
    #1 libusb_get_configuration <null> (libusb-1.0.0.dylib:arm64+0x3b5c)
    #2 reader_func(void*) test1798_point7.cpp:50 (test1798_point7_git_tsan:arm64+0x100000a18)

  Location is heap block of size 200 at 0x00010cb02cb0 allocated by main thread:
    #0 calloc <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x67c40)
    #1 darwin_get_cached_device <null> (libusb-1.0.0.dylib:arm64+0x1b898)
    #2 darwin_scan_devices <null> (libusb-1.0.0.dylib:arm64+0x195e0)
    #3 darwin_init_context <null> (libusb-1.0.0.dylib:arm64+0x191dc)
    #4 darwin_init <null> (libusb-1.0.0.dylib:arm64+0x16a68)
    #5 libusb_init_context <null> (libusb-1.0.0.dylib:arm64+0x60c8)
    #6 libusb_init <null> (libusb-1.0.0.dylib:arm64+0x5a00)
    #7 main test1798_point7.cpp:65 (test1798_point7_git_tsan:arm64+0x100000b94)

  Thread T2 (tid=290452, running) created by main thread at:
    #0 pthread_create <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x33834)
    #1 main test1798_point7.cpp:86 (test1798_point7_git_tsan:arm64+0x100000ccc)

  Thread T3 (tid=290453, running) created by main thread at:
    #0 pthread_create <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x33834)
    #1 main test1798_point7.cpp:87 (test1798_point7_git_tsan:arm64+0x100000ce8)

SUMMARY: ThreadSanitizer: data race (libusb-1.0.0.dylib:arm64+0x177a4) in darwin_set_configuration+0x1f8
==================
Detected Inconsistencies: 1444944
Detected Inconsistencies: 2788755
Detected Inconsistencies: 4118760
Detected Inconsistencies: 5348035
Detected Inconsistencies: 6637053
Detected Inconsistencies: 7892810
Detected Inconsistencies: 9046197
Detected Inconsistencies: 10269945
Detected Inconsistencies: 11460857
Detected Inconsistencies: 12576726
Detected Inconsistencies: 13742244
Detected Inconsistencies: 14908498
Detected Inconsistencies: 16057885
Detected Inconsistencies: 17145265
Detected Inconsistencies: 18298535
Detected Inconsistencies: 19444574
Detected Inconsistencies: 20575847
==================
WARNING: ThreadSanitizer: data race (pid=93951)
  Read of size 1 at 0x00010cb02d69 by thread T3:
    #0 darwin_get_configuration <null> (libusb-1.0.0.dylib:arm64+0x17570)
    #1 libusb_get_configuration <null> (libusb-1.0.0.dylib:arm64+0x3b5c)
    #2 reader_func(void*) test1798_point7.cpp:50 (test1798_point7_git_tsan:arm64+0x100000a18)

  Previous write of size 1 at 0x00010cb02d69 by thread T2:
    #0 darwin_set_configuration <null> (libusb-1.0.0.dylib:arm64+0x177a4)
    #1 libusb_set_configuration <null> (libusb-1.0.0.dylib:arm64+0x3dc8)
    #2 writer_func(void*) test1798_point7.cpp:32 (test1798_point7_git_tsan:arm64+0x1000008a4)

  Location is heap block of size 200 at 0x00010cb02cb0 allocated by main thread:
    #0 calloc <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x67c40)
    #1 darwin_get_cached_device <null> (libusb-1.0.0.dylib:arm64+0x1b898)
    #2 darwin_scan_devices <null> (libusb-1.0.0.dylib:arm64+0x195e0)
    #3 darwin_init_context <null> (libusb-1.0.0.dylib:arm64+0x191dc)
    #4 darwin_init <null> (libusb-1.0.0.dylib:arm64+0x16a68)
    #5 libusb_init_context <null> (libusb-1.0.0.dylib:arm64+0x60c8)
    #6 libusb_init <null> (libusb-1.0.0.dylib:arm64+0x5a00)
    #7 main test1798_point7.cpp:65 (test1798_point7_git_tsan:arm64+0x100000b94)

  Thread T3 (tid=290453, running) created by main thread at:
    #0 pthread_create <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x33834)
    #1 main test1798_point7.cpp:87 (test1798_point7_git_tsan:arm64+0x100000ce8)

  Thread T2 (tid=290452, running) created by main thread at:
    #0 pthread_create <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x33834)
    #1 main test1798_point7.cpp:86 (test1798_point7_git_tsan:arm64+0x100000ccc)

SUMMARY: ThreadSanitizer: data race (libusb-1.0.0.dylib:arm64+0x17570) in darwin_get_configuration+0x70
==================
Detected Inconsistencies: 21634669
Detected Inconsistencies: 22778127
Detected Inconsistencies: 23909604
Final Inconsistency Count: 23909610
RESULT: Point 7 verified. active_config access is NOT synchronized.
ThreadSanitizer: reported 2 warnings
zsh: abort      ./test1798_point7_git_tsan

mcuee in ~/build/libusb/libusb_test/issue1798_point7 % ./test1798_point7_pr1874_tsan
Starting Point 7 Race Test (C++11)...
Targeting unprotected dpriv->active_config access.
Toggling between Config 0 and Config 1.
Detected Inconsistencies: 1334975
Detected Inconsistencies: 2647818
Detected Inconsistencies: 3947257
Detected Inconsistencies: 5169357
Detected Inconsistencies: 6444669
Detected Inconsistencies: 7708514
Detected Inconsistencies: 8893105
Detected Inconsistencies: 10141174
Detected Inconsistencies: 11384299
Detected Inconsistencies: 12546925
Detected Inconsistencies: 13800516
Detected Inconsistencies: 15045033
Detected Inconsistencies: 16284978
Detected Inconsistencies: 17467401
Detected Inconsistencies: 18714953
Detected Inconsistencies: 19950139
Detected Inconsistencies: 21119244
Detected Inconsistencies: 22352393
Detected Inconsistencies: 23593334
Detected Inconsistencies: 24831500
Final Inconsistency Count: 24831538
RESULT: Point 7 verified. active_config access is NOT synchronized.
```

mcuee (approved):

libusb/os/darwin_usb.c
```diff
@@ -1186,17 +1186,24 @@ static enum libusb_error darwin_check_configuration (struct libusb_context *ctx,
...
-      /* devices with more than one configuration should work with GetConfiguration */
-      (*darwin_device)->GetConfiguration (darwin_device, &dev->active_config);
+      atomic_store_explicit(&dev->active_config, atomic_load_explicit(&dev->first_config, memory_order_relaxed), memory_order_relaxed);
+    } else {
+      /* devices with more than one configuration should work with GetConfiguration.
+         active_config is atomic, so read into a local and store only on
+         success */
+      UInt8 active_config;
+      if (kIOReturnSuccess == (*darwin_device)->GetConfiguration (darwin_device, &active_config)) {
+        atomic_store_explicit(&dev->active_config, active_config, memory_order_relaxed);
```
seanm, line 1198: Using atomics here fixes the undefined behaviour of 2 threads reading/writing to the same memory at once, but sometimes that just papers over a deeper problem (maybe the case here, maybe not, I'm not sure).

Can 2 threads enter line 1197 at once?  If so, they can presumably both get a local `active_config` of true, and both enter that `if`. If so, they are still racing, but at least `dev->active_config` will be one or the other, and not torn. Is that OK?

sonatique: Can two threads enter at once? Yes: darwin_check_configuration's only caller is process_new_device, which is reached from both the hotplug thread (darwin_devices_attached, line 689) and from darwin_scan_devices on a user thread during libusb_init (line 1659). So two contexts initializing, or a scan racing an attach for the same cached device, can both run it: both can enter the GetConfiguration branch and both can store.

Is that OK? I believe yes, because both writers are storing fresh answers to the same question read from the same device. In the steady state the answers are identical and the race is idempotent; they can only differ while racing an actual configuration change, and then whichever store lands is a value the device had a moment ago. active_config is a cache and it can be stale the instant after any read, and no reader treats it as authoritative-fresh (libusb_get_configuration is documented as possibly returning a cached value).

The one consumer that needs an ordered value is darwin_restore_state after re-enumeration (the comparison at line 2282), and that pair is not racing: the hotplug thread refreshes active_config inside process_new_device (line 689) before its seq-cst atomic_exchange(&…->in_reenumerate, false) (line 692), and the re-enumerating thread spins on atomic_load(&dpriv->in_reenumerate) (line 2428) and only reads active_config after seeing the flag clear. The seq-cst flag operations provide the happens-before edge, which is exactly why the field accesses themselves can be memory_order_relaxed : the ordering that matters is carried by the handshake, and the relaxed accesses deliberately claim none of their own.

So agreed: the atomics remove exactly the undefined behavior and nothing more, which is all this PR claims. The deeper "should concurrent get/set_configuration be ordered?" question is the #1798 item-7 contract discussion (the Linux backend has the identical pattern). Eliminating even the transient staleness would need a common lock or a don't-clobber-while-open rule on the enumeration write path, but the re-enumeration path requires the refresh (restore_state depends on it), so that surgery is subtle, and for a cache with these semantics I don't think it earns its complexity today.

PR #1875, opened 2026-07-13
core: honor documented atomic ordering in MSVC usbi_atomic_load/store

sonatique:
On MSVC, usbi_atomic_load()/usbi_atomic_store() are plain volatile accesses. The documented contract in libusbi.h says all usbi_atomic operations are ordered with each other, and the GCC-builtin and C11 implementations deliver seq-cst, but volatile only provides ordering on x86/x64 as a side effect of /volatile:ms. 

On MSVC ARM64 (default /volatile:iso) these accesses carry no hardware ordering at all, so the contract is not met.

Implement the load and store with Interlocked operations so the contract holds on every architecture MSVC targets. usbi_atomic_inc()/usbi_atomic_dec() were already Interlocked. 

The load becomes a locked RMW, which is heavier than a plain read, but no current call site is hot enough to matter.

Verified with MSVC Release x64 and Release ARM64 builds.

Related: #1871

### DIFF
--- a/libusb/libusbi.h
+++ b/libusb/libusbi.h
@@ -103,8 +103,8 @@
  */
 #ifdef _MSC_VER
 typedef volatile LONG usbi_atomic_t;
-#define usbi_atomic_load(a)	(*(a))
-#define usbi_atomic_store(a, v)	(*(a)) = (v)
+#define usbi_atomic_load(a)	InterlockedCompareExchange((a), 0, 0)
+#define usbi_atomic_store(a, v)	((void)InterlockedExchange((a), (v)))
 #define usbi_atomic_inc(a)	InterlockedIncrement((a))
 #define usbi_atomic_dec(a)	InterlockedDecrement((a))
 #else


Youw (approved):

Youw:
I checked the existing MSVC call sites and the #1871 consumer. I do not see a practical correctness regression: all destinations are naturally aligned, mutable `LONG` fields; runtime accesses consistently use the atomic helpers; and no caller uses the old store expression's return value. The change only strengthens the documented ordering.

The existing 26 MSVC load/store call sites are cold or coarse-grained. The only plausible hot use is #1871, which loads `completion_port_state` once per IOCP completion. That adds an uncontended RMW/full fence per completion, but it should normally be small beside `GetQueuedCompletionStatus()` and USB I/O, with no steady-state writer contention. #1871 could cache the transition to `WINDOWS_IOCP_RUNNING` if we want to eliminate even that cost.

I do not see a reason to block this PR.

Generated by Codex (GPT-5).

mcuee (approved):

seanm (dismissed):
I think the "generated by AI" really needs to be in the *commit message*, not just the github conversation.   git will outlive github.  Wait, what is generated by AI here?  @Youw's comment?  Or this code change?

sonatique:
@seanm : I think the "Generated by Codex (GPT-5.6 Sol)." otherwise comment by @Youw was like an "AI signature" for the approval. It's not to say the PR is generated by Codex (it is not).

Youw:
Yeah, I asked Claude/Codex to explicitly sign all the comments posted to GitHub by itself, even if asked by me. Inthis particular case I checked PR, but asked Codex to re-check if there are any other implications by the change, and if not - approve the PR. So it did, with signature.

seanm:
I see. It's *very* confusing when you start a sentence with "I" but in fact you are not talking about yourself.  Instead of "I checked..." writing "My LLM checked..." would be clearer.

The 'signature' helps a bit, but English is read top-to-bottom, and the comment is still listed under your account name, and using "I" everywhere.

LLM comments can indeed be helpful, but knowing what's human and what's bot should be very clear (and is getting hard these days).

Can you instruct your LLM to generate text in the 3rd person, not 1st person?

Youw:
> Can you instruct your LLM to generate text in the 3rd person, not 1st person?

Yeat, that I can do. I'll update AGENTS.md with that note too.

PR #1877, opened 2026-07-15
Add thread safety EXCLUDES/REQUIRES to prototypes

seanm:


Youw (approved):
This looks good, but I'm also started AI to check this.

Youw:
One issue from the thread-safety review: Clang treats `REQUIRES` and `EXCLUDES` as inheritable. Keeping each existing definition annotation while adding the same annotation to its forward declaration leaves the definition with both an inherited and an explicit contract. Codex reproduced duplicate `-Wthread-safety` diagnostics for violating calls with Clang 19.1.5 and 22.1.3.

Please retain the annotations on the first declarations, since calls before the definitions need those contracts, and remove the duplicates from the eight definitions: `darwin_exit`, `darwin_get_cached_device`, `darwin_claim_interface`, `darwin_release_interface`, `darwin_clear_halt`, `darwin_reenumerate_device`, `darwin_reset_device`, and `darwin_detach_kernel_driver`. This preserves the intended checking while avoiding each violation being reported twice.

mcuee (approved):

seanm:
@Youw I don't really understand what your LLM means there. Having the decoration on both the prototype and definition as best I can tell is harmless, and in fact beneficial because you don't have to go looking if there's a prototype (not all functions have one).

Youw:
yes, what it says is that if the annotation is present in both function declaration and definition, in case of failure to comply the declaration - the error message will appear twice

seanm:
Ah, well that's fine then.

PR #1878, opened 2026-07-15
Updates to account for C23 changes

seanm:
- use standard [[deprecated]] to indicate deprecation
- never #define static_assert if C23

### DIFF
--- a/libusb/libusb.h
+++ b/libusb/libusb.h
@@ -71,7 +71,9 @@ typedef SSIZE_T ssize_t;
 #endif
 #endif /* _WIN32 || __CYGWIN__ */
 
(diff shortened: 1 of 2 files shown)


Youw (approved):
This looks absolutely fine.
Do we have a way to actually enforce/check this on CI?

sonatique (approved):

seanm:
> Do we have a way to actually enforce/check this on CI?

As I said elsewhere, I don't know anything about github CI, but I presume we could create variants that build as C99, C11, C17, C23, similar to the variant you are working on where we build as C++.  Or maybe it's enough to build oldest (99) and newest (23).

Youw:
ok, I'll preapare a CI update for this

Issue #1786, opened 2026-03-14
Decouple Appveyor MSYS2 and Cygwin build from VS2015


mcuee:
Somehow Appveyor VS2015 build inlcude both Msys 2 build and Cygwin build. I think it is better to decouple.

https://github.com/libusb/libusb/blob/master/appveyor.yml
```
  -
    matrix:
      only:
        - image: Visual Studio 2015
          platform: Win32
          configuration: Release
    environment:
      toolset: v140
    install:
      - cmd: xcopy /S /I "%APPVEYOR_BUILD_FOLDER%" C:\msys64\home\appveyor\libusb
      - cmd: xcopy /S /I "%APPVEYOR_BUILD_FOLDER%" C:\cygwin\home\appveyor\libusb
    build_script:
      - cmd: msbuild "%APPVEYOR_BUILD_FOLDER%\msvc\libusb.sln" /m /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
      - cmd: C:\msys64\usr\bin\bash -l "%APPVEYOR_BUILD_FOLDER%\.private\appveyor_build.sh" MinGW
      - cmd: C:\cygwin\bin\bash -l "%APPVEYOR_BUILD_FOLDER%\.private\appveyor_build.sh" cygwin
    after_build:
      - cmd: 7z a "libusb-build_%APPVEYOR_BUILD_WORKER_IMAGE%_%PLATFORM%_%CONFIGURATION%.7z" tag_* README-build.txt build\%TOOLSET%\%PLATFORM%\%CONFIGURATION%\dll build\%TOOLSET%\%PLATFORM%\%CONFIGURATION%\lib build\%TOOLSET%\%PLATFORM%\%CONFIGURATION%\*.exe C:\msys64\home\appveyor\libusb-MinGW-Win32 C:\cygwin\home\appveyor\libusb-cygwin-Win32

  -
    matrix:
      only:
        - image: Visual Studio 2015
          platform: x64
          configuration: Release
    environment:
      toolset: v140
    install:
      - cmd: xcopy /S /I "%APPVEYOR_BUILD_FOLDER%" C:\msys64\home\appveyor\libusb
      - cmd: xcopy /S /I "%APPVEYOR_BUILD_FOLDER%" C:\cygwin64\home\appveyor\libusb
    build_script:
[...]

PR #1879, opened 2026-07-15
ci: reduce redundant Windows CI builds

Youw:
## Summary

- reduce AppVeyor from 12 jobs to 6 and skip its duplicate branch build when an open pull request exists;
- keep all four Visual Studio 2017/v141 jobs because they retain unique compiler coverage and the Release jobs also run the legacy MinGW and Cygwin builds;
- keep the Visual Studio 2022/v143 Release Win32 and x64 jobs, including their established `.7z` artifacts, because the release process consumes those trusted AppVeyor outputs;
- remove the redundant AppVeyor v143 Debug and v142 jobs;
- unify the GitHub Windows coverage in one matrix whose compact check names identify runner, Visual Studio host, MSVC toolset, architecture, and configuration family;
- pair each related Debug/Release configuration in one job, reducing the Windows matrix from 52 jobs to 26 while preserving all 52 compiler builds and all 52 existing artifact names; and
- cap that matrix at 12 parallel jobs so the other platform workflows retain runner capacity.

## Coverage

| CI host | MSVC toolset | Architectures | Configurations | Jobs |
| --- | --- | --- | --- | ---: |
| AppVeyor Visual Studio 2017 | v141 | Win32, x64 | Debug, Release | 4 |
| AppVeyor Visual Studio 2022 | v143 | Win32, x64 | Release | 2 |
| GitHub `windows-2022`, Visual Studio 2022/MSBuild 17 host | v142 (Visual Studio 2019 toolset) | Win32, x64 | Debug, Release | 2 paired |
| GitHub `windows-2022`, Visual Studio 2022/MSBuild 17 host | v143 (Visual Studio 2022 toolset) | Win32, x64, ARM64 | all 8 existing configurations | 12 paired |
| GitHub `windows-2025-vs2026`, Visual Studio 2026/MSBuild 18 host | v145 (Visual Studio 2026 toolset) | Win32, x64, ARM64 | all 8 existing configurations | 12 paired |

For example, the v142 check is now rendered as:

`Windows / W-2022 | VS 2022 | v142 (VS 2019) | Win32 | Debug + Release`

Each paired job has separate build steps and uploads separate per-configuration artifacts. Both builds are attempted, and the job fails if either build fails.

The AppVeyor Release artifact names used downstream remain unchanged:

- `libusb-build_Visual Studio 2022_Win32_Release.7z`
- `libusb-build_Visual Studio 2022_x64_Release.7z`

## Validation

- built MSVC v142 Debug and Release locally for Win32 and x64 with Visual Studio 2022/MSBuild 17;
- ran actionlint 1.7.12 across all GitHub Actions workflows;
- verified by scripted matrix expansion that GitHub Windows jobs decrease from 52 to 26 while the same 52 compiler/configuration/architecture combinations and the same 52 artifact names remain;
- verified representative full check names are 74, 84, and 96 characters, down from roughly 150 characters;
- verified the AppVeyor matrix has exactly six jobs, every job matches exactly one toolset rule, and both retained v141 Release jobs still run MSBuild, MinGW, and Cygwin; and
- ran `git diff --check`.

Related to #1876.
Related to #1765.
Related to #1786.

Assisted-by: Codex:GPT-5

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -4,26 +4,74 @@ on: [push, pull_request]
 
 jobs:
   build:
+    name: >-
+      ${{ matrix.compiler.runner }} |
+      ${{ matrix.compiler.host }} |
+      MSVC ${{ matrix.compiler.toolset }} (${{ matrix.compiler.toolset_visual_studio }} toolset) |
+      ${{ matrix.architecture }} |
+      ${{ matrix.configuration_family.label }}
     strategy:
+      fail-fast: false
+      # Leave runner capacity for the Linux, macOS, and MSYS2 workflows.
+      max-parallel: 12
       matrix:
-        os:
-          - 'windows-2022'
-          - 'windows-2025'
-        configuration:
-          - 'Debug'
-          - 'Release'
-          - 'Debug-MT'
-          - 'Release-MT'
-          - 'Debug-Hotplug'
-          - 'Release-Hotplug'
-          - 'Debug-Hotplug-MT'
-          - 'Release-Hotplug-MT'
+        compiler:
+          - runner: windows-2022
+            host: Visual Studio 2022/MSBuild 17 host
+            toolset: v143
+            toolset_visual_studio: Visual Studio 2022
+            artifact_prefix: build-windows-2022
+          - runner: windows-2025-vs2026
+            host: Visual Studio 2026/MSBuild 18 host
+            toolset: v145
+            toolset_visual_studio: Visual Studio 2026
+            # Retain the established artifact names despite using the explicit
+            # VS 2026 runner label.
+            artifact_prefix: build-windows-2025
+        configuration_family:
+          - label: Debug + Release
+            first: Debug
+            second: Release
+          - label: Debug-MT + Release-MT
+            first: Debug-MT
+            second: Release-MT
+          - label: Debug-Hotplug + Release-Hotplug
+            first: Debug-Hotplug
+            second: Release-Hotplug
+          - label: Debug-Hotplug-MT + Release-Hotplug-MT
+            first: Debug-Hotplug-MT
+            second: Release-Hotplug-MT
         architecture:
-          - 'Win32'
-          - 'x64'
-          - 'ARM64'
+          - Win32
+          - x64
+          - ARM64
+        include:
+          # VS 2022 hosts the VS 2019-era v142 toolset. Pairing Debug and
+          # Release keeps the four v142 builds in two clearly named jobs.
+          - compiler:
+              runner: windows-2022
+              host: Visual Studio 2022/MSBuild 17 host
+              toolset: v142
+              toolset_visual_studio: Visual Studio 2019
+              artifact_prefix: build-windows-2022-v142
+            configuration_family:
+              label: Debug + Release
+              first: Debug
+              second: Release
+            architecture: Win32
+          - compiler:
+              runner: windows-2022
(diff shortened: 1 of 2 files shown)


seanm:
@Youw I'm not a good person to review this, as I don't know this language/syntax at all...

Youw:
Let's not make it about syntax, but about the behavior overall

seanm:
What is mean is: I don't know how github CI works and don't know what all that stuff in those files is. I'm not qualified to review this.  It's like asking a Python programmer to review C code: sure, he'll recognise words like `if` and `for` but the details are different enough that he's not properly qualified.

appveyor.yml
```diff
@@ -1,14 +1,23 @@
 version: 1.0.{build}
 image:
   - Visual Studio 2017
-  - Visual Studio 2019
```
sonatique, line 4: So, if I understand correctly, the idea is to remove 2019 runs because they are redundant with 2017, correct?

I am wondering: what about 2026? Is it considered redundant with 2022?

Youw: Not because they redundant with 2017, but becase they covered Github Actions builds - a new build variant explicitly specifies PlatformToolset=v142 which is MSVC 2019 toolset (available even from a newer VIsual Studio installation).

MSVC 2026 and MSVC 2022 already covered by existing Github Actions builds.

Youw:
The main reason for this change - reduce the number of Appvoyer builds and effectively reduce the time CI runs (when I'm actively developing something I'm impacient :) ), since Appveyor do not paralelyze builds for free/open source accounts, but GithubActions does.

appveyor.yml
sonatique: Oh, OK,I see, thank you. A bit confusing to see some explicit Visual Studio version mentioned along with toolset versions, but OK.

sonatique (approved):

appveyor.yml
Youw: I'll see how can we make it more clear on the code/script level.

.github/workflows/windows.yml
```diff
@@ -1,29 +1,71 @@
...
 jobs:
   build:
+    name: >-
+      ${{ matrix.compiler.job_label }} |
+      ${{ matrix.architecture }} |
+      ${{ matrix.configuration_family.label }}
     strategy:
+      fail-fast: false
+      # Leave runner capacity for the Linux, macOS, and MSYS2 workflows.
+      max-parallel: 12
```
Youw:
```suggestion
      max-parallel: 16
```

Youw:
Even more updates landed, not the list of jobs under "Checks" is more explicit and visible about what build is run.
I also groupped Release/Debug build to reduce a number of parallel jobs - that actually reduces build time noticable, since GithubActions allowed only 20 build jobs in parallel, and we had 52 Windows jobs before that change (now 26)..

.github/workflows/windows.yml
```diff
@@ -1,29 +1,71 @@
...
 jobs:
   build:
+    name: >-
+      ${{ matrix.compiler.job_label }} |
+      ${{ matrix.architecture }} |
+      ${{ matrix.configuration_family.label }}
     strategy:
+      fail-fast: false
+      # Leave runner capacity for the Linux, macOS, and MSYS2 workflows.
+      max-parallel: 16
```
Youw:
```suggestion
      max-parallel: 18
```

Youw:
Merging this. If team find issues with this - will revert/update as a follow-up.

PR #1741, opened 2026-01-12
DeviceIoControl use async mode

slouken:
DeviceIoControl() will hang indefinitely when the Nacon PS4 Compact Controller is connected, so use async mode to safely call it with a timeout.

We're running into this when adding support for libusb to Steam, so this is a critical fix for us.

This PR is based on https://github.com/libusb/libusb/pull/1127, with the requested changes applied and code reformatted to match libusb style.

Issue #1163, opened 2022-07-07
Windows: libusb_get_device_list deadlock with 8Bitdo Pro2 controller


Megamouse:
When I attach an 8Bitdo Pro2 BT Controller via USB and activate the XInput mode (it can do 4 different modes) then I am stuck in libusb_get_device_list.
I don't have this issue with any other controller. The controller works fine over Bluetooth.

I'm on commit 4239bc3 - libusb 1.0.26
I've also tried 1.0.24, 1.0.25 and latest master... no difference

Curiously it proceeds if you disconnect the controller while it's stuck.

here's a backtrace:
![image](https://user-images.githubusercontent.com/23019877/177836999-7bf34008-36d1-4e23-baf0-a6c1a697227f.png)

Megamouse:
I'm on windows 11.

I hope this is the correct log
<details>
  <summary>Log with controller attached</summary>

  ```
[timestamp] [threadID] facility level [function call] <message>
--------------------------------------------------------------------------------
[ 0.001265] [0000d0ac] libusb: debug [libusb_init] libusb v1.0.26.11724
[ 0.001303] [0000d0ac] libusb: debug [usbi_add_event_source] add HANDLE 00000000000009E8 events 0
[ 0.001331] [0000d0ac] libusb: debug [usbi_io_init] using timer for timeouts
[ 0.001354] [0000d0ac] libusb: debug [usbi_add_event_source] add HANDLE 00000000000009E4 events 0
[ 0.001394] [0000d0ac] libusb: debug [get_windows_version] Windows 8 (or later) 64-bit
[ 0.001419] [0000d0ac] libusb: debug [htab_create] using 1021 entries hash table
[ 0.003035] [0000d0ac] libusb: info [winusbx_init] WinUSB DLL available (with isoch support)
[ 0.003107] [0000d0ac] libusb: info [winusbx_init] libusbK DLL is not available
[ 0.003882] [0000d0ac] libusb: info [windows_init] UsbDk backend is not available
[ 0.003945] [0000d0ac] libusb: warning [libusb_init] installing new context as implicit default
[ 0.004211] [0000d0ac] libusb: debug [libusb_get_device_list]
[ 0.004298] [0000c3fc] libusb: debug [windows_iocp_thread] I/O completion thread started
[ 0.016240] [0000d0ac] libusb: debug [winusb_get_device_list] allocating new device for session [A]
[...]

PR #1127, opened 2022-04-13
DeviceIoControl use async mode

weishaodi2zss:
DeviceIoControl maybe block in long time,we use async mode to control DeviceIoControl wait time.

PR #1882, opened 2026-07-16
windows: bound hub descriptor requests with a timeout

### LINKED ISSUE #1136: Libusb deadlocks with the Nacon wired controller
I'm currently using libusb via the [Cxbx-Reloaded emulator](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded) running under Windows 10, and when I try to configure the input buttons of my Nacon wired controller (VID_146B&PID_0603), the application freezes. This can be worked-around by disconnecting the controller when the freezing occurs, or by simply not having connected it in the first place. I debugged the code and I can confirm the issue is from libusb. When you attempt to open the input configuration GUI of Cxbxr, this code [here](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/5e42d181f2daf4a244450b1535223d4c71ea5e54/src/common/input/LibusbDevice.cpp#L130) is executed. Eventually, it reaches `winusb_get_device_list`, which ultimately calls [`DeviceIoControl`](https://github.com/libusb/libusb/blob/ba698478afc3d3a72644eef9fc4cd24ce8383a4c/libusb/os/windows_winusb.c#L1119). When that function is called on my Nacon controller, it never returns and thus it causes the observed deadlock. Note that this controller works fine with all my other applications, so I don't think it's my controller that's breaking.

PS: I think this issue is the same as [this one](https://github.com/libusb/libusb/issues/804) that was previously reported.

sonatique:
While reviewing #1741, I thought that we could solve pending issues better by starting from scratch based on what we learned.

Since I had not enough time available to start yet another PR, I told my conclusions and ideas to Claude AI , directing him to write coherent code.

This is the result. I reviewed it myself carefully before posting and it seems to do the job, and it really worth being considered as an alternative to #1741, as long as it actually solves the issue obviously : @slouken could please quickly test whether this code solves your issue in a desirable way? Thank you very much in advance for you help.

IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION issues a control transfer on the bus and the hub driver applies no timeout to it, so a device that stops responding on its control pipe hangs enumeration forever with the device list lock held. The Nacon PS4 Compact Controller report (#1136) additionally shows connection-info queries blocking the same way.

Issue the hub IOCTLs used during enumeration and string descriptor reads asynchronously against heap-owned request state and give up after 5 seconds. Since CancelIoEx() is asynchronous and METHOD_BUFFERED completion writes into the output buffer and OVERLAPPED afterwards, a timed-out request is parked on an orphan list and only freed once the kernel has finished with it.

Root hub queries were originally left synchronous on the assumption that they never reach a device, but the Nacon report (#1136) implicates IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, so init_root_hub()'s node and per-port queries now go through the same bounded helper, giving up on the port loops at the first timeout. Devices whose initialization times out are retried with a 30-second backoff instead of stalling every scan, and string descriptor requests get a per-device cooldown after a timeout.

Fixes: #1136
Replaces: #1741

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -1383,6 +1383,9 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 	// full timeout all over again
 	if (hub_timeout_backoff_active(priv->dev_id)) {
 		usbi_dbg(DEVICE_CTX(dev), "skipping initialization of '%s' until its timeout backoff expires", priv->dev_id);
+		// the caller's retained parent reference is only transferred to the
+		// device further down, so it must be released on this early exit
+		libusb_unref_device(parent_dev);
 		return LIBUSB_ERROR_NO_DEVICE;
 	}
 
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -895,8 +895,15 @@ static void winusb_exit(struct libusb_context *ctx)
 
 /*
  * fetch and cache all the config descriptors through I/O
+ *
+ * Returns LIBUSB_ERROR_TIMEOUT if a descriptor request timed out, in which case
+ * probing was abandoned: the control pipe is most likely wedged, every further
+ * request would pay the full timeout, and bNumConfigurations is device-supplied
+ * (not sanitized until usbi_sanitize_device()) so the multiplied stall could
+ * otherwise last for minutes. Ordinary per-descriptor failures are skipped as
+ * before and do not affect the return value.
  */
-static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handle)
+static int cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handle)
 {
 	struct libusb_context *ctx = DEVICE_CTX(dev);
 	struct winusb_device_priv *priv = (struct winusb_device_priv *)usbi_get_device_priv(dev);
@@ -909,14 +916,14 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
 
 	num_configurations = dev->device_descriptor.bNumConfigurations;
 	if (num_configurations == 0)
-		return;
+		return LIBUSB_SUCCESS;
 
 	assert(sizeof(USB_DESCRIPTOR_REQUEST) == USB_DESCRIPTOR_REQUEST_SIZE);
 
 	priv->config_descriptor = (PUSB_CONFIGURATION_DESCRIPTOR *)calloc(num_configurations, sizeof(PUSB_CONFIGURATION_DESCRIPTOR));
 	if (priv->config_descriptor == NULL) {
 		usbi_err(ctx, "could not allocate configuration descriptor array for '%s'", priv->dev_id);
-		return;
+		return LIBUSB_ERROR_NO_MEM;
 	}
 
 	for (i = 0; i <= num_configurations; i++) {
@@ -940,7 +947,10 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
 		// coverity[tainted_data_argument]
 		if (!hub_device_io_control(ctx, hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &cd_buf_short, size,
 			&cd_buf_short, size, &ret_size)) {
-			usbi_info(ctx, "could not access configuration descriptor %u (dummy) for '%s': %s", i, priv->dev_id, windows_error_str(0));
+			DWORD error = GetLastError();
+			usbi_info(ctx, "could not access configuration descriptor %u (dummy) for '%s': %s", i, priv->dev_id, windows_error_str(error));
+			if (error == ERROR_SEM_TIMEOUT)
+				return LIBUSB_ERROR_TIMEOUT;
 			continue;
 		}
 
@@ -966,7 +976,12 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
 
 		if (!hub_device_io_control(ctx, hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, cd_buf_actual, size,
 			cd_buf_actual, size, &ret_size)) {
-			usbi_err(ctx, "could not access configuration descriptor %u (actual) for '%s': %s", i, priv->dev_id, windows_error_str(0));
+			DWORD error = GetLastError();
+			usbi_err(ctx, "could not access configuration descriptor %u (actual) for '%s': %s", i, priv->dev_id, windows_error_str(error));
+			if (error == ERROR_SEM_TIMEOUT) {
+				safe_free(cd_buf_actual);
+				return LIBUSB_ERROR_TIMEOUT;
+			}
 			continue;
 		}
 
@@ -989,6 +1004,8 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
 		priv->config_descriptor[i] = cd_data;
 		cd_buf_actual = NULL;
 	}
+
+	return LIBUSB_SUCCESS;
 }
 
 #define ROOT_HUB_FS_CONFIG_DESC_LENGTH		0x19
@@ -1357,10 +1374,12 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 			dev->device_descriptor.bNumConfigurations, priv->active_config,	priv->dev_id);
 
 		// Cache as many config descriptors as we can
-		cache_config_descriptors(dev, hub_handle);
+		r = cache_config_descriptors(dev, hub_handle);
 
-		// In their great wisdom, Microsoft decided to BREAK the USB speed report between Windows 7 and Windows 8
-		if (windows_version >= WINDOWS_8) {
+		// In their great wisdom, Microsoft decided to BREAK the USB speed report between Windows 7 and Windows 8.
+		// Skip the query when descriptor caching hit a timeout: the same handle serves this request,
+		// so it would most likely just pay another full timeout for nothing.
+		if ((windows_version >= WINDOWS_8) && (r != LIBUSB_ERROR_TIMEOUT)) {
 			conn_info_v2.ConnectionIndex = (ULONG)port_number;
 			conn_info_v2.Length = sizeof(USB_NODE_CONNECTION_INFORMATION_EX_V2);
 			conn_info_v2.SupportedUsbProtocols.Usb300 = 1;
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -721,13 +721,157 @@ static int winusb_init(struct libusb_context *ctx)
 }
 
 /*
+ * Hub IOCTLs that translate into control transfers on the bus - most notably
+ * IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION - are not given a timeout by the
+ * hub driver. A device that stops responding on its control pipe therefore hangs
+ * the calling thread forever, which during enumeration means hanging the whole
+ * library (observed in the wild, e.g. with the Nacon PS4 Compact Controller, #1127).
+ * Such calls are bounded by issuing them asynchronously and giving up after this
+ * many milliseconds. Well-behaved requests complete in milliseconds; 5 seconds
+ * matches the timeout other USB stacks apply to control transfers (e.g. Linux's
+ * USB_CTRL_GET_TIMEOUT) and is generous enough for hubs waking from selective
+ * suspend.
+ */
+#define HUB_IOCTL_TIMEOUT_MS	5000
+
+// After CancelIoEx(), how long to wait for a timed-out request to reach a terminal
+// state before parking it on the orphan list instead of freeing it
+#define HUB_IOCTL_CANCEL_GRACE_MS	250
+
+struct hub_ioctl_request {
+	OVERLAPPED overlapped;	// hEvent holds a per-request manual-reset event
+	struct list_head list;	// linkage in orphaned_requests
+	// the request's in/out data buffer immediately follows the struct
+};
+
+static usbi_mutex_static_t orphaned_requests_lock = USBI_MUTEX_INITIALIZER;
+static struct list_head orphaned_requests = { &orphaned_requests, &orphaned_requests };
+
+// Free timed-out requests whose I/O has since reached a terminal state. Requests
+// still owned by the kernel are left alone - their OVERLAPPED, event and buffer
+// must stay valid until the kernel is done with them.
+static void reap_orphaned_requests(struct libusb_context *ctx)
+{
+	struct hub_ioctl_request *req, *next_req;
+
+	usbi_mutex_static_lock(&orphaned_requests_lock);
+	list_for_each_entry_safe(req, next_req, &orphaned_requests, list, struct hub_ioctl_request) {
+		if (HasOverlappedIoCompleted(&req->overlapped)) {
+			usbi_dbg(ctx, "orphaned hub request %p has completed, freeing it", (void *)req);
+			list_del(&req->list);
+			CloseHandle(req->overlapped.hEvent);
+			free(req);
+		}
+	}
+	usbi_mutex_static_unlock(&orphaned_requests_lock);
+}
+
+/*
+ * Synchronous DeviceIoControl() with a timeout, for hub requests issued outside the
+ * regular transfer path. The handle must have been opened with FILE_FLAG_OVERLAPPED,
+ * which in turn means every DeviceIoControl() on that handle must go through here
+ * (calling it without an OVERLAPPED on an overlapped handle is not allowed).
+ *
+ * The request state (OVERLAPPED, event and data buffer) is heap-allocated and owned
+ * by the request itself: CancelIoEx() only *requests* cancellation, and the
+ * completion of a METHOD_BUFFERED request still writes into the output buffer and
+ * the OVERLAPPED afterwards. On timeout the state is therefore either freed once the
+ * cancellation is observed to complete, or handed over to the orphan list to be
+ * freed later - never released while the kernel may still reference it. This is
+ * also why the caller's buffers are never handed to the kernel directly.
+ *
+ * Returns TRUE on success, with the output copied to out_buf and *bytes_returned
+ * set. Returns FALSE on failure, with *bytes_returned zeroed and the cause
+ * available via GetLastError() (ERROR_SEM_TIMEOUT for a timeout).
+ */
+static BOOL hub_device_io_control(struct libusb_context *ctx, HANDLE handle, DWORD ioctl_code,
+	const void *in_buf, DWORD in_size, void *out_buf, DWORD out_size, DWORD *bytes_returned)
+{
+	struct hub_ioctl_request *req;
+	unsigned char *req_buf;
+	DWORD error, bytes;
+	BOOL ret;
+
+	*bytes_returned = 0;
+
+	reap_orphaned_requests(ctx);
+
+	// All hub IOCTLs are METHOD_BUFFERED: the kernel uses a single intermediate
+	// buffer for input and output, so one request buffer can stand in for both
+	// of the caller's (possibly aliased) buffers
+	assert((ioctl_code & 0x3) == METHOD_BUFFERED);
+
+	req = (struct hub_ioctl_request *)calloc(1, sizeof(*req) + MAX(in_size, out_size));
+	if (req == NULL) {
+		SetLastError(ERROR_NOT_ENOUGH_MEMORY);
+		return FALSE;
+	}
+	req_buf = (unsigned char *)(req + 1);
+
+	// Manual-reset, so that waiting on it does not consume the completion state
+	req->overlapped.hEvent = CreateEventA(NULL, TRUE, FALSE, NULL);
+	if (req->overlapped.hEvent == NULL) {
+		error = GetLastError();
+		free(req);
+		SetLastError(error);
+		return FALSE;
+	}
+
+	if (in_size != 0)
+		memcpy(req_buf, in_buf, in_size);
+
+	if (!DeviceIoControl(handle, ioctl_code, (in_size != 0) ? req_buf : NULL, in_size,
+			req_buf, out_size, NULL, &req->overlapped)) {
+		error = GetLastError();
+		if (error != ERROR_IO_PENDING) {
+			// Failed synchronously - no I/O was ever queued
+			CloseHandle(req->overlapped.hEvent);
+			free(req);
+			SetLastError(error);
+			return FALSE;
+		}
+	}
+	// Even on synchronous success the completion state and event are set,
+	// so a single wait-and-collect path handles both cases
+
+	if (WaitForSingleObject(req->overlapped.hEvent, HUB_IOCTL_TIMEOUT_MS) == WAIT_OBJECT_0) {
+		ret = GetOverlappedResult(handle, &req->overlapped, &bytes, FALSE);
+		error = ret ? ERROR_SUCCESS : GetLastError();
+		if (ret) {
+			if (bytes > out_size)	// should not happen for METHOD_BUFFERED
+				bytes = out_size;
+			memcpy(out_buf, req_buf, bytes);
+			*bytes_returned = bytes;
+		}
+		CloseHandle(req->overlapped.hEvent);
+		free(req);
+		SetLastError(error);
+		return ret;
+	}
+
+	usbi_warn(ctx, "IOCTL 0x%08lX timed out after %d ms, cancelling it",
+		ULONG_CAST(ioctl_code), HUB_IOCTL_TIMEOUT_MS);
+	CancelIoEx(handle, &req->overlapped);
+	if (WaitForSingleObject(req->overlapped.hEvent, HUB_IOCTL_CANCEL_GRACE_MS) == WAIT_OBJECT_0) {
+		CloseHandle(req->overlapped.hEvent);
+		free(req);
+	} else {
+		// The kernel still owns the request state: park it and free it once a
+		// later call (or winusb_exit()) observes it reached a terminal state
+		usbi_mutex_static_lock(&orphaned_requests_lock);
+		list_add_tail(&req->list, &orphaned_requests);
+		usbi_mutex_static_unlock(&orphaned_requests_lock);
+	}
+	SetLastError(ERROR_SEM_TIMEOUT);
+	return FALSE;
+}
+
+/*
 * exit: libusb backend deinitialization function
 */
 static void winusb_exit(struct libusb_context *ctx)
 {
 	int i;
-
-	UNUSED(ctx);
 
 	usbi_mutex_destroy(&autoclaim_lock);
 
@@ -737,6 +881,15 @@ static void winusb_exit(struct libusb_context *ctx)
 	}
 
 	exit_dlls();
+
+	// Last chance to release requests that timed out and were never observed to
+	// complete. Anything still pending is deliberately kept allocated: freeing it
+	// would let the kernel scribble over reused memory when it finally completes.
+	reap_orphaned_requests(ctx);
+	usbi_mutex_static_lock(&orphaned_requests_lock);
+	if (!list_empty(&orphaned_requests))
+		usbi_warn(ctx, "keeping timed-out hub request(s) never released by the kernel");
+	usbi_mutex_static_unlock(&orphaned_requests_lock);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


libusb/os/windows_winusb.c
```diff
@@ -782,8 +935,8 @@ static void cache_config_descriptors(struct libusb_device *dev, HANDLE hub_handl
 		// Dummy call to get the required data size. Initial failures are reported as info rather
 		// than error as they can occur for non-penalizing situations, such as with some hubs.
 		// coverity[tainted_data_argument]
-		if (!DeviceIoControl(hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &cd_buf_short, size,
-			&cd_buf_short, size, &ret_size, NULL)) {
+		if (!hub_device_io_control(ctx, hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &cd_buf_short, size,
+			&cd_buf_short, size, &ret_size)) {
 			usbi_info(ctx, "could not access configuration descriptor %u (dummy) for '%s': %s", i, priv->dev_id, windows_error_str(0));
```
Youw: [P2] Bound the timeout per device, not per configuration

This `continue` makes the five-second deadline apply separately to every configuration. `bNumConfigurations` is device-supplied and is not sanitized until `usbi_sanitize_device()` after this function; a value of 255 can therefore keep one enumeration here for about 22 minutes (including cancellation grace) and can park an orphan request/event on each iteration. After `ERROR_SEM_TIMEOUT`, stop probing this control pipe and propagate that state so `init_device()` does not issue the V2 IOCTL on the same handle; retain `continue` only for ordinary per-descriptor failures. The actual-descriptor failure branch below needs the same treatment.

Generated by Codex (GPT-5).

libusb/os/windows_winusb.c
```diff
@@ -1148,8 +1304,8 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 		conn_info.ConnectionIndex = (ULONG)port_number;
 		// coverity[tainted_data_argument]
 
-		if (!DeviceIoControl(hub_handle, IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, &conn_info, sizeof(conn_info),
-			&conn_info, sizeof(conn_info), &size, NULL)) {
+		if (!hub_device_io_control(ctx, hub_handle, IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, &conn_info, sizeof(conn_info),
```
Youw, line 1462: [P1] Avoid paying this timeout again on every scan

The concrete Nacon trace in #1136 points to this exact `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX` call. On timeout this branch returns before `priv->initialized` is set, while the enumeration caller invokes `init_device()` whenever that flag remains false. A polling application will therefore incur another five-second stall on every `libusb_get_device_list()`, and a request that never reaches terminal cancellation can add another orphan each time. Record a timeout/backoff or known-bad state keyed to the device instance (cleared on removal/reconnection) so later scans skip or throttle the same failed initialization.

Generated by Codex (GPT-5).

libusb/os/windows_winusb.c
```diff
@@ -718,15 +718,159 @@ static int winusb_init(struct libusb_context *ctx)
...
+ * IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION - are not given a timeout by the
+ * hub driver. A device that stops responding on its control pipe therefore hangs
+ * the calling thread forever, which during enumeration means hanging the whole
+ * library (observed in the wild, e.g. with the Nacon PS4 Compact Controller, #1127).
+ * Such calls are bounded by issuing them asynchronously and giving up after this
+ * many milliseconds. Well-behaved requests complete in milliseconds; 5 seconds
+ * matches the timeout other USB stacks apply to control transfers (e.g. Linux's
+ * USB_CTRL_GET_TIMEOUT) and is generous enough for hubs waking from selective
+ * suspend.
+ */
```
Youw, line 735: [P3] Point the rationale at the IOCTL implicated by the Nacon report

The Nacon report is #1136, and its linked blocking source line is `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX`, not the descriptor request. That distinction matters because the node-information timeout takes the repeated-initialization path discussed below.

```suggestion
/*
 * Hub IOCTLs used during enumeration can block indefinitely when a device
 * stops responding. The Nacon controller report (#1136) stalled in
 * IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX, while descriptor requests
 * translate into control transfers for which the hub driver applies no timeout.
 * Such calls are bounded by issuing them asynchronously and giving up after this
 * many milliseconds. Well-behaved requests complete in milliseconds; 5 seconds
 * matches the timeout other USB stacks apply to control transfers (e.g. Linux's
 * USB_CTRL_GET_TIMEOUT) and is generous enough for hubs waking from selective
 * suspend.
 */
```

Generated by Codex (GPT-5).

Youw (changes requested):
Codex reviewed exact head `e8c712d04508d236b2437d60c19c11df8ea376bc` against `master` (`c8da8731eca73d9bb96a95bcd945a18ccf2db477`), including the history in #1127 and #1741, the affected-device reports in #1136 and #1163, and Microsoft's overlapped-I/O/cancellation contracts.

The underlying problem is real and worth fixing: one hub `DeviceIoControl()` can block enumeration while the device-list path is serialized. The Nacon report in #1136 identifies `IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX`; #1163 reports the same unplug-to-unblock behavior with an 8BitDo controller. The first implementation in #1127 stopped the infinite wait for testers, but also produced 15-20 second scans and `stress`/`stress_mt` failures. #1741 retained several correctness blockers: stack-owned `OVERLAPPED` and caller buffers outliving asynchronous `CancelIoEx()`, a shared/auto-reset event race, lost completion byte counts and errors, and an unbounded string-descriptor path.

This rewrite resolves those earlier lifetime and concurrency defects well:

- each request has a private manual-reset event and heap-owned `OVERLAPPED` plus METHOD_BUFFERED data;
- timed-out state remains alive until terminal completion instead of being freed after merely requesting cancellation;
- completion byte counts and `GetLastError()` are propagated correctly;
- configuration and string-descriptor paths are covered; and
- root-hub queries remain synchronous, with a documented five-second rationale.

Two timeout-policy blockers remain in the inline findings. The five-second limit is currently per IOCTL rather than per bad device: it can multiply across configurations within one scan, and the historically implicated node-information failure is retried from scratch on every later device-list scan. If cancellation never reaches terminal state, those retries also accumulate orphan requests and event handles.

Runtime validation is still required after those paths are corrected. CI can compile the code but cannot prove that the affected hub driver returns `ERROR_IO_PENDING`, that the timeout actually fires, or that enumeration and later scans remain responsive. A current-head Nacon/8BitDo debug run should demonstrate the pending -> timeout/cancel path, successful return from enumeration, bounded subsequent scans, and stable orphan/handle counts.

Verification performed by Codex:

- all 57 current GitHub Actions check-runs passed;
- a full VS2022 v143 Debug/Win32 solution build passed with the project's `/W4 /WX` settings;
- direct MinGW-w64 x86_64 and i686 compilation of the changed source passed without warnings in the new code; and
- AppVeyor's legacy MSVC/MinGW/Cygwin jobs were still queued at review time, so that remaining CI gate was not treated as evidence yet.

References: [#1741 lifetime review](https://github.com/libusb/libusb/pull/1741#pullrequestreview-4706897129), [#1741 design review](https://github.com/libusb/libusb/pull/1741#pullrequestreview-4713246915), [CancelIoEx](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-cancelioex), and [DeviceIoControl](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-deviceiocontrol).

Generated by Codex (GPT-5).

libusb/os/windows_winusb.c
sonatique: A per-device backoff keyed by device instance ID: when initialization fails on a timeout (the conn-info path from #1136, or a root-hub timeout), the device is skipped for 30 seconds instead of stalling every libusb_get_device_list() by the full timeout, which also bounds orphan-request accumulation (secondary concern). 
 
 The backoff is not cleared on reconnection, because a replug can't cheaply be distinguished from a still-wedged device, so a replugged healthy device can stay invisible for up to 30 s. Successful initialization clears the entry, and winusb_exit() frees the list.

libusb/os/windows_winusb.c
sonatique: cache_config_descriptors() now abandons the device at the first request timeout instead of continuing through up to 255 device-supplied configurations,  the 22-minute  case is gone; the max is one timeout per device. It returns LIBUSB_ERROR_TIMEOUT on abandonment, and init_device() uses that to skip the V2 speed query it would otherwise issue. Ordinary per-descriptor failures keep their skip-and-continue behavior, as asked.

libusb/os/windows_winusb.c
sonatique: The comment now names IOCTL_USB_GET_NODE_CONNECTION_INFORMATION_EX as the request implicated by #1136. 
But, it has more serious implications: that report discard the idea of leaving init_root_hub() synchronous ("root hub queries never reach a device"), because a Nacon plugged directly into a root-hub port (common for a game controller) would hang the per-port speed loops identically. 

So init_root_hub() now goes through the bounded helper too, its two speed loops give up on the first timeout instead of probing every remaining port, and a node-info timeout returns LIBUSB_ERROR_TIMEOUT so the caller can distinguish it from an access failure.

sonatique:
@Youw :  pushed 3 new commit addressing your comments.
The #1136 evidence prompted extending coverage to init_root_hub() (beyond what Codex literally asked at P3)
Codex asked for "skip or throttle" (P1): this is the throttle variant.

mcuee:
I like this idea even though I may not be qualified to review the code changes.

mcuee (approved):

libusb/os/windows_winusb.c
```diff
@@ -1094,6 +1376,14 @@ static int init_device(struct libusb_device *dev, struct libusb_device *parent_d
 	if (priv->initialized)
 		return LIBUSB_SUCCESS;
 
+	// Skip devices whose recent hub IOCTL timeout put them into their backoff
+	// window: retrying immediately would stall every device list scan by the
+	// full timeout all over again
+	if (hub_timeout_backoff_active(priv->dev_id)) {
+		usbi_dbg(DEVICE_CTX(dev), "skipping initialization of '%s' until its timeout backoff expires", priv->dev_id);
+		return LIBUSB_ERROR_NO_DEVICE;
```
Youw, line 1397: [P2] Release the caller-owned parent reference on the backoff exit

`get_ancestor()` gives the GEN_PASS caller a retained reference. Normal initialization transfers that reference at `dev->parent_dev = parent_dev`, while the already-initialized branch explicitly releases it. This new early return runs before either path, and the GEN_PASS failure cleanup does not unref `parent_dev`; every device-list scan during the 30-second backoff can therefore increment the parent's refcount and eventually pin the hub/device after disconnect. Release the caller-owned reference before returning (`libusb_unref_device()` is NULL-safe):

```suggestion
		libusb_unref_device(parent_dev);
		return LIBUSB_ERROR_NO_DEVICE;
```

Generated by Codex (GPT-5.6 Sol), assisted by Claude Code (Claude Fable 5).

Youw (changes requested):
Codex (GPT-5.6 Sol) and Claude Code (Claude Fable 5) independently reviewed exact head `2591c3a56e7c356d6062586f0d8572e8aa69097f` against the PR base, then cross-checked the results against the caller and core refcount paths.

The cross-check retained one actionable P2 blocker: the new backoff early return runs before `init_device()` takes ownership of the caller's retained `parent_dev` reference. A polling application can therefore add another parent reference on every scan during the backoff window and pin the parent hub/device after disconnect. The inline finding includes the minimal fix.

The three existing review concerns are addressed by the current commits and can be resolved:

- Configuration probing stops at the first descriptor timeout, skips the V2 request on that handle, and does not multiply the timeout across all configurations.
- Repeated enumeration is throttled per device for 30 seconds, satisfying the requested skip-or-throttle behavior instead of paying the timeout on every scan.
- Root-hub handles are opened for overlapped I/O, all relevant root-hub IOCTLs use the bounded helper, timeout exits stop the port loops, and the rationale now names the historically implicated IOCTL accurately.

Fable 5 also raised two low-severity hypotheses that the combined cross-check did not retain as findings. Hub-wide throttling is not justified without hardware evidence because the IOCTL targets a specific child port and a hub-wide backoff could hide healthy siblings. The string path cannot pay two timeout waits in one call: failure of the language-ID request returns immediately before the requested-string IOCTL. Those behaviors remain suitable targets for the affected-device runtime test rather than code-review blockers.

Static verification covered the full diff, helper lifetime/cancellation, all hub-handle call sites, initialization and enumeration ownership, backend init/exit lifetime, and `git diff --check`. Current GitHub Actions checks and the PR AppVeyor status pass; the separate AppVeyor branch status remains failed. Hardware validation with a wedged Nacon/8BitDo-class device is still required to observe pending -> timeout/cancel behavior, bounded later scans, sibling-port behavior, and stable orphan/handle counts.

Related to #1127, #1136, #1163, and #1741.

Generated by Codex (GPT-5.6 Sol), assisted by Claude Code (Claude Fable 5).

libusb/os/windows_winusb.c
sonatique: Confirmed and fixed. I verified the contract independently: the GEN_PASS caller's retained reference is only transferred at dev->parent_dev = parent_dev, the already-initialized case releases it in the caller, and nothing releases it when init_device() fails: so this exit leaked one parent reference per scan during the backoff window. Applied your suggested fix.

Verifying this also turned up the same leak in four pre-existing early returns of init_device() (parent not a hub, first-ancestor mismatch, which additionally leaks the ancestor reference itself, ancestor not found, bus number not found). Those predate this PR and are cold paths, which is presumably why they were never noticed. Fixed in a separate commit so it can be dropped from this PR if we prefer it.

mcuee:
I do not have the right device to trigger the issue mentioned in PR #1741. But at least there is no issue.

```
 libusb_pr1882  $Env:LIBUSB_DEBUG=3
 libusb_pr1882  .\build\v145\x64\Release-Hotplug-MT\listdevs.exe -v
libusb: info [winusbx_init] WinUSB DLL available (with isoch support)
libusb: info [windows_init] UsbDk backend is not available
24ae:4057 (bus 3, device 13) path: 2.3.2.3
    manufacturer = SEMICO
    product = USB Keyboard
7768:642c (bus 3, device 11) path: 2.3.2.1
    manufacturer = Dongle
    product = Inphic IN9 SLE
    serial_number = 20240401
1a2c:7fff (bus 3, device 12) path: 2.3.2.2
    manufacturer =
    product = 2.4G Wireless Device
    serial_number = 8545124150344637
047f:c056 (bus 1, device 3) path: 6
    manufacturer = Plantronics
    product = Plantronics Blackwire 3220 Series
    serial_number = BB305534D79EBD418B6757528EAC0C19
413c:b06e (bus 3, device 7) path: 2.5libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_413C&PID_B06E\2110': [31] A device attached to the system is not functioning.
libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_413C&PID_B06E\2110': [31] A device attached to the system is not functioning.

413c:b06f (bus 3, device 10) path: 2.3.5libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_413C&PID_B06F\0123456789ABCDEF': [31] A device attached to the system is not functioning.
libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_413C&PID_B06F\0123456789ABCDEF': [31] A device attached to the system is not functioning.

1bcf:2bb4 (bus 1, device 1) path: 3
    manufacturer = CN03N9728LU00548DCG8A00
    product = Integrated_Webcam_FHD
    serial_number = 01.00.00
0bda:402e (bus 3, device 9) path: 2.3.4
    manufacturer = Generic
    product = USB Audio
    serial_number = 200901010001
0424:2740 (bus 3, device 14) path: 2.3.2.5
    manufacturer = Microchip Tech
    product = Hub Controller
8087:0033 (bus 1, device 2) path: 10
0b95:772b (bus 1, device 5) path: 7.1
    manufacturer = ASIX Elec. Corp.
    product = AX88772C
    serial_number = C1B13D
0bda:8153 (bus 3, device 3) path: 4.4
    manufacturer = Realtek
    product = USB 10/100/1000 LAN
    serial_number = 111000001
16c0:05dc (bus 1, device 6) path: 7.4
    manufacturer = www.fischl.de
    product = USBasp
    serial_number = 0001
0bda:0413 (bus 3, device 2) path: 4.3libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_0BDA&PID_0413\9&18D5C933&0&3': [31] A device attached to the system is not functioning.
libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_0BDA&PID_0413\9&18D5C933&0&3': [31] A device attached to the system is not functioning.

0bda:0487 (bus 3, device 1) path: 4
    manufacturer = Dell Inc.
    product = Dell dock
05e3:0608 (bus 1, device 4) path: 7
    product = USB2.0 Hub
8086:15f0 (bus 3, device 0)
0bda:5487 (bus 3, device 4) path: 2
    manufacturer = Dell Inc.
    product = Dell dock
8086:7ec0 (bus 2, device 0)
0424:2744 (bus 3, device 8) path: 2.3.2
    manufacturer = Microchip Tech
    product = USB2744
8086:7e7d (bus 1, device 0)
0bda:5413 (bus 3, device 6) path: 2.3
    manufacturer = Dell Inc.
    product = Dell dock
0424:5744 (bus 3, device 5) path: 4.3.2libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_0424&PID_5744\A&3AD159B5&0&2': [31] A device attached to the system is not functioning.
libusb: warning [winusb_get_device_string] could not retrieve language IDs for 'USB\VID_0424&PID_5744\A&3AD159B5&0&2': [31] A device attached to the system is not functioning.
```

libusb/os/windows_winusb.c
```diff
@@ -2429,8 +2744,8 @@ static int winusb_fetch_string_descriptor(libusb_device *dev,
 	sd.req.SetupPacket.wIndex = priv->langid;
 	sd.req.SetupPacket.wLength = (USHORT)sizeof(sd.desc);
 
-	BOOL rv = DeviceIoControl(hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &sd, size,
-		&sd, size, &ret_size, NULL);
+	BOOL rv = hub_device_io_control(ctx, hub_handle, IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION, &sd, size,
+		&sd, size, &ret_size);
 	CloseHandle(hub_handle);
 	if (!rv) {
```
Youw, line 2780: [P2] Throttle timed-out actual string descriptor requests

Once the language ID has been cached, this failure path only returns `LIBUSB_ERROR_IO`; it does not set `langid_unavailable` or any timeout cooldown. The core caches only successful device strings, and config/interface strings also reach this helper directly, so every later manufacturer/product/serial/config/interface lookup can issue another five-second IOCTL. If cancellation never reaches terminal completion, each retry can also retain another request/event on the orphan list. Capture the error before `CloseHandle()` and, for `ERROR_SEM_TIMEOUT`, record a per-device descriptor cooldown or unavailable state checked before reopening the hub; ordinary descriptor failures can remain retryable.

libusb/os/windows_winusb.c
```diff
@@ -720,15 +720,251 @@ static int winusb_init(struct libusb_context *ctx)
...
+// Whether initialization of this device should be skipped because a recent
+// hub IOCTL timeout put it into its backoff window. Expired entries are
+// dropped, so the next attempt after the window goes through again.
+static bool hub_timeout_backoff_active(const char *dev_id)
+{
+	struct hub_timeout_backoff *backoff;
+	bool active = false;
+
+	usbi_mutex_static_lock(&hub_timeout_backoffs_lock);
+	backoff = hub_timeout_backoff_find(dev_id);
```
Youw: [P3] Prune expired entries that never reappear

This removes an expired entry only when the same device ID is queried again. If a timed-out device is unplugged permanently, its backoff allocation remains in the global list until the last Windows backend exit; a long-running process that sees many distinct failed devices therefore accumulates entries indefinitely, and every lookup becomes progressively more expensive. Prune all expired entries while traversing the list under the lock (or from an appropriate removal path), while retaining the existing 30-second behavior for matching devices.

Youw (changes requested):
Codex reviewed current head 3cf9bac8df93828cf02d8d464208c57ddeeb121c against base 6bb97402df0ec0c09defcbd4f5146447c7f7cc53 and cross-checked the result against the complete existing conversation.

All four earlier inline findings are addressed and remain resolved. Two new issues are left inline: timed-out actual string descriptor requests are not throttled after the language ID is cached, and expired backoff entries for devices that never reappear are retained until backend exit.

All current CI checks pass and `git diff --check` is clean. Affected-device validation with a Nacon/8BitDo-class device is still outstanding; the available runtime report covers healthy devices only.

The PR description also needs an update: it still says root-hub queries remain synchronous, whereas the later commits correctly route root-hub IOCTLs through the overlapped timeout helper.

Related to #1136 and #1741.

sonatique:
@Youw 
> Codex reviewed current head [3cf9bac]

Both inline findings addressed on the newer commits.

Strings: a timed-out request now arms a per-device-object cooldown (string_backoff_expiry) checked before opening the hub. Deliberately separate from the enumeration backoff, so string trouble never hides an otherwise enumerable device from the list.
Backoff list: hub_timeout_backoff_active() now prunes every expired entry it traverses, not just a matching one.

Please re-review

Youw:
@sonatique this PR is for `sonatique:windows-hub-ioctl-timeout`, but the two commits with fixes you've mentioned are pushed to `libusb:windows-hub-ioctl-timeout`.

To make this work you need either create a PR into `libsub:master` from `libusb:windows-hub-ioctl-timeout` or push the fixes to `sonatique:windows-hub-ioctl-timeout`.

sonatique:
@Youw :
> @sonatique this PR is for `sonatique:windows-hub-ioctl-timeout`, but the two commits with fixes you've mentioned are pushed to `libusb:windows-hub-ioctl-timeout`.

Sorry: I pushed to the wrong repos. Corrected now, thanks.

libusb/os/windows_winusb.c
```diff
@@ -2377,8 +2697,22 @@ static int winusb_fetch_string_descriptor(libusb_device *dev,
...
 	}
 
+	// Likewise during the cooldown that follows a string request timeout. The
+	// core caches only successful lookups and config/interface strings reach
+	// this helper directly, so without a cooldown a wedged control pipe would
+	// stall every string query by the full request timeout. The cooldown is
+	// per device object rather than the enumeration backoff on purpose: a
+	// device without strings can still be enumerable and usable, so it must
+	// not disappear from the device list.
+	if ((priv->string_backoff_expiry != 0) && (GetTickCount64() < priv->string_backoff_expiry)) {
```
Youw: [P2] Synchronize the string cooldown state

Codex found that this ULONGLONG deadline is also written below without synchronization. libusb promises complete thread safety, and the device/config/interface string getters can be called concurrently for the same device; on Win32 an ordinary 64-bit load/store may tear, and in C this is a data race. A caller can therefore miss an armed cooldown and issue another five-second IOCTL (and potentially retain another orphan), which is the retry this field is intended to prevent. Please protect the language-ID/cooldown/request state with a per-device lock, or use properly aligned interlocked accesses for the deadline; the pre-existing racy langid cache does not make this access safe.

Youw (changes requested):
Codex reviewed exact head a288bdb06bceff32edb4db5b144f9b3bded74281 against its merge base and current master, including the complete prior discussion and the newly included fixes.

The two July 29 findings are addressed and their review threads have been resolved:

- timed-out actual string descriptor requests now arm a per-device cooldown before later hub opens; and
- backoff lookup now prunes every expired entry it traverses.

One remaining P2 concurrency issue is left inline: the new 64-bit cooldown deadline is read and written without synchronization even though these public getters may run concurrently. That can tear on Win32 and can miss the throttle the field is meant to enforce.

Repository attribution also needs cleanup before merge. The eight commits after 90b7f996 lack the required Assisted-by trailer, and the PR body credits only Claude even though several commits explicitly implement Codex review findings. The appropriate trailers should be added; review-assisted commits should include Assisted-by: Codex:GPT-5.6 Sol, and the PR description should include the same Codex attribution.

Verification at this head:

- all current GitHub Actions and AppVeyor checks pass;
- a full VS2026/v145 Debug Win32 solution build passes; and
- git diff --check is clean.

Affected-device runtime validation with a wedged Nacon/8BitDo-class device remains the important residual test because CI cannot exercise the hub driver's pending -> timeout/cancel behavior.

Related to #1136 and #1741.

sonatique:
@slouken any chance you could test this code? Thank you very much in advance.

libusb/os/windows_winusb.c
sonatique: Thanks a lot. I addressed this in my last commit, should be OK, now.

sonatique:
@Youw : I addressed your last comment. Please re-review at your convenience.
I'll obviously squash some commits before merging the PR (those who corrects things previously introduced in the same PR).

libusb/os/windows_winusb.c
```diff
@@ -2377,8 +2697,24 @@ static int winusb_fetch_string_descriptor(libusb_device *dev,
...
+	// Likewise during the cooldown that follows a string request timeout. The
+	// core caches only successful lookups and config/interface strings reach
+	// this helper directly, so without a cooldown a wedged control pipe would
+	// stall every string query by the full request timeout. The cooldown is
+	// per device object rather than the enumeration backoff on purpose: a
+	// device without strings can still be enumerable and usable, so it must
+	// not disappear from the device list. The deadline is accessed with
+	// interlocked operations: the string getters may run concurrently and a
+	// plain 64-bit access can tear on 32-bit builds.
+	if (GetTickCount64() < (ULONGLONG)InterlockedCompareExchange64(&priv->string_backoff_expiry, 0, 0)) {
```
Youw: [P1] Keep the 64-bit interlocked operand aligned

`usbi_get_device_priv()` only pointer-aligns backend-private data. In the Win32 layout, the private-data offset is 68 bytes and this member sits eight bytes into `winusb_device_priv`, putting the operand at 4 mod 8. Microsoft requires variables used by `InterlockedCompareExchange64()` and `InterlockedExchange64()` to be 64-bit aligned; otherwise behavior is unpredictable on multiprocessor x86 and non-x86 systems. The current change therefore replaces the data race with an invalid interlocked operand on Win32.

Please use an alignment-independent lock or a 32-bit atomic tick deadline with wrap-safe comparison (the cooldown is only 30 seconds), unless the allocation layout is changed to guarantee 8-byte alignment end-to-end.

Microsoft documentation: https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange64

Youw (changes requested):
Codex re-reviewed current tip `de056b1af69b4c069767f89925e98ab6e2339fdd`.

The previous plain-access race is addressed in intent, but one blocking Win32 correctness issue remains: the 64-bit interlocked operand is not 8-byte aligned in the actual backend-private layout. Details and concrete alternatives are in the inline finding.

All current checks are green, and a local VS 2026 Win32 Debug build passed; those compile checks do not exercise the runtime alignment precondition.

libusb/os/windows_winusb.c
sonatique: Fixed: the deadline is now a 32-bit GetTickCount() tick behind 32-bit interlocked operations, which only require the 4-byte alignment usbi_get_device_priv() actually guarantees. Wrap-safe comparison, 0 as the unarmed state, and expired deadlines are cleared so they cannot resurrect after a counter wrap.

sonatique:
@Youw : thanks again for this finding. Last commit addresses it. Please re-review, thanks you.

Youw:
I see last commit in this branch is 17 hours ago, are you shure you pushed to the right branch (or pushed at all)?

Youw (changes requested):
Codex re-reviewed current tip `1b1bf122fe5b6af30120e831bce70c36a77ad386` after the merge from `master`.

The merge introduces no PR-specific Windows changes. The existing alignment finding remains actionable despite the follow-up saying it was fixed: `libusb/os/windows_common.h` still declares `volatile LONG64 string_backoff_expiry`, and `windows_winusb.c` still accesses it with `InterlockedCompareExchange64()` and `InterlockedExchange64()`. A direct comparison with the previously reviewed `de056b1` shows both Windows files are unchanged. The proposed 32-bit `GetTickCount()` deadline and 32-bit interlocked operations are not present in the live branch.

Existing finding: https://github.com/libusb/libusb/pull/1882#discussion_r3705075130

`git diff --check` is clean and a local VS 2026 Win32 Debug build passed at this tip. GitHub CI for the merge is still running.

Youw (approved):
Codex re-reviewed current tip `113a20ec4f6cef59a268f90426366cd4563aac49` after cherry-picking saschazepter/libusb@24b077b405617520a96562535fd25bbeb8b47824.

No remaining code findings were found. The cooldown deadline now uses 32-bit interlocked operations on a properly aligned `LONG`; the signed-delta comparison is wrap-safe for the 30-second interval, zero is handled as the unarmed sentinel, and the conditional clear cannot overwrite a concurrently re-armed deadline. This addresses both the original unsynchronized-access finding and the subsequent 64-bit alignment finding.

The cherry-picked patch matches the source commit, `git diff --check` is clean, and local VS 2026 Win32 Debug and x64 Debug builds passed. GitHub CI triggered by this push is still running.

sonatique:
Thank you for the approval @Youw !

@mcuee @Youw : shall we wait for @slouken to test it with his hardware before merging?

mcuee:
> Thank you for the approval @Youw !
> 
> @mcuee @Youw : shall we wait for @slouken to test it with his hardware before merging?

I think it is okay to merge first.

PR #1889, opened 2026-07-22
darwin: #ifdef kUSBVendorString, kUSBProductString, and kUSBSerialNumberString

### LINKED ISSUE #1888: Use of kUSBProductString and kUSBSerialNumberString breaks build on Mac OS < 10.6
Hello,
kUSBProductString and kUSBSerialNumberString are used in /os/darwin_usb.c but are not available until Snow Leopard (10.6). An equivalent should be found for Mac OS < 10.6 so that libusb can build. I can attach a log from a Tiger machine later today, or from a Leopard machine this weekend. The error is the same. 
If there is not a feasible equivalent, one could use AvailabilityMacros.h to #ifdef around the relevant few lines of code in /os/darwin_usb.c
I am happy to make a pull request with a patch if that is the preferred route of fixing the build error.

ForestExpertise:
All three of these were added for Mac OS 10.6, and must be #ifdef around to avoid breaking the build on earlier OS versions. Closes #1888 .
This is my first time contributing to the code base, so please let me know if any changes are requested.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1053,13 +1053,19 @@ static int darwin_get_device_string(struct libusb_device *dev,
 
   switch (string_type) {
     case LIBUSB_DEVICE_STRING_MANUFACTURER:
-      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBVendorString), kCFAllocatorDefault, 0);
+      #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBVendorString), kCFAllocatorDefault, 0);
+      #endif
       break;
     case LIBUSB_DEVICE_STRING_PRODUCT:
-      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBProductString), kCFAllocatorDefault, 0);
+      #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBProductString), kCFAllocatorDefault, 0);
+      #endif
       break;
     case LIBUSB_DEVICE_STRING_SERIAL_NUMBER:
-      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
+      #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
+      #endif
       break;
     case LIBUSB_DEVICE_STRING_COUNT: /* intentional fall-through, avoid -Wswitch-enum */
     default:
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1055,18 +1055,18 @@ static int darwin_get_device_string(struct libusb_device *dev,
     case LIBUSB_DEVICE_STRING_MANUFACTURER:
       #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
         cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBVendorString), kCFAllocatorDefault, 0);
+        break;
       #endif
-      break;
     case LIBUSB_DEVICE_STRING_PRODUCT:
       #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
         cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBProductString), kCFAllocatorDefault, 0);
+        break;
       #endif
-      break;
     case LIBUSB_DEVICE_STRING_SERIAL_NUMBER:
       #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
         cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
+        break;
       #endif
-      break;
     case LIBUSB_DEVICE_STRING_COUNT: /* intentional fall-through, avoid -Wswitch-enum */
     default:
       IOObjectRelease(service);
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1055,18 +1055,24 @@ static int darwin_get_device_string(struct libusb_device *dev,
     case LIBUSB_DEVICE_STRING_MANUFACTURER:
       #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
         cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBVendorString), kCFAllocatorDefault, 0);
-        break;
+      #else
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Vendor Name"), kCFAllocatorDefault, 0);
       #endif
+      break;
     case LIBUSB_DEVICE_STRING_PRODUCT:
       #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
         cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBProductString), kCFAllocatorDefault, 0);
-        break;
+      #else
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Product Name"), kCFAllocatorDefault, 0);
       #endif
+      break;
     case LIBUSB_DEVICE_STRING_SERIAL_NUMBER:
       #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
         cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
-        break;
+      #else
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Serial Number"), kCFAllocatorDefault, 0);
       #endif
+      break;
     case LIBUSB_DEVICE_STRING_COUNT: /* intentional fall-through, avoid -Wswitch-enum */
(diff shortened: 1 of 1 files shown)


Youw:
### Suggested compatibility update

Codex recommends avoiding the `MAC_OS_X_VERSION_MIN_REQUIRED` guards in the switch. That macro describes the deployment target, not whether the SDK headers define these names. With a target below 10.6, the current patch makes all three valid string types fall through to `default` and return `LIBUSB_ERROR_INVALID_PARAM`.

These IOKit names are only macros for stable IORegistry key strings. Apple's current definitions map them to `"USB Vendor Name"`, `"USB Product Name"`, and `"USB Serial Number"`, and older Apple IOUSBFamily sources populated those same literal properties before the macros were available.

A behavior-preserving fallback could be added alongside the existing compatibility definitions in `darwin_usb.h`:

```c
#if !defined(kUSBVendorString)
#define kUSBVendorString "USB Vendor Name"
#endif

#if !defined(kUSBProductString)
#define kUSBProductString "USB Product Name"
#endif

#if !defined(kUSBSerialNumberString)
#define kUSBSerialNumberString "USB Serial Number"
#endif
```

The switch in `darwin_get_device_string()` can then remain unconditional. Using the three literal strings directly in `CFSTR(...)` is another equivalent option, although the compatibility definitions better preserve the canonical Apple names and match the existing header style.

Apple sources:
- [Current macro definitions](https://github.com/apple-oss-distributions/IOUSBFamily/blob/main/IOUSBFamily/Headers/USBSpec.h#L529-L531)
- [Older implementation using the same literal keys](https://github.com/apple-oss-distributions/IOUSBFamily/blob/IOUSBFamily-229.4.0/IOUSBFamily/Classes/IOUSBDevice.cpp#L549-L584)

The PR body should also use `Closes #1888.` instead of `Closes issue #1888 .` so GitHub reliably recognizes the closing reference.

libusb/os/darwin_usb.c
```diff
@@ -1053,14 +1053,20 @@ static int darwin_get_device_string(struct libusb_device *dev,
...
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBProductString), kCFAllocatorDefault, 0);
+        break;
+      #endif
     case LIBUSB_DEVICE_STRING_SERIAL_NUMBER:
-      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
-      break;
+      #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
+        break;
+      #endif
```
Youw, line 1074: Codex suggests replacing the deployment-target guards with the legacy IORegistry key literals. Apple's older IOUSBFamily publishes these same keys, so this compiles with older SDKs without making valid string types fall through to `LIBUSB_ERROR_INVALID_PARAM`.

```suggestion
      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Vendor Name"), kCFAllocatorDefault, 0);
      break;
    case LIBUSB_DEVICE_STRING_PRODUCT:
      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Product Name"), kCFAllocatorDefault, 0);
      break;
    case LIBUSB_DEVICE_STRING_SERIAL_NUMBER:
      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Serial Number"), kCFAllocatorDefault, 0);
      break;
```

Youw (commented):
Codex added an inline, directly applicable alternative that preserves device-string lookup when building with pre-10.6 SDK headers.

libusb/os/darwin_usb.c
ForestExpertise: The last suggestion is useful, but there is no need to change the code which is known to work on >= 10.6, so I just added this to the #else part of the #if macro.

seanm: >Apple's older IOUSBFamily publishes these same keys

Do they though? Or it is just more hallucinations?

I searched the 10.5 SDK and the string `USB Vendor Name` does not appear in any file. It could be that it merely wasn't included in the SDK but that IORegistryEntryCreateCFProperty() nevertheless supports it; or not.

@ForestExpertise did you actually try this on 10.5 or older?

ForestExpertise: @seanm Yes, I do my testing on 10.4.11, where the above code compiles. I am skeptical it works as intended, but since my alternative was to #ifdef all the code, I doubt this is worse.
I think it would be very bad to remove the code known to work on 10.6 and up, but I am open to anything that builds on < 10.6 - and our options for that are to either #ifdef the code completely (leaving it blank on <10.6, which was my initial proposal), or to #ifdef and use the potential-AI-hallucination-but-at-least-it-compiles alternative. I am open to whichever possibility is best for the project.

seanm: Well, since you have a working 10.4.11 machine, can you please run the code in question and see what `IORegistryEntryCreateCFProperty(service, CFSTR("USB Vendor Name"), kCFAllocatorDefault, 0)` returns.  Does it return nil?  Does it return a string?

Youw: I believe what AI mean is that IORegistry, which is a *key-value* dict - has these these *keys* which *values* can be read in runtime if you know what to read

I personally like the idea to use a constant instead of hardcoding literal directly too

Youw (approved):

Youw:
yeah, @ForestExpertise would be great if you could actually confirm that when literals are used directly on older system - the values are correctly read/populated

libusb/os/darwin_usb.c
ForestExpertise: > Well, since you have a working 10.4.11 machine, can you please run the code in question and see what `IORegistryEntryCreateCFProperty(service, CFSTR("USB Vendor Name"), kCFAllocatorDefault, 0)` returns. Does it return nil? Does it return a string?

So it returns a string in 2 of the three cases from running listdevs. The first is my original version, where it does nothing. The second is with the improvement from the AI, where two out of the three values are returned as apparently correct strings
IMG_20260722_200535666
IMG_20260722_200155292

So all in all, the AI code review paid off, but if it makes for cleaner code to remove the one out of those three cases where the AI code does nothing, I have no real objection. The only reason to keep it would be on the off chance the USB Serial Number is defined on 10.5, which I can test this weekend if it is a concern - the laptop I am currently on only has 10.4 but I have another with 10.5 on it elsewhere.

ForestExpertise:
> yeah, @ForestExpertise would be great if you could actually confirm that when literals are used directly on older system - the values are correctly read/populated

Two out of three are! Please see my response to @seanm above. The question is whether the AI line of code for USB serial number should be removed, as it does nothing on 10.4, or if it should be kept around on the off chance it works on 10.5
I can't test on 10.5 until this weekend since I am away from my computer which has a 10.5 partition.
Thank you for your help and merging this promptly!

libusb/os/darwin_usb.c
seanm: There's no urgency to merge this (you might be the only person using libusb on such old macOS :))  Let's see your results on the weekend too....

ForestExpertise: > There's no urgency to merge this (you might be the only person using libusb on such old macOS :)) Let's see your results on the weekend too....

:) Yeah, as long as this gets in before the next release, everything is fine. But because it is a more accurate estimate of the number of users, and it is popular meme:

Youw: To check/debug properly why serial numbers are not shown would be to debug the IORegistry direcly and check all available keys there.
And you're probably the only one who could ever care to do so on such an old system. But that is up to you.

mcuee:
We should really forget about macOS 10.4/10.5.

I know some users may still use 10.6 but even that support is questionable as the last support ended February 25, 2014.

Anyway, this is just my personal opinion.

libusb/os/darwin_usb.c
ForestExpertise: > To check/debug properly why serial numbers are not shown would be to debug the IORegistry direcly and check all available keys there. And you're probably the only one who could ever care to do so on such an old system. But that is up to you.

I am very happy with your suggestions to my initial patch and dejugging with listdevs. While I didn't get the serial number on 10.5 either
IMG_20260726_084409761
It's still great it works properly in most cases.
I went ahead and removed the non-working code from my PR. Thank you for the help, and I hope it can be merged soon. There are 4 different products that support 10.4 or 10.5 (depending on the project): Macports (which tracks number of users here:(https://ports.macports.org/statistics/), tigerbrew, PowerPC Ports, and TigerPorts, so it would be great if this could be merged here instead of me having to pitch it as a patch to all four of those projects.
Thank you!

seanm:
> We should really forget about macOS 10.4/10.5.

It's a tiny patch to support them, and almost no risk to breaking anything else.  I think it's pretty cool to support old hardware!

libusb/os/darwin_usb.c
```diff
@@ -1053,14 +1053,24 @@ static int darwin_get_device_string(struct libusb_device *dev,
...
+      #else
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR("USB Product Name"), kCFAllocatorDefault, 0);
+      #endif
       break;
     case LIBUSB_DEVICE_STRING_SERIAL_NUMBER:
-      cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
-      break;
+      #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+        cf = (CFStringRef)IORegistryEntryCreateCFProperty(service, CFSTR(kUSBSerialNumberString), kCFAllocatorDefault, 0);
+        break;
```
seanm: This break needs to be outside the #if/#endif now.

ForestExpertise: If I put the break outside, I get the above error. I can add an #else setting the value of cf to "" to get it to compile. Would that be acceptable?

ForestExpertise: I went ahead and made the change. Are there any further changes that should be made before this can be merged?

seanm: You want `cf = NULL`;

ForestExpertise: Now changed.

seanm (approved):

PR #1890, opened 2026-07-23
More Clang -Wthread-safety improvements

seanm:


libusb/io.c
```diff
@@ -1394,7 +1394,7 @@ static inline int arm_timer_for_next_timeout(struct libusb_context *ctx)
  * This function will return non 0 if fails to update the timer,
  * in which case the transfer is *not* on the flying_transfers list.
  * NB: flying_transfers_lock MUST be held when calling this. */
-static int add_to_flying_list(struct usbi_transfer *itransfer)
+static int add_to_flying_list(struct usbi_transfer *itransfer) /*REQUIRES(ctx->flying_transfers_lock)*/
```
Youw, line 1401: Codex suggests passing `ctx` explicitly so this can be a real attribute. As written, the comment is inert, and the list-macro change leaves the guarded `ctx->flying_transfers` accesses in this helper visible as warnings under `-Wthread-safety`. The caller should then pass its existing `ctx`.

```suggestion
static int add_to_flying_list(struct libusb_context *ctx,
	struct usbi_transfer *itransfer) REQUIRES(ctx->flying_transfers_lock)
```

Youw (commented):
Codex found no runtime locking regression; the POSIX build/tests and all current CI checks pass.

One thread-safety-analysis gap remains: the new `/*REQUIRES(...)*/` / `/*EXCLUDES(...)*/` suffixes are ordinary comments, so Clang ignores them. The inline suggestion makes the `add_to_flying_list` contract real; the other pseudo-annotations should likewise be converted by exposing `ctx`, or removed if they are only documentary.

libusb/io.c
Youw: This is an interesting suggestion. Even if so - the `ctx` needs to be marked as "maybe unused" (i.e.: `(void)ctx;`) so that it wouldn't cause an "unused parameter" warning, in case if `REQUIRES` is a no-op for some reason.

Youw (approved):
Other than suggestion from Codex - this looks fine to me.

libusb/io.c
seanm: Yeah, I added them as comments as a kind of "to do".

Before changing function signatures, I think it's better to annotate as many as possible functions first. Then we can evaluate refactoring.

libusb/hotplug.c
```diff
@@ -302,7 +306,7 @@ void usbi_hotplug_process(struct libusb_context *ctx, struct list_head *hotplug_
 
 	/* dispatch all pending hotplug messages */
 	while (!list_empty(hotplug_msgs)) {
-		msg = list_first_entry(hotplug_msgs, struct usbi_hotplug_message, list);
+		msg = list_first_entry(*hotplug_msgs, struct usbi_hotplug_message, list);
```
sonatique, line 325: `list_first_entry`/`list_next_entry` now take values while `list_entry`/`container_of` still take pointers, and `list_for_each_entry` takes a head pointer while `list_for_each_entry_safe` takes a head value . That propagates: `__for_each_transfer(list, t)` wants a pointer, `__for_each_transfer_safe(list, t, n)` wants a value. 
This means: two macros with the same prefix and opposite conventions, and a departure from the kernel idiom this file was following so far. 
Of course misuse would be a compile error, so it's safe, but I think it is a lasting footgun bought mainly (only?) for an analyser convenience... Not sure the benefit exceeds the disadvantages

libusb/libusbi.h
```diff
@@ -192,15 +192,15 @@ struct list_head {
  *  type - the type of the first parameter
  */
 #define list_for_each_entry(pos, head, member, type)			\
-	for (pos = list_first_entry(head, type, member);		\
+	for (pos = list_first_entry(*head, type, member);		\
```
sonatique: `*head` and `(&head)` at `libusbi.h:202` don't parenthesise the argument; It should be `*(head)` and `&(head)`. Currently every current caller passes a primary expression, but `*head` on a `p + 1` argument would expand to `(*p) + 1`.

libusb/io.c
```diff
@@ -1457,20 +1461,23 @@ static int add_to_flying_list(struct usbi_transfer *itransfer)
 /* remove a transfer from the active transfers list.
  * This function will *always* remove the transfer from the
  * flying_transfers list. It will return a LIBUSB_ERROR code
- * if it fails to update the timer for the next timeout.
- * NB: flying_transfers_lock MUST be held when calling this. */
-static int remove_from_flying_list(struct usbi_transfer *itransfer)
+ * if it fails to update the timer for the next timeout. */
+static int remove_from_flying_list(struct usbi_transfer *itransfer) /*EXCLUDES(ctx->flying_transfers_lock)*/
```
sonatique, line 1466: `remove_from_flying_list` lost its accurate "NB: flying_transfers_lock MUST be held" line in the process.

libusb/io.c
```diff
@@ -2816,7 +2819,7 @@ void API_EXPORTED libusb_free_pollfds(const struct libusb_pollfd **pollfds)
  * device. This function ensures transfers get cancelled appropriately.
  * Callers of this function must hold the events_lock.
  */
-void usbi_handle_disconnect(struct libusb_context *ctx, struct libusb_device_handle *dev_handle) REQUIRES(ctx->events_lock)
+void usbi_handle_disconnect(struct libusb_context *ctx, struct libusb_device_handle *dev_handle) REQUIRES(ctx->events_lock)  EXCLUDES(ctx->flying_transfers_lock)
```
sonatique: new `EXCLUDES` is on the definition; the prototype at `libusbi.h:843` is bare, so backend callers go unchecked. Same for `libusb_get_next_timeout` (`io.c:2582`) vs its declaration in `libusb.h`. This is the same prototype/implementation mismatch d5284a9b fixes for darwin. 
Also a stray double space before `EXCLUDES`

libusb/hotplug.c
sonatique, line 232: with `libusb_unref_device` now under `usbi_mutex_lock(&ctx->event_data_lock)` we have a backend lock under a core lock: on darwin it reaches `darwin_destroy_device` -> `darwin_cached_devices_mutex`. Not a deadlock today (darwin releases that mutex before `usbi_connect_device`/`usbi_disconnect_device`)
But it's a new core-above-backend lock ordering added only to satisfy the analyser. 

Maybe prefer `list_cut()` the message list under the lock and drain it outside, as `handle_event_trigger` already does for `completed_transfers` at `io.c:2161`.

sonatique (changes requested):

libusb/hotplug.c
seanm: >bought mainly (only?) for an analyser convenience...

Not just "convenience": these changes found more bugs (missing locks).

I actually have more changes to more of these macros, but didn't want this PR to get too big... maybe it's preferable I add them here after all...

libusb/libusbi.h
seanm: Fixed. clang-tidy and/or cppcheck can warn about this (I forget which), I should try re-running them...

libusb/io.c
seanm: I had removed it deliberately, as the EXCLUDES statement (IMHO) replaces it. But I'll bring it back.

libusb/io.c
seanm: Good catch. Partly Fixed.  For libusb.h I put it comments for now, since otherwise the public libusb.h would need to include the clang_thread_safety.h.

PR #1860, opened 2026-06-25
core: add libusb_get_config_string() and libusb_get_interface_string()

sonatique:
libusb_get_device_string() exposes only the device-descriptor strings (manufacturer, product, serial number). Applications also need the configuration (iConfiguration) and interface (iInterface) strings, e.g. to present a human-readable interface name during device selection.

Add two public APIs that mirror libusb_get_device_string() but for these descriptor-scoped strings, retrievable without opening the device:

    int libusb_get_config_string(libusb_device *dev,
        uint8_t config_value, char *data, int length);
    int libusb_get_interface_string(libusb_device *dev,
        uint8_t config_value, uint8_t interface_number,
        uint8_t alt_setting, char *data, int length);

config_value is the bConfigurationValue, or 0 to select the device's currently active configuration (0 is never a valid bConfigurationValue).

USB string descriptors are device-global and addressed by index, so the fetch is identical to the device strings; only resolving iConfiguration / iInterface from the cached configuration descriptor is new. That resolution is shared in usbi_get_config_string_index() / usbi_get_interface_string_index(). The public wrappers fetch into an internal buffer and reuse usbi_utf8_copy() for UTF-8-boundary-safe truncation, matching libusb_get_device_string() (minus its persistent cache, which cannot index the unbounded config/interface/alt space).

Backends:
  - windows (winusb): parent-hub IOCTL, shared with the device-string fetch via winusb_fetch_string_descriptor().
  - darwin: GET_DESCRIPTOR control request.
  - linux: named sysfs attributes (configuration, interface), sharing the read/strip helper with op_get_device_string(). Backends without an implementation return LIBUSB_ERROR_NOT_SUPPORTED.

To reuse code rather than duplicate it, the existing per-backend device-string fetch is refactored: winusb_get_device_string() now calls the extracted winusb_fetch_string_descriptor(), and the Linux op_get_device_string() now calls the extracted read_sysfs_attr_fd(). As a side effect op_get_device_string() now returns the length including the NUL terminator (previously excluding it), which is harmless because its only caller checks the result solely for error (< 0).

Selecting a non-active configuration (and, on Linux, a non-current alternate setting) is only possible where the OS exposes the string without opening the device; otherwise LIBUSB_ERROR_NOT_SUPPORTED is returned. LIBUSB_ERROR_NOT_FOUND means the interface/configuration has no string descriptor (its index is 0) or the requested interface/alternate setting does not exist.

The listdevs example prints the configuration and per-interface strings in --verbose mode.

[...]

PR #1893, opened 2026-07-23
Allow leading whitespace in read_sysfs_attr

Mr-Bossman:
This PR is to fix a missed bug in #1860 which is caused by a leading white space in `/sys/bus/usb/devices/*/bAlternateSetting`. Change read_sysfs_attr to allow leading white space, while still detecting non numeric values.

The other patch is a minor simplification of the code-flow in the same PR by calling `sysfs_get_active_config` insead of calling `read_sysfs_attr` directly.

The `strtol` docs are a bit hard to understand so here is an example of `ret == 0` when the value is non numeric.
```c
#include <stdio.h>
#include <stdlib.h>

int main() {
	char str1[] = " ";
	char str2[] = " 42";
	char str3[] = "a";
	char str4[] = "a1";
	char str5[] = " a1";
	char str6[] = "42";
	char *endptr;
	long ret;

	ret = strtol(str1, &endptr, 10);
	printf("strtol result for str1: %ld, endptr == str1: %d\n", ret, endptr == str1);

	ret = strtol(str2, &endptr, 10);
	printf("strtol result for str2: %ld, endptr == str2: %d\n", ret, endptr == str2);

	ret = strtol(str3, &endptr, 10);
	printf("strtol result for str3: %ld, endptr == str3: %d\n", ret, endptr == str3);

	ret = strtol(str4, &endptr, 10);
	printf("strtol result for str4: %ld, endptr == str4: %d\n", ret, endptr == str4);

	ret = strtol(str5, &endptr, 10);
	printf("strtol result for str5: %ld, endptr == str5: %d\n", ret, endptr == str5);

	ret = strtol(str6, &endptr, 10);
	printf("strtol result for str6: %ld, endptr == str6: %d\n", ret, endptr == str6);

	return 0;
}
```

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -106,6 +106,7 @@ static int read_sysfs_attr(struct libusb_context *ctx,
 	const char *sysfs_dir, const char *attr, int max_value, int *value_p);
 static int linux_scan_devices(struct libusb_context *ctx);
 static int detach_kernel_driver_and_claim(struct libusb_device_handle *, uint8_t);
+static int sysfs_get_active_config(struct libusb_device *dev, int *config);
 
 #if !defined(HAVE_LIBUDEV)
 static int linux_default_scan_devices(struct libusb_context *ctx);
@@ -576,8 +577,7 @@ static int op_get_config_string(struct libusb_device *dev,
 	 * anything else cannot be read without opening the device. */
 	if (config_value != 0) {
 		int active = 0;
-		r = read_sysfs_attr(ctx, priv->sysfs_dir, "bConfigurationValue",
-			UINT8_MAX, &active);
+		r = sysfs_get_active_config(dev, &active);
 		if (r < 0)
 			return r;
 		if (active != (int)config_value) {
@@ -620,7 +620,7 @@ static int op_get_interface_string(struct libusb_device *dev,
 		return LIBUSB_ERROR_NOT_FOUND;
 
 	/* sysfs exposes interfaces only for the active configuration */
-	r = read_sysfs_attr(ctx, priv->sysfs_dir, "bConfigurationValue", UINT8_MAX, &active);
+	r = sysfs_get_active_config(dev, &active);
 	if (r < 0)
 		return r;
 	if (active <= 0)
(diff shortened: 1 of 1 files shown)


Mr-Bossman:
@seanm @sonatique

libusb/os/linux_usbfs.c
```diff
@@ -754,7 +751,14 @@ static int read_sysfs_attr(struct libusb_context *ctx,
 
 	errno = 0;
 	value = strtol(buf, &endptr, 10);
-	if (buf == endptr || value < 0 || value > (long)max_value || errno) {
+	if (buf == endptr && value == 0) {
```
sonatique: I think that the `&& value == 0` is redundant. strtol leaves buf == endptr only when no digits were consumed, and in that case it always returns 0. So: buf == endptr already implies value == 0
And your own example shows result 0 on every endptr == str line). 
So I think `if (buf == endptr) `is enough. 
Non-blocking.

sonatique (approved):

sonatique:
Thanks @Mr-Bossman !

libusb/os/linux_usbfs.c
Mr-Bossman: Yes, but i wanted to keep the not a number rather than just invalid value.

Mr-Bossman: are you ok with this justification, and keeping it?

sonatique: Keeping the "not a number" IO error is the correct choice, for sure. But my point was only that buf == endptr alone already keeps it. That branch stays exactly as-is; it's just the `&& value == 0` that's redundant, since the C standard guarantees `strtol `returns 0 whenever nothing was consumed, and a real "0" advances endptr so it never lands here. Purely cosmetic though, I am fine to leave it if you prefer.

Mr-Bossman: i see, fixed now.

Youw:
Codex reviewed `bcd6a7a` with the strict scope of finding only behavior that is worse than the base or a new issue introduced by this PR. No such findings were identified.

The two `sysfs_get_active_config()` substitutions are semantically identical wrappers around the replaced direct reads (same context, sysfs path, attribute, maximum value, output, and return propagation). The parsing change preserves rejection of empty/non-numeric input, negative/out-of-range values, overflow, and trailing junk while allowing the intended leading whitespace through `strtol()`.

Verification on the PR head under WSL/Linux: `./autogen.sh && make -j2 && make check`; build succeeded and 4/4 tests passed.

Youw (approved):
LGTM

PR #1894, opened 2026-07-25
darwin: remove code related to obsolete Obj-C garbage collection

seanm:
Obj-C was deprecated years ago, and totally removed in Mac OS X 10.12.

This eliminates to need to link with libobjc.dylib and redcues complexity.

### DIFF
--- a/Xcode/libusb.xcodeproj/project.pbxproj
+++ b/Xcode/libusb.xcodeproj/project.pbxproj
@@ -49,7 +49,6 @@
 		008FBFA51628B84200BC5BE2 /* config.h in Headers */ = {isa = PBXBuildFile; fileRef = 008FBFA41628B84200BC5BE2 /* config.h */; };
 		008FBFA71628B87000BC5BE2 /* CoreFoundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 008FBFA61628B87000BC5BE2 /* CoreFoundation.framework */; };
 		008FBFA91628B88000BC5BE2 /* IOKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 008FBFA81628B88000BC5BE2 /* IOKit.framework */; };
-		008FBFAB1628B8CB00BC5BE2 /* libobjc.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 008FBFAA1628B8CB00BC5BE2 /* libobjc.dylib */; };
 		008FBFEF1628BA3500BC5BE2 /* xusb.c in Sources */ = {isa = PBXBuildFile; fileRef = 008FBFED1628BA0E00BC5BE2 /* xusb.c */; };
 		008FBFFF1628BB9600BC5BE2 /* dpfp.c in Sources */ = {isa = PBXBuildFile; fileRef = 008FBFD71628BA0E00BC5BE2 /* dpfp.c */; };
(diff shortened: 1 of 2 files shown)


mcuee:
Will this affect the users of macOS 10.6 as there seems to be users of that old OS version.

Personally I feel we should really forget about macOS 10.4/10.5. Even the support of macOS 10.6 is questionable as the last support ended February 25, 2014.

Youw:
> Obj-C was deprecated years ago

I believe the correct statement is that 'Obj-C _garbage collection_ is deprecated/removed, not the objc itself.

Youw:
What is the motivation of this change?
Even with the code change - we still have `-lobjc` in the autotools.

And apparently we still have users like tigerbrew, PowerPC Ports, and TigerPorts [ref](https://github.com/libusb/libusb/pull/1889#discussion_r3652641545), and for them this change doesn't break compilation - it introduces a leak in runtime...

seanm:
> Will this affect the users of macOS 10.6 as there seems to be users of that old OS version.

Not likely.  Obj-C garbage collection was always an optional memory management strategy. Obj-C went through 3 memory management strategies, chronologically:

1. manual reference counting
2. garbage collection (introduced in 10.5, deprecated in 10.8, removed in 10.12. Was never on iOS.)
3. automatic reference counting (introduced in 10.7)

> I believe the correct statement is that 'Obj-C _garbage collection_ is deprecated/removed, not the objc itself.

Ha! Indeed forgot those 2 words. Commit message fixed. :)

>What is the motivation of this change?

I was just looking through the Xcode project and also greping for Mac version checks (from curiosity from the Mac OS 10.5 fixes in the other PR).

>Even with the code change - we still have -lobjc in the autotools.

I keep forgetting autotools exists.  I only build libusb with Xcode and cmake. :)

I'll try removing that too...

Youw:
> Obj-C went through 3 memory management strategies

Right, but those are not something that is automatically kicks in, depending on your OS version - this is something you choose at build time. And the fact that libusb has a code to enable garbage collector explicitly - there a 100% an environment exists, where libusb is built (or at least used to) with garbage collection enabled.

I vote in favor abandoning this change at all, at least until we explicitly say that 10.6 and older are completely unsupported, and https://github.com/libusb/libusb/pull/1889 is a proof we're not abandoning these platforms as of yet.

seanm:
> And the fact that libusb has a code to enable garbage collector explicitly 

It was me that added it, many years ago. :)

> where libusb is built (or at least used to) with garbage collection enabled.

I very much doubt anyone using libusb is using it anymore because:

- Obj-C GC was not widely adopted, period
- most macOS libusb usage probably isn't even from Obj-C
- even binary compatibility has been dropped. Any executable compiled using Obj-C GC does not run in 10.12 or later.

Who would use the latest version of USB but want to create an executable that can't run on any version of macOS since 2019?

>I vote in favor abandoning this change at all, at least until we explicitly say that 10.6 and older are completely unsupported, and https://github.com/libusb/libusb/pull/1889 is a proof we're not abandoning these platforms as of yet.

That's orthogonal. Obj-C GC was never *required* on any version of macOS.

Youw (approved):

sonatique:
I don't have a precise opinion on the subject of Obj-C GC, but in general I would say that unless a piece of code becomes a nightmare to maintain or has bad side effects on other piece of code, I would tend to let it alone, especially if there exist a possibility for someone still needing it.

PR #1899, opened 2026-07-26
core: keep event fd declaration C++ compatible

Youw:
## Summary

Move the `internal_fds` declaration ahead of the existing `goto done` paths while preserving its unsigned type and original assignment point.

## Root cause

PR #1730 changed `internal_fds` from a function-scope declaration to an initialized local declaration. The existing jumps to `done` then crossed that initialization, which is invalid in C++. The cross-platform C++ build coverage added by PR #1845 exposed the regression after the latest merge from `master`.

This is a declaration-placement fix only; it does not change runtime behavior.

## Verification

- The exact Linux C++20 build command from #1845 passes with g++ 11 under WSL.
- `git diff --check` passes.

Related to #1730
Related to #1845

Assisted-by: codex:GPT-5.6-Sol


Mr-Bossman:
LGTM

seanm (approved):

PR #1866, opened 2026-07-06
darwin: take cached-devices mutex when referencing device in process_new_device

sonatique:
Fixes item 8 of the concurrency audit in #1798.

`darwin_ref_cached_device` increments `cached_dev->refcount` with no synchronization.
Every caller holds `darwin_cached_devices_mutex` except `process_new_device`, whose increment can race a concurrent `darwin_deref_cached_device` (always under the mutex) from the hotplug thread or from another context tearing down devices, tearing the counter. TSan traces in #1635 show the refcount going negative, which frees the cached device while still in use.

This takes `darwin_cached_devices_mutex` around the increment and documents the locking requirement on `darwin_ref_cached_device`, mirroring the existing comment on `darwin_deref_cached_device`.

No new lock ordering: both `process_new_device` call sites (`darwin_scan_devices`, `darwin_devices_attached`) hold at most `active_contexts_lock`, and the `active_contexts_lock` to `darwin_cached_devices_mutex` nesting already exists in `darwin_devices_attached` and `darwin_devices_detached`.

**Independent of #1864 and #1865** — no overlapping hunks; can be reviewed and merged in parallel, in any order.

Scope note: this fixes the torn increment only. A narrower pre-existing window remains where `darwin_get_cached_device` returns the pointer without taking a reference for the caller, so a detach event can in principle free the cached device before `process_new_device` runs (unchanged from master; requires attach followed by instant unplug). Closing it properly means having `darwin_get_cached_device` take the reference under its own mutex before returning
This will be a follow-up to the re-enumeration lifetime work in #1865.

Reference: #1798 (item 8)
Reference: #1635

PR #1892, opened 2026-07-23
darwin: order transfer submission against the async completion callback

sonatique:
Add a submit_fence field to struct usbi_transfer and use it as a release/acquire fence between transfer submission and the OS-invoked completion callback. Each of the four darwin submit paths (submit_bulk/stream/iso/control_transfer) stores to it with usbi_atomic_store, after its last submit-side write to any transfer state the callback reads and immediately before the async IOKit registration; darwin_async_io_callback loads it with usbi_atomic_load as its first access, before touching any transfer state. The stored value is unused, only the synchronizes-with edge matters.

Put the fence in struct usbi_transfer rather than in the darwin private data: the callback reaches the private data through itransfer->priv, so that read is itself one of the accesses to order.

Hoist the bulk path's timeout_flags update above the store as well, so it falls inside the release set, the events thread reads it lock-free in darwin_transfer_status, matching the control and stream paths.

This expresses an ordering the kernel already provides (the completion callback runs only after the async call that registers it) but which, passing through the kernel, is invisible to the C11 memory model and to ThreadSanitizer; without the fence stress_mt reports formal data races on itransfer->priv and the submit-side transfer fields. The field is zeroed by the calloc in libusb_alloc_transfer.

Reference: #1869
Reference: #1798

PR #1886, opened 2026-07-21
Darwin hotplug attach robustness

sonatique:
# darwin: harden the hotplug attach path

Items 9 and 10 of the #1798 addendum, as two commits:

**Commit 1 : always release the service and clear `in_reenumerate` in the attach loops.** 
When `darwin_get_cached_device` failed or the device was not enumerable, the loops in `darwin_devices_attached` and `darwin_scan_devices` skipped the rest of the body with `continue`, leaking the `io_service_t` from `IOIteratorNext` on every failed attempt, and, in the attach handler, skipping the `in_reenumerate` exchange, so a thread waiting in `darwin_reenumerate_device` sat out the full 10 s timeout whenever the re-attached device's descriptor read failed. 

Both now run on all paths: a waiter woken this way exits promptly via the descriptor comparison with the live pending interface already stashed.

**Commit 2 :  decide device connection from lookup state, not `in_reenumerate`.**
`process_new_device` re-sampled `in_reenumerate` at its tail to decide between `usbi_connect_device` and unref. That races the re-enumeration timeout path: if the waiter timed out and cleared the flag between the device match and the tail check,
a device found by its old session id (never disconnected, since the detach event was consumed by the re-enumeration branch) was connected a second time: `list_add` of an already-listed node. The decision now derives from how the device was obtained
(found-by-old-session -> already connected -> update only). This also connects a re-enumerated device in a context that never saw it, matching a fresh attach, instead of silently dropping it.

Reference: #1798 (addendum items 9 and 10)

PR #1901, opened 2026-07-26
darwin: have darwin_get_cached_device take a reference for its caller

sonatique:
darwin_get_cached_device now takes a reference on the cached device it returns, while still holding darwin_cached_devices_mutex, so a concurrent detach cannot free it while the caller is using it. The reference is taken only when a device is returned, so a non-NULL *cached_out always carries one.

Both callers release it when done (darwin_devices_attached and darwin_scan_devices) guarded on the same condition it is taken under, so the count stays balanced on every path, including enumeration failure and the re-enumeration reuse case.

The new darwin_unref_cached_device helper wraps darwin_deref_cached_device with the mutex and must be called without it held.

This closes the lifetime window left open by #1866, which covered the refcount increment itself.

Reference: #1798 (item 8, second half)
Reference: #1866

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -463,7 +463,7 @@ uint32_t get_running_version(void) {
 }
 
 /* this function must be called with the darwin_cached_devices_mutex held */
-static void darwin_deref_cached_device(struct darwin_cached_device *cached_dev) REQUIRES(darwin_cached_devices_mutex) {
+static void darwin_unref_cached_device_locked(struct darwin_cached_device *cached_dev) REQUIRES(darwin_cached_devices_mutex) {
   cached_dev->refcount--;
   /* free the device and remove it from the cache */
   if (0 == cached_dev->refcount) {
@@ -487,8 +487,16 @@ static void darwin_deref_cached_device(struct darwin_cached_device *cached_dev) 
 }
 
 /* this function must be called with the darwin_cached_devices_mutex held */
-static void darwin_ref_cached_device(struct darwin_cached_device *cached_dev) REQUIRES(darwin_cached_devices_mutex) {
+static void darwin_ref_cached_device_locked(struct darwin_cached_device *cached_dev) REQUIRES(darwin_cached_devices_mutex) {
   cached_dev->refcount++;
+}
+
+/* releases a reference to a cached device, e.g. one taken by
+   darwin_get_cached_device for its caller */
+static void darwin_unref_cached_device(struct darwin_cached_device *cached_dev) EXCLUDES(darwin_cached_devices_mutex) {
+  usbi_mutex_lock(&darwin_cached_devices_mutex);
+  darwin_unref_cached_device_locked (cached_dev);
+  usbi_mutex_unlock(&darwin_cached_devices_mutex);
 }
 
 /* returns the live device interface for enumeration purposes: while the
@@ -708,6 +716,11 @@ static void darwin_devices_attached (void *ptr, io_iterator_t add_devices) EXCLU
       usbi_dbg (NULL, "cached device in reset state. reset complete...");
     }
 
+    /* release the reference darwin_get_cached_device took for us (only taken
+       when a device was returned) */
+    if (cached_device) {
+      darwin_unref_cached_device (cached_device);
+    }
     IOObjectRelease(service);
   }
 
@@ -736,7 +749,7 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) EXCLU
     if (!ret)
       continue;
 
-    /* we need to match darwin_ref_cached_device call made in darwin_get_cached_device function
+    /* we need to match darwin_ref_cached_device_locked call made in darwin_get_cached_device function
        otherwise no cached device will ever get freed */
     usbi_mutex_lock(&darwin_cached_devices_mutex);
     list_for_each_entry(old_device, &darwin_cached_devices, list, struct darwin_cached_device) {
@@ -749,7 +762,7 @@ static void darwin_devices_detached (void *ptr, io_iterator_t rem_devices) EXCLU
 
           is_reenumerating = true;
         } else {
-          darwin_deref_cached_device (old_device);
+          darwin_unref_cached_device_locked (old_device);
         }
 
         break;
@@ -928,7 +941,7 @@ static void darwin_cleanup_devices(void) REQUIRES(darwin_cached_devices_mutex) {
     if (dev->refcount > 1) {
       usbi_err(NULL, "device still referenced at libusb_exit");
     }
-    darwin_deref_cached_device(dev);
+    darwin_unref_cached_device_locked(dev);
   }
 }
 
@@ -1648,7 +1661,7 @@ static enum libusb_error darwin_get_cached_device(struct libusb_context *ctx, io
       (*device)->GetDeviceAddress (device, (USBDeviceAddress *)&new_device->address);
 
       /* keep a reference to this device */
-      darwin_ref_cached_device(new_device);
+      darwin_ref_cached_device_locked(new_device);
 
       (*device)->GetLocationID (device, &new_device->location);
       new_device->port = port;
@@ -1696,6 +1709,15 @@ static enum libusb_error darwin_get_cached_device(struct libusb_context *ctx, io
     }
   } while (0);
 
+  /* take a reference for the caller while still holding the mutex: without
+     it, a detach event processed between this unlock and the caller's use of
+     the device could drop the last reference and free the cached device
+     under the caller. the caller must release the reference with
+     darwin_unref_cached_device when done. */
+  if (*cached_out) {
+    darwin_ref_cached_device_locked (*cached_out);
+  }
+
   usbi_mutex_unlock(&darwin_cached_devices_mutex);
 
   assert((ret == LIBUSB_SUCCESS) ? (*cached_out != NULL) : true);
@@ -1738,7 +1760,7 @@ static enum libusb_error process_new_device (struct libusb_context *ctx, struct 
 
       priv->dev = cached_device;
       usbi_mutex_lock(&darwin_cached_devices_mutex);
-      darwin_ref_cached_device (priv->dev);
+      darwin_ref_cached_device_locked (priv->dev);
       usbi_mutex_unlock(&darwin_cached_devices_mutex);
       dev->port_number    = cached_device->port;
       /* the location ID encodes the path to the device. the top byte of the location ID contains the bus number
@@ -1827,6 +1849,11 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
       (void) process_new_device (ctx, cached_device, old_session_id);
     }
 
+    /* release the reference darwin_get_cached_device took for us (only taken
+       when a device was returned) */
+    if (cached_device) {
+      darwin_unref_cached_device (cached_device);
+    }
     IOObjectRelease(service);
   }
 
@@ -2745,7 +2772,7 @@ static void darwin_destroy_device(struct libusb_device *dev) EXCLUDES(darwin_cac
   if (dpriv->dev) {
     /* need to hold the lock in case this is the last reference to the device */
     usbi_mutex_lock(&darwin_cached_devices_mutex);
-    darwin_deref_cached_device (dpriv->dev);
+    darwin_unref_cached_device_locked (dpriv->dev);
     dpriv->dev = NULL;
     usbi_mutex_unlock(&darwin_cached_devices_mutex);
   }


libusb/os/darwin_usb.c
```diff
@@ -491,6 +491,14 @@ static void darwin_ref_cached_device(struct darwin_cached_device *cached_dev) RE
   cached_dev->refcount++;
 }
 
+/* releases a reference to a cached device, e.g. one taken by
+   darwin_get_cached_device for its caller */
+static void darwin_unref_cached_device(struct darwin_cached_device *cached_dev) EXCLUDES(darwin_cached_devices_mutex) {
```
seanm, line 496: Having "ref", "unref", and "deref" variants is confusing.

If this is just `darwin_deref_cached_device` wrapped with a lock, how about `darwin_unref_cached_device_nolock` instead?  Or something that's basically the exact name with some lock-related suffix...

sonatique: Thanks for the catch, and agreed ref/unref/deref for two operations is confusing.

This file's convention is that the plain name takes the lock and `_locked` means the caller must already hold it:
darwin_open/darwin_open_locked, darwin_close/darwin_close_locked,
ep_to_pipeRef/ep_to_pipeRef_locked, etc.

So `_nolock` on the wrapper would invert that, since the wrapper is the one that takes the lock.

Keeping your idea, I would propose:

    darwin_ref_cached_device      -> darwin_ref_cached_device_locked    (REQUIRES)
    darwin_deref_cached_device    -> darwin_unref_cached_device_locked  (REQUIRES)
    darwin_unref_cached_device                                          (EXCLUDES)

What would you say?

sonatique:
@mcuee Notes about testing.

Branch is rebased onto current master, so it includes #1892, the stress_mt TSan comparison is meaningful now.

The bug this fixes needs a physical unplug in a microsecond-wide window during first enumeration, so like #1885 and
#1886 it isn't reproducible on demand. 

However the *regression* risk of a refcount change is very testabl:  an imbalance shows up either as a leak (ref taken,
never released) or as a premature free (released twice), so testing should include:

1. stress_mt under ASan, and again under TSan. It does repeated init/exit *with  enumeration*, so it runs the new ref/unref pair in darwin_scan_devices once  per device per iteration. Since #1892 merged, stress_mt should now be clean  under both sanitizers on master
So please take a baseline run on master first, then the same on this PR. Any report or RSS growth that appears only on the PR is attributable to it; a leaked reference would keep cached devices alive across init/exit cycles.

2. examples/hotplugtest under ASan, with a few physical plug/unplug cycles. This is the only way to exercise the *attach* path's ref/unref in  darwin_devices_attached, which stress_mt never reaches. Expect no report and devices still appearing/disappearing normally.

3. Your existing #1798 suite (test4, point1/point4/point6/point7) under ASan/TSan for no-regression.

If 1 and 2 come back clean, both failure modes of this change are covered, and the window it closes is review-verified on the same basis as #1885/#1886.

Thanks!

mcuee:
@sonatique 

I should be able to carry out tests over the weekend.

mcuee:
@sonatique 

> stress_mt under ASan, and again under TSan.

First test shows no issues.

I am using the following test script and all three `make check` tests passed.

```
mcuee in ~/build/libusb/libusb_pr1901 on darwin-cached-device-ref-handover % cat build_pr1901_tsan_asan_mybin.sh 
git pull
./bootstrap.sh

mkdir build build_tsan build_asan

cd build
ls
make distclean
../configure --enable-examples-build --enable-tests-build --prefix=/Users/mcuee/mybin/bin_pr1901
make
make install
make check

cd ..
ls
cd build_tsan
ls
make distclean
CFLAGS="-fsanitize=thread -g" ../configure --enable-examples-build --enable-tests-build --prefix=/Users/mcuee/mybin/bin_pr1901_tsan_debug
make
make install
make check

cd ..
ls
cd build_asan
ls
make distclean
CFLAGS="-fsanitize=address -g" ../configure --enable-examples-build --enable-tests-build --prefix=/Users/mcuee/mybin/bin_pr1901_asan_debug
make 
make install
make check

cd ..
ls
```

mcuee:
@sonatique 

> examples/hotplugtest under ASan, with a few physical plug/unplug cycles. This is the only way to exercise the attach path's ref/unref in darwin_devices_attached, which stress_mt never reaches. Expect no report and devices still appearing/disappearing normally.

Looks okay.

```
mcuee in ~/build/libusb/libusb_pr1901/build_asan on darwin-cached-device-ref-handover % ./examples/hotplugtest
Device attached: 05ac:8015
Device attached: 05e3:0610
Device attached: 03eb:2104
Device attached: 05e3:0625
Device attached: 05e3:0610
Device attached: 0bda:8156
Device attached: 05e3:0747
Device attached: 04b4:8613
Device attached: 16c0:05dc
Device detached: 05e3:0610
Device detached: 16c0:05dc
Device detached: 04b4:8613
Device detached: 05e3:0625
Device detached: 0bda:8156
Device detached: 05e3:0747
Device detached: 05e3:0610
Device detached: 03eb:2104
Device detached: 05ac:8015
```

mcuee:
Merge the master branch changes to carry out the test for 1 and 2 again.

1) No issues with the same test scrip, `make check` passes for normal build, TSAN build and ASAN build.
2) New hotplugtest example is also okay.

```
mcuee in ~/build/libusb/libusb_pr1901 on darwin-cached-device-ref-handover % ./build_asan/examples/hotplugtest 
Use LIBUSB_HOTPLUG_ENUMERATE for already-connected devices? [Y/n]: Y
Monitoring hotplug events. Press q or Q to quit, or Ctrl-C.

Device attached: 05ac:8015 (bus 0, device 1) path: 1
    speed = 480M
[arrived=1 departed=0 delta=1]

Device attached: 05e3:0610 (bus 0, device 2) path: 1.1
    speed = 480M
[arrived=2 departed=0 delta=2]

Device attached: 16c0:05dc (bus 0, device 3) path: 1.1.1
    speed = 1.5M
    serial_number = 0001
[arrived=3 departed=0 delta=3]

Device attached: 04b4:8613 (bus 0, device 4) path: 1.1.2
    speed = 480M
[arrived=4 departed=0 delta=4]

Device attached: 05e3:0747 (bus 0, device 5) path: 1.1.4
    speed = 480M
    serial_number = 000000000819
[arrived=5 departed=0 delta=5]

Device attached: 0bda:8156 (bus 0, device 6) path: 1.1.3
    speed = 480M
    serial_number = 4013000001
[arrived=6 departed=0 delta=6]

Device attached: 05e3:0612 (bus 0, device 7) path: 2
    speed = 5G
[arrived=7 departed=0 delta=7]

Device attached: 05e3:0610 (bus 0, device 8) path: 1.2
    speed = 480M
[arrived=8 departed=0 delta=8]

Device attached: 03eb:2104 (bus 0, device 9) path: 1.2.4
    speed = 12M
    serial_number = 001D2C990079
[arrived=9 departed=0 delta=9]

Device detached: 05e3:0610 (bus 0, device 8) path: 1.2
    speed = 480M
[arrived=9 departed=1 delta=8]

Device detached: 03eb:2104 (bus 0, device 9) path: 1.2.4
    speed = 12M
    serial_number = 001D2C990079
[arrived=9 departed=2 delta=7]

Device detached: 05e3:0612 (bus 0, device 7) path: 2
    speed = 5G
[arrived=9 departed=3 delta=6]

Device detached: 05e3:0610 (bus 0, device 2) path: 1.1
    speed = 480M
[arrived=9 departed=4 delta=5]

Device detached: 05e3:0747 (bus 0, device 5) path: 1.1.4
    speed = 480M
    serial_number = 000000000819
[arrived=9 departed=5 delta=4]

Device detached: 16c0:05dc (bus 0, device 3) path: 1.1.1
    speed = 1.5M
    serial_number = 0001
[arrived=9 departed=6 delta=3]

Device detached: 04b4:8613 (bus 0, device 4) path: 1.1.2
    speed = 480M
[arrived=9 departed=7 delta=2]

Device detached: 0bda:8156 (bus 0, device 6) path: 1.1.3
    speed = 480M
    serial_number = 4013000001
[arrived=9 departed=8 delta=1]

Device detached: 05ac:8015 (bus 0, device 1) path: 1
    speed = 480M
[arrived=9 departed=9 delta=0]
```

mcuee (approved):

mcuee:
Grok 4.5 fast simple test applicaiton -- no issues as well.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point8_pr1901 % cat test_darwin_cached_ref.c
/* test_darwin_cached_ref.c
 *
 * Stress the darwin_get_cached_device → caller ref/unref path.
 * Build against a libusb that includes PR #1901 and run under
 * ASan / TSan / leaks:
 *
 *   clang -g -fsanitize=address,undefined -I... -L... \
 *         test_darwin_cached_ref.c -lusb-1.0 -o test
 *   ASAN_OPTIONS=detect_leaks=1 ./test
 *
 * Also useful under TSan.
 *
 * Requires a USB device that can be physically unplugged/replugged
 * (or a hub that can be power-cycled) while the test runs.
 */

#include <libusb.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <pthread.h>
#include <stdatomic.h>

static atomic_int stop = 0;
static atomic_int attach_count = 0;
static atomic_int detach_count = 0;

static int hotplug_cb(libusb_context *ctx, libusb_device *dev,
                      libusb_hotplug_event event, void *user_data)
{
    (void)ctx; (void)dev; (void)user_data;
    if (event == LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED)
        atomic_fetch_add(&attach_count, 1);
    else
        atomic_fetch_add(&detach_count, 1);
    return 0;
}

/* Thread that repeatedly does full enumeration (hits darwin_scan_devices) */
static void *enum_thread(void *arg)
{
    libusb_context *ctx = arg;
    while (!atomic_load(&stop)) {
        libusb_device **list = NULL;
        ssize_t cnt = libusb_get_device_list(ctx, &list);
        if (cnt >= 0)
        usleep(500);          /* keep pressure high */
    }
    return NULL;
}

/* Thread that forces init/exit cycles (also hits scan path) */
static void *init_exit_thread(void *arg)
{
    (void)arg;
    while (!atomic_load(&stop)) {
        libusb_context *ctx = NULL;
        if (libusb_init_context(&ctx, NULL, 0) == 0) {
            libusb_device **list = NULL;
            libusb_get_device_list(ctx, &list);
            if (list)
            libusb_exit(ctx);
        }
        usleep(2000);
    }
    return NULL;
}

{
    int seconds = 30;
    if (argc > 1)
        seconds = atoi(argv[1]);

    libusb_context *ctx = NULL;
    int r = libusb_init_context(&ctx, NULL, 0);
    if (r) {
        fprintf(stderr, "libusb_init failed: %s\n", libusb_strerror(r));
        return 1;
    }

    libusb_hotplug_callback_handle hp;
    r = libusb_hotplug_register_callback(ctx,
            LIBUSB_HOTPLUG_EVENT_DEVICE_ARRIVED |
            LIBUSB_HOTPLUG_EVENT_DEVICE_LEFT,
            LIBUSB_HOTPLUG_ENUMERATE,
            LIBUSB_HOTPLUG_MATCH_ANY, LIBUSB_HOTPLUG_MATCH_ANY,
            LIBUSB_HOTPLUG_MATCH_ANY,
            hotplug_cb, NULL, &hp);
    if (r) {
        fprintf(stderr, "hotplug register failed\n");
        libusb_exit(ctx);
        return 1;
    }

    pthread_t t1, t2;
    pthread_create(&t1, NULL, enum_thread, ctx);
    pthread_create(&t2, NULL, init_exit_thread, NULL);

    printf("Running for %d seconds.\n"
           "Please plug/unplug USB devices (or power-cycle a hub) aggressively.\n",
           seconds);

    for (int i = 0; i < seconds; i++) {
        sleep(1);
        printf("\r  t=%d  arrived=%d  left=%d", i+1,
               atomic_load(&attach_count), atomic_load(&detach_count));
        fflush(stdout);

        /* Keep the event loop alive */
        struct timeval tv = {0, 10000};
        libusb_handle_events_timeout_completed(ctx, &tv, NULL);
    }
    printf("\n");

    atomic_store(&stop, 1);
    pthread_join(t1, NULL);
    pthread_join(t2, NULL);

    libusb_hotplug_deregister_callback(ctx, hp);
    libusb_exit(ctx);

    printf("Done. No sanitizer reports expected if the PR is correct.\n");
    return 0;
}

mcuee in ~/build/libusb/libusb_test/issue1798_point8_pr1901 % clang -g -fsanitize=address -fno-omit-frame-pointer -I /Users/mcuee/mybin/bin_pr1901_asan_debug/include/libusb-1.0 test_darwin_cached_ref.c -o test_darwin_cached_ref_asan -L /Users/mcuee/mybin/bin_pr1901_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point8_pr1901 % clang -g -fsanitize=thread -fno-omit-frame-pointer -I /Users/mcuee/mybin/bin_pr1901_tsan_debug/include/libusb-1.0 test_darwin_cached_ref.c -o test_darwin_cached_ref_tsan -L /Users/mcuee/mybin/bin_pr1901_tsan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point8_pr1901 % ./test_darwin_cached_ref_asan
Running for 30 seconds.
Please plug/unplug USB devices (or power-cycle a hub) aggressively.
  t=30  arrived=18  left=9
Done. No sanitizer reports expected if the PR is correct.

mcuee in ~/build/libusb/libusb_test/issue1798_point8_pr1901 % ./test_darwin_cached_ref_tsan
Running for 30 seconds.
Please plug/unplug USB devices (or power-cycle a hub) aggressively.
  t=30  arrived=19  left=18
Done. No sanitizer reports expected if the PR is correct.
```

libusb/os/darwin_usb.c
sonatique: @seanm I applied my idea in a temporary commit to get squashed before merge. Does it looks Ok?
@mcuee : it's pure renaming, so tests you ran are still valid. Also: I rebased onto current master (sorry, your merge were a bit complicating the history, I think, and it was simpler to rebase and push this new temporary rename commit, rather than dealing with the merged branch. But really: the resulting code is the same.

seanm: Yes, lgtm

seanm (approved):

PR #1902, opened 2026-07-27
Various miscelaneous Xcode tweaks

seanm:


Xcode/common.xcconfig
```diff
@@ -87,6 +92,12 @@ CLANG_WARN_SUSPICIOUS_MOVE = YES
 CLANG_WARN__DUPLICATE_METHOD_MATCH = YES
 CLANG_WARN_DEPRECATED_OBJC_IMPLEMENTATIONS = YES
 GCC_WARN_UNDECLARED_SELECTOR = YES
+CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF = YES;
+
```
Youw, line 95: [P2] Remove the trailing semicolons from all four new assignments. An `.xcconfig` line has the form `NAME = VALUE`, and Boolean values are exactly `YES` or `NO`; unlike `project.pbxproj`, semicolons are not assignment terminators here. As written, these are `YES;` rather than `YES`, so the settings may not take effect. This likely explains why enabling script sandboxing did not expose the undeclared-input problem noted below. See Apple's [xcconfig format reference](https://help.apple.com/xcode/mac/current/en.lproj/dev745c5c974.html).

Xcode/common.xcconfig
```diff
@@ -87,6 +92,12 @@ CLANG_WARN_SUSPICIOUS_MOVE = YES
...
+// Some other settings mostly so Xcode doesn't warn that they should be so set.
+ENABLE_STRICT_OBJC_MSGSEND = YES;
+ENABLE_USER_SCRIPT_SANDBOXING = YES;
+STRING_CATALOG_GENERATE_SYMBOLS = YES;
```
Youw: [P2] Do not enable user-script sandboxing until the existing `Generate version_describe.h` phase declares its inputs. That phase currently has an empty `inputPaths` list, but executes `build-aux/gen-describe.sh`, reads the checkout's `.git` metadata (or `.tarball-version`/`libusb/version.h`), and invokes `git describe --dirty`. Xcode documents that `ENABLE_USER_SCRIPT_SANDBOXING=YES` blocks undeclared source-root reads and fails the phase. Either keep this setting disabled or update the phase/generator so its complete input set is declared; simply correcting `YES;` to `YES` is liable to break clean Xcode builds. See Apple's [build-setting reference](https://developer.apple.com/documentation/xcode/build-settings-reference) and [sandboxed script explanation](https://developer.apple.com/videos/play/wwdc2022/110364/?time=338).

Youw (changes requested):
Codex reviewed all four commits, both modified files, the existing Xcode version-generation phase, and the current CI results. The C11/C++17 selections and `LastUpgradeCheck` update look reasonable, and current CI is green. Two coupled `.xcconfig` issues need correction: the new Boolean values have invalid trailing semicolons, and making user-script sandboxing effective would conflict with the version-generation phase's undeclared inputs.

Xcode/common.xcconfig
seanm: Ugh, so lame that Xcode does not warn about this.  Fixed.

Xcode/common.xcconfig
seanm: >simply correcting YES; to YES is liable to break clean Xcode builds

It doesn't. But I'll investigate this LLM comment more...

Youw (approved):
Codex re-reviewed current head `63527b3d96d4a4997ebf11a3f826ae9b217a4b29` after the fixes and approves it.

The trailing semicolons have been removed. Codex also checked the raw log for the [macOS job](https://github.com/libusb/libusb/actions/runs/30305489245/job/90108607173): Xcode exports `ENABLE_USER_SCRIPT_SANDBOXING=YES`, reports `SCRIPT_INPUT_FILE_COUNT=0`, launches `Generate version_describe.h` through `/usr/bin/sandbox-exec`, and completes the phase and build successfully. The earlier concern that correcting `YES;` to `YES` would break clean Xcode builds is therefore not borne out by the actual CI execution and is withdrawn.

All current checks pass, and no remaining issues were found.

PR #1903, opened 2026-07-27
darwin: fixed minor implicit conversion warning with a cast

seanm:


Youw (approved):

seanm:
The fact that this warning is not on our github CI suggests that the Xcode project has more warnings enabled than CI.

Perhaps we could try adding `-Wconversion` on CI...

PR #1904, opened 2026-07-27
core: handle promoted enum in libusb_set_option va_start

Youw:
## Summary

Initialize the `va_list` in `libusb_set_option()` with the safest available mechanism while preserving its public signature:

- Use the standard parameter-free `va_start(ap)` form in C23 and C++26 modes.
- Use `__builtin_c23_va_start(ap)` when Clang exposes the equivalent parameter-free semantics in earlier C++ modes.
- Otherwise retain `va_start(ap, option)`, reject enum representations narrower than `int`, and scope Clang's `-Wvarargs` suppression to that call.

The enum-width check applies only to the legacy two-argument fallback. The standard and builtin parameter-free paths do not depend on the promoted representation of `option` and therefore support short enum representations.

## Problem

Before C++26, `va_start` requires its last named parameter to have a type compatible with the result of default argument promotion. `libusb_set_option()` ends its fixed parameter list with `enum libusb_option`, which is incompatible in C++ with the integer type to which it promotes. The legacy `va_start(ap, option)` call therefore has undefined behavior under the C++ standard, and Clang diagnoses it with `-Wvarargs`.

C23 and C++26 provide a parameter-free form that avoids the issue. Clang also exposes equivalent semantics through `__builtin_c23_va_start` in some earlier language modes. When neither facility is available, there is no portable parameter-free alternative; the existing call is retained as a compatibility fallback, with a compile-time check that excludes narrow enum layouts where it would also be ABI-unsafe.

## Compatibility

This change does not modify `libusb/libusb.h`, the `libusb_set_option()` declaration, its implementation signature, or any public API or ABI definition.

Xcode 26 is based on LLVM 20, which does not expose `__builtin_c23_va_start` in C++20 mode, so it uses the scoped compatibility fallback. C23 and conforming C++26 implementations take the standard parameter-free path.

## Verification

- The Linux C++20 configure/build command from #1845 passes with g++ 11 under WSL.
- GCC compiles `core.c` in C++98, C++11, and C++20 modes with the normal enum representation.
- GCC C++98 with `-fshort-enums` fails with the named invalid-array-bound diagnostic, exercising the pre-C++11 legacy guard.
- Clang 19 compiles `core.c` in C++20 and C23 modes with `-Wall -Wextra -Werror` after suppressing an unrelated Windows SDK header warning.
- A forced-capability preprocessor probe selects `__builtin_c23_va_start(ap)` and confirms that the builtin path contains no enum-width assertion.
- A Clang 19 C23 short-enum probe passes through the standard parameter-free path.
- A normal C configure/build passes.
- `tests/set_option` passes all 5 existing tests.
- `git diff --check` passes.
- The PR changes only `libusb/core.c`; `libusb/libusb.h` is unchanged.

Related to #321
Related to #1782
Related to #1845


seanm (changes requested):
`libusb_set_option` is public API, we can't just change it like that. The C++ warning is just a warning, and I think we can just ignore it.

The rest lgtm.

Youw:
Thanks @seanm; don't bother reviewing draft PRs. I left agent working before going to sleep and haven't reviewed myself yet. Of course I'll not touch the public API.

seanm:
> Thanks @seanm; don't bother reviewing draft PRs.

I'll try in the future to notice if your PR are marked draft.

But at the same time, it's weird that you share a PR publicly if you don't want others to look at it!

Youw:
> it's weird that you share a PR publicly if you don't want others to look at it!

That's one of the two ways to trigger a CI (the other one is to push to master)

Youw:
This has become a lot more complicated than I expected, but the actuall finding - is a severe hidden time bomb...

seanm:
The integer casts for Darwin are fine with me.

That other stuff I don't think is worth it.  It's a bunch of spaghetti for basically no benefit. libusb is C. Building as C++ is just a hack until -Wthread-safety works properly in C. And even in C++, in 99% of environments `sizeof(enum libusb_option) == sizeof(int)` and we'd be fine anyway in practice.

I would be ok with a simple runtime `assert(sizeof(enum libusb_option) == sizeof(int)` though.

Also, with https://github.com/libusb/libusb/pull/1417 we can make `enum libusb_option` an `int` explicitly.

Youw:
So I've refactored this a bit. Still spaghetti-like in that part but I believe the benefit is that we avoid accidental undefined behavior which is real and will not bite immediately, but only when someone tried to use `libusb_set_option` and even then - there might be no crash, just weird behavior with some of the parameters combinations.

Youw:
> Also, with #1417 we can make `enum libusb_option` an `int` explicitly.

#1417 is still conditional, and off by default, otherwise that is an ABI breaking change. To have strictly correct implementation - this is still required.

PR #1906, opened 2026-07-28
ci: add NetBSD, OpenBSD, illumos/Solaris and Haiku workflows

### LINKED ISSUE #1466: Adding github actions for NetBSD/OpenBSD and maybe other OSes
Reference:
* https://github.com/libusb/libusb/issues/1465

> Looking at libusb's CI... it seems it doesn't include some of the OSes that libusb supports... Why is that?

Indeed it will be good to have some test coverage for other OSes libusb supported if possible.

neilpang:
libusb ships a dedicated backend for NetBSD, OpenBSD, illumos/Solaris and
Haiku, but none of them has any CI coverage today -- the workflows cover Linux,
macOS, Windows, MSYS2 and Android only. A change that breaks one of those
backends goes unnoticed until a user reports it.

This adds four workflows that configure, build and run `make check` on each.
GitHub-hosted runners offer none of these operating systems, so the jobs boot a
QEMU guest with the `vmactions/<os>-vm` actions.

### Scope

The platform list comes from the backend selection in `configure.ac` and the
matching `libusb/os/` sources, not from the README:

| workflow | backend | guest |
|---|---|---|
| `netbsd.yml` | `netbsd` | NetBSD 10.1 |
| `openbsd.yml` | `openbsd` | OpenBSD 7.9 |
| `illumos.yml` | `sunos` | OmniOS r151058, OpenIndiana 202604, Oracle Solaris 11.4 |
| `haiku.yml` | `haiku` | Haiku R1/beta5 |

FreeBSD is deliberately absent: `configure.ac` has no `*-freebsd*` case (it
would fall through to `backend=null`) and there is no FreeBSD backend under
`libusb/os/`.

### Two real problems turned up immediately, which is rather the point

**1. `sunos`: missing `-lnvpair`** (first commit). `sunos_usb.c` includes
`<sys/nvpair.h>` and calls `nvlist_alloc()`, `nvlist_add_int32()`,
`nvlist_pack()` and `nvlist_free()`, but the `sunos` branch of `configure.ac`
only added `-ldevinfo`. The Solaris link editor tolerates undefined symbols
when building a shared object, so `libusb-1.0.so` linked fine and this stayed
hidden; the static link that `tests/Makefile.am` asks for via
`AM_LDFLAGS = -static` fails:

```
Undefined                       first referenced
 symbol                             in file
nvlist_free       ../libusb/.libs/libusb-1.0.a(sunos_usb.o)
nvlist_add_int32  ../libusb/.libs/libusb-1.0.a(sunos_usb.o)
nvlist_pack       ../libusb/.libs/libusb-1.0.a(sunos_usb.o)
nvlist_alloc      ../libusb/.libs/libusb-1.0.a(sunos_usb.o)
ld: fatal: symbol referencing errors
```

So `make check` has never been able to link on illumos or Solaris. Fixed here,
since CI cannot be green without it.

**2. Haiku: `stress_mt` fails.** It builds and links, but the test reports:

```
Device count mismatch: Thread 2 discovered 1 devices instead of 0
Device count mismatch: Thread 3 discovered 0 devices instead of 1
```

The per-thread enumeration results disagree, which looks like a genuine bug in
the Haiku backend rather than a CI artefact, and fixing it is well outside this
PR. Rather than paper over it, that one test is excluded from the Haiku run
with the reason recorded next to it; the other three still run. Happy to open a
separate issue for it.

### Design notes

- `autoreconf` runs on the Ubuntu runner, not in the guest. That keeps the
  guest package set down to a compiler, GNU make and git. `configure` still
  runs inside the guest, so platform detection is genuine.
- `git` is installed in the guests because `configure.ac` invokes
  `build-aux/gen-describe.sh --classify`, which fails hard when `.git` is
  present but `git` is not on `PATH`. The job logs confirm the intended path is
  taken: `checking source identification for LIBUSB_DESCRIBE... git`.
- OmniOS and OpenIndiana need `system/header/header-ugen` and
  `system/header/header-usb` installed; their images ship no USB headers. The
  Solaris image already has them.
- `cache-after-prepare: true` caches the prepared VM image so later runs skip
  the package install.
- Failures `cat` `config.log` / `tests/test-suite.log` into the job log rather
  than uploading an artifact: when a guest command fails the action does not
  copy files back, so an artifact upload would find nothing.

### CI evidence

All runs below are on `b114fa11`, the exact head of this branch:

| workflow | result | run |
|---|---|---|
| NetBSD | `backend=NetBSD`, 24 CC / 26 CCLD, tests 4/4 PASS | https://github.com/neilpang/libusb/actions/runs/30352693083 |
| OpenBSD | `backend=OpenBSD`, 24 CC / 26 CCLD, tests 4/4 PASS | https://github.com/neilpang/libusb/actions/runs/30352693124 |
| illumos | `backend=SunOS` x3, 24 CC / 26 CCLD each, tests 4/4 PASS each | https://github.com/neilpang/libusb/actions/runs/30352693084 |
| Haiku | `backend=Haiku`, 23 CC / 26 CCLD, tests 3/3 PASS | https://github.com/neilpang/libusb/actions/runs/30352693117 |

The existing linux / macOS / Windows / MSYS2 x3 / Android workflows also pass
on that same commit, so the `configure.ac` change is not a regression.

### Cost

With the prepared-image cache warm, the new jobs take 1m30s (NetBSD), 1m38s
(OmniOS), 1m43s (OpenBSD), 2m04s (Haiku), 2m42s (Solaris) and 3m05s
(OpenIndiana) -- five of the six finish faster than the existing `linux` job
(3m57s) on the same run. The first run on a fresh cache key is slower; the
worst observed was 7m45s. This is the concern raised in #1466 and in
libusb/hidapi#477, so the numbers seemed worth stating.

### Things for you to decide

- These add six VM jobs to every pull request. If that is too much on the PR
  path, the `pull_request` trigger can be dropped so they only run on `master`
  pushes and tags, or they can be moved to a schedule.
- They use `actions/checkout@v6` to match the existing workflows. v7 is out;
  bumping all of them together felt like a separate change rather than
  something to bundle here.
- No README badges are added -- the README has none today, and adding a block
  of ten seemed a bigger stylistic change than this PR should make. Easy to add
  if you want them.

Closes: #1466

### DIFF
--- a/.github/workflows/haiku.yml
+++ b/.github/workflows/haiku.yml
@@ -1,0 +1,52 @@
+name: Haiku
+
+# Run branch pushes only on master; pull requests continue to run against all
+# branches.
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '**'
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      # Full history + tags are needed by build-aux/gen-describe.sh.
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      # autoreconf runs on the runner rather than in the guest. The Haiku image
+      # does ship autotools, but keeping the bootstrap on the runner makes this
+      # workflow behave the same way as the other VM ones.
+      - name: bootstrap
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends autoconf automake libtool m4
+          ./bootstrap.sh
+
+      # No prepare step: the Haiku image already ships gcc, GNU make and git.
+      - name: build and test in Haiku
+        uses: vmactions/haiku-vm@v1
+        with:
+          usesh: true
+          run: |
+            set -e
+            uname -a
+            gcc --version | head -1
+            mkdir build && cd build
+            ../configure --enable-examples-build --enable-tests-build \
+              || { cat config.log; exit 1; }
+            make -j4
+            # stress_mt is skipped on Haiku via an OS_HAIKU branch in
+            # tests/Makefile.am; see libusb/libusb#1907.
+            make check || { cat tests/test-suite.log; exit 1; }
--- a/.github/workflows/illumos.yml
+++ b/.github/workflows/illumos.yml
@@ -1,0 +1,131 @@
+name: illumos
+
+# The sunos backend is shared by illumos distributions and Oracle Solaris, so
+# all three are covered here.
+#
+# The OmniOS and OpenIndiana images do not ship the USB headers that
+# sunos_usb.c includes, so system/header/header-ugen (for
+# sys/usb/clients/ugen/usb_ugen.h) and system/header/header-usb (for
+# sys/usb/usba.h) are installed explicitly. The Solaris image already has them.
+#
+# Run branch pushes only on master; pull requests continue to run against all
+# branches.
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '**'
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  omnios:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      # Full history + tags are needed by build-aux/gen-describe.sh.
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      # autoreconf runs on the runner rather than in the guest: it keeps the
+      # guest package set down to a compiler, GNU make and git, which saves
+      # several minutes of emulated-VM time per run.
+      - name: bootstrap
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends autoconf automake libtool m4
+          ./bootstrap.sh
+
+      - name: build and test in OmniOS
+        uses: vmactions/omnios-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          prepare: |
+            pkg install developer/gcc14 developer/build/gnu-make developer/versioning/git \
+              system/header/header-ugen system/header/header-usb
+          run: |
+            set -e
+            PATH=/opt/gcc-14/bin:/usr/gnu/bin:$PATH
+            export PATH
+            uname -a
+            gcc --version | head -1
+            mkdir build && cd build
+            CC=gcc ../configure --enable-examples-build --enable-tests-build \
+              || { cat config.log; exit 1; }
+            gmake -j4
+            gmake check || { cat tests/test-suite.log; exit 1; }
+
+  openindiana:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: bootstrap
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends autoconf automake libtool m4
+          ./bootstrap.sh
+
+      - name: build and test in OpenIndiana
+        uses: vmactions/openindiana-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          prepare: |
+            pkg install developer/gcc-14 developer/build/gnu-make developer/versioning/git \
+              system/header/header-ugen system/header/header-usb
+          run: |
+            set -e
+            PATH=/usr/gnu/bin:$PATH
+            export PATH
+            uname -a
+            gcc --version | head -1
+            mkdir build && cd build
+            CC=gcc ../configure --enable-examples-build --enable-tests-build \
+              || { cat config.log; exit 1; }
+            gmake -j4
+            gmake check || { cat tests/test-suite.log; exit 1; }
+
+  solaris:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: bootstrap
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends autoconf automake libtool m4
+          ./bootstrap.sh
+
+      - name: build and test in Solaris
+        uses: vmactions/solaris-vm@v1
+        with:
+          usesh: true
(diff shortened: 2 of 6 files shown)


Youw (commented):
Codex found one additional coverage-maintenance issue beyond the existing Copilot threads and left it inline on the Haiku test invocation. The current upstream run was also checked against the latest PR head; the new VM jobs completed so far are passing.

.github/workflows/haiku.yml
```diff
@@ -0,0 +1,57 @@
...
+            ../configure --enable-examples-build --enable-tests-build \
+              || { cat config.log; exit 1; }
+            make -j4
+            # stress_mt is excluded: it fails on Haiku with "Device count
+            # mismatch: Thread N discovered X devices instead of Y", i.e. the
+            # per-thread enumeration results disagree. That looks like a real
+            # bug in the Haiku backend rather than a CI problem, so it is left
+            # out here instead of being papered over; the other three tests do
+            # run. Drop this override once it is fixed.
+            make check TESTS='stress set_option init_context' \
```
Youw: The command-line TESTS override replaces the canonical TESTS list in tests/Makefile.am, which currently expands from noinst_PROGRAMS. Any future test added there will therefore be silently absent from Haiku CI, exactly where this PR is intended to catch platform regressions. Please make stress_mt report an Automake skip on Haiku (exit 77), or add a project-level exclusion mechanism that removes only stress_mt, so newly added tests remain included automatically.

Youw:
@neilpang thanks for the contribution!

I'm sure all the review comments is something Opus can easily fix. Once done - we should be goot to merge.

neilpang:
Thanks @Youw -- all three addressed, and re-verified end to end.

**`illumos.yml`: Solaris didn't install GNU make** -- correct, and a good
catch. The job worked only because the Solaris 11.4 image happens to ship
`/usr/bin/gmake`; relying on that is exactly the kind of thing that breaks
silently later. `developer/build/gnu-make` is now installed explicitly.

**`netbsd.yml`: `pkg_add -u`** -- changed to plain `pkg_add`, since consistency
with the OpenBSD job is worth having. One factual note for the record, though:
the stated failure mode did not apply here. The NetBSD 10.1 base image has
neither `gmake` nor `git` (I probed it before writing the workflow), and
`/usr/sbin/pkg_add -u gmake git-base` installed both and produced a green run
-- NetBSD's `-u` installs a package that is not present rather than erroring.
So this is a readability change, not a bug fix.

**`haiku.yml`: the `TESTS` override** -- you were right, and thanks for
catching it; a hardcoded list is a denylist written as an allowlist and would
have silently dropped future tests on exactly the platform this PR exists to
watch.

I took your first suggestion: `stress_mt` now returns 77 on Haiku
(`tests/stress_mt.c`, new commit), so `TESTS = $(noinst_PROGRAMS)` is untouched
and anything added later is picked up automatically. The workflow is back to a
plain `make check`.

I did try the project-level exclusion first, but `$(filter-out ...)` in
`tests/Makefile.am` makes automake warn on every platform's bootstrap, not just
Haiku:

```
tests/Makefile.am:54: warning: filter-out stress_mt,$(noinst_PROGRAMS: non-POSIX variable name
tests/Makefile.am:54: (probably a GNU make extension)
```

That seemed a poor trade for everyone running `bootstrap.sh`, so I dropped it.
Happy to switch to `XFAIL_TESTS = stress_mt` instead if you would rather the
test keep running and turn the suite red the day Haiku starts passing it --
that self-alerts when #1907 is fixed, at the cost of going red if the failure
turns out to be intermittent.

Haiku now reports it as a skip, with the count intact:

```
PASS: stress
SKIP: stress_mt
PASS: set_option
PASS: init_context
# TOTAL: 4
# PASS:  3
# SKIP:  1
# FAIL:  0
```

`# TOTAL: 4` rather than 3 is the part that matters -- the test list is still
derived from `noinst_PROGRAMS`.

All four workflows green on `3ca212c0`:

| workflow | run |
|---|---|
| NetBSD | https://github.com/neilpang/libusb/actions/runs/30440862236 |
| OpenBSD | https://github.com/neilpang/libusb/actions/runs/30440862215 |
| illumos | https://github.com/neilpang/libusb/actions/runs/30440862210 |
| Haiku | https://github.com/neilpang/libusb/actions/runs/30440862167 |

.github/workflows/haiku.yml
```diff
@@ -0,0 +1,52 @@
...
+      # workflow behave the same way as the other VM ones.
+      - name: bootstrap
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends autoconf automake libtool m4
+          ./bootstrap.sh
+
+      # No prepare step: the Haiku image already ships gcc, GNU make and git.
+      - name: build and test in Haiku
+        uses: vmactions/haiku-vm@v1
```
Youw, line 39: [P1] Restrict the write-enabled `GITHUB_TOKEN`

The repository's Actions settings currently report `default_workflow_permissions: write`. Because these workflows declare no `permissions`, this third-party action inherits that write-enabled default on master pushes and same-repository PR runs; GitHub actions can access `github.token` even when it is not passed explicitly. These jobs only check out, build, and test, so add top-level `permissions: contents: read` and apply the same restriction to all four new workflows. Pinning each `vmactions/*@v1` reference to the full commit SHA exercised by the green run would further reduce the mutable-tag supply-chain risk.

Youw (changes requested):
Codex re-reviewed current head `3ca212c0419e00301bab08bcc043932dfad6b9e3` against base `6bb97402df0ec0c09defcbd4f5146447c7f7cc53` and checked the complete existing review conversation.

All three earlier findings are fixed: NetBSD now uses the requested install form, Solaris explicitly installs GNU make, and Haiku preserves the canonical test list while reporting `stress_mt` as an Automake skip. Exact-head logs confirm all 11 fork workflows passed; NetBSD and Solaris each ran 4/4 tests, and Haiku reported 3 passes plus the intended single skip. `git diff --check` is clean. Upstream AppVeyor is still queued.

One new workflow-security issue remains inline. The repository currently grants workflows a write-enabled default `GITHUB_TOKEN`, while all four new workflows invoke third-party VM actions without narrowing `permissions`. Least-privilege token permissions should be added before approval.

Two non-blocking documentation cleanups are also worth making: the Haiku workflow comment points to an `OS_HAIKU` branch in `tests/Makefile.am`, but the skip is implemented with `__HAIKU__` in `tests/stress_mt.c`; and the PR description's CI table still labels `b114fa11` as the exact head and describes the old 3/3 Haiku run instead of the current 3-pass/1-skip run on `3ca212c0`.

.github/workflows/haiku.yml
Youw: postpone for now

Youw (approved):
This is fine at this point.
As for access restriction issue found by codex - that is not unique to this PR, other existing worflows also needs fixing, I suggest doing so in a separate PR, for all workflows at once.

Youw:
@mcuee, @sonatique unless you have anything to add here - I'll merge this soon

Issue #1781, opened 2026-03-10
stress_mt fails under Haiku OS


mcuee:
stress_mt fails under Haiku OS.

R1 beta 5
```
~/build/libusb_pr1779> ./examples/listdevs
0000:0000 (bus 1, device 101)
04b4:8613 (bus 1, device 1)
0000:0000 (bus 0, device 101)
16c0:05dc (bus 0, device 1)
~/build/libusb_pr1779> ./examples/testlibusb
Dev (bus 1, device 101): 0000 - 0000 speed: Unknown
  Manufacturer:              HAIKU Inc.
  Product:                   EHCI RootHub
Dev (bus 1, device 1): 04B4 - 8613 speed: Unknown
Dev (bus 0, device 101): 0000 - 0000 speed: Unknown
  Manufacturer:              HAIKU Inc.
  Product:                   OHCI RootHub
Dev (bus 0, device 1): 16C0 - 05DC speed: Unknown
  Manufacturer:              www.fischl.de
  Product:                   USBasp
~/build/libusb_pr1779> make check
Making check in libusb
make[1]: Entering directory '/boot/home/build/libusb_pr1779/libusb'
make[1]: Nothing to be done for 'check'.
make[1]: Leaving directory '/boot/home/build/libusb_pr1779/libusb'
Making check in examples
make[1]: Entering directory '/boot/home/build/libusb_pr1779/examples'
make[1]: Nothing to be done for 'check'.
make[1]: Leaving directory '/boot/home/build/libusb_pr1779/examples'
Making check in tests
make[1]: Entering directory '/boot/home/build/libusb_pr1779/tests'
make  check-TESTS
make[2]: Entering directory '/boot/home/build/libusb_pr1779/tests'
make[3]: Entering directory '/boot/home/build/libusb_pr1779/tests'
PASS: stress
FAIL: stress_mt
PASS: set_option
PASS: init_context
============================================================================
Testsuite summary for libusb-1.0 1.0.30-rc1
============================================================================
# TOTAL: 4
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See tests/test-suite.log for debugging.
Some test(s) failed.  Please report this to libusb-devel@lists.sourceforge.net,
together with the test-suite.log file (gzipped) and your system
information.  Thanks.
============================================================================
make[3]: *** [Makefile:808: test-suite.log] Error 1
[...]

PR #1909, opened 2026-07-30
DO NOT MERGE: baseline for the Haiku stress_mt failure

Youw:
Throwaway branch, opened only to record the "before" half of the evidence for the Haiku enumeration fix, on libusb's own runner rather than a fork's.

No backend change. It removes the `__HAIKU__` skip that #1906 added to `tests/stress_mt.c`, so the Haiku job runs the test against the current backend, and adds a `listdevs` check first so a green run could not be mistaken for a vacuous one.

Expected result: red, with `Device count mismatch: Thread N discovered 0 devices instead of 1`.

Will be closed as soon as the run finishes.

### DIFF
--- a/.github/workflows/haiku.yml
+++ b/.github/workflows/haiku.yml
@@ -52,4 +52,16 @@ jobs:
             # USB device at all. Fail loudly instead of testing nothing.
             ./examples/listdevs | tee listdevs.out
             test -s listdevs.out || { echo "no USB devices in the guest"; exit 1; }
+            # TEMPORARY (drop before merge): the failure is scheduling
+            # dependent, so a single run proves little either way. Repeat it
+            # inside the guest that is already booted and report the rate;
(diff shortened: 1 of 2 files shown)


Youw:
Recorded: **10 failures in 25 runs** of `stress_mt` on the current backend, plus a `make check` failure with `Device count mismatch: Thread 1 discovered 0 devices instead of 1` — https://github.com/libusb/libusb/actions/runs/30533153444

Worth noting for anyone reading #1781: on the CI guest, which publishes a single root hub, the failure is scheduling dependent rather than deterministic. The [first run here](https://github.com/libusb/libusb/actions/runs/30532892243) passed with no fix at all, which is why the rate is measured over 25 repeats instead of trusting one run. mcuee's 4-device hardware is a much stronger reproducer.

Baseline captured; closing. The fix is #1910.

PR #1910, opened 2026-07-30
haiku: enumerate existing devices into every new context

### LINKED ISSUE #1781: stress_mt fails under Haiku OS

### LINKED ISSUE #1907: Haiku: stress_mt reports device count mismatches between threads
Splitting this out of #1906, which adds CI for the Haiku backend. The new
Haiku job builds and links fine, but `make check` fails in `stress_mt`.

### What happens

```
Running multithreaded init/exit test...
Starting 8 threads
Running multithreaded init/exit test with enumeration...
Starting 8 threads
Device count mismatch: Thread 2 discovered 1 devices instead of 0
Device count mismatch: Thread 3 discovered 0 devices instead of 1
Device count mismatch: Thread 4 discovered 1 devices instead of 0
Device count mismatch: Thread 5 discovered 0 devices instead of 1
All done, 4 errors
FAIL stress_mt (exit status: 1)
```

The first phase (init/exit without enumeration) passes; only the phase with
enumeration fails. Each of the 8 threads calls `libusb_init()` and
`libusb_get_device_list()` and records its device count, and
`tests/stress_mt.c:238` compares each thread's count with the previous
thread's. On Haiku those counts are not stable across threads.

One observation, offered as data rather than a diagnosis: the counts alternate
cleanly (0, 1, 0, 1, ...) rather than varying randomly, and threads 6-8 agreed
with their predecessors, giving exactly 4 mismatches. That regularity suggests
something more structured than a plain race, but I have not investigated the
Haiku backend.

The other three tests (`stress`, `set_option`, `init_context`) pass.

### Environment

- Haiku `hrev57937+113`, x86_64, running under QEMU (the `vmactions/haiku-vm`
  image, R1/beta5)
- libusb `1.0.30` at master `af42f358`
- gcc, built with `./configure --enable-examples-build --enable-tests-build`

### Reproducing

Full job log (the run predates the workaround, so the failure is visible):
https://github.com/neilpang/libusb/actions/runs/30346905816

Or locally on any Haiku box:

```bash
./bootstrap.sh && mkdir build && cd build
../configure --enable-examples-build --enable-tests-build
make -j4 && make check
```

### Relation to #1906

Because this looks like a genuine backend bug rather than a CI artefact, #1906
does not paper over it -- the Haiku job runs the other three tests and skips
`stress_mt`, with the reason recorded next to the exclusion:

```yaml
make check TESTS='stress set_option init_context'
```

That override should be dropped once this is fixed.

Youw:
`stress_mt` fails on Haiku because contexts disagree about how many devices
exist:

```
Device count mismatch: Thread 2 discovered 1 devices instead of 0
Device count mismatch: Thread 3 discovered 0 devices instead of 1
```

### Why

`haiku_init()` starts the roster on the `gInitCount` 0->1 transition and does
nothing else. The only code that ever puts a device into a context is the
`WatchedEntry` constructor, which walks `for_each_context()` once — while the
roster is building its tree. The backend has no `get_device_list` and no
`hotplug_poll`, so `libusb_get_device_list()` only walks `ctx->usb_devs`.

A context created while the roster is already running therefore never receives
the devices that are already there, and nothing ever refills its list. It sees
zero, permanently. `stress_mt` keeps up to eight contexts alive at once, so
whichever one happened to start the roster sees the devices and the rest see
none — which is also why its first, non-enumerating phase passes.

The other hotplug backends all split "start the global monitor once" from
"scan into this context, every time": `linux_usbfs.c` (`op_init` ->
`linux_scan_devices`), `windows_common.c` (`windows_init` ->
`windows_initial_scan_devices`), `darwin_usb.c` (`darwin_init_context` ->
`darwin_scan_devices`). Haiku only had the first half.

### Fix

The roster keeps a registry of the devices it knows about, and
`USBRoster::Init()` replays that registry into each new context. Device
arrival, device removal and this initial scan all go through the same
per-(device, context) helpers, under one lock, so the two paths cannot
disagree or add a device twice.

`get_device_list` and `hotplug_poll` stay `NULL`, so
`libusb_has_capability(LIBUSB_CAP_HAS_HOTPLUG)` is unchanged and there is no
API or ABI change. Hotplug still works: `ctx->hotplug_ready` is still 0 while
`backend.init` runs, so the initial scan produces no spurious ARRIVED events,
exactly as on the other backends.

A handful of smaller defects in the functions this touches are fixed in the
same commit — roster start/stop serialization, a use-after-free in
`RosterLooper::Stop()` (the entry tree was deleted before `Quit()`, which
still dispatches queued messages that walk it), a leaked looper on the
`Start()` failure path, and the device-list reference that nothing inherits
when a context is not yet hotplug-ready. They are listed in the commit
message.

### Testing

`tests/stress_mt.c` loses the `__HAIKU__` skip added in #1906.

The test only compares the per-thread counts against each other, so it also
passes when every thread sees nothing — the workflow now runs `listdevs`
first and fails if the guest publishes no USB device, so a green run cannot be
vacuous.

A caveat that shaped how this was measured: on the CI guest, which publishes
a single root hub, the failure is scheduling dependent rather than
deterministic. [The first baseline run](https://github.com/libusb/libusb/actions/runs/30532892243)
removed the skip with no backend change and *passed*, so one green run would
have proved nothing. Both sides therefore run `stress_mt` 25 times inside the
already-booted guest and report the failure rate. (mcuee's four-device
hardware in #1781 is a considerably stronger reproducer than CI.)

Before and after, same runner, same measurement:

| | `stress_mt`, 25 in-guest runs | `make check` | run |
|---|---|---|---|
| before, no backend change | **10 failures** | FAIL, `Thread 1 discovered 0 devices instead of 1` | [30533153444](https://github.com/libusb/libusb/actions/runs/30533153444) |
| after, run 1 | **0 failures** | PASS | [30533222480 attempt 1](https://github.com/libusb/libusb/actions/runs/30533222480/attempts/1) |
| after, run 2 | **0 failures** | PASS | [attempt 2](https://github.com/libusb/libusb/actions/runs/30533222480/attempts/2) |
| after, run 3 | **0 failures** | PASS | [attempt 3](https://github.com/libusb/libusb/actions/runs/30533222480/attempts/3) |

75 consecutive passes against a measured 40% per-run failure rate. The commit
carrying that loop has been dropped again, so what is proposed here runs
`stress_mt` once, as every other platform does.

Off Haiku, all three backend sources were checked with
`g++ -fsyntax-only -Wall -Wextra` against hand-written Be API stubs; WSL
cannot run Haiku binaries, so that only covers the build.

### Not fixed here

`USBDevice` ownership. A raw `USBDevice*` is stored in every context's device
priv, `.destroy_device` is `NULL`, and `~WatchedEntry` deletes it — so
unplugging a device while an application still holds a reference or an open
handle reads freed memory. That is a design change rather than a bug fix, it
is not made worse by this commit, and it wants its own pull request. Same for
the per-device state (`fClaimedInterfaces`, `fActiveConfiguration`) that is
shared process-wide instead of per context, and for the transfer path in
`haiku_usb_backend.cpp`. I will open an issue with the details.

Fixes: #1781
Closes: #1907

### DIFF
--- a/libusb/os/haiku_pollfs.cpp
+++ b/libusb/os/haiku_pollfs.cpp
@@ -18,6 +18,29 @@
 #include <Path.h>
 #include <cstring>
 
+class WatchedEntry;
+
+/* Guards gDeviceRegistry and, with it, the transitions that add a device to a
+   context or take it away again. Held by the roster looper thread (device
+   arrival and removal) and by USBRoster::EnumerateInto() (a context being
+   created), so that those two paths cannot interleave for the same device.
+
+   Lock order is gInitLock -> gRosterLock -> active_contexts_lock ->
+   ctx->usb_devs_lock. The looper thread never takes gInitLock, which is what
+   makes it safe to hold gInitLock across the blocking RosterLooper::Stop(). */
+static usbi_mutex_static_t gRosterLock = USBI_MUTEX_INITIALIZER;
+
+/* Every device the roster currently knows about, in discovery order, so that
+   a context populated from the registry ends up with the same device order as
+   one populated by the constructor loop below. */
+static WatchedEntry *gDeviceRegistry GUARDED_BY(gRosterLock) = NULL;
+
+/* Guards the roster reference count and USBRoster::fLooper. Never taken by
+   the looper thread; see the lock order above. */
+static usbi_mutex_static_t gInitLock = USBI_MUTEX_INITIALIZER;
+static int gInitCount GUARDED_BY(gInitLock) = 0;
+
+
 class WatchedEntry {
 public:
 			WatchedEntry(BMessenger *, entry_ref *);
@@ -26,13 +49,25 @@ public:
 	bool		EntryRemoved(ino_t node);
 	bool		InitCheck();
 
+	/* Called for a single context, with gRosterLock held. */
+	void		AddToContext(struct libusb_context *ctx);
+	void		RemoveFromContext(struct libusb_context *ctx);
+	WatchedEntry*	RegistryLink() const { return fRegistryLink; }
+
 private:
+	void		Register() REQUIRES(gRosterLock);
+	void		Unregister() REQUIRES(gRosterLock);
+
 	BMessenger*	fMessenger;
 	node_ref	fNode;
 	bool		fIsDirectory;
 	USBDevice*	fDevice;
+	unsigned long	fSessionID;
+	uint8		fBusNumber;
+	uint8		fDeviceAddress;
 	WatchedEntry*	fEntries;
 	WatchedEntry*	fLink;
+	WatchedEntry*	fRegistryLink;
 	bool		fInitCheck;
 };
 
@@ -52,12 +87,102 @@ private:
 };
 
 
+void
+WatchedEntry::Register()
+{
+	WatchedEntry **tail = &gDeviceRegistry;
+	while (*tail != NULL)
+		tail = &(*tail)->fRegistryLink;
+	*tail = this;
+}
+
+
+void
+WatchedEntry::Unregister()
+{
+	WatchedEntry **link = &gDeviceRegistry;
+	while (*link != NULL && *link != this)
+		link = &(*link)->fRegistryLink;
+	if (*link == this)
+		*link = fRegistryLink;
+	fRegistryLink = NULL;
+}
+
+
+void
+WatchedEntry::AddToContext(struct libusb_context *ctx)
+{
+	struct libusb_device *dev = usbi_get_device_by_session_id(ctx, fSessionID);
+	if (dev != NULL) {
+		usbi_dbg(ctx, "using previously allocated device with location %lu", fSessionID);
+		libusb_unref_device(dev);
+		return;
+	}
+
+	usbi_dbg(ctx, "allocating new device with location %lu", fSessionID);
+	dev = usbi_alloc_device(ctx, fSessionID);
+	if (dev == NULL) {
+		/* Logged as an error rather than debug: a context that silently
+		   ends up with fewer devices than its siblings is exactly the
+		   symptom this backend used to have. */
+		usbi_err(ctx, "device allocation failed");
+		return;
+	}
+
+	*((USBDevice **)usbi_get_device_priv(dev)) = fDevice;
+	dev->bus_number = fBusNumber;
+	dev->device_address = fDeviceAddress;
+
+	static_assert(sizeof(dev->device_descriptor) == sizeof(usb_device_descriptor),
+		      "mismatch between libusb and OS device descriptor sizes");
+	memcpy(&dev->device_descriptor, fDevice->Descriptor(), LIBUSB_DT_DEVICE_SIZE);
+	usbi_localize_device_descriptor(&dev->device_descriptor);
+
+	if (usbi_sanitize_device(dev) < 0) {
+		usbi_err(ctx, "device sanitization failed");
+		libusb_unref_device(dev);
+		return;
+	}
+
+	usbi_connect_device(dev);
+}
+
+
+void
+WatchedEntry::RemoveFromContext(struct libusb_context *ctx)
+{
+	struct libusb_device *dev = usbi_get_device_by_session_id(ctx, fSessionID);
+	if (dev == NULL) {
+		usbi_dbg(ctx, "device with location %lu not found", fSessionID);
+		return;
+	}
+
+	if (usbi_atomic_load(&ctx->hotplug_ready)) {
+		/* The device-left message takes over the reference held by the
+		   context's device list. */
+		usbi_disconnect_device(dev);
+	} else {
+		/* A context is only enumerated but not yet hotplug-ready while
+		   it is inside libusb_init_context(); no message is posted
+		   there, so nothing would inherit that reference. */
+		usbi_detach_device(dev);
+		libusb_unref_device(dev);
+	}
+
+	libusb_unref_device(dev);
+}
+
+
 WatchedEntry::WatchedEntry(BMessenger *messenger, entry_ref *ref)
 	:	fMessenger(messenger),
 		fIsDirectory(false),
 		fDevice(NULL),
+		fSessionID(0),
+		fBusNumber(0),
+		fDeviceAddress(0),
 		fEntries(NULL),
 		fLink(NULL),
+		fRegistryLink(NULL),
 		fInitCheck(false)
 {
 	BEntry entry(ref);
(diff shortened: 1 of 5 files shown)


Youw:
Verification is complete.

The failure is scheduling dependent on the CI guest, which publishes a single root hub, so a single run settles nothing — [the first baseline run](https://github.com/libusb/libusb/actions/runs/30532892243) passed with no fix at all. Both sides were therefore measured as a rate over 25 in-guest repeats:

| | `stress_mt`, 25 runs | `make check` |
|---|---|---|
| no backend change | **10 failures** | FAIL |
| with the fix, x3 | **0 failures** each | PASS |

75 consecutive passes against a measured 40% per-run failure rate. The temporary commit that carried the repeat loop has been dropped; [the final run of what is proposed here](https://github.com/libusb/libusb/actions/runs/30534078635) is green with all five tests passing.

The `solaris` check is red for an unrelated pre-existing reason — `memset` undeclared in `examples/hotplugtest.c` — which also fails on master and is fixed by #1911. Everything else is green.

Follow-ups deliberately left out of this pull request are collected in #1912.

libusb/os/haiku_pollfs.cpp
```diff
@@ -18,6 +18,29 @@
...
+class WatchedEntry;
+
+/* Guards gDeviceRegistry and, with it, the transitions that add a device to a
+   context or take it away again. Held by the roster looper thread (device
+   arrival and removal) and by USBRoster::EnumerateInto() (a context being
+   created), so that those two paths cannot interleave for the same device.
+
+   Lock order is gInitLock -> gRosterLock -> active_contexts_lock ->
+   ctx->usb_devs_lock. The looper thread never takes gInitLock, which is what
+   makes it safe to hold gInitLock across the blocking RosterLooper::Stop(). */
```
Youw: Nit: the documented chain skips a lock that actually participates. The `RosterLooper` BLooper lock sits between `gInitLock` and `gRosterLock` on every path that takes both — looper dispatch holds it before taking `gRosterLock`, the constructor's initial tree build holds it (the `Lock()`/`Unlock()` around the `WatchedEntry` walk) before the ctors take `gRosterLock`, and `Stop()` acquires it under `gInitLock` before the tree teardown takes `gRosterLock`. Its position is load-bearing for the deadlock-freedom argument this comment makes: the invariant that must hold is that nothing acquires the looper lock while holding `gRosterLock`. Since this comment is the documentation of record for the ordering, worth spelling that out:

```suggestion
   Lock order is gInitLock -> RosterLooper's BLooper lock -> gRosterLock ->
   active_contexts_lock -> ctx->usb_devs_lock. The BLooper lock sits between
   the first two on every path that takes both: the looper thread holds it
   while dispatching device arrival/removal, the initial tree build in the
   RosterLooper constructor holds it around the WatchedEntry walk, and
   RosterLooper::Stop() takes it before deleting the tree. Nothing may lock
   the looper while holding gRosterLock. The looper thread never takes
   gInitLock, which is what makes it safe to hold gInitLock across the
   blocking RosterLooper::Stop(). */
```

*— suggestion by Claude*

libusb/os/haiku_pollfs.cpp
```diff
@@ -26,13 +49,25 @@ class WatchedEntry {
 	bool		EntryRemoved(ino_t node);
 	bool		InitCheck();
 
+	/* Called for a single context, with gRosterLock held. */
+	void		AddToContext(struct libusb_context *ctx);
+	void		RemoveFromContext(struct libusb_context *ctx);
```
Youw: Optional: the precondition lives in prose here while `Register()`/`Unregister()` get real `REQUIRES` annotations. Since `gRosterLock` is declared above the class, clang Thread Safety Analysis can enforce this contract too — it would flag any future caller that forgets the lock (the exact interleaving bug this PR is curing):

```suggestion
	/* Called for a single context, with gRosterLock held. */
	void		AddToContext(struct libusb_context *ctx) REQUIRES(gRosterLock);
	void		RemoveFromContext(struct libusb_context *ctx) REQUIRES(gRosterLock);
```

*— suggestion by Claude*

Youw (commented):
> **🤖 Automated review by Claude** — this review and its inline `suggestion` blocks were generated by Claude (model `claude-fable-5`, via Claude Code) at a maintainer's request. Findings are advisory; I could not run the code on Haiku.

## Summary
The implementation is sound. The registry + replay design is the right shape — it gives Haiku the same "start the monitor once / scan into every context" split every other hotplug backend has, and routing arrival, removal and initial scan through the same per-(device, context) helpers under one lock is what makes the two paths unable to disagree. I found **no correctness issues**; two documentation/hardening nits inline.

## Verified
- **Locking**: every path nests consistently — looper dispatch (BLooper lock → `gRosterLock` → `active_contexts_lock`), initial tree build in the `RosterLooper` constructor (which takes the looper lock around the `WatchedEntry` walk, so `fRoot` is published before any dispatch can run — no dropped-message or torn-tree window), `EnumerateInto` (`gInitLock` → `gRosterLock`), and `Stop()` (`gInitLock` → looper lock → `gRosterLock`). The deadlock-freedom argument for holding `gInitLock` across the blocking `Stop()` holds: `gRosterLock` is only ever held for bounded, non-looper-touching work, and the looper thread never takes `gInitLock`.
- **Refcounting in `RemoveFromContext()`**: matches core exactly. `usbi_hotplug_notification()` early-returns when `!ctx->hotplug_ready`, so for a context still inside `libusb_init_context()` nothing would ever consume the device-list reference — the `usbi_detach_device()` + extra unref branch is genuinely required, and the `usbi_disconnect_device()` branch correctly hands the list reference to the DEVICE_LEFT message (`handle_hotplug_msgs` unrefs `msg->device`). The `hotplug_ready` read cannot race the exit path: 1→0 happens in `usbi_hotplug_exit()` only *after* `libusb_exit()` unlinks the context under `active_contexts_lock`, which the remover holds for the whole walk.
- **No missed/duplicated devices**: a context being created is already in `active_contexts` during `haiku_init()`, so an arriving device is added by whichever of the looper loop or `EnumerateInto()` runs second — and the session-id dedup in `AddToContext()` makes that idempotent. Session-id (heap pointer) reuse is safe because removal from every context happens under `gRosterLock` before `~WatchedEntry` frees the `USBDevice`, i.e. before the address can recycle.
- **`gInitCount` removal**: no references remain in `haiku_usb_backend.cpp` or `haiku_usb.h` — the linkage change is safe. The old `atomic_add` counter also left `Start()`/`Stop()` unserialized against each other (a 1→0 exit racing a 0→1 init could tear `fLooper`), which `gInitLock` now fixes properly — worth having even though core's `default_context_lock` incidentally serializes backend init/exit today, since `~USBRoster()` runs outside that lock.
- **Lifecycle fixes**: the `Stop()` use-after-free fix is correct (`fRoot` cleared under the looper lock happens-before any post-`Quit()` dispatch, which the new NULL guard then catches; the messenger is copied to a local because `Quit()` frees `this`). The `Start()` failure path now reaps the already-running looper thread via `Stop()` instead of leaking it, and `EntryCreated()` no longer links init-failed children into the tree.
- **CI guard**: the `listdevs`-must-be-non-empty check closes the vacuous-pass hole in `stress_mt` (it only compares per-thread counts, so all-zeros passes); with the skip removed this makes a green Haiku run meaningful. The 25-run before/after methodology in the PR body is the right response to a scheduling-dependent failure with a measured ~40% baseline rate.

## Nits (inline)
1. The lock-order comment omits the `RosterLooper` BLooper lock, which in reality sits between `gInitLock` and `gRosterLock` on every path that takes both — and whose position is load-bearing (nothing may lock the looper while holding `gRosterLock`). Suggestion inline.
2. `AddToContext()`/`RemoveFromContext()` document their `gRosterLock` precondition in prose; annotating them `REQUIRES(gRosterLock)` lets clang Thread Safety Analysis enforce what the comment promises, matching `Register()`/`Unregister()`.

The deferred `USBDevice` ownership problem (raw pointer in device priv, `.destroy_device == NULL`, freed by `~WatchedEntry` regardless of outstanding app references) is real and correctly scoped out — it predates this change and wants the promised separate issue.

Youw:
Both review rounds addressed in 83974b0d. Second round was Codex `gpt-5.6-sol` at xhigh reasoning.

**Codex, major — `BLooper::Run()` failure was ignored.** This is real and I have verified it against [Haiku's `Looper.cpp`](https://github.com/haiku/haiku/blob/master/src/kits/app/Looper.cpp): when `spawn_thread()` fails or `fMsgPort < B_OK`, `Run()` returns early **without calling `Unlock()`** and **without setting `fRunCalled`**.

The consequence is worse than a missed error. A `BLooper` is constructed locked by the creating thread, so with `Run()` failed that lock is still held — which means the `Lock()` further down the constructor *nests* rather than blocks, the tree gets built, and `fInitCheck` is set. `Start()` reports success for a roster that has no thread behind it and never delivers a hotplug event. Then the first `Stop()` from any other thread blocks forever in `Lock()` **while holding `gInitLock`**, hanging every subsequent `libusb_init_context()` and `libusb_exit()` in the process.

Now checked. `Start()` takes the object down via `Stop()`, where `Quit()` sees `fRunCalled` unset and deletes it outright, and `gInitCount` is untouched so the next context retries.

**Codex, minor — DEVICE_LEFT allocation failure leaks the list reference.** Agreed and correctly characterised: if the `calloc()` in `usbi_hotplug_notification()` fails, nothing inherits the reference `usbi_disconnect_device()` handed over. That is core-wide and identical for the linux and windows backends, so it does not belong in a Haiku fix. Noted rather than papered over — the "nothing leaks on either path" claim in the earlier discussion holds only modulo that allocation failing.

**Claude, nit 1 — lock order comment.** Applied. The `RosterLooper` BLooper lock is now named, along with the invariant the argument actually rests on: nothing may lock the looper while holding `gRosterLock`.

**Claude, nit 2 — `REQUIRES` annotations.** Applied to `AddToContext()`/`RemoveFromContext()`, matching `Register()`/`Unregister()`.

Off-Haiku syntax check still clean (`g++ -fsyntax-only -Wall -Wextra` against Be API stubs). Re-running Haiku CI.

libusb/os/haiku_pollfs.cpp
```diff
@@ -52,12 +93,102 @@ class RosterLooper : public BLooper {
...
+	while (*link != NULL && *link != this)
+		link = &(*link)->fRegistryLink;
+	if (*link == this)
+		*link = fRegistryLink;
+	fRegistryLink = NULL;
+}
+
+
+void
+WatchedEntry::AddToContext(struct libusb_context *ctx)
```
seanm, line 119: Personally I like the idea of repeating the `REQUIRES(gRosterLock)` on the method implementation too, so one doesn't have to scroll up a lot to see the prototype.

seanm (commented):
lgtm at a glance, but I don't know or use haiku....

mcuee:
Let me try to test over the weekend or next week.

PR #1913, opened 2026-07-30
ci: gate testcore stress test

Youw:
## Summary

- Keep building `testcore`, but exclude it from the default `make check` test list.
- Add `--enable-testcore` to include the stress test explicitly.
- Enable `testcore` automatically for `master` builds and for pull requests carrying either the `force_testcore` or `force_long_tests` label.
- Trigger on PR label additions while skipping unrelated label-triggered workflow runs.

## Rationale

`testcore` performs 10,000 concurrent enumeration loops and adds roughly 70–110 seconds to each MSYS2 test configuration. Running it on every pull request substantially increases CI time without removing coverage from `master`; the labels preserve an immediate opt-in path for relevant PRs.

## Validation

- `actionlint` 1.7.12 on all workflow files
- `bash -n .private/ci-build.sh`
- Autotools bootstrap and default/forced out-of-tree configurations
- Default configuration generated four tests without `testcore`
- `--enable-testcore` generated the fifth test and `testcore` passed
- `.private/ci-build.sh` forwarded `TESTCORE_CONFIGURE_FLAG=--enable-testcore`

Related to #1809

Assisted-by: codex:gpt-5.6-sol

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,14 +19,18 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  cancel-in-progress: >-
+    ${{ github.event_name == 'pull_request' &&
+        (github.event.action != 'labeled' ||
+         github.event.label.name == 'force_testcore' ||
+         github.event.label.name == 'force_long_tests') }}
 
 jobs:
   c23:
     if: >-
       github.event.action != 'labeled' ||
-      contains(github.event.pull_request.labels.*.name, 'force_testcore') ||
-      contains(github.event.pull_request.labels.*.name, 'force_long_tests')
+      github.event.label.name == 'force_testcore' ||
+      github.event.label.name == 'force_long_tests'
     name: GCC 16 / C23
     runs-on: ubuntu-latest
     container: gcc:16
@@ -59,8 +63,8 @@ jobs:
   build:
     if: >-
       github.event.action != 'labeled' ||
-      contains(github.event.pull_request.labels.*.name, 'force_testcore') ||
-      contains(github.event.pull_request.labels.*.name, 'force_long_tests')
+      github.event.label.name == 'force_testcore' ||
+      github.event.label.name == 'force_long_tests'
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -7,6 +7,15 @@ on:
     tags:
       - '**'
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
+
+# Keep the long-running stress test on master unless a PR explicitly opts in.
+env:
+  TESTCORE_CONFIGURE_FLAG: >-
+    ${{ (github.ref == 'refs/heads/master' ||
+         contains(github.event.pull_request.labels.*.name, 'force_testcore') ||
+         contains(github.event.pull_request.labels.*.name, 'force_long_tests'))
+        && '--enable-testcore' || '' }}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -14,6 +23,10 @@ concurrency:
 
 jobs:
   c23:
+    if: >-
+      github.event.action != 'labeled' ||
+      github.event.label.name == 'force_testcore' ||
+      github.event.label.name == 'force_long_tests'
     name: GCC 16 / C23
     runs-on: ubuntu-latest
     container: gcc:16
@@ -44,6 +57,10 @@ jobs:
           make -j"$(nproc)"
 
   build:
+    if: >-
+      github.event.action != 'labeled' ||
+      github.event.label.name == 'force_testcore' ||
+      github.event.label.name == 'force_long_tests'
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,8 +25,8 @@ jobs:
   c23:
     if: >-
       github.event.action != 'labeled' ||
-      github.event.label.name == 'force_testcore' ||
-      github.event.label.name == 'force_long_tests'
+      contains(github.event.pull_request.labels.*.name, 'force_testcore') ||
+      contains(github.event.pull_request.labels.*.name, 'force_long_tests')
     name: GCC 16 / C23
     runs-on: ubuntu-latest
     container: gcc:16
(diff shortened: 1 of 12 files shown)


.github/workflows/linux.yml
```diff
@@ -7,13 +7,26 @@ on:
     tags:
       - '**'
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
Youw, line 10: 🟠 **Adding any label mid-run cancels in-flight CI.** The `labeled` event creates a new workflow run in the same concurrency group (`<workflow>-<PR number>`), and `cancel-in-progress` is `true` for every `pull_request` event — including `labeled`. So the sequence “CI running → maintainer adds an unrelated label (`windows`, `core`, …)” cancels the real run, and the replacement run then skips itself via the job `if`. The commit's checks end as cancelled + skipped until someone re-runs or pushes.

The `cancel-in-progress` line sits just outside this diff hunk, so no one-click suggestion — but the fix is one expression, in all nine workflows. Cancelling only when the *added* label is one of the force labels keeps the snappy “label it and CI restarts with testcore” behavior while making unrelated labels harmless:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: >-
    ${{ github.event_name == 'pull_request' &&
        (github.event.action != 'labeled' ||
         github.event.label.name == 'force_testcore' ||
         github.event.label.name == 'force_long_tests') }}
```

(`github.event.label` is the label that was just added; it is null for non-labeled events, where the comparison safely evaluates false.)

*— review by Claude*

.github/workflows/linux.yml
```diff
@@ -7,13 +7,26 @@ on:
...
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   c23:
+    if: >-
+      github.event.action != 'labeled' ||
+      contains(github.event.pull_request.labels.*.name, 'force_testcore') ||
+      contains(github.event.pull_request.labels.*.name, 'force_long_tests')
```
Youw: Related refinement: keying the guard on `github.event.label.name` (the label *just added*) instead of `contains()` over all current labels avoids a wasteful full re-run when an **unrelated** label is added to a PR that already carries `force_testcore` — with `contains()`, that labeled event passes the guard and duplicates the entire matrix for no new information. With `event.label.name`, only adding one of the two force labels retriggers; everything else skips.

Same change applies to the equivalent `if:` blocks in the other workflows (and pairs with the `cancel-in-progress` fix above).

```suggestion
      github.event.action != 'labeled' ||
      github.event.label.name == 'force_testcore' ||
      github.event.label.name == 'force_long_tests'
```

*— suggestion by Claude*

Youw (commented):
> **🤖 Automated review by Claude** — this review and its inline `suggestion` blocks were generated by Claude (model `claude-fable-5`, via Claude Code) at a maintainer's request.

## Summary
The gating design is right: `testcore` stays built everywhere (so it can't bitrot), the automake conditional only controls the `TESTS` list, `master` keeps full coverage, and the labels give PRs an opt-in. The `ci-build.sh` forwarding and the textual `${{ env.… }}` interpolation into the vmactions guests are both correct — workflow env does not propagate into the QEMU guests, so interpolating at parse time is the only approach that works there.

**One real interaction bug** (inline, with fix): the new `labeled` trigger joins the same concurrency group as code-push runs with `cancel-in-progress: true`, so adding *any* unrelated label to a PR while its CI is running **cancels the in-flight run** and replaces it with a skipped one. Given how routinely PRs get labeled in this repo (`windows`, `core`, `BSD`…), this will regularly kill half-finished CI. Two smaller suggestions ride along.

## Verified
- `tests/Makefile.am`: the explicit `TESTS` list covers every program the old `TESTS=$(noinst_PROGRAMS)` ran (stress, stress_mt, set_option, init_context, macos, umockdev) — nothing silently dropped; `testcore` remains in `noinst_PROGRAMS` unconditionally so it still compiles on every platform and config.
- `configure.ac`: `--enable-testcore` defaults to no, `AM_CONDITIONAL` shape matches the neighbors.
- `ci-build.sh`: array-based arg handling quotes correctly and stays inert when the variable is empty/unset (`${TESTCORE_CONFIGURE_FLAG:-}`), and both MSYS2 invocations inherit it.
- Empty-flag interpolation in the vmactions `run:` scripts leaves a bare line-continuation, which the shell joins harmlessly.
- The job-level `if` correctly passes for `push` events (`action` is empty) and for the default PR event types.
- Skipped label-triggered runs still satisfy required-check semantics (skipped = neutral), so the `if` guard does not wedge merges.

## Notes (no action needed)
- The explicit `TESTS` list means future test programs must be added in two places; the `TESTS +=` lines sitting directly next to their `noinst_PROGRAMS +=` counterparts makes that reasonably hard to miss.
- This touches `haiku.yml` in the same `run:` block that #1910 rewrites — whichever lands second needs a trivial rebase.
- The 9-block duplication is the idiomatic GitHub Actions tax (no YAML anchors); a reusable workflow would be more churn than it saves.

.github/workflows/linux.yml
```diff
@@ -7,13 +7,26 @@ on:
     tags:
       - '**'
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
+
+# Keep the long-running stress test on master unless a PR explicitly opts in.
+env:
+  TESTCORE_CONFIGURE_FLAG: >-
+    ${{ (github.ref == 'refs/heads/master' ||
```
Youw, line 15: Question + optional simplification: `github.ref == 'refs/heads/master'` excludes **tag pushes**, although these workflows also trigger on `tags: ['**']` — so release-tag builds, arguably the runs that deserve the stress test most, skip `testcore`. If that wasn't a deliberate choice, `github.event_name == 'push'` covers both (the `on.push` filter already restricts pushes to master and tags) and reads simpler. If it *was* deliberate, ignore this. Same expression in the other eight workflows.

```suggestion
    ${{ (github.event_name == 'push' ||
```

*— suggestion by Claude*

.github/workflows/linux.yml
Youw: Fixed in `2d146d1c`. Codex updated all nine workflows so an unrelated `labeled` event no longer cancels the in-flight run; ordinary PR updates and additions of either force label still cancel and replace older runs. All workflows pass `actionlint`.

.github/workflows/linux.yml
Youw: Addressed in `2d146d1c`. Codex restored `github.event.label.name` for the event-retrigger guard. The separate `TESTCORE_CONFIGURE_FLAG` expression still checks the complete current label array, so a multi-label PR runs `testcore` whenever either force label is present, while adding an unrelated label does not duplicate the matrix.

.github/workflows/linux.yml
Youw: Codex kept the tag behavior unchanged intentionally: the requested default scope is specifically the `master` branch, so `github.ref == 'refs/heads/master'` excludes tag pushes. Expanding the policy to release tags can be considered separately if desired.

Youw:
I'm done withe agentic coding/review for this one. Once approved by maintainers - ready for merge.

mcuee (approved):

sonatique (approved):

PR #1915, opened 2026-08-03
core: bound the fill loop in libusb_get_port_numbers

sonatique:
libusb_get_port_numbers walks the parent chain twice without holding a lock: once to count the depth, and again to fill the caller's array from the end backwards. A backend that updates a published device can change the chain between the two passes: the darwin backend does exactly that when it reuses a libusb_device for a re-enumerated device.

If the chain is longer during the second pass than it was during the first, the fill index runs past the start of the array and writes before port_numbers[0], corrupting the caller's memory.

Solution: Stop the loop when the index reaches zero, so a chain that grew is reported as the shorter depth that was actually validated against the caller's buffer size rather than overflowing it.

A chain that shrank between the passes already fails safe: the leading entries are left untouched and the returned depth simply exceeds what was written.

### DIFF
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1013,9 +1013,13 @@ int API_EXPORTED libusb_get_port_numbers(libusb_device *dev,
 		return LIBUSB_ERROR_OVERFLOW;
 	}
 
-	/* fill port numbers directly into the correct positions */
+	/* Fill port numbers directly into the correct positions. The parent
+	 * chain is walked without holding a lock, so a backend updating a
+	 * published device can make it longer between the two passes; bound the
+	 * index so a chain that grew cannot write before the start of the
+	 * caller's array. */
 	i = depth;
-	while (dev && dev->port_number != 0) {
+	while (dev && dev->port_number != 0 && i > 0) {
 		port_numbers[--i] = dev->port_number;
 		dev = dev->parent_dev;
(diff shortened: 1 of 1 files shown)


Youw (approved):

seanm:
@sonatique @Youw guys, I used to lament that libusb was too slow in merging PRs (it could take weeks/months), but now you're creating and merging PRs before I even know they exist, I think we are going too fast now.

I received a "@sonatique requested your review" email yesterday (a holiday here) and 25 hours later this was merged.

Do we have a policy, written or informal, about when a PR should be merged?

Youw:
The only thing we have is an "unwritten policy", and if I remember correctly it is something like"make sure other maintainers check/approve before merge, give it more time for bigger changes (at least a few days/weeks), etc.".

For a change like this - small, isolated and obviously not incorrect - I don't see reasons to wait for everyone to review it.

sonatique:
@seanm : it's mainly my fault: I had a free time window in which I hoped I could complete #1798 so I kind of rushed doing PRs. Some of them where depending on others and its obviously simpler to have the first merged before doing the second rather than maintaining a stack of PR. 

I tried to wait proportionally to the PR complexity and risk, and this one in particular had low complexity and low risk, so I merged it quickly.

PR #1916, opened 2026-08-03
windows: propagate the auto-claim status on control transfers

sonatique:
On Windows, a control transfer that isn't already bound to a claimed interface goes through auto_claim(), which walks the device's interfaces and tries libusb_claim_interface() on each one of the matching API type. Whatever the reason a claim failed, the result reported to the application was LIBUSB_ERROR_NOT_FOUND.

That matters because libusb_claim_interface() returns LIBUSB_ERROR_NO_DEVICE once the libusb_device is no longer attached, which is exactly what happens when a device is unplugged or re-enumerates (firmware renumeration, driver reload): the enumeration pass detaches the old libusb_device while the application still holds a handle on it. From that point on every control
transfer on that handle failed with NOT_FOUND, indistinguishable from a device that simply exposes no interface of the requested API type. 

The only hint was the "could not auto-claim any interface" log line, so applications were left guessing whether to reopen the device or give up.

auto_claim() now keeps the status of the last claim attempt and returns it, and both winusbx_submit_control_transfer() and hid_submit_control_transfer() forward it instead of overwriting it with NOT_FOUND. 

A disconnected or re-enumerated device surfaces LIBUSB_ERROR_NO_DEVICE, the documented signal to close and reopen the handle, while NOT_FOUND keeps its narrower meaning. The scan also stops at the first NO_DEVICE rather than making 31 further attempts that cannot succeed, and the log line names the status being returned.

### DIFF
--- a/libusb/os/windows_winusb.c
+++ b/libusb/os/windows_winusb.c
@@ -644,20 +644,36 @@ static int auto_claim(struct libusb_transfer *transfer, int *interface_number, i
 
 	usbi_mutex_lock(&autoclaim_lock);
 	if (current_interface < 0) { // No serviceable interface was found
+		// Status of the last claim attempt, returned to the caller when the scan
+		// comes up empty: LIBUSB_ERROR_NO_DEVICE for a device that has been
+		// unplugged or re-enumerated, LIBUSB_ERROR_NOT_FOUND when the device
+		// exposes no interface of this API type.
+		int claim_status = LIBUSB_ERROR_NOT_FOUND;
+
 		for (current_interface = 0; current_interface < USB_MAXINTERFACES; current_interface++) {
 			// Must claim an interface of the same API type
-			if ((priv->usb_interface[current_interface].apib->id == api_type)
-					&& (libusb_claim_interface(transfer->dev_handle, current_interface) == LIBUSB_SUCCESS)) {
+			if (priv->usb_interface[current_interface].apib->id != api_type)
+				continue;
+
+			claim_status = libusb_claim_interface(transfer->dev_handle, current_interface);
+			if (claim_status == LIBUSB_SUCCESS) {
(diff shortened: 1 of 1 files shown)


Youw (approved):
This looks absolutely fine.
I don't suppose we have a way to test this, like a unit test or so?

sonatique:
> This looks absolutely fine. I don't suppose we have a way to test this, like a unit test or so?

Not I am aware of, no, unfortunately. 

However I did test it with a simple program and hardware: open a device by VID/PID, claim no interface, and issue a string descriptor request twice a second until one fails. 

Claiming nothing is the point: it leaves every `interface_handle[].api_handle` invalid, so `get_valid_interface()` returns -1
and each control transfer goes through `auto_claim()`. 

Then I pulled the cable mid-loop.

On master:
opened device, no interface claimed, requesting string descriptor 3
[   0] 20    ok
...
[   9] 20     ok
[  10] 20     ok
[  11] -5    LIBUSB_ERROR_NOT_FOUND

first hard failure: LIBUSB_ERROR_NOT_FOUND

this PR:
opened device, no interface claimed, requesting string descriptor 3
[   0] 20    ok
...
[   8] 20    ok
[   9] -4    LIBUSB_ERROR_NO_DEVICE

first hard failure: LIBUSB_ERROR_NO_DEVICE

So the disconnect now surfaces as LIBUSB_ERROR_NO_DEVICE, which is what `libusb_submit_transfer()` already documents for a disconnected device, instead of a NOT_FOUND the caller can't act on.

sonatique:
@mcuee : would you like more time to review this one as well? Thanks!

mcuee (approved):

mcuee:
> @mcuee : would you like more time to review this one as well? Thanks!

The change looks good to me. I have approved the PR.

sonatique:
@seanm : would you like more time to review this one? Thanks!

seanm:
I don't know much about WIndows, but at a glance, this lgtm

PR #1780, opened 2026-03-10
darwin: avoid hotplug shutdown deadlock

sonatique:

Issue #1711, opened 2025-10-17
crash in darwin_open


wws13:
It crashed on some devices on macOS, when application in sandbox 

Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Codes:   0x0000000000000001, 0x0000000000000000

0   libusb-1.0.dylib              	       0x100fe72d0 darwin_open + 104
1   libusb-1.0.dylib              	       0x100fe54d0 libusb_open + 140

The proposed patch fixes the crash.

[sandbox.patch](https://github.com/user-attachments/files/22966811/sandbox.patch)

seanm:
Would you be interested/willing to create a PR (pull request)?

wws13:
ok, https://github.com/libusb/libusb/pull/1713

mcuee:
Comments from the PR (merged).

https://github.com/libusb/libusb/pull/1713#issuecomment-3867699934
>> All, do you think we might have a problem elsewhere in the code, since the user ends up with a device handle for which there is no priv device? Should it have been filtered out earlier before coming that far? It is of course fine to have these null checks, but it could be indicative of something else being broken.
>
> Yes, this is a fix for the symptom, not the root cause. Unfortunately, I can't reproduce this crash. I fixed it based on user-reported crashes, and with this fix, there were fewer crashes. Since the crash only occurs on the macOS app in the sandbox, it's possible that the system is preventing some devices from opening, but this hasn't been verified.

mcuee:
#1713 has been merged. But there are other issues as mention in #1798.

mcuee:
Seems to have no issues with libusb git.

```
mcuee in ~/build/libusb/libusb_test/issue1711 % cat usb.sb 
(version 1)
(allow default)

;; Allow communication with the USB HID and hardware stack
(allow iokit-open
    (iokit-user-client-class "IOUSBDeviceUserClientV2")
    (iokit-user-client-class "IOUSBInterfaceUserClientV2")
    (iokit-user-client-class "AppleUSBHostDeviceUserClient")
    (iokit-user-client-class "AppleUSBHostInterfaceUserClient")
)

;; Allow libusb to see the device tree
(allow iokit-get-properties)

mcuee in ~/build/libusb/libusb_test/issue1711 % sandbox-exec -f usb.sb ./libusb_git/examples/listdevs
[...]

PR #1919, opened 2026-08-04
darwin: hold retained run loop references instead of rereading the global

sonatique:
# darwin: hold retained run loop references instead of rereading the global

Completes item 4 of #1798; #1780 fixed the exit-path half.

This one was a bit tricky, and I feared I would miss some substitution, so I asked Claude to help me.
I did a careful review and adjustments before and after Claude work. The results looks OK to me.

`darwin_open` and `darwin_claim_interface` store a retained reference to the event
thread run loop (read under `libusb_darwin_at_mutex`) and register their async event
sources on it; close and release remove the sources from that reference instead of
rereading the global. Besides the unlocked reads, this also makes the retain and
release symmetric: on master both target *the global*, so a handle that outlives the
event thread releases a different run loop than it retained. Every access to
`libusb_darwin_acfl`/`libusb_darwin_acfls` is now under the mutex, so both are marked
`GUARDED_BY`.

Also fixed on the same paths: the device source is removed from the mode it was added
to (`kCFRunLoopCommonModes`, not `DefaultMode`); `darwin_restore_state` releases an
interface whose re-claim failed; `darwin_set_interface_altsetting` stops after a failed
re-claim instead of dereferencing the interface it just released; and re-claiming an
interface tears down the previous event source instead of orphaning it.

**Scope:** this is hardening, not a live-crash fix — the original race needs an
open/close to run while the last context is torn down, and #1711's reproducer has been
quiet since #1780. Verification is review plus no-regression.

**Known pre-existing, deliberately out of scope:** re-claiming an interface still leaks
the previous *plug-in* (`QueryInterface` overwrites it while open) — fixing that means
close-then-reopen, which changes behaviour on the macOS workaround path and needs
device testing; and the device event source is created by whichever handle sees
`open_count == 0` but stored per-handle, so with two handles the source, its run loop
reference and the `USBDeviceClose` can all be skipped.

Reference: #1798 (item 4)

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -1968,7 +1968,7 @@ static void darwin_close_locked (struct libusb_device_handle *dev_handle) {
   if (0 == dpriv->open_count) {
     /* delete the device's async event source */
     if (priv->cfSource) {
-      CFRunLoopRemoveSource (libusb_darwin_acfl, priv->cfSource, kCFRunLoopDefaultMode);
+      CFRunLoopRemoveSource (libusb_darwin_acfl, priv->cfSource, kCFRunLoopCommonModes);
       CFRelease (priv->cfSource);
       priv->cfSource = NULL;
       CFRelease (libusb_darwin_acfl);
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -60,10 +60,15 @@ static const mach_port_t darwin_default_master_port = 0;
 static usbi_mutex_static_t libusb_darwin_at_mutex = USBI_MUTEX_INITIALIZER;
 static usbi_cond_t libusb_darwin_at_cond = USBI_COND_INITIALIZER;
 
+/* not defined in SDKs before macOS 10.7 */
+#ifndef CF_RETURNS_RETAINED
+#define CF_RETURNS_RETAINED
+#endif
+
 #define LIBUSB_DARWIN_STARTUP_FAILURE ((CFRunLoopRef) -1)
 
-static CFRunLoopRef libusb_darwin_acfl = NULL; /* event cf loop */
-static CFRunLoopSourceRef libusb_darwin_acfls = NULL; /* shutdown signal for event cf loop */
+static CFRunLoopRef libusb_darwin_acfl GUARDED_BY(libusb_darwin_at_mutex); /* event cf loop */
+static CFRunLoopSourceRef libusb_darwin_acfls GUARDED_BY(libusb_darwin_at_mutex); /* shutdown signal for event cf loop */
 
 /* Lock ordering: the per-device lock (struct darwin_cached_device, dpriv->lock)
    may be held when acquiring darwin_cached_devices_mutex (e.g. via
@@ -77,7 +82,9 @@ static CFRunLoopSourceRef libusb_darwin_acfls = NULL; /* shutdown signal for eve
    before dpriv->lock. Consequently, code holding dpriv->lock must not call
    public libusb_* functions that take locks (e.g. libusb_release_interface).
    Lock-free ones such as libusb_get_active_config_descriptor are tolerated
-   but discouraged. */
+   but discouraged. libusb_darwin_at_mutex is the innermost lock. darwin_exit
+   holds it across pthread_join, so the event thread must not take it after it
+   signals shutdown. */
 static usbi_mutex_t darwin_cached_devices_mutex = USBI_MUTEX_INITIALIZER;
 static struct list_head darwin_cached_devices GUARDED_BY(darwin_cached_devices_mutex);
 static int init_count GUARDED_BY(darwin_cached_devices_mutex);
@@ -107,6 +114,9 @@ static int darwin_clear_halt_locked(struct libusb_device_handle *dev_handle, uns
 static int darwin_reset_device(struct libusb_device_handle *dev_handle) REQUIRES(dev_handle->lock);
 static int darwin_detach_kernel_driver (struct libusb_device_handle *dev_handle, uint8_t interface) REQUIRES(dev_handle->lock);
 static int darwin_detach_kernel_driver_locked (struct libusb_device_handle *dev_handle, uint8_t interface) REQUIRES(dev_handle->lock);
+#if MAX_INTERFACE_VERSION >= 700
+static int darwin_capture_release_interface(struct libusb_device_handle *dev_handle, uint8_t iface) REQUIRES(dev_handle->lock);
+#endif
 static void darwin_async_io_callback (void *refcon, IOReturn result, void *arg0);
 
 static enum libusb_error darwin_scan_devices(struct libusb_context *ctx);
@@ -1880,6 +1890,50 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
   return LIBUSB_SUCCESS;
 }
 
+static void darwin_device_release_event_source (struct darwin_device_handle_priv *priv) {
+  if (priv->cfSource) {
+    CFRunLoopRemoveSource (priv->runloop, priv->cfSource, kCFRunLoopCommonModes);
+    CFRelease (priv->cfSource);
+    priv->cfSource = NULL;
+  }
+  if (priv->runloop) {
+    CFRelease (priv->runloop);
+    priv->runloop = NULL;
+  }
+}
+
+static void darwin_interface_release_event_source (struct darwin_interface *cInterface) {
+  if (cInterface->cfSource) {
+    CFRunLoopRemoveSource (cInterface->runloop, cInterface->cfSource, kCFRunLoopDefaultMode);
+    CFRelease (cInterface->cfSource);
+    cInterface->cfSource = NULL;
+  }
+  if (cInterface->runloop) {
+    CFRelease (cInterface->runloop);
+    cInterface->runloop = NULL;
+  }
+}
+
+/* Returns NULL if the event thread published no run loop. Callers keep the
+   reference for as long as their source is registered, because libusb_exit
+   can clear the global at any time. The run loop stays valid, but it does not
+   stay running: a source added during libusb_exit may never run. */
+static CFRunLoopRef darwin_retain_event_runloop (void) CF_RETURNS_RETAINED EXCLUDES(libusb_darwin_at_mutex) {
+  CFRunLoopRef runloop;
+
+  usbi_mutex_lock (&libusb_darwin_at_mutex);
+  runloop = libusb_darwin_acfl;
+  if (LIBUSB_DARWIN_STARTUP_FAILURE == runloop) {
+    runloop = NULL;
+  }
+  if (runloop) {
+    CFRetain (runloop);
+  }
+  usbi_mutex_unlock (&libusb_darwin_at_mutex);
+
+  return runloop;
+}
+
 /* must be called while holding dpriv->lock */
 static int darwin_open_locked (struct libusb_device_handle *dev_handle) {
   struct darwin_device_handle_priv *priv = (struct darwin_device_handle_priv *)usbi_get_device_handle_priv(dev_handle);
@@ -1906,11 +1960,29 @@ static int darwin_open_locked (struct libusb_device_handle *dev_handle) {
       priv->is_open = true;
     }
 
+    priv->runloop = darwin_retain_event_runloop ();
+    if (NULL == priv->runloop) {
+      usbi_err (usbi_handle_ctx (dev_handle), "no event thread run loop, cannot open device");
+
+      if (priv->is_open) {
+        (*dpriv->device)->USBDeviceClose (dpriv->device);
+      }
+
+      priv->is_open = false;
+
+      return LIBUSB_ERROR_OTHER;
+    }
+
     /* create async event source */
     kresult = (*dpriv->device)->CreateDeviceAsyncEventSource (dpriv->device,
                                                                                 &priv->cfSource);
     if (kresult != kIOReturnSuccess) {
       usbi_err (usbi_handle_ctx (dev_handle), "CreateDeviceAsyncEventSource: %s", darwin_error_str(kresult));
+
+      /* the handle can still be closed, so leave no source behind */
+      priv->cfSource = NULL;
+      CFRelease (priv->runloop);
+      priv->runloop = NULL;
 
       if (priv->is_open) {
         (*dpriv->device)->USBDeviceClose (dpriv->device);
@@ -1921,10 +1993,8 @@ static int darwin_open_locked (struct libusb_device_handle *dev_handle) {
       return darwin_to_libusb (kresult);
     }
 
-    CFRetain (libusb_darwin_acfl);
-
     /* add the cfSource to the async run loop */
-    CFRunLoopAddSource(libusb_darwin_acfl, priv->cfSource, kCFRunLoopCommonModes);
+    CFRunLoopAddSource(priv->runloop, priv->cfSource, kCFRunLoopCommonModes);
   }
 
   /* device opened successfully */
@@ -1962,17 +2032,16 @@ static void darwin_close_locked (struct libusb_device_handle *dev_handle) {
   dpriv->open_count--;
   if (NULL == dpriv->device) {
     usbi_warn (usbi_handle_ctx (dev_handle), "darwin_close device missing IOService");
+
+    /* the core frees the handle after close returns, so remove the source and
+       decrement the run loop reference count here */
+    darwin_device_release_event_source (priv);
     return;
   }
 
   if (0 == dpriv->open_count) {
     /* delete the device's async event source */
-    if (priv->cfSource) {
-      CFRunLoopRemoveSource (libusb_darwin_acfl, priv->cfSource, kCFRunLoopCommonModes);
-      CFRelease (priv->cfSource);
-      priv->cfSource = NULL;
-      CFRelease (libusb_darwin_acfl);
-    }
+    darwin_device_release_event_source (priv);
 
     if (priv->is_open) {
       /* close the device */
@@ -2277,12 +2346,27 @@ static int darwin_claim_interface(struct libusb_device_handle *dev_handle, uint8
     return ret;
   }
 
-  cInterface->cfSource = NULL;
+  /* a re-claim still holds the previous source: remove it, or it stays in the
+     run loop */
+  darwin_interface_release_event_source (cInterface);
+
+  cInterface->runloop = darwin_retain_event_runloop ();
+  if (NULL == cInterface->runloop) {
+    usbi_err (ctx, "no event thread run loop, cannot claim interface");
+
+    /* can't continue without an async event source */
+    (void)darwin_release_interface (dev_handle, iface);
+
+    return LIBUSB_ERROR_OTHER;
+  }
 
   /* create async event source */
   kresult = (*IOINTERFACE(cInterface))->CreateInterfaceAsyncEventSource (IOINTERFACE(cInterface), &cInterface->cfSource);
   if (kresult != kIOReturnSuccess) {
     usbi_err (ctx, "could not create async event source");
+
+    /* the release below must not use a source that was never created */
+    cInterface->cfSource = NULL;
 
     /* can't continue without an async event source */
     (void)darwin_release_interface (dev_handle, iface);
@@ -2291,7 +2375,7 @@ static int darwin_claim_interface(struct libusb_device_handle *dev_handle, uint8
   }
 
   /* add the cfSource to the async thread's run loop */
-  CFRunLoopAddSource(libusb_darwin_acfl, cInterface->cfSource, kCFRunLoopDefaultMode);
+  CFRunLoopAddSource(cInterface->runloop, cInterface->cfSource, kCFRunLoopDefaultMode);
 
   usbi_dbg (ctx, "interface opened");
 
@@ -2305,6 +2389,10 @@ static int darwin_release_interface(struct libusb_device_handle *dev_handle, uin
   /* current interface */
   struct darwin_interface *cInterface = &priv->interfaces[iface];
 
+  /* do this before the check below: a failed re-claim can leave a source with
+     no interface to close */
+  darwin_interface_release_event_source (cInterface);
+
   /* Check to see if an interface is open */
   if (!IOINTERFACE(cInterface)) {
     return LIBUSB_SUCCESS;
@@ -2312,13 +2400,6 @@ static int darwin_release_interface(struct libusb_device_handle *dev_handle, uin
 
   /* clean up endpoint data */
   cInterface->num_endpoints = 0;
-
-  /* delete the interface's async event source */
-  if (cInterface->cfSource) {
-    CFRunLoopRemoveSource (libusb_darwin_acfl, cInterface->cfSource, kCFRunLoopDefaultMode);
-    CFRelease (cInterface->cfSource);
-    cInterface->cfSource = NULL;
-  }
 
   kresult = (*IOINTERFACE(cInterface))->USBInterfaceClose(IOINTERFACE(cInterface));
   if (kresult != kIOReturnSuccess)
@@ -2413,8 +2494,19 @@ static int darwin_set_interface_altsetting_locked(struct libusb_device_handle *d
     /* For some reason we need to reclaim the interface after the pipe error with some versions of macOS */
     ret = darwin_claim_interface (dev_handle, iface);
     if (LIBUSB_SUCCESS != ret) {
+      /* clear the bit first: the release below can re-enumerate the device,
+         and the restore re-claims every interface still marked claimed */
+      dev_handle->claimed_interfaces &= ~(1U << iface);
+
+      /* release as the public API does, so an auto-detached kernel driver is
+         attached again */
+#if MAX_INTERFACE_VERSION >= 700
+      darwin_capture_release_interface (dev_handle, iface);
+#else
       darwin_release_interface (dev_handle, iface);
-      usbi_err (usbi_handle_ctx (dev_handle), "could not reclaim interface: %s", darwin_error_str(kresult));
+#endif
+      usbi_err (usbi_handle_ctx (dev_handle), "could not reclaim interface: %d", ret);
+      return ret;
     }
     ret = check_alt_setting_and_clear_halt(dev_handle, altsetting, cInterface);
   }
@@ -2515,6 +2607,13 @@ static enum libusb_error darwin_restore_state (struct libusb_device_handle *dev_
 
       ret = darwin_claim_interface (dev_handle, iface);
       if (LIBUSB_SUCCESS != ret) {
+        /* the bits of the interfaces not claimed again are clear, so close
+           skips them: release them here */
+        for (uint8_t i = iface ; i < USB_MAXINTERFACES ; ++i) {
+          if (claimed_interfaces & (1U << i)) {
+            darwin_release_interface (dev_handle, i);
+          }
+        }
         usbi_dbg (ctx, "darwin/restore_state: could not claim interface %u", iface);
         return LIBUSB_ERROR_NOT_FOUND;
       }
--- a/libusb/os/darwin_usb.h
+++ b/libusb/os/darwin_usb.h
@@ -147,6 +147,7 @@ struct darwin_device_priv {
 struct darwin_interface {
   usb_interface_t      interface;
   uint8_t              num_endpoints;
+  CFRunLoopRef         runloop;  /* retained run loop cfSource is registered on */
   CFRunLoopSourceRef   cfSource;
   uint64_t             frames[256];
   uint8_t              endpoint_addrs[USB_MAXENDPOINTS];
@@ -154,6 +155,7 @@ struct darwin_interface {
 
 struct darwin_device_handle_priv {
   bool                 is_open;
+  CFRunLoopRef         runloop;  /* retained run loop cfSource is registered on */
   CFRunLoopSourceRef   cfSource;
 
   struct darwin_interface interfaces[USB_MAXINTERFACES];


libusb/os/darwin_usb.c
```diff
@@ -2398,8 +2483,10 @@ static int darwin_set_interface_altsetting_locked(struct libusb_device_handle *d
     /* For some reason we need to reclaim the interface after the pipe error with some versions of macOS */
     ret = darwin_claim_interface (dev_handle, iface);
     if (LIBUSB_SUCCESS != ret) {
+      /* the interface is released and cannot be used any further */
       darwin_release_interface (dev_handle, iface);
-      usbi_err (HANDLE_CTX (dev_handle), "could not reclaim interface: %s", darwin_error_str(kresult));
+      usbi_err (HANDLE_CTX (dev_handle), "could not reclaim interface: %d", ret);
+      return ret;
```
Youw, line 2509: [P2] Keep `claimed_interfaces` in sync after the failed re-claim

`darwin_release_interface()` nulls `IOINTERFACE(cInterface)` and zeroes `num_endpoints`, but this path returns while `dev_handle->claimed_interfaces` still contains `iface`. The core will therefore continue to treat the interface as claimed: a later `libusb_claim_interface()` returns success without invoking the backend, while endpoint I/O cannot resolve a pipe. Since this function already holds `dev_handle->lock`, please clear the bit when tearing the interface down (or otherwise restore a consistent claimed interface) before returning.

libusb/os/darwin_usb.c
```diff
@@ -2500,6 +2587,9 @@ static enum libusb_error darwin_restore_state (struct libusb_device_handle *dev_
 
       ret = darwin_claim_interface (dev_handle, iface);
       if (LIBUSB_SUCCESS != ret) {
+        /* the bit is never set in claimed_interfaces, so nothing else
+           would release what the failed claim left behind */
+        darwin_release_interface (dev_handle, iface);
```
Youw: [P2] Unwind the originally claimed interfaces that have not been visited

At entry, `claimed_interfaces` saves the old mask and `dev_handle->claimed_interfaces` is cleared. If re-claiming interface N fails, this releases only N; any originally claimed higher-numbered interfaces still retain their old `IOINTERFACE`, `cfSource`, and now their retained `runloop`, while their claim bits remain clear. `darwin_close()` snapshots only the rebuilt bits, so it will skip those entries before the handle storage is freed, leaking the interface/source/run-loop resources. Please release every remaining entry from the saved mask (or clean all old entries before rebuilding) before this early return.

Youw (commented):
Codex reviewed the Darwin run-loop lifetime changes and the affected error paths. CI is green, including macOS, but two cleanup/state-consistency issues remain in the newly touched failure handling; details are inline.

sonatique:
@youw thanks for the review. Your comments have been addressed, please re-review.

libusb/os/darwin_usb.c
sonatique: Good catch, fixed. The core never touches claimed_interfaces around
libusb_set_interface_alt_setting, it checks the bit, unlocks, then calls the
backen,  so nothing else would have cleared it, leaving the core treating the
interface as claimed while num_endpoints was zeroed and IOINTERFACE NULL. 
A later libusb_claim_interface would then tell success without
reaching the backend..

libusb/os/darwin_usb.c
sonatique: Agreed, : my change released only the interface that failed, which made the path look handled when it wasn't.

restore_state zeroes dev_handle->claimed_interfaces at entry and re-sets bits
only on success, so on an early return the not-yet-revisited entries still hold
their pre-reset IOINTERFACE, cfSource and retained runloop with their bits
clear, and darwin_close iterates only the rebuilt bits.

libusb/os/darwin_usb.c
```diff
@@ -2398,8 +2483,13 @@ static int darwin_set_interface_altsetting_locked(struct libusb_device_handle *d
     /* For some reason we need to reclaim the interface after the pipe error with some versions of macOS */
     ret = darwin_claim_interface (dev_handle, iface);
     if (LIBUSB_SUCCESS != ret) {
+      /* the interface is released and cannot be used any further: clear its
+         claimed bit too, or the core keeps treating it as claimed while no
+         pipe can be resolved for it */
       darwin_release_interface (dev_handle, iface);
-      usbi_err (HANDLE_CTX (dev_handle), "could not reclaim interface: %s", darwin_error_str(kresult));
+      dev_handle->claimed_interfaces &= ~(1U << iface);
```
Youw, line 2499: [P2] Preserve the auto-detach release when clearing the claim bit

For `MAX_INTERFACE_VERSION >= 700`, a public claim can run through `darwin_capture_claim_interface()` and increment `dpriv->capture_count` after detaching the kernel driver. This failure path calls only the lower-level `darwin_release_interface()` and then clears the claim bit. The caller can no longer reach `darwin_capture_release_interface()` (`libusb_release_interface()` returns NOT_FOUND and `darwin_close()` skips the cleared bit), so `capture_count` is never decremented and `darwin_attach_kernel_driver_locked()` is never invoked. Please perform the capture/auto-attach side of the release before making this interface unclaimed, without re-entering the public API while `dev_handle->lock` is held.

Youw (commented):
Codex re-reviewed head `d225a3e6`. The two earlier findings are addressed: the failed alt-setting re-claim now updates the claim mask, and restore failure unwinds not-yet-revisited interfaces. One auto-detach cleanup issue remains in the new bookkeeping fix; details are inline.

The current NetBSD failure is unrelated infrastructure (`ftp.netbsd.org` refused package downloads before compilation); macOS and the other GitHub Actions workflows pass.

libusb/os/darwin_usb.c
sonatique: Correct, thank you. This comes from the claim-bit clear I added for your previous finding.

The failure path now releases through the same function the public API would:

    #if MAX_INTERFACE_VERSION >= 700
          darwin_capture_release_interface (dev_handle, iface);
    #else
          darwin_release_interface (dev_handle, iface);
    #endif
          dev_handle->claimed_interfaces &= ~(1U << iface);

with a guarded forward declaration next to the existing ones, since the definition is further down in the same #if block. Locking is unchanged.

Note: a failed re-claim on an auto-detach handle can now trigger a re-enumeration as part of the unwind. That is heavy, but it is exactly what a normal libusb_release_interface would do, and the alternative is a permanently detached kernel driver.

sonatique:
@Youw thanks again for review. Your comment has been addressed, please re-review.

libusb/os/darwin_usb.c
```diff
@@ -2398,8 +2486,19 @@ static int darwin_set_interface_altsetting_locked(struct libusb_device_handle *d
...
+         cleared below, nothing else can do it. the bit must be cleared too,
+         or the core keeps treating the interface as claimed while no pipe
+         can be resolved for it */
+#if MAX_INTERFACE_VERSION >= 700
+      darwin_capture_release_interface (dev_handle, iface);
+#else
       darwin_release_interface (dev_handle, iface);
-      usbi_err (HANDLE_CTX (dev_handle), "could not reclaim interface: %s", darwin_error_str(kresult));
+#endif
+      dev_handle->claimed_interfaces &= ~(1U << iface);
```
Youw: [P2] Clear the claim bit before auto-attach can restore this interface

When `capture_count` falls to zero, `darwin_capture_release_interface()` calls `darwin_attach_kernel_driver_locked()`, which re-enumerates the device. `darwin_reenumerate_device()` snapshots `dev_handle->claimed_interfaces` before that happens; because `iface` is still set until this line, `darwin_restore_state()` can successfully re-claim it and create a fresh interface event source/run-loop retain. This line then clears the bit, so `darwin_close()` skips that newly restored interface and its resources are orphaned. Please clear/exclude `iface` before invoking the capture-aware release (or otherwise prevent the attach re-enumeration from restoring the interface being released).

Youw (commented):
Codex re-reviewed head `ef50dfb0`. Calling the capture-aware release fixes the previous `capture_count`/kernel-driver reattachment omission, and the other earlier findings remain addressed. One ordering issue remains in that new unwind; details are inline.

All current GitHub Actions workflows pass, including macOS and NetBSD. AppVeyor is still pending.

libusb/os/darwin_usb.c
Youw: Addressed in `ade253ec`. Codex verified that the failed interface is removed from `claimed_interfaces` before the capture-aware release, so any kernel-driver reattach re-enumeration cannot restore that interface. The branch is also based directly on current `master` (`1110f3a2`).

Youw (commented):
Codex re-reviewed head `ade253ec` against current `master` (`1110f3a2`). The final ordering issue is addressed: the failed interface is removed from `claimed_interfaces` before the capture-aware release can re-enumerate and restore state. The three earlier findings remain addressed, and Codex found no new in-scope correctness issues in the refreshed diff.

The macOS job and the completed platform jobs pass. Fuzzing, Linux, MSYS2 aarch64, Solaris, and AppVeyor are still pending.

Youw (approved):
I'm too unfamiliar with darwin implementation to properly reason if this change is correct, I can only trust at this point that gpt did a thorow review and didn't find any (unfixed) issues with the implmenetation.

sonatique:
@youw: thank you for your comments and quick turn-around.

Found a related pre-existing issue while checking the restore_state unwind. 

The issue is: if `darwin_restore_state` fails to re-claim an interface, the `capture_count` increments from the original auto-detach claims can never be undone: restore_state leaves those claim bits clear, so `darwin_close` skips them and `darwin_capture_release_interface` never runs. 

`capture_count` lives in `darwin_cached_device`, not the handle, so the effect outlives `libusb_close` :  the kernel driver stays detached and the device is unusable by the OS until it is physically replugged.

It seems to comes from a0b5d27f ("darwin: fix reset device", v1.0.23) then abd788c2 ("darwin: use detach kernel APIs for capture", v1.0.25). 
So it has shipped in every release since v1.0.25, unreported, this means the trigger (a re-claim failing during restore after a reset) is evidently rare...

This PR does not make is worse: that path previously released nothing at all, leaking the interface and its event source too; this PR recovers those. Only the capture bookkeeping is left.

Should we fix this here? Or later?

The thing is: the unwind loop calls `darwin_release_interface` (line 2605), the low-level one, which does no capture handling. 

It cannot simply call `darwin_capture_release_interface` instead, because that function starts by taking `dpriv->lock` (line 3658), and `restore_state` is already running with `dpriv->lock` , so it would deadlock. 

What we could do is inline the body of lines 3659 to 3665 without the surrounding lock and unlock, since `darwin_attach_kernel_driver_locked` (line 3595) does not take `dpriv->lock` itself:
```c
darwin_release_interface (dev_handle, i);
if (dev_handle->auto_detach_kernel_driver && dpriv->capture_count > 0) {
  darwin_attach_kernel_driver_locked (dev_handle, i);   /* dpriv->lock already held */
}
```
That is not obviously correct either. `darwin_attach_kernel_driver_locked` decrements at line 3603 and, when the count reaches zero, calls `darwin_reenumerate_device` at line 3611. 

We are already inside a re-enumeration at that point, so the `in_reenumerate` guard makes the nested call return immediately: the count would be corrected without the reattach actually happening.

It is not obvious that this is the right outcome, and there is no reproducer to test it against.

This issue is five years old, and unreported, and not strictly related to this PR. I would rather do it separately than add another behavior change here (or not doing it at all?), but I can add a separated commit as well, while we're at it.

What do you think?

Youw:
> Should we fix this here? Or later?

I vote for later.

sonatique:
> > Should we fix this here? Or later?
> 
> I vote for later.

What do you think? Shall I open an issue, or directly try with a PR when I have time?

Youw:
Open an issue now and a PR when you have time

sonatique:
@Youw : issue is #1923

seanm:
Please give me time to review this before merging.

mcuee:
Let me test this PR over the weekend.

mcuee:
First test -- `make check` passes the normal build, TSAN debug build and ASAN debug build.

mcuee:
Test code created by Claude Sonnet 5 Medium (free Web Chatbot).

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % cat issue1798_point4_runloop.c 
/*
 * issue1798_point4_runloop.c
 *
 * Reproducer for libusb/libusb#1798, item 4:
 *   "libusb_darwin_acfl accessed without libusb_darwin_at_mutex"
 *
 * darwin_open / darwin_close / darwin_claim_interface / darwin_release_interface
 * touch the global CFRunLoopRef libusb_darwin_acfl (CFRetain/CFRunLoopAddSource
 * on open+claim, CFRunLoopRemoveSource/CFRelease on close+release) WITHOUT
 * holding libusb_darwin_at_mutex. That mutex is only held on the event-thread
 * side when acfl is set (darwin_event_thread_main) and NULLed (event thread
 * shutdown, reached via the last libusb_exit()).
 *
 * This test intentionally races libusb_exit() against a pool of worker
 * threads that are mid-flight in open/claim/release/close on the SAME
 * context, to try to catch:
 *   - a NULL CFRunLoopRef dereference (acfl read as non-NULL, then NULLed,
 *     then used), or
 *   - a use-after-free / CF crash (acfl CFRelease'd by the exit path while
 *     a worker is still doing CFRetain/CFRunLoopAddSource/RemoveSource on it).
 *
 * IMPORTANT: this test deliberately does something the libusb docs do not
 * clearly forbid but also do not clearly allow -- calling libusb_exit()
 * while other threads still have in-flight calls on the same context. That
 * ambiguity is exactly what item 4 (and the item 7 discussion in #1798) is
 * about. Treat any crash/TSan report here as evidence for the audit, not
 * as "misuse" -- if this is misuse, it should be documented as such.
 *
 * Build (TSan strongly recommended; ideally against a libusb built with
 * -fsanitize=thread too, since the race is inside libusb, not this test):
 *
 *       $(pkg-config --cflags libusb-1.0) \
 *       -o issue1798_point4_runloop issue1798_point4_runloop.c \
 *       $(pkg-config --libs libusb-1.0) -lpthread
 *
 * Usage:
 *   ./issue1798_point4_runloop <vid_hex> <pid_hex> [iface=0] [workers=8] [pre_exit_ms=50]
 *
 * Because libusb_exit() is called while workers may still be inside libusb
 * calls, this single run only covers ONE shutdown race window. Run it in a
 * loop (see the companion issue1798_point4_stress.sh) to cover many windows
 * and increase the odds of hitting the narrow race.
 */

#include <stdio.h>
#include <stdlib.h>
#include <stdatomic.h>
#include <string.h>
#include <unistd.h>
#include <pthread.h>

static libusb_context      *g_ctx        = NULL;
static libusb_device       *g_dev        = NULL;
static int                  g_iface      = 0;
static atomic_int           g_stop       = 0;
static atomic_long          g_open_ok    = 0;
static atomic_long          g_open_fail  = 0;
static atomic_long          g_claim_ok   = 0;
static atomic_long          g_claim_fail = 0;

/* Each worker hammers open -> claim -> (tiny hold) -> release -> close,
 * back to back, with no delay, so that at any given instant several
 * threads are likely to be inside darwin_open/darwin_close/
 * darwin_claim_interface/darwin_release_interface -- exactly the four
 * functions item 4 flags as touching libusb_darwin_acfl unlocked. */
static void *worker_main(void *arg)
{
    (void)arg;
        int rc = libusb_open(g_dev, &handle);
        if (rc != LIBUSB_SUCCESS || handle == NULL) {
            atomic_fetch_add(&g_open_fail, 1);
            /* Context may already be torn down; back off briefly instead
             * of busy-spinning once libusb_exit() has clearly won. */
            usleep(200);
            continue;
        }
        atomic_fetch_add(&g_open_ok, 1);

        rc = libusb_claim_interface(handle, g_iface);
        if (rc == LIBUSB_SUCCESS) {
            atomic_fetch_add(&g_claim_ok, 1);
            /* Give the race window a little more room without stalling
             * the whole loop; keep it tiny so we cycle fast. */
            usleep(50);
            libusb_release_interface(handle, g_iface);
        } else {
            atomic_fetch_add(&g_claim_fail, 1);
        }

        libusb_close(handle);
    }
    return NULL;
}

{
    if (argc < 3) {
        fprintf(stderr,
            "usage: %s <vid_hex> <pid_hex> [iface=0] [workers=8] [pre_exit_ms=50]\n",
            argv[0]);
        return 2;
    }

    unsigned vid = (unsigned)strtoul(argv[1], NULL, 16);
    unsigned pid = (unsigned)strtoul(argv[2], NULL, 16);
    g_iface        = (argc > 3) ? atoi(argv[3]) : 0;
    int workers    = (argc > 4) ? atoi(argv[4]) : 8;
    int pre_exit_ms = (argc > 5) ? atoi(argv[5]) : 50;

    int rc = libusb_init(&g_ctx);
    if (rc != LIBUSB_SUCCESS) {
        return 1;
    }

    libusb_device **list = NULL;
    if (n < 0) {
                libusb_error_name((int)n));
        libusb_exit(g_ctx);
        return 1;
    }

            continue;
        if (desc.idVendor == vid && desc.idProduct == pid) {
            /* Take our own reference so the device survives
             * libusb_free_device_list()'s unref below. */
            g_dev = libusb_ref_device(list[i]);
            break;
        }
    }

    if (g_dev == NULL) {
        fprintf(stderr, "device %04x:%04x not found\n", vid, pid);
        libusb_exit(g_ctx);
        return 1;
    }

    fprintf(stderr,
        "point4: %d workers hammering open/claim(iface=%d)/release/close, "
        "libusb_exit() fires after ~%d ms\n",
        workers, g_iface, pre_exit_ms);

    pthread_t *tids = calloc((size_t)workers, sizeof(pthread_t));
    for (int i = 0; i < workers; i++)
        pthread_create(&tids[i], NULL, worker_main, NULL);

    /* Let workers ramp up so several are guaranteed to be mid-call
     * (inside darwin_open / darwin_claim_interface / darwin_close /
     * darwin_release_interface) when we pull the rug out from under
     * them below. */
    usleep((useconds_t)pre_exit_ms * 1000);

    /* This is the race: libusb_exit() runs darwin_exit(), which stops the
     * event thread and (per the audit) NULLs/CFReleases libusb_darwin_acfl
     * under libusb_darwin_at_mutex -- while worker threads above may be
     * inside darwin_open/darwin_close/darwin_claim_interface/
     * darwin_release_interface touching the same CFRunLoopRef WITHOUT that
     * mutex. */
    fprintf(stderr, "point4: calling libusb_exit() now (race point)\n");
    libusb_exit(g_ctx);
    fprintf(stderr, "point4: libusb_exit() returned (no crash inside exit itself)\n");

    /* Tell workers to stop and give them a moment to notice; any calls
     * they make on the now-dead context are exactly what we're trying to
     * catch, so we intentionally do NOT synchronize this stop with the
     * exit above. */
    atomic_store(&g_stop, 1);
    usleep(200 * 1000);

    for (int i = 0; i < workers; i++)
        pthread_join(tids[i], NULL);

    fprintf(stderr,
        "point4: done. open ok=%ld fail=%ld, claim ok=%ld fail=%ld "
        "(fails after the exit point are expected; a crash or TSan report "
        "anywhere above is the actual finding)\n",
        (long)g_open_ok, (long)g_open_fail, (long)g_claim_ok, (long)g_claim_fail);

    libusb_unref_device(g_dev);
    free(tids);
    return 0;
}
```

mcuee:
@sonatique 

Test results are different between libusb git and this PR.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_git_asan_debug/include issue1798_point4_runloop_pr1919.c -o issue1798_point4_runloop_git_asan -L /Users/mcuee/mybin/bin_git_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1919_asan_debug/include issue1798_point4_runloop_pr1919.c -o issue1798_point4_runloop_pr1919_asan -L /Users/mcuee/mybin/bin_pr1919_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_git_asan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
AddressSanitizer:DEADLYSIGNAL
=================================================================
==15970==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x0001904dd8b4 bp 0x00016c1528f0 sp 0x00016c152870 T7)
==15970==The signal is caused by a READ memory access.
==15970==Hint: address points to the zero page.
point4: libusb_exit() returned (no crash inside exit itself)
AddressSanitizer:DEADLYSIGNAL
    #0 0x0001904dd8b4 in __CFCheckCFInfoPACSignature+0x4 (CoreFoundation:arm64e+0x1508b4)
    #1 0x000104160474 in darwin_claim_interface darwin_usb.c:2279
    #2 0x00010415120c in darwin_capture_claim_interface darwin_usb.c:3541
    #3 0x00010412e2fc in libusb_claim_interface core.c:1867
    #4 0x00010407d4e8 in worker_main issue1798_point4_runloop_pr1919.c:112
    #5 0x000104915d64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #6 0x00019034dc54 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c54)
    #7 0x000190348c18 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1c18)

==15970==Register values:
 x[0] = 0x0000000000000000   x[1] = 0x0000610000031140   x[2] = 0x00000001fdbf2c28   x[3] = 0x000054030002180f  
 x[4] = 0x00000b3200000000   x[5] = 0x0000540300000001   x[6] = 0x00000000000010bc   x[7] = 0x0000000000000000  
 x[8] = 0x0000000000000008   x[9] = 0x00000001f5b72158  x[10] = 0x0000610000031140  x[11] = 0x0000610000031148  
x[12] = 0x000022e08c002e88  x[13] = 0x0000000000000000  x[14] = 0x0000000000000000  x[15] = 0x000000016c15286c  
x[16] = 0x00000001903e204c  x[17] = 0x00000001fc5764c8  x[18] = 0x0000000000000000  x[19] = 0x0000000000000000  
x[20] = 0x0000610000031140  x[21] = 0x00000001fdbf2c28  x[22] = 0x00000001fc577000  x[23] = 0x0000000000000000  
x[24] = 0x0000000000000000  x[25] = 0x0000000000000000  x[26] = 0x0000000000000000  x[27] = 0x0000000000000000  
x[28] = 0x0000000000000000     fp = 0x000000016c1528f0     lr = 0x00000001903e20a8     sp = 0x000000016c152870  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (CoreFoundation:arm64e+0x1508b4) in __CFCheckCFInfoPACSignature+0x4
Thread T7 created by T0 here:
    #0 0x000104911198 in pthread_create+0x60 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x39198)
    #1 0x00010407d068 in main issue1798_point4_runloop_pr1919.c:184
    #2 0x00018ff8fdfc in start+0x1b4c (dyld:arm64e+0x1fdfc)

==15970==ABORTING
zsh: abort      ./issue1798_point4_runloop_git_asan 04b4 8613

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1919_asan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
point4: libusb_exit() returned (no crash inside exit itself)
=================================================================
==15975==ERROR: AddressSanitizer: heap-use-after-free on address 0x617000000080 at pc 0x0001049dd9ac bp 0x00016b7261b0 sp 0x00016b7261a8
READ of size 4 at 0x617000000080 thread T3
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
    #0 0x0001049dd9a8 in log_v core.c:2949
    #1 0x0001049d5024 in usbi_log core.c:3047
    #2 0x000104a0c904 in darwin_claim_interface darwin_usb.c:2344
    #3 0x0001049fd20c in darwin_capture_claim_interface darwin_usb.c:3646
    #4 0x0001049da2fc in libusb_claim_interface core.c:1867
    #5 0x0001048794e8 in worker_main issue1798_point4_runloop_pr1919.c:112
    #6 0x000104f61d64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #7 0x00019034dc54 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c54)
    #8 0x000190348c18 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1c18)

0x617000000080 is located 0 bytes inside of 752-byte region [0x617000000080,0x617000000370)
freed by thread T0 here:
    #0 0x000104f65258 in free+0x7c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x41258)
    #1 0x0001049dd7d4 in libusb_exit core.c:2822
    #2 0x0001048790d4 in main issue1798_point4_runloop_pr1919.c:199
    #3 0x00018ff8fdfc in start+0x1b4c (dyld:arm64e+0x1fdfc)

    #0 0x000104f65450 in calloc+0x80 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x41450)
    #1 0x0001049dc7dc in libusb_init_context core.c:2628
    #2 0x0001049dc688 in libusb_init core.c:2584
    #3 0x000104878c4c in main issue1798_point4_runloop_pr1919.c:143
    #4 0x00018ff8fdfc in start+0x1b4c (dyld:arm64e+0x1fdfc)

Thread T3 created by T0 here:
    #0 0x000104f5d198 in pthread_create+0x60 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x39198)
    #1 0x000104879068 in main issue1798_point4_runloop_pr1919.c:184
    #2 0x00018ff8fdfc in start+0x1b4c (dyld:arm64e+0x1fdfc)

SUMMARY: AddressSanitizer: heap-use-after-free core.c:2949 in log_v
Shadow bytes around the buggy address:
  0x616ffffffe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616ffffffe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616fffffff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616fffffff80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x617000000000: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x617000000080:[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000100: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000180: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000200: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000280: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000300: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==15975==ABORTING
zsh: abort      ./issue1798_point4_runloop_pr1919_asan 04b4 8613

```

mcuee:
@sonatique 

Test setup -- two USB devices (04b4:8613 and 16c0:05dc) behind the external USB 3.2 Type C multifunction hub (with USB Ethernet adapter and USB Storage device built-in)

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % lsusb_macos 
Bus 000 Device 001: ID 05ac:8015 Apple USB2.0 Hub  Serial: Not Provided
Bus 000 Device 003: ID 05e3:0610 GenesysLogic USB2.1 Hub  Serial: Not Provided
Bus 000 Device 006: ID 16c0:05dc www.fischl.de USBasp  Serial: 0001
Bus 000 Device 007: ID 04b4:8613  Unnamed Device  Serial: Not Provided
Bus 000 Device 002: ID 05e3:0625 GenesysLogic USB3.2 Hub  Serial: Not Provided
Bus 000 Device 004: ID 0bda:8156 Realtek USB 10/100/1G/2.5G LAN  Serial: 4013000001
Bus 000 Device 005: ID 05e3:0747 Generic USB Storage  Serial: 000000000819
Bus 000 Device 001: ID 05ac:8007 Apple Inc. XHCI Root Hub USB 2.0 Simulation 

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % lsusb_macos -t
/:  Bus 000.Dev 001: XHCI Root Hub USB 2.0 Simulation, 5000M
    |__ Bus 000.Dev 001: USB2.0 Hub, 480Mb/s
        |__ Bus 000.Dev 003: USB2.1 Hub, 480Mb/s
            |__ Bus 000.Dev 006: USBasp, 1.5Mb/s
            |__ Bus 000.Dev 007: Unnamed Device, 480Mb/s
    |__ Bus 000.Dev 002: USB3.2 Hub, 10Gb/s
        |__ Bus 000.Dev 004: USB 10/100/1G/2.5G LAN, 5Gb/s
        |__ Bus 000.Dev 005: USB Storage, 5Gb/s
```

sonatique:
@mcuee thank you for building the reproducer and running it both ways, this is exactly the evidence the PR needed!

I think the two results are not the same bug. Checking the line numbers against the sources:

master, darwin_usb.c:2279:
  CFRunLoopAddSource(libusb_darwin_acfl, cInterface->cfSource, kCFRunLoopDefaultMode);
with x[0] = 0x0 in the register dump. That is the global run loop read without libusb_darwin_at_mutex after libusb_exit unset it, dereferenced inside CoreFoundation. That is item 4 of #1798, reproduced. Thank you.

PR, darwin_usb.c:2344:
  usbi_err (ctx, "no event thread run loop, cannot claim interface");
That is the new guard working as intended: darwin_acquire_event_runloop() read the global under the mutex, saw NULL, and claim returned LIBUSB_ERROR_OTHER without touching CoreFoundation. No CF crash, no NULL run loop use.

What ASAN reports on the PR build is the next statement: log_v reads the libusb_context that libusb_exit already freed (core.c:2822 in the trace, the region allocated by libusb_init). T

hat is the test calling into a context concurrently with its own libusb_exit, which is a use after free regardless of backend or code path, since every libusb entry point dereferences ctx for its locks and its logging. 

Nothing the darwin backend can do about it, and it is also what the test's own header comment flags as the undocumented part.

So I read this as the PR passing for item 4: the libusb bug is gone, and the remaining report belongs to the "libusb_exit while calls are in flight" question.

That one is a documented-contract gap rather than a darwin defect, so I would rather open a separate issue for it than widen this PR.

Two more runs would round out the verification, if you have the time:
1. the same binary under TSAN against the PR, to confirm no race is reported on libusb_darwin_acfl;
2. a variant that sets g_stop and joins the workers before libusb_exit, to confirm the normal lifecycle is clean under ASAN with the PR.
Thanks!

mcuee:
Okay, I should be able to carry out the test over the weekend.

mcuee:
@sonatique 

TSan output.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1919_tsan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
Assertion failed: (pthread_mutex_lock(mutex) == 0), function usbi_mutex_lock, file threads_posix.h, line 49.
zsh: abort      ./issue1798_point4_runloop_pr1919_tsan 04b4 8613
```

mcuee:
I will carry out the second test later.

Updated test code from Claude.
```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % cat issue1798_point4_clean_lifecycle_pr1919.c 
/*
 * issue1798_point4_clean_lifecycle_pr1919.c
 *
 * Companion to issue1798_point4_runloop_pr1919.c, requested by the PR #1919
 * author: that test intentionally races libusb_exit() against in-flight
 * calls on the same context, which surfaces a UAF on ctx itself (every
 * libusb entry point dereferences ctx for its locks/logging) -- a separate,
 * documented-contract question, not a darwin backend defect. This variant
 * removes that race entirely: workers are stopped and joined BEFORE
 * libusb_exit() is ever called, so there is no concurrent use of ctx across
 * the exit boundary. Purpose: confirm the PR's retained-runloop mechanism
 * (per-handle/per-interface CFRetain in darwin_acquire_event_runloop, freed
 * in darwin_close_locked/darwin_release_interface) is clean under ASan in
 * completely ordinary, non-racing usage -- no double-free, no UAF, and
 * (informally, via repeated iterations) no obvious retain/release
 * imbalance in the open/claim/release/close path.
 *
 * This test runs MANY init/open/claim/release/close/exit cycles in a single
 * process (fresh context each iteration) specifically because it's now
 * data-race-free and safe to loop -- that also gives ASan/leak-checking
 * more chances to catch a slow retain leak than a single iteration would.
 *
 * Build libusb pinned to the PR #1919 commit, with ASan (not TSan -- the
 * two sanitizers can't be combined in one binary; TSan verification of the
 * racy test is the companion run, see pr1919_verify_item4.sh):
 *
 *   cd libusb
 *   git fetch origin pull/1919/head:pr1919
 *   git checkout 203529f1fd5f1a3ad021e584d20bc4113835a7ac
 *   ./bootstrap.sh
 *   CC=clang CFLAGS="-g -O1 -fsanitize=address" \
 *   LDFLAGS="-fsanitize=address" \
 *
 * Then build this test against that tree directly (not pkg-config):
 *
 *   clang -Wall -Wextra -g -O1 -fsanitize=address \
 *       -o issue1798_point4_clean_lifecycle_pr1919 issue1798_point4_clean_lifecycle_pr1919.c \
 *       -framework CoreFoundation -framework IOKit -lpthread
 *
 * Note on leak checking: LeakSanitizer (the detect_leaks side of ASan) has
 * only partial/unreliable support on macOS. This test's repeated-iteration
 * design is a low-effort substitute -- if you want a stronger leak signal,
 * run it for a large --iterations count under `leaks --atExit -- ./binary`
 * or watch RSS growth (e.g. `footprint` / Activity Monitor) across the run;
 * steady-state RSS across iterations is the thing to look for, since a
 * per-iteration CFRetain leak would show up as monotonic growth.
 *
 * Usage:
 *   ./issue1798_point4_clean_lifecycle_pr1919 <vid_hex> <pid_hex> \
 *       [iface=0] [workers=8] [hold_ms=50] [iterations=50]
 */

#include <stdio.h>
#include <stdlib.h>
#include <stdatomic.h>
#include <string.h>
#include <unistd.h>
#include <pthread.h>

static int                  g_iface      = 0;
static atomic_int           g_stop       = 0;
static atomic_long          g_open_ok    = 0;
static atomic_long          g_open_fail  = 0;
static atomic_long          g_claim_ok   = 0;
static atomic_long          g_claim_fail = 0;

typedef struct {
    libusb_device *dev; /* per-iteration device pointer, threads read only */
} worker_ctx_t;

/* Same open -> claim -> (tiny hold) -> release -> close hammering as the
 * racy test, so this exercises the exact same darwin_open/darwin_claim_
 * interface/darwin_release_interface/darwin_close paths -- the only
 * difference from the racy variant is WHEN g_stop is observed relative to
 * libusb_exit(): here every worker has already returned from its last
 * libusb call, and been joined, before main ever calls libusb_exit(). */
static void *worker_main(void *arg)
{
    worker_ctx_t *wc = (worker_ctx_t *)arg;
        int rc = libusb_open(wc->dev, &handle);
        if (rc != LIBUSB_SUCCESS || handle == NULL) {
            atomic_fetch_add(&g_open_fail, 1);
            continue;
        }
        atomic_fetch_add(&g_open_ok, 1);

        rc = libusb_claim_interface(handle, g_iface);
        if (rc == LIBUSB_SUCCESS) {
            atomic_fetch_add(&g_claim_ok, 1);
            usleep(50);
            libusb_release_interface(handle, g_iface);
        } else {
            atomic_fetch_add(&g_claim_fail, 1);
        }

        libusb_close(handle);
    }
    return NULL;
}

/* One full, ordinary lifecycle: init -> find device -> spawn workers ->
 * hold_ms -> stop+JOIN workers (no calls in flight anymore) -> exit.
 * Returns 0 on success, nonzero on any unexpected libusb error (a crash
 * or ASan report will terminate the process before this returns at all,
 * which is the actual thing being tested). */
static int run_one_cycle(unsigned vid, unsigned pid, int workers, int hold_ms)
{
    libusb_context *ctx = NULL;
    int rc = libusb_init(&ctx);
    if (rc != LIBUSB_SUCCESS) {
        return 1;
    }

    libusb_device **list = NULL;
    ssize_t n = libusb_get_device_list(ctx, &list);
    if (n < 0) {
                libusb_error_name((int)n));
        libusb_exit(ctx);
        return 1;
    }

    libusb_device *dev = NULL;
            continue;
        if (desc.idVendor == vid && desc.idProduct == pid) {
            dev = libusb_ref_device(list[i]);
            break;
        }
    }

    if (dev == NULL) {
        fprintf(stderr, "device %04x:%04x not found\n", vid, pid);
        libusb_exit(ctx);
        return 1;
    }

    atomic_store(&g_stop, 0);
    worker_ctx_t wc = { .dev = dev };

    pthread_t *tids = calloc((size_t)workers, sizeof(pthread_t));
    for (int i = 0; i < workers; i++)
        pthread_create(&tids[i], NULL, worker_main, &wc);

    usleep((useconds_t)hold_ms * 1000);

    /* Stop and JOIN before touching exit -- by the time libusb_exit() is
     * called below, every worker thread has already returned from its
     * last libusb_* call. No thread is inside libusb when ctx goes away. */
    atomic_store(&g_stop, 1);
    for (int i = 0; i < workers; i++)
        pthread_join(tids[i], NULL);

    libusb_unref_device(dev);

    /* Only the main thread touches ctx from here on, and only sequentially. */
    libusb_exit(ctx);

    free(tids);
    return 0;
}

{
    if (argc < 3) {
        fprintf(stderr,
            "usage: %s <vid_hex> <pid_hex> [iface=0] [workers=8] [hold_ms=50] [iterations=50]\n",
            argv[0]);
        return 2;
    }

    unsigned vid   = (unsigned)strtoul(argv[1], NULL, 16);
    unsigned pid   = (unsigned)strtoul(argv[2], NULL, 16);
    g_iface        = (argc > 3) ? atoi(argv[3]) : 0;
    int workers    = (argc > 4) ? atoi(argv[4]) : 8;
    int hold_ms    = (argc > 5) ? atoi(argv[5]) : 50;
    int iterations = (argc > 6) ? atoi(argv[6]) : 50;

    fprintf(stderr,
        "clean-lifecycle: %d iterations, %d workers, hold_ms=%d, iface=%d, "
        "fresh libusb_init/libusb_exit each iteration, workers always "
        "joined before exit\n",
        iterations, workers, hold_ms, g_iface);

    for (int it = 0; it < iterations; it++) {
        int rc = run_one_cycle(vid, pid, workers, hold_ms);
        if (rc != 0) {
            fprintf(stderr, "clean-lifecycle: iteration %d returned error %d\n",
                    it, rc);
            return rc;
        }
        if ((it % 10) == 0)
            fprintf(stderr, "clean-lifecycle: iteration %d/%d ok "
                "(cumulative open ok=%ld fail=%ld, claim ok=%ld fail=%ld)\n",
                it, iterations, (long)g_open_ok, (long)g_open_fail,
                (long)g_claim_ok, (long)g_claim_fail);
    }

    fprintf(stderr,
        "clean-lifecycle: all %d iterations completed cleanly. "
        "open ok=%ld fail=%ld, claim ok=%ld fail=%ld. "
        "A clean exit here means: no double-free/UAF on the retained "
        "CFRunLoopRef path under ordinary (non-racing) open/claim/release/"
        "close/exit usage. If you ran under `leaks` or watched RSS, flat "
        "steady-state memory across iterations is the retain/release-"
        "balance signal; monotonic growth would suggest a leaked CFRetain "
        "somewhere in darwin_acquire_event_runloop's callers.\n",
        iterations, (long)g_open_ok, (long)g_open_fail,
        (long)g_claim_ok, (long)g_claim_fail);

    return 0;
}

```

mcuee:
@sonatique 

I think your analsys is good.

Test results with the updated test code. No more issues with this PR and libusb git.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1919_asan_debug/include issue1798_point4_clean_lifecycle_pr1919.c -o issue1798_point4_clean_lifecycle_pr1919_asan -L /Users/mcuee/mybin/bin_pr1919_asan_debug/lib -lusb-1.0 -pthread 

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_git_asan_debug/include issue1798_point4_clean_lifecycle_pr1919.c -o issue1798_point4_clean_lifecycle_git_asan -L /Users/mcuee/mybin/bin_git_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_clean_lifecycle_git_asan 04b4 8613
clean-lifecycle: 50 iterations, 8 workers, hold_ms=50, iface=0, fresh libusb_init/libusb_exit each iteration, workers always joined before exit
clean-lifecycle: iteration 0/50 ok (cumulative open ok=656 fail=0, claim ok=162 fail=494)
clean-lifecycle: iteration 10/50 ok (cumulative open ok=2889 fail=0, claim ok=1870 fail=1019)
clean-lifecycle: iteration 20/50 ok (cumulative open ok=4695 fail=0, claim ok=3676 fail=1019)
clean-lifecycle: iteration 30/50 ok (cumulative open ok=6536 fail=0, claim ok=5517 fail=1019)
clean-lifecycle: iteration 40/50 ok (cumulative open ok=8361 fail=0, claim ok=7342 fail=1019)
clean-lifecycle: all 50 iterations completed cleanly. open ok=9958 fail=0, claim ok=8939 fail=1019. A clean exit here means: no double-free/UAF on the retained CFRunLoopRef path under ordinary (non-racing) open/claim/release/close/exit usage. If you ran under `leaks` or watched RSS, flat steady-state memory across iterations is the retain/release-balance signal; monotonic growth would suggest a leaked CFRetain somewhere in darwin_acquire_event_runloop's callers.

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_clean_lifecycle_pr1919_asan 04b4 8613
clean-lifecycle: 50 iterations, 8 workers, hold_ms=50, iface=0, fresh libusb_init/libusb_exit each iteration, workers always joined before exit
clean-lifecycle: iteration 0/50 ok (cumulative open ok=720 fail=0, claim ok=181 fail=539)
clean-lifecycle: iteration 10/50 ok (cumulative open ok=3005 fail=0, claim ok=1986 fail=1019)
clean-lifecycle: iteration 20/50 ok (cumulative open ok=4795 fail=0, claim ok=3776 fail=1019)
clean-lifecycle: iteration 30/50 ok (cumulative open ok=6617 fail=0, claim ok=5598 fail=1019)
clean-lifecycle: iteration 40/50 ok (cumulative open ok=8452 fail=0, claim ok=7433 fail=1019)
clean-lifecycle: all 50 iterations completed cleanly. open ok=10106 fail=0, claim ok=9087 fail=1019. A clean exit here means: no double-free/UAF on the retained CFRunLoopRef path under ordinary (non-racing) open/claim/release/close/exit usage. If you ran under `leaks` or watched RSS, flat steady-state memory across iterations is the retain/release-balance signal; monotonic growth would suggest a leaked CFRetain somewhere in darwin_acquire_event_runloop's callers.
```

mcuee (approved):

mcuee:
@sonatique 

I have approved this PR.

Pending review from @seanm.

libusb/os/darwin_usb.c
```diff
@@ -1865,6 +1870,40 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
   return LIBUSB_SUCCESS;
 }
 
+/* removes the interface's event source and drops both references. Safe when
+   there is no source: cfSource is only ever set together with runloop. */
```
seanm: I think it would be a bit clearer for other Apple programmers to speak of "reference counts" here.  ex:

```
/* removes the interface's event source and decrement the reference count of both objects. Safe when
   there is no source: cfSource is only ever set together with runloop. */
```

libusb/os/darwin_usb.c
```diff
@@ -1947,16 +2003,31 @@ static void darwin_close_locked (struct libusb_device_handle *dev_handle) {
   dpriv->open_count--;
   if (NULL == dpriv->device) {
     usbi_warn (HANDLE_CTX (dev_handle), "darwin_close device missing IOService");
+
+    /* the core frees the handle once darwin_close returns: last chance to
+       unregister the source and drop the run loop retain */
```
seanm:
```
    /* the core frees the handle once darwin_close returns: last chance to
       remove the source and decrement the run loop reference count. */
```

libusb/os/darwin_usb.c
```diff
@@ -1865,6 +1870,40 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
...
+    CFRelease (cInterface->runloop);
+    cInterface->runloop = NULL;
+  }
+}
+
+/* returns a retained reference to the event thread run loop, or NULL if none
+   is published; the caller must CFRelease it. The reference keeps the
+   CFRunLoop valid but not running: removal is always safe, while a source
+   added during libusb_exit may never fire. */
+static CFRunLoopRef darwin_acquire_event_runloop (void) EXCLUDES(libusb_darwin_at_mutex) {
```
seanm: You could tag with function declaration with `CF_RETURNS_RETAINED`, which will help the static analyzer detect reference counting errors.

I would also consider renaming it `darwin_retain_event_runloop` to better match the CFRetain/CFRelease naming pair.

libusb/os/darwin_usb.c
```diff
@@ -1865,6 +1870,40 @@ static enum libusb_error darwin_scan_devices(struct libusb_context *ctx) {
   return LIBUSB_SUCCESS;
 }
 
+/* removes the interface's event source and drops both references. Safe when
+   there is no source: cfSource is only ever set together with runloop. */
+static void darwin_interface_drop_event_source (struct darwin_interface *cInterface) {
```
seanm: I would also consider renaming it `darwin_interface_release_event_source` to better match the CFRetain/CFRelease naming pair, instead of the verb "drop".

seanm (changes requested):
This looks generally good, nice work.  Just a few little suggestions.

Also, could you split the kCFRunLoopCommonModes change into a separate commit, I think that might be useful in future bisecting.

libusb/os/darwin_usb.c
sonatique: darwin_acquire_event_runloop -> darwin_retain_event_runloop, tagged CF_RETURNS_RETAINED

libusb/os/darwin_usb.c
sonatique: darwin_interface_drop_event_source -> darwin_interface_release_event_source

libusb/os/darwin_usb.c
sonatique: changed in your way

libusb/os/darwin_usb.c
sonatique: both comments now speak of reference counts

sonatique:
@seanm Thanks, all your requested applied in the new head.

The kCFRunLoopCommonModes fix is now the first commit, before the run loop work. Good call, it is independent and bisects better alone.

Also rebased on current master: #1917 and the other darwin changes touch the same functions.

sonatique:
@mcuee thanks for the runs and the approval. Two things.

The TSan assert (pthread_mutex_lock, threads_posix.h:49) is the same problem as the ASan report, not a new one: after libusb_exit the worker locks a mutex that libusb_exit already destroyed. TSan reported no race on libusb_darwin_acfl before it, which was the point of that run.

I rebased the PR on current master and applied @seanm's review. The rebase changed darwin_open_locked and darwin_close_locked, so could you re-run both tests on the new head when you have time?

mcuee:
@sonatique 

Yes, I should be able to carry out the test over the weekend.

mcuee:
No regressions with the latest PR #1919.

Same results with the seond test for latest git or this PR.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1919_asan_debug/include issue1798_point4_clean_lifecycle_pr1919.c -o issue1798_point4_clean_lifecycle_pr1919_asan -L /Users/mcuee/mybin/bin_pr1919_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_git_asan_debug/include issue1798_point4_clean_lifecycle_pr1919.c -o issue1798_point4_clean_lifecycle_git_asan -L /Users/mcuee/mybin/bin_git_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_clean_lifecycle_git_asan 04b4 8613
clean-lifecycle: 50 iterations, 8 workers, hold_ms=50, iface=0, fresh libusb_init/libusb_exit each iteration, workers always joined before exit
clean-lifecycle: iteration 0/50 ok (cumulative open ok=664 fail=0, claim ok=167 fail=497)
clean-lifecycle: iteration 10/50 ok (cumulative open ok=2920 fail=0, claim ok=1901 fail=1019)
clean-lifecycle: iteration 20/50 ok (cumulative open ok=4748 fail=0, claim ok=3729 fail=1019)
clean-lifecycle: iteration 30/50 ok (cumulative open ok=6528 fail=0, claim ok=5509 fail=1019)
clean-lifecycle: iteration 40/50 ok (cumulative open ok=8284 fail=0, claim ok=7265 fail=1019)
clean-lifecycle: all 50 iterations completed cleanly. open ok=9817 fail=0, claim ok=8798 fail=1019. A clean exit here means: no double-free/UAF on the retained CFRunLoopRef path under ordinary (non-racing) open/claim/release/close/exit usage. If you ran under `leaks` or watched RSS, flat steady-state memory across iterations is the retain/release-balance signal; monotonic growth would suggest a leaked CFRetain somewhere in darwin_acquire_event_runloop's callers.

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_clean_lifecycle_pr1919_asan 04b4 8613
clean-lifecycle: 50 iterations, 8 workers, hold_ms=50, iface=0, fresh libusb_init/libusb_exit each iteration, workers always joined before exit
clean-lifecycle: iteration 0/50 ok (cumulative open ok=711 fail=0, claim ok=179 fail=532)
clean-lifecycle: iteration 10/50 ok (cumulative open ok=2948 fail=0, claim ok=1929 fail=1019)
clean-lifecycle: iteration 20/50 ok (cumulative open ok=4714 fail=0, claim ok=3695 fail=1019)
clean-lifecycle: iteration 30/50 ok (cumulative open ok=6499 fail=0, claim ok=5480 fail=1019)
clean-lifecycle: iteration 40/50 ok (cumulative open ok=8270 fail=0, claim ok=7251 fail=1019)
clean-lifecycle: all 50 iterations completed cleanly. open ok=9845 fail=0, claim ok=8826 fail=1019. A clean exit here means: no double-free/UAF on the retained CFRunLoopRef path under ordinary (non-racing) open/claim/release/close/exit usage. If you ran under `leaks` or watched RSS, flat steady-state memory across iterations is the retain/release-balance signal; monotonic growth would suggest a leaked CFRetain somewhere in darwin_acquire_event_runloop's callers.
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % 

```

mcuee:
For the first test.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_git_asan_debug/include issue1798_point4_runloop_pr1919.c -o issue1798_point4_runloop_git_asan -L /Users/mcuee/mybin/bin_git_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1919_asan_debug/include issue1798_point4_runloop_pr1919.c -o issue1798_point4_runloop_pr1919_asan -L /Users/mcuee/mybin/bin_pr1919_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_git_asan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
AddressSanitizer:DEADLYSIGNAL
=================================================================
==9541==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x000182087954 bp 0x00016b742970 sp 0x00016b7428f0 T8)
==9541==The signal is caused by a READ memory access.
==9541==Hint: address points to the zero page.
point4: libusb_exit() returned (no crash inside exit itself)
    #0 0x000182087954 in __CFCheckCFInfoPACSignature+0x4 (CoreFoundation:arm64e+0x150954)
    #1 0x000104bfaafc in darwin_claim_interface darwin_usb.c:2294
    #2 0x000104bebfd4 in darwin_capture_claim_interface darwin_usb.c:3556
    #3 0x000104bc9ed0 in libusb_claim_interface core.c:1867
    #4 0x000104b194e8 in worker_main issue1798_point4_runloop_pr1919.c:112
    #5 0x0001054c1d64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #6 0x000181ef7c54 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c54)
    #7 0x000181ef2c18 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1c18)

==9541==Register values:
 x[0] = 0x0000000000000000   x[1] = 0x0000610000039a40   x[2] = 0x00000001effd76b0   x[3] = 0x0000400f00022a03  
 x[4] = 0x00000b3200000000   x[5] = 0x0000400f00000001   x[6] = 0x00000000000010bc   x[7] = 0x0000000000000000  
 x[8] = 0x0000000000000008   x[9] = 0x00000001e7f42558  x[10] = 0x0000610000039a40  x[11] = 0x0000610000039a48  
x[12] = 0x00002d087e002e88  x[13] = 0x0000000000000000  x[14] = 0x0000000000000000  x[15] = 0x000000016b7428ec  
x[16] = 0x0000000181f8c084  x[17] = 0x00000001ee9564c8  x[18] = 0x0000000000000000  x[19] = 0x0000000000000000  
x[20] = 0x0000610000039a40  x[21] = 0x00000001effd76b0  x[22] = 0x00000001ee957000  x[23] = 0x0000000000000000  
x[24] = 0x0000000000000000  x[25] = 0x0000000000000000  x[26] = 0x0000000000000000  x[27] = 0x0000000000000000  
x[28] = 0x0000000000000000     fp = 0x000000016b742970     lr = 0x0000000181f8c0e0     sp = 0x000000016b7428f0  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (CoreFoundation:arm64e+0x150954) in __CFCheckCFInfoPACSignature+0x4
Thread T8 created by T0 here:
    #0 0x0001054bd198 in pthread_create+0x60 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x39198)
    #1 0x000104b19068 in main issue1798_point4_runloop_pr1919.c:184
    #2 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

==9541==ABORTING
zsh: abort      ./issue1798_point4_runloop_git_asan 04b4 8613

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1919_asan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
point4: libusb_exit() returned (no crash inside exit itself)
AddressSanitizer:DEADLYSIGNAL
=================================================================
==9546==ERROR: AddressSanitizer: SEGV on unknown address 0x0000000001d0 (pc 0x0001042b53b0 bp 0x00016bd0ad80 sp 0x00016bd0ad60 T2)
==9546==The signal is caused by a READ memory access.
==9546==Hint: address points to the zero page.
AddressSanitizer:DEADLYSIGNAL
    #0 0x0001042b53b0 in usbi_handling_events libusbi.h:495
    #1 0x0001042b4fb8 in libusb_close core.c:1651
    #2 0x000104209560 in worker_main issue1798_point4_runloop_pr1919.c:123
    #3 0x000104c41d64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #4 0x000181ef7c54 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c54)
    #5 0x000181ef2c18 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1c18)

==9546==Register values:
 x[0] = 0x0000000000000000   x[1] = 0x0000000000000000   x[2] = 0xffffffffffffffe4   x[3] = 0x000000702d7c1524  
 x[4] = 0x000000702d7c1540   x[5] = 0x00000000000000a8   x[6] = 0x000000016bc88000   x[7] = 0x0000000000000001  
 x[8] = 0x0000000000000000   x[9] = 0x00000000000001d0  x[10] = 0x000000016bd0ac80  x[11] = 0x000000702d7c1594  
x[12] = 0x000000016bd0a668  x[13] = 0x000000016bd0a660  x[14] = 0x000000016bd0ac80  x[15] = 0x00007fffffffffff  
x[16] = 0x0000000181f01090  x[17] = 0x0000000104cb0558  x[18] = 0x0000000000000000  x[19] = 0x000000016bd0aea0  
x[20] = 0x000000016bd0b000  x[21] = 0x0000000104209324  x[22] = 0x0000000000000000  x[23] = 0x0000000000000000  
x[24] = 0x0000000000000000  x[25] = 0x0000000000000000  x[26] = 0x0000000000000000  x[27] = 0x0000000000000000  
x[28] = 0x0000000000000000     fp = 0x000000016bd0ad80     lr = 0x00000001042b4fbc     sp = 0x000000016bd0ad60  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV libusbi.h:495 in usbi_handling_events
Thread T2 created by T0 here:
    #0 0x000104c3d198 in pthread_create+0x60 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x39198)
    #1 0x000104209068 in main issue1798_point4_runloop_pr1919.c:184
    #2 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

==9546==ABORTING
zsh: abort      ./issue1798_point4_runloop_pr1919_asan 04b4 8613
```

mcuee:
@seanm 

Please review this PR again. Thanks.

mcuee:
For the first test, I also tested PR #1942.

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1942_asan_debug/include issue1798_point4_runloop_pr1919.c -o issue1798_point4_runloop_pr1942_asan -L /Users/mcuee/mybin/bin_pr1942_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1942_asan
usage: ./issue1798_point4_runloop_pr1942_asan <vid_hex> <pid_hex> [iface=0] [workers=8] [pre_exit_ms=50]
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1942_asan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
point4: libusb_exit() returned (no crash inside exit itself)
=================================================================
==57905==ERROR: AddressSanitizer: heap-use-after-free on address 0x617000000080 at pc 0x00010029d074 bp 0x00016ff56210 sp 0x00016ff56208
READ of size 4 at 0x617000000080 thread T6
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
    #0 0x00010029d070 in log_v core.c:2954
    #1 0x000100294ff4 in usbi_log core.c:3052
    #2 0x0001002cabf0 in darwin_claim_interface darwin_usb.c:2366
    #3 0x0001002bbdcc in darwin_capture_claim_interface darwin_usb.c:3675
    #4 0x000100299e8c in libusb_claim_interface core.c:1867
    #5 0x0001001ed4e8 in worker_main issue1798_point4_runloop_pr1919.c:112
    #6 0x000100bbdd64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #7 0x000181ef7c54 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c54)
    #8 0x000181ef2c18 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1c18)

0x617000000080 is located 0 bytes inside of 752-byte region [0x617000000080,0x617000000370)
freed by thread T0 here:
    #0 0x000100bc1258 in free+0x7c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x41258)
    #1 0x00010029ce9c in libusb_exit core.c:2827
    #2 0x0001001ed0d4 in main issue1798_point4_runloop_pr1919.c:199
    #3 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

    #0 0x000100bc1450 in calloc+0x80 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x41450)
    #1 0x00010029bea4 in libusb_init_context core.c:2633
    #2 0x00010029bd50 in libusb_init core.c:2589
    #3 0x0001001ecc4c in main issue1798_point4_runloop_pr1919.c:143
    #4 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

Thread T6 created by T0 here:
    #0 0x000100bb9198 in pthread_create+0x60 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x39198)
    #1 0x0001001ed068 in main issue1798_point4_runloop_pr1919.c:184
    #2 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

SUMMARY: AddressSanitizer: heap-use-after-free core.c:2954 in log_v
Shadow bytes around the buggy address:
  0x616ffffffe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616ffffffe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616fffffff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616fffffff80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x617000000000: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x617000000080:[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000100: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000180: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000200: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000280: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000300: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==57905==ABORTING
zsh: abort      ./issue1798_point4_runloop_pr1942_asan 04b4 8613
```

sonatique:
@seanm this PR is blocked by your change request that I addressed. Could you please check again? Thanks!

sonatique:
> @seanm this PR is blocked by your change request that I addressed. Could you please check again? Thanks!

@seanm : ping. Thank you!

sonatique:
@seanm : Do you think you'll have time to re-review soon?

seanm:
I guess github status doesn't make it obvious, but I've been on vacation for the last 2.5 weeks, just back now... will try to catch up on libusb this week...

PR #1914, opened 2026-08-02
darwin: fix Apple SDK enum conversion warnings

Youw:
## Summary

Cast the three Apple request-type inputs to `UInt8` before `USBmakebmRequestType` combines them.

This is the remaining `libusb/os/darwin_usb.c` portion extracted from #1904. The language-ID conversion that was also present there has already landed through #1903, so it is not duplicated here.

## Root cause

Xcode 26's SDK defines the request values and masks used by `USBmakebmRequestType` in different anonymous enums. Clang therefore diagnoses the macro's cross-enum bitwise operations at the Darwin string-descriptor call sites added by #1860.

The inputs are byte-sized USB request fields. Casting each input to `UInt8` before the macro combines them makes the intended representation explicit without changing request values or runtime behavior.

## Impact

This removes the Apple SDK anonymous-enum warnings exposed by the macOS C++ build in #1845. The change is confined to `libusb/os/darwin_usb.c` and does not alter the public API or ABI.

## Verification

- The macOS workflow passes on the current head.
- `git diff --check` passes.
- The branch is based on current `origin/master`.
- The diff changes only three `USBmakebmRequestType` call sites in `libusb/os/darwin_usb.c`.

Related to #1845
Related to #1860
Related to #1903
Related to #1904

Assisted-by: Codex:GPT-5.6 Sol

PR #1921, opened 2026-08-05
docs: clarify C and C++ build contexts

Youw:
## Summary

- Document that libusb's primary use and build context is as a C library compiled with a C compiler.
- Recognize that the C sources are sometimes compiled in non-primary contexts, including as C++.
- Require context-specific changes to name that context, explain why the change is needed there, and state whether the primary C library build is unaffected in both commit messages and PR descriptions.

## Why

Without explicit context, a compiler-specific compatibility change can appear to be a general C or libusb build fix. This guidance keeps the scope and intent of future changes clear.

This is a documentation-only change and does not affect library builds, the public API, or the ABI.

## Validation

- `git diff --check origin/master...HEAD`
- Reviewed the one-file Markdown diff for scope and wording.

Related to #1845
Related to #1914

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,22 @@ with their own certificate.
 - Branch names should be explicit and brief about what is being done or asked — e.g. `add-commit-attribution`, `fix-win-hang`, etc.
 - Prefer kebab-case.
 
+## Primary C Build and Other Compilation Contexts
+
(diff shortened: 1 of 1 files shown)


sonatique (approved):

mcuee (approved):

seanm (approved):
No idea how useful this actually is, but sure, why not.

PR #1924, opened 2026-08-07
darwin: emit the ADD_ZERO_PACKET terminator with an async write

pradlol:
## Summary

On macOS the terminating zero-length packet (ZLP) for `LIBUSB_TRANSFER_ADD_ZERO_PACKET` is sent from `darwin_async_io_callback()` with the **synchronous** `WritePipe(..., 0)`. IOKit accepts that call and returns `kIOReturnSuccess`, but it never places a zero-length packet on the bus. A device that relies on the terminator to end a transfer whose length is an exact multiple of `wMaxPacketSize` therefore never sees it and stalls waiting for the rest of a request that never arrives — in practice the next IN transfer times out and the device appears hung.

## Root cause

The synchronous `WritePipe(pipe, buf, 0)` returns `kIOReturnSuccess` but does not put a zero-length packet on the bus. This was confirmed with a USB protocol analyzer (Total Phase Beagle): on the unmodified backend no ZLP is transmitted after a bulk OUT transfer whose length is a multiple of `wMaxPacketSize`, and instrumenting the call showed it returning success on every terminator regardless. Only the asynchronous pipe write (`WritePipeAsync` / `WritePipeAsyncTO`) with a length of 0 actually emits the packet on the wire.

This is a different failure mode from #678, which addresses an incorrectly computed `maxPacketSize` (so the `length % maxPacketSize` guard clears the flag and no ZLP is *attempted*). Here the flag survives and the write *is* attempted — it just never reaches the bus. The two are complementary.

## Change

Queue the terminator as a second asynchronous write instead of the synchronous one:

- `darwin_async_io_callback()` records the data-phase result/size and, when a terminator is due, hands off to `darwin_submit_zero_packet()`, which issues `WritePipeAsyncTO` (or `WritePipeAsync` for interrupt endpoints, which do not support timeouts) with a dedicated completion handler, `darwin_zlp_io_callback()`.
- The write reuses the interface plug-in already pinned at submission, so the event-thread callback still does not take `dev_handle->lock` via `ep_to_pipeRef`.
- `darwin_zlp_io_callback()` drops the pin and signals the transfer complete. Because the terminator carries no data, the data-phase `result`/`size` are left untouched, so `actual_length` is unaffected.
- If the terminator cannot be submitted, the pin is released and the transfer is completed with the data-phase result, matching the previous best-effort behaviour. A small `darwin_release_zlp_interface()` helper centralises dropping the pin.

No public API or ABI change.

### DIFF
--- a/libusb/os/darwin_usb.c
+++ b/libusb/os/darwin_usb.c
@@ -108,6 +108,7 @@ static int darwin_reset_device(struct libusb_device_handle *dev_handle) REQUIRES
 static int darwin_detach_kernel_driver (struct libusb_device_handle *dev_handle, uint8_t interface) REQUIRES(dev_handle->lock);
 static int darwin_detach_kernel_driver_locked (struct libusb_device_handle *dev_handle, uint8_t interface) REQUIRES(dev_handle->lock);
 static void darwin_async_io_callback (void *refcon, IOReturn result, void *arg0);
+static void darwin_zlp_io_callback (void *refcon, IOReturn result, void *arg0);
 
 static enum libusb_error darwin_scan_devices(struct libusb_context *ctx);
 static enum libusb_error process_new_device (struct libusb_context *ctx, struct darwin_cached_device *cached_device,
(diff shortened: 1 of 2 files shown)


sonatique:
Hello @pradlol , 
Thank you for the PR. The Beagle capture seems to be  a solid diagnosis and the direction of the fix looks right. 

If I am right, your solution is not exactly equivalent to a working-as-expected sync Write, therefore, a few questions:

1. Is there a guarantee that WritePipeAsyncTO for the ZLP, submitted from within the data-phase completion, is serialized after the data write on the wire? For transfers that already get chunked internally, could the ZLP risk overtaking a not-yet-flushed fragment, or is submission-from-completion sufficient to guarantee ordering here?

2. The old design signaled transfer-complete in one shot from a single callback. This introduces a second completion stage, with the transfer effectively half-complete between the two callbacks. What happens if libusb_cancel_transfer() is called, or the device disconnects, in that window? Does darwin_release_zlp_interface() get invoked correctly in both cases, and is there any risk of the transfer completing twice or not at all?

3. WritePipeAsync (no timeout variant) is used for interrupt endpoints. If the ZLP write on an interrupt endpoint never completes, does the transfer hang indefinitely, whereas the old (broken) sync path would have at least returned promptly? Worth confirming that this is an accepted tradeoff.

4. Has this been exercised beyond your precise case, ideally against a second device/controller combination, given it touches the core async completion path for every OUT/interrupt transfer with ADD_ZERO_PACKET set, not just the previously-broken case?

5. I think that the old code had a second, independent problem beyond the ZLP not appearing: it was blocking the event thread with a sync call from an async completion context, which is generally an anti-pattern. The new code doesn't have that problem. In that sense the fix is better than a hypothetical working synchronous write would have been, not just equivalent to one. Do you confirm this analysis?

6. Does the fact that the sync write does not actually do something constitutes a bug in the framework in your opinion? Shall we fill a bug report somehow?

Thanks!

PR #1926, opened 2026-08-09
linux: read a wrapped device's active configuration from the kernel

### LINKED ISSUE #1925: Linux: libusb_wrap_sys_device() sends a GET_CONFIGURATION request on the bus, contrary to its documented contract
### Summary

The doxygen for `libusb_wrap_sys_device()` states:

> This is a non-blocking function; no requests are sent over the bus.

On Linux this does not hold. The wrap path issues a real `GET_CONFIGURATION`
control transfer to the device before returning.

The two ways a `libusb_device` gets initialized are asymmetric:

|                                          | `libusb_open()` (enumerated) | `libusb_wrap_sys_device()`             |
| ---------------------------------------- | ---------------------------- | -------------------------------------- |
| `sysfs_dir` given to `initialize_device()` | the device's sysfs directory | `NULL`                                 |
| active configuration read from            | sysfs `bConfigurationValue`  | live `GET_CONFIGURATION` on the wire   |

`libusb/os/linux_usbfs.c`:

```c
/* op_wrap_sys_device() */
r = initialize_device(dev, busnum, devaddr, NULL, fd);
```

```c
/* initialize_device() */
if (sysfs_dir) {
	/* sysfs descriptors are in bus-endian format */
	usbi_localize_device_descriptor(&dev->device_descriptor);
	return LIBUSB_SUCCESS;
}

/* cache active config */
...
r = usbfs_get_active_config(dev, fd);
```

and `usbfs_get_active_config()` performs an `IOCTL_USBFS_CONTROL` with
`bRequest = LIBUSB_REQUEST_GET_CONFIGURATION`.

### Why this is more than a documentation bug

Sending the request exposes a branch that the enumerated path can never reach:

```c
	} else if (active_config == 0) {
		if (dev_has_config0(dev)) {
			/* some buggy devices have a configuration 0, but we're
			 * reaching into the corner of a corner case here. */
			priv->active_config = 0;
		} else {
			priv->active_config = -1;
		}
	} else {
```

A device that **successfully** answers `0`, and has no configuration with
`bConfigurationValue == 0`, ends up at `active_config = -1`. From then on
`op_get_active_config_descriptor()` logs `device unconfigured` and returns
`LIBUSB_ERROR_NOT_FOUND`.

The state is sticky: of the seven assignments to `priv->active_config` in
`linux_usbfs.c`, the only one that can repair it after initialization is in
`op_set_configuration()` — so nothing recovers unless the caller explicitly
calls `libusb_set_configuration()`.

Note that the neighbouring error path — same function, a few lines above —
already treats this situation as recoverable:

```c
		/* assume the current configuration is the first one if we have
		 * the configuration descriptors, otherwise treat the device
		 * as unconfigured. */
		if (priv->config_descriptors)
			priv->active_config = (int)priv->config_descriptors[0].desc->bConfigurationValue;
		else
			priv->active_config = -1;
```

A device that fails to answer gets `config_descriptors[0]`; a device that
answers `0` gets `-1`. The second case seems no less recoverable than the first.

### Measured

Intel AX210 Bluetooth, `8087:0032`, Linux 7.1.4 x86_64, libusb 1.0.30.

A minimal program sending exactly the request libusb sends
(`bmRequestType=0x80, bRequest=0x08, wValue=0, wIndex=0, wLength=1,
timeout=1000`) through `USBDEVFS_CONTROL`, with
`/sys/module/usbcore/parameters/usbfs_snoop` enabled:

```
usb 1-4: control urb: bRequestType=80 bRequest=08 wValue=0000 wIndex=0000 wLength=0001
usb 1-4: ep0 ctrl-in, length 1, timeout 1000
usb 1-4: ep0 ctrl-in, actual_length 1, status 0
usb 1-4: data: 00
```

`status 0` — the device answers, it does not stall or time out. At the same
time the kernel's own view is:

```
/sys/bus/usb/devices/1-4/bConfigurationValue  ->  1
/sys/bus/usb/devices/1-4/bNumConfigurations   ->  1
```

So on this device the two entry points disagree: `libusb_open()` reports
configuration `1` from sysfs, while `libusb_wrap_sys_device()` reaches
`active_config = -1` — and per the doxygen, it should not have asked at all.

### Where it surfaced

QEMU's `usb-host` device takes the wrap path for its `hostdevice=` addressing
mode and the enumerated path for `hostbus=`/`hostaddr=`. With `hostdevice=`,
this radio enumerates in the guest under its correct name and
`SET_CONFIGURATION` succeeds, but every bulk/interrupt transfer stalls. The
same device on the same host works with `hostbus=`/`hostaddr=`. That
difference is what led here; QEMU's handling of `LIBUSB_ERROR_NOT_FOUND` is a
separate matter and not what this report is about.

### Suggested directions

Either one would resolve the contract mismatch:

1. Correct the doxygen for `libusb_wrap_sys_device()` — on Linux, one control
   transfer is issued.
2. Or make the `active_config == 0` branch fall back the same way the error
   path above it does, using `config_descriptors[0].desc->bConfigurationValue`
   when the device has no configuration `0`.

Glad to send a patch for whichever you prefer.

---

drpars:
Read a wrapped device's active configuration from the sysfs bConfigurationValue instead of asking the device. Against master
`c75357cd`: two files, 144 insertions / 13 deletions — `libusb/os/linux_usbfs.c` and one Doxygen note in `libusb/core.c`.

## The defect

`libusb_wrap_sys_device()` passes `sysfs_dir = NULL` to `initialize_device()`, so the sysfs path is skipped and the active
configuration is asked of the device with a `GET_CONFIGURATION` request. An Intel 8087:0032 Bluetooth radio answers **0** to
that request while its sysfs bConfigurationValue is **1** and `bNumConfigurations` is 1; the device declares no configuration 0,
so `priv->active_config` becomes `-1` and stays there. Measured on an ASUS G513RM, radio at `/dev/bus/usb/001/003` (sysfs
`1-4`), always with `LIBUSB_OPTION_NO_DEVICE_DISCOVERY` — which is what `libusb_wrap_sys_device()` exists for and what silences
the init-time sysfs scan, so every sysfs access is the code under test:

| | baseline | with this patch |
|---|---|---|
| `libusb_get_configuration()` | **0** (wrong) | **1** |
| `libusb_get_active_config_descriptor()` | **-5 NOT_FOUND** (wrong) | 0, `bConfigurationValue = 1`, `bNumInterfaces = 2` |
| control transfers on the wire, per run | 2 | **1** |
| ASUS N-KEY keyboard `1-3`, regression control | 1 / OK | 1 / OK, unchanged |

Correct in both call orders and under a descriptor re-check after `get_configuration()`. Two transfers become one, not zero: the
wrap-time transfer is gone, the one inside `libusb_get_configuration()` stays on purpose.

## What the patch does

`wrapped_fd_get_active_config()`, gated on `sysfs_available`, `fstat()`s the wrapped descriptor and requires `S_ISCHR`; opens
`/sys/dev/char/<major>:<minor>/bConfigurationValue` **through the symlink**; reads it with `parse_sysfs_attr_fd()`, the numeric
parser `read_sysfs_attr()` now shares; and accepts the value only if `IOCTL_USBFS_GET_CAPABILITIES` on the wrapped fd then
answers 0, or `ENOTTY` from a kernel older than the ioctl. The directory name is never copied and never resolved to a name
first: the kernel can give a name like `1-4` to a new device as soon as this one disconnects, so the link must resolve at open
time.

On a live fd the ioctl returned 0 and usbmon
captured zero URBs from it; on a staged disconnect it returned `ENODEV`, as did `CONNECTINFO`, `GET_SPEED`, `GETDRIVER` and
`CONTROL` on the same kept fd; on an fd opened `O_RDONLY` every usbfs ioctl tried returned `EPERM`, with the device attached and
again after it was gone, so `EPERM` separates nothing.

An empty attribute means an unconfigured device and arrives as `-1`. A real `0` is resolved the way the request path resolves
it, `if (active == 0 && !dev_has_config0(dev)) active = -1;`. Every failure at every step returns `LIBUSB_ERROR_NOT_SUPPORTED`,
and both call sites test `== LIBUSB_SUCCESS` only, so none reaches a user.

Also in this branch: `read_sysfs_attr()` splits into `open_sysfs_attr()` plus `parse_sysfs_attr_fd()` so the wrapped path can
read an attribute from a descriptor it opened itself, and the side that opens the fd closes it; `open_sysfs_attr()` fails
instead of truncating a path that does not fit, as a separate commit; `priv->sysfs_dir` is never set, no field is added, and
`op_get_active_config_descriptor()` is untouched, so its zero-I/O property holds; the `libusb_wrap_sys_device()` Doxygen note
names the one exception to "sends no requests over the bus"; and `<sys/sysmacros.h>` is a first use in the tree —
bionic ships `major()`/`minor()` as plain macros, read from its source.

## Scope: why `op_get_configuration()` is also touched

The arm in `op_get_configuration()` is the second place this branch touches. The narrower scope — seed the cache in
`initialize_device()` only, leave that arm out — was built and measured on the same radio. In the default call order it is
byte-identical to the unpatched baseline:

| call | unpatched | seed only | seed + the arm |
|---|---|---|---|
| `libusb_get_active_config_descriptor()` | NOT_FOUND | OK | OK |
| `libusb_get_configuration()` | 0 | 0 | 1 |
| the same descriptor call again | NOT_FOUND | **NOT_FOUND** | OK |

`usbfs_get_active_config()` writes the device's answer into `priv->active_config`, the cache `op_get_active_config_descriptor()`
reads; a correct seed that the first `libusb_get_configuration()` erases is not a fix. So the arm is cache protection,
not a fix to that function. The control request stays where it is: it is what reports `LIBUSB_ERROR_NO_DEVICE` on disconnect.
The sysfs bConfigurationValue replaces the device's answer only after that request returns `LIBUSB_SUCCESS` and only when
`hpriv->fd_keep` says the handle came through `libusb_wrap_sys_device()` — set at exactly one place, `op_wrap_sys_device()`.

## The minor-recycle race, realized on hardware

Staged with a `dummy_hcd` configfs gadget: device A (configuration 1, char device 189:1153) is wrapped, then unbind +
`dummy_hcd` reload + rebind enumerate device B as `10-1` with **A's minor** and configuration 2. Open-by-name on A's dead fd
reads **2**, B's value; this design opens through the link, reads 2, probes, gets `ENODEV`, and rejects it. Name-only reuse
(rebind, no reload) also reads B's value, which open-by-name survives only because the minor dangles.

## Limitation: an unprivileged wrap gets no saving

On an `O_RDONLY` wrapped fd the sysfs bConfigurationValue is read correctly and then discarded: every usbfs ioctl on such an fd
returned `EPERM`, measured with the device attached and again after it was gone, so the accept rule can never be met. The
`GET_CONFIGURATION` that follows returns `EPERM` too, and libusb answers from `config_descriptors[0]`. Probe tool
(`disctest`), fd opened `O_RDONLY` on the live `dummy_hcd` gadget used for the staged run above:

```
ioctl GET_CAPABILITIES rc=-1 errno=1 (Operation not permitted)
ioctl CONTROL(GET_CFG) rc=-1 errno=1 (Operation not permitted)
```

The answer is right there only because the radio has one configuration; the saving applies to a writable wrapped fd.

## Degradation, unconfigured, and multi-configuration

Faults injected into unmodified binaries with a user + mount namespace: tmpfs over `/sys` and over `/sys/dev/char` are
byte-identical to the unpatched baseline, and tmpfs over `/sys/bus/usb/devices` still answers from the sysfs bConfigurationValue
in 1 transfer, because the link resolves under `/sys/devices`. `echo 0 > .../1-4/bConfigurationValue` left the radio really
unconfigured — attribute empty, `hci0` gone — and both call orders gave `config = 0` and `NOT_FOUND` before and after, in 1
transfer instead of 2. A two-configuration `dummy_hcd` gadget (`1d6b:0104`, minor `1153 = (10-1)*128 + (2-1)`) matches the
baseline's values in 1 transfer at active = 1, active = 2 and unconfigured; active = 2 is decisive, since a guess answers 1.

## Build and test

16 gcc 16.2.1 `-Werror` configurations per tree, cell by cell: four logging modes × udev and no-udev at `-O2`; default and
`--disable-log` at `-O3`; `-O0`, `-O1` and `-Os` with default and disabled logging. Master `c75357cd` and this branch are clean
in all 16; `3649cc72`, the head before the truncation fix, fails 4 — `-O2 --disable-log` with and without udev, both `-O3`.

`make check` with `--enable-tests-build`: 5/5 PASS on master and on this branch. Under `.private/ci-build.sh --werror --
--enable-udev` the `umockdev` case fails on **both** trees with a byte-identical ASan exit-time leak summary, 115933 bytes in
8213 allocations; with `ASAN_OPTIONS=detect_leaks=0` it is 5/5 on both. `make -C doc` exits 0 on both trees, with the same
pre-existing doxygen warnings and none at a changed line.

## Intentional behaviour changes

1. A device whose `GET_CONFIGURATION` request fails no longer keeps the `config_descriptors[0]` guess when the sysfs
   bConfigurationValue is available. Reasoned from the code; no such device was observed.
2. A device that is unconfigured **and** declares a configuration 0 now reports unconfigured, per the documented contract.

## Not measured

A device with a real configuration 0: the `active == 0 && !dev_has_config0(dev)` arm has never executed, because configfs refuses
`configs/c.0` on this kernel. A stalling device whose request fails while the sysfs read succeeds — change 1 above. Android and
SELinux end to end (waived), an NDK compile of the two new includes, big-endian, and the non-Linux backends this does not touch.
32-bit is compile-tested only.

Fixes: #1925

### DIFF
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -714,8 +714,7 @@ static int open_sysfs_attr(struct libusb_context *ctx,
 	return fd;
 }
 
-/* Note only suitable for attributes which always read >= 0, < 0 is error.
- * Takes ownership of fd and closes it. */
+/* fd stays open: the caller that opened it also closes it. */
 static int parse_sysfs_attr_fd(struct libusb_context *ctx, int fd,
 	const char *attr, int max_value, int *value_p)
 {
@@ -728,13 +727,11 @@ static int parse_sysfs_attr_fd(struct libusb_context *ctx, int fd,
 	r = read(fd, buf, sizeof(buf) - 1);
 	if (r < 0) {
 		r = errno;
-		close(fd);
 		if (r == ENODEV)
 			return LIBUSB_ERROR_NO_DEVICE;
 		usbi_err(ctx, "attribute %s read failed, errno=%zd", attr, r);
 		return LIBUSB_ERROR_IO;
 	}
-	close(fd);
 
 	if (r == 0) {
 		/* Certain attributes (e.g. bConfigurationValue) are not
@@ -789,13 +786,16 @@ static int parse_sysfs_attr_fd(struct libusb_context *ctx, int fd,
 static int read_sysfs_attr(struct libusb_context *ctx,
 	const char *sysfs_dir, const char *attr, int max_value, int *value_p)
 {
-	int fd;
+	int fd, r;
 
 	fd = open_sysfs_attr(ctx, sysfs_dir, attr);
 	if (fd < 0)
 		return fd;
 
-	return parse_sysfs_attr_fd(ctx, fd, attr, max_value, value_p);
+	r = parse_sysfs_attr_fd(ctx, fd, attr, max_value, value_p);
+	close(fd);
+
+	return r;
 }
 
 static int sysfs_scan_device(struct libusb_context *ctx, const char *devname)
@@ -819,33 +819,26 @@ static int sysfs_get_active_config(struct libusb_device *dev, int *config)
 			UINT8_MAX, config);
 }
 
-/* read the bConfigurationValue for a device we only have a usbfs file
- * descriptor for, i.e. one that came in through libusb_wrap_sys_device().
+/* Read the sysfs bConfigurationValue of a device that came in through
+ * libusb_wrap_sys_device(). Such a device has no priv->sysfs_dir. A usbfs
+ * node is a character device, so /sys/dev/char/<major>:<minor> links to the
+ * owning device's sysfs directory. Open the attribute through that link,
+ * never through a copy of the directory name: the kernel can give a name
+ * like "1-4" to a new device as soon as this one disconnects.
  *
- * Such a device has no priv->sysfs_dir, so the active configuration is
- * normally asked of the device itself with a GET_CONFIGURATION request. The
- * kernel already knows the answer and publishes the mapping needed to find
- * it: a usbfs node is a character device, and /sys/dev/char/<major>:<minor>
- * is a symlink to the owning device's sysfs directory. The attribute is
- * opened through that link, never through a copy of the directory name: the
- * kernel can give the "1-4" style name to a new device the moment this one
- * disconnects, so the link must be resolved at open time.
+ * Trust the value only if IOCTL_USBFS_GET_CAPABILITIES on the wrapped fd
+ * then answers 0, or ENOTTY, which initialize_handle() already treats as a
+ * kernel without the ioctl; no such kernel was tested here. The ioctl is
+ * issued for that answer, not for the capabilities. Observed here: a live fd
+ * answered 0 with no bus traffic in usbmon, and an fd held open across a
+ * disconnect answered ENODEV, as did every other usbfs ioctl tried on it.
+ * "Anything but ENODEV" would not do: an fd opened without write access
+ * answered EPERM both while the device was attached and after it was gone.
+ * The probe runs after the read, so a value from a recycled link is rejected.
  *
- * The value read is trusted only if USBDEVFS_GET_CAPABILITIES on the
- * wrapped fd then proves the device is still connected. The kernel answers
- * that ioctl from memory, without bus traffic. usbfs checks the connection
- * before it decodes the command, so success is proof, and so is ENOTTY
- * from a kernel that predates the ioctl. Any other result is not: ENODEV
- * means the device is gone, and EPERM means the fd was opened without
- * write access and can prove nothing. A usbfs fd never comes back once its
- * device disconnects, so an fd that is connected after the read was
- * connected throughout it, and the value is its own.
- *
- * Any failure at any step is not an error: the caller falls back to asking
- * the device, which is what it did before, and for a disconnected fd that
- * request path is what reports LIBUSB_ERROR_NO_DEVICE. All failures collapse
- * to LIBUSB_ERROR_NOT_SUPPORTED so no caller is tempted to report one.
- */
+ * Every failure returns LIBUSB_ERROR_NOT_SUPPORTED, never a device error. The
+ * caller then asks the device, and that path reports LIBUSB_ERROR_NO_DEVICE
+ * for a disconnected fd. */
 static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
 	int *config)
 {
@@ -861,6 +854,8 @@ static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
 	if (fstat(fd, &st) < 0 || !S_ISCHR(st.st_mode))
 		return LIBUSB_ERROR_NOT_SUPPORTED;
 
+	/* No truncation check: the longest path this can build is 55
+	 * characters, two 10-digit numbers included, and path holds 64. */
 	snprintf(path, sizeof(path),
 		SYSFS_MOUNT_PATH "/dev/char/%u:%u/bConfigurationValue",
 		major(st.st_rdev), minor(st.st_rdev));
@@ -871,6 +866,7 @@ static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
 
 	r = parse_sysfs_attr_fd(ctx, attr_fd, "bConfigurationValue", UINT8_MAX,
 		&active);
+	close(attr_fd);
 	if (r < 0)
 		return LIBUSB_ERROR_NOT_SUPPORTED;
 
@@ -879,9 +875,8 @@ static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
 		return LIBUSB_ERROR_NOT_SUPPORTED;
 
 	/* An unconfigured device leaves the attribute empty, which
-	 * parse_sysfs_attr_fd() already reports as -1. Reaching this line with
-	 * a 0 means the kernel named configuration 0, so decide it the way the
-	 * request path decides the same answer. */
+	 * parse_sysfs_attr_fd() reports as -1. Map a real 0 as
+	 * usbfs_get_active_config() maps it. */
 	if (active == 0 && !dev_has_config0(dev))
 		active = -1;
 
@@ -1311,20 +1306,20 @@ static int initialize_device(struct libusb_device *dev, uint8_t busnum,
 
 	/* cache active config */
 	if (wrapped_fd >= 0) {
-		/* The device came in through libusb_wrap_sys_device(), which
-		 * is documented not to send requests over the bus when the
-		 * kernel can describe the wrapped device. Ask the kernel, and
-		 * only fall back to asking the device if it cannot be. */
+		/* libusb_wrap_sys_device() is documented not to send bus
+		 * requests when the kernel can describe the device. Read the
+		 * sysfs bConfigurationValue first; ask the device only if that
+		 * read fails. */
 		int active_config;
 
 		if (wrapped_fd_get_active_config(dev, wrapped_fd,
 				&active_config) == LIBUSB_SUCCESS) {
-			usbi_dbg(ctx, "active configuration %d for wrapped device from sysfs",
+			usbi_dbg(ctx, "active configuration %d for wrapped device from sysfs bConfigurationValue",
 				 active_config);
 			priv->active_config = active_config;
 			return LIBUSB_SUCCESS;
 		}
-		usbi_dbg(ctx, "no sysfs answer for wrapped device, asking the device");
+		usbi_dbg(ctx, "no sysfs bConfigurationValue for wrapped device, asking the device");
 	}
 
 	if (wrapped_fd < 0)
@@ -1789,19 +1784,26 @@ static int op_get_configuration(struct libusb_device_handle *handle,
 
 			active_config = priv->active_config;
 
-			/* The request stays: it is what reports a disconnect,
-			 * and this function is documented to do that. For a
-			 * wrapped descriptor its answer is not the one to keep
-			 * though -- initialize_device() already preferred the
-			 * kernel's, and this call would otherwise write the
-			 * device's answer back over it, which is the cache
-			 * op_get_active_config_descriptor() goes on to use. */
+			/* usbfs_get_active_config() has already written
+			 * priv->active_config: the device's answer, or a
+			 * config_descriptors[0] guess when the request fails
+			 * with anything but ENODEV. This arm overwrites it
+			 * with the sysfs bConfigurationValue; it does not
+			 * prevent the write. priv->active_config is the cache
+			 * op_get_active_config_descriptor() reads.
+			 *
+			 * The request itself stays. It is what reports a
+			 * disconnect, and libusb_get_configuration() is
+			 * documented to return LIBUSB_ERROR_NO_DEVICE when the
+			 * device has been disconnected. */
 			if (hpriv->fd_keep &&
 			    wrapped_fd_get_active_config(handle->dev, hpriv->fd,
 					&kernel_config) == LIBUSB_SUCCESS) {
 				if (kernel_config != active_config)
 					usbi_dbg(usbi_handle_ctx(handle),
-						 "device reports configuration %d but the kernel reports %d, keeping the kernel's",
+						 "device reports active configuration %d, "
+						 "sysfs bConfigurationValue reports %d, "
+						 "keeping the sysfs value",
 						 active_config, kernel_config);
 				priv->active_config = kernel_config;
 				active_config = kernel_config;
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -710,18 +710,16 @@ static int open_sysfs_attr(struct libusb_context *ctx,
 	return fd;
 }
 
-/* Note only suitable for attributes which always read >= 0, < 0 is error */
-static int read_sysfs_attr(struct libusb_context *ctx,
-	const char *sysfs_dir, const char *attr, int max_value, int *value_p)
+/* Note only suitable for attributes which always read >= 0, < 0 is error.
+ * Takes ownership of fd and closes it. */
+static int parse_sysfs_attr_fd(struct libusb_context *ctx, int fd,
+	const char *attr, int max_value, int *value_p)
 {
 	char buf[20], *endptr;
 	long value;
 	ssize_t r;
-	int fd;
 
-	fd = open_sysfs_attr(ctx, sysfs_dir, attr);
-	if (fd < 0)
-		return fd;
+	UNUSED(attr);  /* only referenced by usbi_err(), a no-op without logging */
 
 	r = read(fd, buf, sizeof(buf) - 1);
 	if (r < 0) {
@@ -783,6 +781,19 @@ static int read_sysfs_attr(struct libusb_context *ctx,
 	return 0;
 }
 
+/* Note only suitable for attributes which always read >= 0, < 0 is error */
+static int read_sysfs_attr(struct libusb_context *ctx,
+	const char *sysfs_dir, const char *attr, int max_value, int *value_p)
+{
+	int fd;
+
+	fd = open_sysfs_attr(ctx, sysfs_dir, attr);
+	if (fd < 0)
+		return fd;
+
+	return parse_sysfs_attr_fd(ctx, fd, attr, max_value, value_p);
+}
+
 static int sysfs_scan_device(struct libusb_context *ctx, const char *devname)
 {
 	uint8_t busnum, devaddr;
@@ -811,26 +822,34 @@ static int sysfs_get_active_config(struct libusb_device *dev, int *config)
  * normally asked of the device itself with a GET_CONFIGURATION request. The
  * kernel already knows the answer and publishes the mapping needed to find
  * it: a usbfs node is a character device, and /sys/dev/char/<major>:<minor>
- * is a symlink to the owning device's sysfs directory, whose basename is the
- * same "1-4" form priv->sysfs_dir holds. The node's device numbers come from
- * the sysfs "dev" attribute udev read when it created the node, and
- * linux_get_device_address() already trusts this file descriptor's identity
- * when it reads the bus number and address out of the node's name, so
- * resolving the same file descriptor's device numbers through the kernel's
- * own mapping introduces no new source of trust.
+ * is a symlink to the owning device's sysfs directory. The attribute is
+ * opened through that link, never through a copy of the directory name: the
+ * kernel can give the "1-4" style name to a new device the moment this one
+ * disconnects, so the link must be resolved at open time.
+ *
+ * The value read is trusted only if USBDEVFS_GET_CAPABILITIES on the
+ * wrapped fd then proves the device is still connected. The kernel answers
+ * that ioctl from memory, without bus traffic. usbfs checks the connection
+ * before it decodes the command, so success is proof, and so is ENOTTY
+ * from a kernel that predates the ioctl. Any other result is not: ENODEV
+ * means the device is gone, and EPERM means the fd was opened without
+ * write access and can prove nothing. A usbfs fd never comes back once its
+ * device disconnects, so an fd that is connected after the read was
+ * connected throughout it, and the value is its own.
  *
  * Any failure at any step is not an error: the caller falls back to asking
- * the device, which is what it did before. All failures collapse to
- * LIBUSB_ERROR_NOT_SUPPORTED so no caller is tempted to report one.
+ * the device, which is what it did before, and for a disconnected fd that
+ * request path is what reports LIBUSB_ERROR_NO_DEVICE. All failures collapse
+ * to LIBUSB_ERROR_NOT_SUPPORTED so no caller is tempted to report one.
  */
 static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
 	int *config)
 {
-	struct libusb_context *ctx = DEVICE_CTX(dev);
-	char link[64], target[PATH_MAX], *name;
+	struct libusb_context *ctx = usbi_device_ctx(dev);
+	char path[64];
 	struct stat st;
-	ssize_t len;
-	int active, r;
+	uint32_t caps;
+	int active, attr_fd, r;
 
 	if (!sysfs_available)
 		return LIBUSB_ERROR_NOT_SUPPORTED;
@@ -838,26 +857,26 @@ static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
 	if (fstat(fd, &st) < 0 || !S_ISCHR(st.st_mode))
 		return LIBUSB_ERROR_NOT_SUPPORTED;
 
-	snprintf(link, sizeof(link), "/sys/dev/char/%u:%u",
+	snprintf(path, sizeof(path),
+		SYSFS_MOUNT_PATH "/dev/char/%u:%u/bConfigurationValue",
 		major(st.st_rdev), minor(st.st_rdev));
 
-	len = readlink(link, target, sizeof(target) - 1);
-	if (len < 0)
-		return LIBUSB_ERROR_NOT_SUPPORTED;
-	target[len] = '\0';
-
-	name = strrchr(target, '/');
-	name = name ? name + 1 : target;
-	if (*name == '\0')
+	attr_fd = open(path, O_RDONLY | O_CLOEXEC);
+	if (attr_fd < 0)
 		return LIBUSB_ERROR_NOT_SUPPORTED;
 
-	r = read_sysfs_attr(ctx, name, "bConfigurationValue", UINT8_MAX, &active);
+	r = parse_sysfs_attr_fd(ctx, attr_fd, "bConfigurationValue", UINT8_MAX,
+		&active);
 	if (r < 0)
 		return LIBUSB_ERROR_NOT_SUPPORTED;
 
+	r = ioctl(fd, IOCTL_USBFS_GET_CAPABILITIES, &caps);
+	if (r < 0 && errno != ENOTTY)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
 	/* An unconfigured device leaves the attribute empty, which
-	 * read_sysfs_attr() already reports as -1. Reaching this line with a 0
-	 * means the kernel named configuration 0, so decide it the way the
+	 * parse_sysfs_attr_fd() already reports as -1. Reaching this line with
+	 * a 0 means the kernel named configuration 0, so decide it the way the
 	 * request path decides the same answer. */
 	if (active == 0 && !dev_has_config0(dev))
 		active = -1;
@@ -1777,7 +1796,7 @@ static int op_get_configuration(struct libusb_device_handle *handle,
 			    wrapped_fd_get_active_config(handle->dev, hpriv->fd,
 					&kernel_config) == LIBUSB_SUCCESS) {
 				if (kernel_config != active_config)
-					usbi_dbg(HANDLE_CTX(handle),
+					usbi_dbg(usbi_handle_ctx(handle),
 						 "device reports configuration %d but the kernel reports %d, keeping the kernel's",
 						 active_config, kernel_config);
 				priv->active_config = kernel_config;
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -693,9 +693,13 @@ static int open_sysfs_attr(struct libusb_context *ctx,
 	const char *sysfs_dir, const char *attr)
 {
 	char filename[256];
-	int fd;
+	int fd, r;
 
-	snprintf(filename, sizeof(filename), SYSFS_DEVICE_PATH "/%s/%s", sysfs_dir, attr);
+	r = snprintf(filename, sizeof(filename), SYSFS_DEVICE_PATH "/%s/%s", sysfs_dir, attr);
+	if (r < 0 || r >= (int)sizeof(filename)) {
+		usbi_err(ctx, "sysfs attribute path %s/%s too long", sysfs_dir, attr);
+		return LIBUSB_ERROR_OTHER;
+	}
 	fd = open(filename, O_RDONLY | O_CLOEXEC);
 	if (fd < 0) {
 		if (errno == ENOENT) {
--- a/libusb/os/linux_usbfs.c
+++ b/libusb/os/linux_usbfs.c
@@ -36,6 +36,8 @@
 #include <string.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
 #include <sys/utsname.h>
 #include <sys/vfs.h>
 #include <unistd.h>
@@ -798,6 +800,69 @@ static int sysfs_get_active_config(struct libusb_device *dev, int *config)
 			UINT8_MAX, config);
 }
 
+/* read the bConfigurationValue for a device we only have a usbfs file
+ * descriptor for, i.e. one that came in through libusb_wrap_sys_device().
+ *
+ * Such a device has no priv->sysfs_dir, so the active configuration is
+ * normally asked of the device itself with a GET_CONFIGURATION request. The
+ * kernel already knows the answer and publishes the mapping needed to find
+ * it: a usbfs node is a character device, and /sys/dev/char/<major>:<minor>
+ * is a symlink to the owning device's sysfs directory, whose basename is the
+ * same "1-4" form priv->sysfs_dir holds. The node's device numbers come from
+ * the sysfs "dev" attribute udev read when it created the node, and
+ * linux_get_device_address() already trusts this file descriptor's identity
+ * when it reads the bus number and address out of the node's name, so
+ * resolving the same file descriptor's device numbers through the kernel's
+ * own mapping introduces no new source of trust.
+ *
+ * Any failure at any step is not an error: the caller falls back to asking
+ * the device, which is what it did before. All failures collapse to
+ * LIBUSB_ERROR_NOT_SUPPORTED so no caller is tempted to report one.
+ */
+static int wrapped_fd_get_active_config(struct libusb_device *dev, int fd,
+	int *config)
+{
+	struct libusb_context *ctx = DEVICE_CTX(dev);
+	char link[64], target[PATH_MAX], *name;
+	struct stat st;
+	ssize_t len;
+	int active, r;
+
+	if (!sysfs_available)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	if (fstat(fd, &st) < 0 || !S_ISCHR(st.st_mode))
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	snprintf(link, sizeof(link), "/sys/dev/char/%u:%u",
+		major(st.st_rdev), minor(st.st_rdev));
+
+	len = readlink(link, target, sizeof(target) - 1);
+	if (len < 0)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+	target[len] = '\0';
+
+	name = strrchr(target, '/');
+	name = name ? name + 1 : target;
+	if (*name == '\0')
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	r = read_sysfs_attr(ctx, name, "bConfigurationValue", UINT8_MAX, &active);
+	if (r < 0)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	/* An unconfigured device leaves the attribute empty, which
+	 * read_sysfs_attr() already reports as -1. Reaching this line with a 0
+	 * means the kernel named configuration 0, so decide it the way the
+	 * request path decides the same answer. */
+	if (active == 0 && !dev_has_config0(dev))
+		active = -1;
+
+	*config = active;
+
+	return LIBUSB_SUCCESS;
+}
+
 int linux_get_device_address(struct libusb_context *ctx, int detached,
 	uint8_t *busnum, uint8_t *devaddr, const char *dev_node,
 	const char *sys_name, int fd)
@@ -1218,6 +1283,23 @@ static int initialize_device(struct libusb_device *dev, uint8_t busnum,
 	}
 
 	/* cache active config */
+	if (wrapped_fd >= 0) {
+		/* The device came in through libusb_wrap_sys_device(), which
+		 * is documented not to send requests over the bus when the
+		 * kernel can describe the wrapped device. Ask the kernel, and
+		 * only fall back to asking the device if it cannot be. */
+		int active_config;
+
+		if (wrapped_fd_get_active_config(dev, wrapped_fd,
+				&active_config) == LIBUSB_SUCCESS) {
+			usbi_dbg(ctx, "active configuration %d for wrapped device from sysfs",
+				 active_config);
+			priv->active_config = active_config;
+			return LIBUSB_SUCCESS;
+		}
+		usbi_dbg(ctx, "no sysfs answer for wrapped device, asking the device");
+	}
+
 	if (wrapped_fd < 0)
 		fd = get_usbfs_fd(dev, O_RDWR, 1);
 	else
@@ -1675,8 +1757,29 @@ static int op_get_configuration(struct libusb_device_handle *handle,
 		struct linux_device_handle_priv *hpriv = (struct linux_device_handle_priv *)usbi_get_device_handle_priv(handle);
 
 		r = usbfs_get_active_config(handle->dev, hpriv->fd);
-		if (r == LIBUSB_SUCCESS)
+		if (r == LIBUSB_SUCCESS) {
+			int kernel_config;
+
 			active_config = priv->active_config;
+
+			/* The request stays: it is what reports a disconnect,
+			 * and this function is documented to do that. For a
+			 * wrapped descriptor its answer is not the one to keep
+			 * though -- initialize_device() already preferred the
+			 * kernel's, and this call would otherwise write the
+			 * device's answer back over it, which is the cache
+			 * op_get_active_config_descriptor() goes on to use. */
+			if (hpriv->fd_keep &&
+			    wrapped_fd_get_active_config(handle->dev, hpriv->fd,
+					&kernel_config) == LIBUSB_SUCCESS) {
+				if (kernel_config != active_config)
+					usbi_dbg(HANDLE_CTX(handle),
+						 "device reports configuration %d but the kernel reports %d, keeping the kernel's",
+						 active_config, kernel_config);
+				priv->active_config = kernel_config;
+				active_config = kernel_config;
+			}
+		}
 	}
 	if (r < 0)
 		return r;
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1410,11 +1410,11 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
  * available to you through libusb_get_device(). This device is destroyed
  * during libusb_close(). The device shall not be opened through libusb_open().
  *
- * No requests are sent over the bus and the function does not block, with
- * one exception: on Linux, if the active configuration of the wrapped
- * device cannot be read from the kernel, it is asked of the device itself
- * with a single GET_CONFIGURATION control request, sent with a 1000
- * millisecond timeout.
+ * This function does not block and sends no requests over the bus. On Linux
+ * one exception is possible: a single GET_CONFIGURATION control request to
+ * the wrapped device, with a 1000 millisecond timeout. The kernel usually
+ * supplies the active configuration of that device. This request is sent
+ * only when the kernel cannot supply it.
  *
  * Since version 1.0.23, \ref LIBUSB_API_VERSION >= 0x01000107
  *
--- a/libusb/core.c
+++ b/libusb/core.c
@@ -1410,7 +1410,11 @@ void API_EXPORTED libusb_unref_device(libusb_device *dev)
  * available to you through libusb_get_device(). This device is destroyed
  * during libusb_close(). The device shall not be opened through libusb_open().
  *
- * This is a non-blocking function; no requests are sent over the bus.
+ * No requests are sent over the bus and the function does not block, with
+ * one exception: on Linux, if the active configuration of the wrapped
+ * device cannot be read from the kernel, it is asked of the device itself
+ * with a single GET_CONFIGURATION control request, sent with a 1000
+ * millisecond timeout.
  *
  * Since version 1.0.23, \ref LIBUSB_API_VERSION >= 0x01000107
  *


drpars:
# Evidence — libusb #1925, `linux-wrap-sys-device-active-config`

Patch: `libusb-1925.patch`, 109 insertions / 2 deletions, two files
(`libusb/os/linux_usbfs.c`, `libusb/core.c`), against master `5a6a8d3`; applies clean with
`--whitespace=error-all`.

Machine: ASUS G513RM, Linux 7.1.4-1-g14, x86_64. Device under test: Intel 8087:0032 Bluetooth
radio, bus 1 device 3, sysfs `1-4` — it reproduces the issue (its GET_CONFIGURATION answer is 0
while the kernel says 1). Regression control: ASUS N-KEY keyboard, bus 1 device 2, sysfs `1-3`,
which answers correctly. Harness `wraptest.c` + `ioctl_spy.c` (`harness/run.sh`), every run with
`LIBUSB_OPTION_NO_DEVICE_DISCOVERY`, so each sysfs access provably belongs to the code under
test. usbfs `nr=0` is a CONTROL transfer on the bus; `nr=31`/`nr=26` are served from memory.

---

## 1. The defect

Unpatched `5a6a8d3`, radio, default call order — verbatim from `run-base.log`:

```
--- BT radio 1-4  | default order (get_configuration first) ---
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[MARK] ===== libusb_wrap_sys_device =====
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
  [wire control transfers] = 2
```

The device is configured — the kernel says so in the first two lines — and libusb reports it
unconfigured, having spent two control transfers to get there.

## 2. The fix, measured

Same harness, patched tree. Default order, verbatim from `run-m2.log`:

```
--- BT radio 1-4  | default order (get_configuration first) ---
  sysfs bConfigurationValue  = 1
[MARK] ===== libusb_wrap_sys_device =====
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
  [wire control transfers] = 1
```

Reverse order with a descriptor re-check after `libusb_get_configuration()` — the run that
separates this patch from the narrower ones, verbatim from `run-m2.log`:

```
--- BT radio 1-4  | reverse order + re-check (kills m1) ---
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== libusb_get_configuration =====
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor (again) =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
  [wire control transfers] = 1
```

All five runs in `run-m2.log` (default, default+recheck, revorder, revorder+recheck, keyboard)
give `config = 1`, a valid descriptor (radio `bNumInterfaces = 2`, keyboard 1) and
`[wire control transfers] = 1`. The keyboard run is unchanged in every value; only its transfer
count drops.

## 3. Why the second hunk exists — the m1 exhibit

`m1` is the suggested narrower scope: seed the cache in `initialize_device()` only, leave
`op_get_configuration()` alone. Preserved as
`rejected-designs/m1-maintainer-scope-insufficient.patch`. Measured on the same hardware,
**m1 in the default call order is byte-identical to the unpatched baseline** — the defect is not
fixed at all. The decisive three-call sequence:

| call | unpatched | m1 | v4 |
|---|---|---|---|
| `libusb_get_active_config_descriptor` | NOT_FOUND | **OK** (`bNumInterfaces = 2`) | OK |
| `libusb_get_configuration` | `config = 0` | `config = 0` | `config = 1` |
| the same descriptor call again | NOT_FOUND | **NOT_FOUND** | OK |

The cause is in the call, not the seeding: `usbfs_get_active_config()` writes the *device's*
answer back over `priv->active_config`, which is exactly the cache
`op_get_active_config_descriptor()` reads. A correct seed that any later
`libusb_get_configuration()` erases is not a fix. The hunk is therefore **cache protection, not
a fix to `op_get_configuration()`**: the ioctl is retained (§4), and the kernel's answer replaces
the device's only after that ioctl has returned `LIBUSB_SUCCESS`, and only when `hpriv->fd_keep`
says the handle came through `libusb_wrap_sys_device()` (`fd_keep` is set at exactly one place,
`op_wrap_sys_device`).

## 4. Wire transfers: 2 → 1, not 0

`USBDEVFS nr=0` lines counted mechanically per run: **2 in every one of the five baseline runs,
1 in every one of the five patched runs** (`run-base.log` / `run-m2.log`), radio and keyboard
alike, in both call orders. The one that disappears is the transfer `initialize_device()` sent
at wrap time, which
contradicted `libusb_wrap_sys_device()`'s documented non-blocking contract; the `core.c` doc
hunk now states the remaining exception precisely. The one that stays is inside
`libusb_get_configuration()`, kept deliberately: it is what reports `LIBUSB_ERROR_NO_DEVICE` on
disconnect, and that function's doxygen documents it. `op_get_active_config_descriptor()` is
untouched and still does zero I/O.

## 5. A genuinely unconfigured device

`echo 0 > /sys/bus/usb/devices/1-4/bConfigurationValue` left the radio really unconfigured
(attribute empty, `hci0` gone) — the first execution of the helper's empty-attribute → `-1`
branch on real hardware.

Before, verbatim from `unconf-base-default.log`:

```
  sysfs bConfigurationValue  = <empty>
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
```

After, verbatim from `unconf-m2-default.log`:

```
  sysfs bConfigurationValue  = <empty>
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] readlink /sys/dev/char/189:2 -> ../../devices/pci0000:00/0000:00:08.1/0000:05:00.3/usb1/1-4
[spy] open /sys/bus/usb/devices/1-4/bConfigurationValue
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
```

Both call orders, both with re-checks (`unconf-{base,m2}-{default,revorder}.log`): user-visible
results **identical** — `config = 0`, descriptor `NOT_FOUND` — reached with 1 transfer instead
of 2. The documented unconfigured contract (core.c "return 0 if unconfigured", descriptor.c
"NOT_FOUND if unconfigured") is preserved. Radio restored afterwards (`echo 1`, `hci0` back).

Those two `[spy]` lines are also the direct answer to "please confirm the `/sys/dev/char`
symlink": it resolves to the owning device's sysfs directory, whose basename is the same `1-4`
form `priv->sysfs_dir` holds. Checked for all 12 USB nodes on this machine; each node's
major:minor matches `/sys/bus/usb/devices/<name>/dev`; `CONFIG_SYSFS_DEPRECATED` absent.

## 6. Degradation matrix

Method: **unmodified binaries** — no rebuild, no `#define` edit — run inside a user + mount
namespace (`unshare -r -m --propagation private`) that hides part of sysfs (`harness/degrade.sh`).

| view | what it models | patched result | transfers |
|---|---|---|---|
| D0 normal | namespace control | `config = 1`, descriptor OK | 1 |
| D1 tmpfs over `/sys` | `sysfs_available = 0`, the Android shape | `config = 0`, NOT_FOUND | 2 |
| D2 tmpfs over `/sys/dev/char` | node → sysfs mapping missing | `config = 0`, NOT_FOUND | 2 |
| D3 tmpfs over `/sys/bus/usb/devices` | attribute unreachable | `config = 0`, NOT_FOUND | 2 |

The D1–D3 blocks of `degrade-m2.log` are **byte-identical** to the unpatched ones in
`degrade-base.log` (`diff` over those blocks: no output): every degraded path lands exactly
where the unpatched library lands — old behaviour, not a new failure mode.

## 7. Project CI and umockdev

`.private/ci-build.sh --werror` (ASan + extra warnings, `HAVE_LIBUDEV=1`):

* Before umockdev was installed here: exit 0, `# TOTAL: 4 / # PASS: 4 / # FAIL: 0`.
* With umockdev installed, the patched tree (`ci-v4.log`) and a pristine `5a6a8d3` tree
  (`ci-base2.log`) both run all ten test bodies successfully — `ok 1 /libusb/open-close` …
  `ok 10 /libusb/hotplug/add-remove` in both logs — and both then fail the suite on ASan's
  leak check at exit:

  ```
  SUMMARY: AddressSanitizer: 115933 byte(s) leaked in 8213 allocation(s).
  FAIL umockdev (exit status: 1)
  ```

  Byte-identical in the two trees, as is the leak list (leaks in `libumockdev` plus one
  pre-existing allocation on the enumeration path). Zero delta from the patch.
* With `ASAN_OPTIONS=detect_leaks=0`: 5/5 PASS in both trees.

## 8. Doxygen

`make -C doc` exits 0 (`doc-v4.log`). Every warning is pre-existing: obsolete tags in the
checked-in `doxygen.cfg` (`OUTPUT_TEXT_DIRECTION`, `HTML_TIMESTAMP`, …) and the parser's
`Found ')' without opening '(' for trailing return type ' -> lock)...'` noise on old code in
`core.c`, `hotplug.c`, `io.c`. None at a changed line. The new sentence renders in
`doc/api-1.0/group__libusb__dev.html`.

---

## Measured

* Defect and fix on real hardware, both call orders, with and without a descriptor re-check,
  plus an unaffected device as regression control.
* Wire transfers 2 → 1, counted from the ioctl trace, in every run.
* The m1 (initialize-device-only) scope failing the three-call sequence.
* A genuinely unconfigured device: identical user-visible results before and after.
* Degradation D1–D3 with unmodified binaries: byte-identical to unpatched.
* Project CI with ASan and `--werror`; umockdev bodies 10/10 in both trees with an identical
  leak summary; doxygen build.
* `/sys/dev/char` symlink shape for all 12 USB nodes on this machine.
* bionic (Android libc) ships `sys/sysmacros.h` with `major()`/`minor()` as plain macros — read
  from the bionic source, **not** compiled with an NDK.

## Not measured

* A multi-configuration device, and a device with a real configuration 0: the
  `active == 0 && !dev_has_config0(dev)` arm of the new helper has never executed — code-read
  only. Closable with `dummy_hcd` + a configfs gadget; not done.
* Android / SELinux end to end (waived: "I guess no need to cover"), and an NDK compile of the
  two new includes.
* Big-endian and 32-bit builds.
* A real hotplug / minor-recycle race. The retained ioctl gates the override, so a device that
  has gone away fails the call before the kernel's answer can be applied — reasoned from usbfs
  semantics, not measured.

## Known cost, disclosed

When `/sys/bus/usb/devices/<name>/bConfigurationValue` is unreadable (SELinux, say) while
`/sys/dev/char` is readable, every `libusb_get_configuration()` on a wrapped device logs one
`usbi_err` line from `read_sysfs_attr()` and then succeeds through the fallback. Accepted:
per-call resolution (one `fstat`, one `readlink`, one attribute read) is the price of not
caching a directory for wrapped devices.

## Two intentional behaviour changes

1. A stalling device whose answer disagrees with the kernel no longer wins: the guessed
   `config_descriptors[0]` is replaced by the kernel's real answer.
2. A device that is unconfigured *and* declares a configuration 0 now reports unconfigured, per
   the documented contract, instead of a full descriptor.

drpars:
Follow-up: the multi-configuration gap is now measured, with `dummy_hcd` + a configfs gadget as you suggested.

Gadget: `1d6b:0104`, two configurations (values 1 and 2), one Loopback function each, bound to `dummy_udc.0`; enumerated by this host as `10-1`, usbfs node `/dev/bus/usb/010/002`. The `/sys/dev/char` symlink holds on a bus with no PCI host controller behind it too, and the node's minor is `1153 = (10-1)*128 + (2-1)`.

Default call order, with a descriptor re-check after `get_configuration`, unpatched baseline vs this branch:

| gadget state | baseline | patched |
|---|---|---|
| active = 1 (the first) | `config = 1`, descriptor value 1 | same values, 2 → 1 transfers |
| active = 2 — while the *first* descriptor's value is 1 | `config = 2`, descriptor value 2 | same values, 2 → 1 transfers |
| unconfigured (`echo 0`; no config 0 exists) | `config = 0`, `NOT_FOUND` | identical, 2 → 1 transfers |

The second row is the one that matters: the sysfs read reports the **active** configuration, not the first one — a `config_descriptors[0]`-style guess would have answered 1 there.

One negative result, recorded so nobody retries it: a **real configuration numbered 0** turned out not to be constructible on this kernel via configfs — `mkdir configs/c.0` is refused. So the `active == 0 && dev_has_config0()` arm remains code-read only (it mirrors the request path's arm verbatim); hardware in the Broadcom 5880 family appears to be the only way to execute it.

Teardown verified: gadget removed, `dummy_hcd` unloaded, the machine's own devices unaffected.

drpars:
CI note on the AppVeyor red: the failing job is `stress.exe` (Windows backend, MinGW-Win32 run under Wine) segfaulting in its `default_context_change` case — a binary this diff is not compiled into (`linux_usbfs.c` is Linux-only, and the `core.c` hunk is a doc comment). The identical commit also ran on the project's second AppVeyor instance and passed all six jobs there ([54513062](https://ci.appveyor.com/project/LudovicRousseau/libusb/builds/54513062) success vs [54513061](https://ci.appveyor.com/project/dickens/libusb/builds/54513061)). Happy to rebase to retrigger if useful.

drpars:
CI note on the remaining red, `linux / build`: it is the Emscripten job, and it is pre-existing on master rather than caused by this branch.

It fails at link time with ~143 `wasm-ld: undefined symbol` errors, all coming from `emscripten_webusb.o` — `vtable for std::bad_optional_access`, `std::__2::bad_function_call`, `std::logic_error::logic_error(char const*)`, `__cxa_throw` — with emcc's own hint in the log: *"link failed with undefined C++ symbols. Try linking with 'em++' or passing '-sDEFAULT_TO_CXX'"*.

The same job fails the same way on master itself: run [31329931669](https://github.com/libusb/libusb/actions/runs/31329931669) (push, `daedf276c38a544dba86d408a930fa8983c42f00`) — same count, same symbols, same emcc hint. The run on this PR is [31295681714](https://github.com/libusb/libusb/actions/runs/31295681714).

This diff also cannot reach that build: `linux_usbfs.c` is not compiled for the Emscripten backend, and the `core.c` hunk is a documentation comment with no code change.

Everything else on this commit is green — `GCC 16 / C23`, `Fuzzing`, and all 27 MSVC/Windows jobs (29 of 30 checks). Together with the AppVeyor note above, both remaining reds are environmental.

Youw (commented):
Codex found one correctness issue in the wrapped-file-descriptor sysfs lookup. The inline comment gives the required fix direction.

Review scope: PR #1926, which fixes #1925.

libusb/os/linux_usbfs.c
```diff
@@ -802,6 +804,69 @@ static int sysfs_get_active_config(struct libusb_device *dev, int *config)
...
+	if (len < 0)
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+	target[len] = '\0';
+
+	name = strrchr(target, '/');
+	name = name ? name + 1 : target;
+	if (*name == '\0')
+		return LIBUSB_ERROR_NOT_SUPPORTED;
+
+	r = read_sysfs_attr(ctx, name, "bConfigurationValue", UINT8_MAX, &active);
```
Youw: [P2] Keep the sysfs read bound to the wrapped file descriptor

This code gets a topology name such as `1-4` from `/sys/dev/char/<major>:<minor>`. It then opens the attribute through that topology name. Linux can reuse the topology name after a device disconnects. If a new device connects between these operations, libusb can store the new device's configuration with descriptors from the old file descriptor.

The wrap path has a second problem. `initialize_device()` returns after this helper succeeds. It does not do the previous usbfs ioctl that returned `ENODEV` for a disconnected file descriptor.

Open `bConfigurationValue` through the `/sys/dev/char/<major>:<minor>` link. Then run a usbfs ioctl that checks whether the original file descriptor is still connected and does not send a bus request. Reject the sysfs value or use the fallback when the ioctl returns `ENODEV`. Apply the same post-read check in `op_get_configuration()`. Its control ioctl runs before this helper, so it does not validate the later sysfs read.

drpars: Fixed in 672c3814. Both problems are closed. The failure and the fix were
both measured on staged hardware (`dummy_hcd` gadget, Linux 7.1.8); logs
below.

What changed:

- The attribute is now opened through the link itself:
  `/sys/dev/char/<major>:<minor>/bConfigurationValue`. The kernel resolves
  the owner at open time. No topology name is copied. `read_sysfs_attr()`
  is split so its parser can run on the link-opened fd
  (`parse_sysfs_attr_fd()`).
- After the read, `USBDEVFS_GET_CAPABILITIES` runs on the wrapped fd. The
  value is trusted only when the result proves the device is still
  connected: 0, or ENOTTY from a kernel that predates the ioctl. Both are
  proof because usbfs checks `connected()` before it decodes the command,
  and it answers this ioctl from memory. Every other result rejects the
  value: ENODEV means the device is gone, and EPERM means the fd was opened
  without write access, so nothing can be proven. On rejection the callers
  fall back to the request path, so wrapping a disconnected fd reports
  `LIBUSB_ERROR_NO_DEVICE` again.
- `op_get_configuration()`: the post-read check lives inside the helper,
  and that function calls the helper after its control ioctl. The diff
  shows no new logic in the function itself — the protection is in the
  shared helper. The helper never returns success without the probe, so a
  foreign value cannot overwrite the cache at that call site either.

Why `GET_CAPABILITIES`: `initialize_handle()` already issues it on every
opened fd and on every wrapped fd, with an ENOTTY branch for old kernels.
`CONNECTINFO` would also work; this choice matches the existing idiom.

Measurements (dummy_hcd gadget `1d6b:0104`, two configurations, Loopback
function; same harness as the issue):

1. Disconnected fd: `GET_CAPABILITIES`, `CONNECTINFO`, `GET_SPEED`,
   `GETDRIVER`, `CONTROL` all return ENODEV.
2. usbmon on the gadget bus during the live probe battery: the only URBs
   are the `CONTROL` GET_CONFIGURATION pair. The probe makes no bus
   traffic.
3. The reuse race, staged end to end: gadget A enumerated (config 1,
   char device 189:1153), fd kept open, A disconnected, `dummy_hcd`
   reloaded, gadget B enumerated with the same name `10-1` and the same
   189:1153, B set to config 2. The old sequence (readlink, reopen by
   name) returned B's value 2 through A's dead fd. The new sequence read
   the same 2 and rejected it: the probe returned ENODEV. The name-only
   variant (rebind without module reload) fails the same way through the
   readlink-then-open window.
4. O_RDONLY fd: every usbfs ioctl returns EPERM, live or dead — the
   FMODE_WRITE check runs before `connected()`. The probe therefore
   rejects the value for read-only fds and the pre-PR behavior applies.
5. Regression matrix on real devices (wire-transfer counter): unchanged —
   one control transfer per `libusb_get_configuration`, none at wrap.
   Degraded-sysfs matrix: unchanged, with one improvement — with
   `/sys/bus/usb/devices` unreadable but `/sys/dev/char` intact, the link
   path still answers from the kernel; the old code fell back to the wire.

<details>
<summary>Log excerpts</summary>

Live fd battery (usbmon capturing bus 10 at the same time):

```
ioctl GET_CAPABILITIES rc=0        caps=0x1ff
ioctl CONNECTINFO      rc=0        devnum=2 slow=0
ioctl GET_SPEED        rc=3
ioctl GETDRIVER        rc=-1 errno=61 (No data available)
ioctl CONTROL(GET_CFG) rc=1        wire config=1
```

usbmon, complete capture for the battery above — only the CONTROL pair:

```
ffff8b76bdac1e00 1032349177 S Ci:10:002:0 s 80 08 0000 0000 0001 1 <
ffff8b76bdac1e00 1032349364 C Ci:10:002:0 0 1 = 01
```

Disconnected fd battery:

```
ioctl GET_CAPABILITIES rc=-1 errno=19 (No such device)
ioctl CONNECTINFO      rc=-1 errno=19 (No such device)
ioctl GET_SPEED        rc=-1 errno=19 (No such device)
ioctl GETDRIVER        rc=-1 errno=19 (No such device)
ioctl CONTROL(GET_CFG) rc=-1 errno=19 (No such device)
```

Minor-reuse staging (A dead, B live and owning A's 189:1153, B at config 2;
fd is still A's):

```
old sequence: fstat 189:1153 -> readlink -> name '10-1'
              open-by-name value='2'      <-- another device's value
new sequence: open-via-link value='2'
              liveness GET_CAPABILITIES rc=-1 errno=19 -> value REJECTED
```

O_RDONLY fd on a live device:

```
ioctl GET_CAPABILITIES rc=-1 errno=1 (Operation not permitted)
ioctl CONTROL(GET_CFG) rc=-1 errno=1 (Operation not permitted)
```

</details>

libusb/os/linux_usbfs.c
```diff
@@ -708,18 +710,14 @@ static int open_sysfs_attr(struct libusb_context *ctx,
...
-static int read_sysfs_attr(struct libusb_context *ctx,
-	const char *sysfs_dir, const char *attr, int max_value, int *value_p)
+/* Note only suitable for attributes which always read >= 0, < 0 is error.
+ * Takes ownership of fd and closes it. */
+static int parse_sysfs_attr_fd(struct libusb_context *ctx, int fd,
+	const char *attr, int max_value, int *value_p)
 {
 	char buf[20], *endptr;
 	long value;
 	ssize_t r;
```
Youw, line 723: [P1] Keep `attr` used when logging is disabled

With `--disable-log`, the log macros discard their arguments. The new `attr` parameter is then unused, and the required Linux `--werror` job fails with `-Werror=unused-parameter`. Codex reproduced the same failure locally. Mark the parameter as used, as `read_sysfs_attr_fd()` already does.

```suggestion
	ssize_t r;

	UNUSED(attr);
```

drpars: Applied in `3649cc72`.

I kept the trailing comment that `read_sysfs_attr_fd()` carries, since that is the sibling you pointed at:

```c
	UNUSED(attr);  /* only referenced by usbi_err(), a no-op without logging */
```

Reproduced and verified with the job's own commands, on the PR head:

* `.private/ci-build.sh --werror --build-dir build-nolog -- --disable-log` — before: fails at `linux_usbfs.c:716:21` with the identical message; after: passes, `make check` included.
* `--enable-debug-log --werror`: passes. The cast is inert when logging is on.
* `--enable-udev --werror`: the umockdev test fails here either way, with a byte-identical ASan leak summary in both trees (115933 bytes in 8213 allocations) — my container-less environment, not the diff.

So that this is not a one-cell answer, I swept the rest of the matrix as well: every logging mode (default, `--enable-debug-log`, `--disable-log`, `--enable-system-log`) against both `--enable-udev` and `--disable-udev` under gcc 16.2, plus clang 22.1 and the sanitized recipe from the workflow (`-fsanitize=address,undefined -fstack-protector-all -ftrivial-auto-var-init=pattern -Wwrite-strings -Werror`). All clean, with no diagnostic naming `linux_usbfs.c` or `core.c`. The logging axis was the only one that interacted with the change at all.

One note on what the next run will show: the job stopped at `disable-log`, so the Emscripten step below it did not execute this time. It is the pre-existing red that #1936 addresses, and #1936 has not merged yet, so I expect it back.

libusb/os/linux_usbfs.c
```diff
@@ -708,18 +710,16 @@ static int open_sysfs_attr(struct libusb_context *ctx,
 	return fd;
 }
 
-/* Note only suitable for attributes which always read >= 0, < 0 is error */
-static int read_sysfs_attr(struct libusb_context *ctx,
-	const char *sysfs_dir, const char *attr, int max_value, int *value_p)
+/* Note only suitable for attributes which always read >= 0, < 0 is error.
+ * Takes ownership of fd and closes it. */
+static int parse_sysfs_attr_fd(struct libusb_context *ctx, int fd,
+	const char *attr, int max_value, int *value_p)
```
Youw, line 719: this function used to have a temporary internal `fd` which it owned and managed

now it requires a calling side to open an `fd` and takes ownerwhip (closes) of it which is not great design

this function should not take ownership of `fd` and the calling side who opened `fd` should remain responsible to close it

libusb/os/linux_usbfs.c
```diff
@@ -1222,6 +1306,23 @@ static int initialize_device(struct libusb_device *dev, uint8_t busnum,
...
+		int active_config;
+
+		if (wrapped_fd_get_active_config(dev, wrapped_fd,
+				&active_config) == LIBUSB_SUCCESS) {
+			usbi_dbg(ctx, "active configuration %d for wrapped device from sysfs",
+				 active_config);
+			priv->active_config = active_config;
+			return LIBUSB_SUCCESS;
+		}
+		usbi_dbg(ctx, "no sysfs answer for wrapped device, asking the device");
```
Youw:
```suggestion
		usbi_dbg(ctx, "no sysfs bConfigurationValue for wrapped device, asking the device");
```

libusb/os/linux_usbfs.c
```diff
@@ -1679,8 +1780,29 @@ static int op_get_configuration(struct libusb_device_handle *handle,
...
+			if (hpriv->fd_keep &&
+			    wrapped_fd_get_active_config(handle->dev, hpriv->fd,
+					&kernel_config) == LIBUSB_SUCCESS) {
+				if (kernel_config != active_config)
+					usbi_dbg(usbi_handle_ctx(handle),
+						 "device reports configuration %d but the kernel reports %d, keeping the kernel's",
+						 active_config, kernel_config);
+				priv->active_config = kernel_config;
+				active_config = kernel_config;
+			}
```
Youw, line 1810: I'm not entirely following when this practically happen; was it caught during testing or was it claude who assumed this scenario?

Youw (changes requested):
Now that AIs finished its "conversation", time for a human review.

@drpars ask claude to re-write comments/etc. as per latest policy introduced in https://github.com/libusb/libusb/commit/c75357cd70393b24e0416f664a09f0e97aced934

libusb/os/linux_usbfs.c
drpars: Done, as you describe. `parse_sysfs_attr_fd()` no longer closes the fd; its only two callers,
`read_sysfs_attr()` and `wrapped_fd_get_active_config()`, each close what they open.

errno ordering, since it is easy to break later: nothing runs between the failing `read()` and the
`r = errno;` capture, and in `wrapped_fd_get_active_config()` the `close()` runs before
`IOCTL_USBFS_GET_CAPABILITIES`, so it cannot overwrite the errno `ENOTTY` is tested on.

One divergence stays, stated so it is not a surprise: `read_sysfs_attr_fd()` still closes the fd
its caller opened, the opposite of `parse_sysfs_attr_fd()`. It predates this branch
and is outside this change's scope, so I left it; I will unify it in a separate commit if you want
that.

I also built the alternative shape — `read_sysfs_attr()` left exactly as master has it, the wrapped
path given its own reader — and measured it locally, so it is a push away. Do you want that one
instead?

libusb/os/linux_usbfs.c
drpars: Applied, byte for byte. The two sibling messages on this path now carry your term as well: the
`initialize_device()` success line, where it replaces a bare "sysfs", and the
`op_get_configuration()` divergence line, where it replaces "the kernel".

libusb/os/linux_usbfs.c
drpars: Part of it was assumed, and that part is what you are pointing at.

**Assumed.** A device that fails the `GET_CONFIGURATION` request, so
`usbfs_get_active_config()` writes the `config_descriptors[0]` guess and the sysfs
bConfigurationValue then corrects it: that was reasoned from the code, never observed. No comment on
this branch ever claimed it; the claim is in the PR description, "Intentional behaviour changes"
item 1, and the replacement description says there that it was reasoned from the code and that no
such device was seen. The one failing `GET_CONFIGURATION` I did measure is the `O_RDONLY` wrap in
the PR description, where the arm cannot run at all.

**Measured.** On the Intel 8087:0032 radio of issue #1925, every run I made read 0 from
`GET_CONFIGURATION` and 1 from sysfs bConfigurationValue. I also built your narrower scope — seed
the cache in `initialize_device()` only, leave `op_get_configuration()` alone — and ran it there:

| call | unpatched | seed only | seed + the arm |
|---|---|---|---|
| `libusb_get_active_config_descriptor()` | NOT_FOUND | OK | OK |
| `libusb_get_configuration()` | 0 | 0 | 1 |
| the same descriptor call again | NOT_FOUND | **NOT_FOUND** | OK |

The third row is the point: the first `libusb_get_configuration()` erases the seed, so the narrow
scope does not hold on this device. The mechanism is in the PR description, under "Scope".

The comment above the arm now states what the code does: `usbfs_get_active_config()` has already
written `priv->active_config` on every path it returns success from, and the arm overwrites that
value rather than preventing the write. It also keeps why the control request stays: that request
is what reports a disconnect.

Your call: keep the arm or drop it?

drpars:
## A correction first

In the `:720` thread I said I had swept the rest of the build matrix and that it was "All clean, with no diagnostic
naming `linux_usbfs.c` or `core.c`." It was narrower than I described: every cell went through `.private/ci-build.sh`,
which defaults `asan=yes`, and ASan *suppresses* the `-Wformat-truncation` in `open_sysfs_attr()` rather than exposing
it: the same cell exits 0 with ASan on and fails with it off:

```
.private/ci-build.sh --werror           --build-dir ci-asan   --no-test -- --disable-log  -> exit 0
.private/ci-build.sh --werror --no-asan --build-dir ci-noasan --no-test -- --disable-log  -> exit 2

In function 'open_sysfs_attr',
    inlined from 'read_sysfs_attr' at ../../libusb/os/linux_usbfs.c:790:7,
    inlined from 'op_get_interface_string' at ../../libusb/os/linux_usbfs.c:641:6:
../../libusb/os/linux_usbfs.h:30:33: error: '%s' directive output may be truncated writing
up to 255 bytes into a region of size 235 [-Werror=format-truncation=]
```

gcc 16.2.1 20260810. It is an inliner heuristic, so another gcc version may draw the line elsewhere.

## Comments rewritten per the policy introduced in `c75357cd`

Your review body asked for this; it is the bulk of the second commit. Comments, log strings and one
Doxygen note only, no logic changed.

* `wrapped_fd_get_active_config()`'s contract comment goes from 27 lines to 19. It was the longest
  comment block in `linux_usbfs.c`; it no longer is. What it keeps is stated as observed behaviour,
  not as usbfs internals.
* `op_get_configuration()`'s comment now states what the code does instead of what it protects, and
  keeps why the control request stays. Detail in the `:1804` thread.
* One term for one concept: sysfs bConfigurationValue for the kernel's answer, the device for the
  control request — in the three debug messages and in the comments around them.

## The truncation fix, as its own commit

`open_sysfs_attr()` discards its `snprintf()` return value. The commit consumes it and returns
`LIBUSB_ERROR_OTHER` when the path does not fit, in the form `read_sysfs_string_attr()` already uses here. It
touches pre-existing code, so the reason is on record: master never trips the defect. Splitting
`read_sysfs_attr()` for your `[P2]` point is what makes `open_sysfs_attr()` small enough for gcc to inline it
into `read_sysfs_attr()`, and that into `op_get_interface_string()`, where `intf_dir[256]` becomes the `%s`
argument. This PR creates the failure, so this PR fixes it. Over 16 gcc `-Werror` configurations per tree —
four logging modes × udev and no-udev at `-O2`, default and `--disable-log` at `-O3`, `-O0`/`-O1`/`-Os` with
default and disabled logging — master and this branch are clean in all 16 and `3649cc72` fails 4.

## One gap the run above exposes

The `-Werror` legs of `linux / build` — netlink, udev, debug-log and disable-log — all run
`.private/ci-build.sh`, which fixes `-O2` and leaves ASan on. The legs that do not (Release build, emscripten,
the umockdev container run, the clang ASan/UBSan rebuild, and the separate `GCC 16 / C23` job) do not build
with `-Werror`, so a warning does not fail them either, and there is no `-O3` leg anywhere. A class of gcc
diagnostics therefore cannot appear in that job. One non-ASan `-Werror` leg would close it; I can open a
separate issue for it rather than bundle it here.

## Three notes

The `O_RDONLY` wrap gets no transfer saving — measured, under "Limitation" in the PR description.

The two commits sit on top of `3649cc72` rather than replacing it, so `3649cc72..HEAD` is the delta.

Both commits now carry `claude-code:claude-opus-5` only; earlier ones carried a second `claude-code` trailer.

libusb/os/linux_usbfs.c
Youw: > Your call: keep the arm or drop it?

If this fixes observer unintended behavior with a real device - that is a good reason to keep it.

I do see the analysis from claude, the question here: @drpars do you understand its reasoning, implementation and the fix and do you agree that all of this is needed?

The author is reponsible for the code it contributes, not the AI/model and not the reviewer who tries to make sense of it.

Youw:
@seanm, @sonatique please take a look as well

libusb/os/linux_usbfs.c
drpars: The measurements in this PR were taken on my own machine, an ASUS G513RM, on the Intel 8087:0032 radio from #1925. I personally guarantee its accuracy - the work was done here, on this hardware, and I checked it. I guarantee the accuracy of all measurements on the G513RM. I accept the patch as is and take responsibility for it. Whether libusb will carry this is up to you, and I don't want you to rely on my word for the parts read from the code.

Part of this area was extracted from the code instead of being observed, and this is already stated in the PR description.

I should say who I am, since you don't know me. I am an amateur learning to read code. I ran into this while setting up VFIO dGPU passthrough with QEMU on this laptop, and followed it down to here. That is all.

drpars:
## Measurement

```
Command: uname -rm
Output:
7.1.9-zen1-2-zen x86_64

Command: gcc --version | head -1
Output:
gcc (GCC) 16.2.1 20260810

Command: lsusb -d 8087:0032
Output:
Bus 001 Device 003: ID 8087:0032 Intel Corp. AX210 Bluetooth

Command: lsusb -d 0b05:19b6
Output:
Bus 001 Device 002: ID 0b05:19b6 ASUSTek Computer, Inc. N-KEY Device

Command: git -C libusb log --oneline -1
Output:
9cc77779 ci: Do not trigger workflows on PR label events (#1947)

Command: git -C libusb worktree list
Output:
libusb      9cc77779 [master]
libusb-base 3601696f (detached HEAD)
libusb-head e2129540 (detached HEAD)

Command: git -C base rev-parse HEAD
Output:
3601696fae591a4236a5030ea7b42c717e610aa5

Command: git -C head rev-parse HEAD
Output:
e2129540487778c9ebb88ffda6126da08fc5e024

Command: git -C libusb diff --stat 3601696f e2129540
Output:
 libusb/core.c           |   6 +-
 libusb/os/linux_usbfs.c | 152 ++++++++++++++++++++++++++++++++++++++++++++----
 2 files changed, 145 insertions(+), 13 deletions(-)

Command: grep -m1 "  \$ " head/b/config.log
Output:
  $ ../configure --disable-udev

Command: cd base/b && make V=1 CFLAGS="-O2 -Wall -Wextra -Werror" 2>&1 | grep -c "warning:"
Output:
0

Command: cd head/b && make V=1 CFLAGS="-O2 -Wall -Wextra -Werror" 2>&1 | grep -c "warning:"
Output:
0

Command: ldd base/wraptest | grep libusb-1.0
Output:
libusb-1.0.so.0 => libusb-base/b/libusb/.libs/libusb-1.0.so.0 (0x00007f903f550000)

Command: ldd head/wraptest | grep libusb-1.0
Output:
libusb-1.0.so.0 => libusb-head/b/libusb/.libs/libusb-1.0.so.0 (0x00007fb779b73000)

```

```
Command: cd base && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
[MARK] ===== teardown =====

Command: cd base && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc recheck
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
[MARK] ===== libusb_get_active_config_descriptor (again) =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
[MARK] ===== teardown =====

Command: cd base && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc revorder
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== teardown =====

Command: cd base && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc revorder recheck
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 0
[MARK] ===== libusb_get_active_config_descriptor (again) =====
  libusb_get_active_config_desc -> -5 (LIBUSB_ERROR_NOT_FOUND)
[MARK] ===== teardown =====

Command: cd head && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== teardown =====

Command: cd head && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc recheck
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== libusb_get_active_config_descriptor (again) =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== teardown =====

Command: cd head && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc revorder
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== teardown =====

Command: cd head && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/003 1-4 nodisc revorder recheck
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/003

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] open /sys/dev/char/189:2/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor (again) =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 2
[MARK] ===== teardown =====

```

```
Command: cd base && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/002 1-3 nodisc
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/002

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 1
[MARK] ===== teardown =====

Command: cd head && LD_PRELOAD=./ioctl_spy.so ./wraptest /dev/bus/usb/001/002 1-3 nodisc
Output:
== kernel's view ==
  sysfs bConfigurationValue  = 1
  sysfs bNumConfigurations   = 1
[spy] open /dev/bus/usb/001/002

== libusb's view (fd 3, O_RDWR) ==
[MARK] ===== libusb_init_context (NO_DEVICE_DISCOVERY) =====
[spy] opendir /dev/bus/usb
[MARK] ===== libusb_wrap_sys_device =====
[spy] ioctl fd=3 USBDEVFS nr=31 size=0
[spy] open /sys/dev/char/189:1/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_wrap_sys_device        -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED)
[MARK] ===== libusb_get_configuration =====
[spy] ioctl fd=3 USBDEVFS nr=0 size=24   <-- CONTROL TRANSFER (goes on the bus)
[spy] open /sys/dev/char/189:1/bConfigurationValue
[spy] ioctl fd=3 USBDEVFS nr=26 size=4
  libusb_get_configuration      -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), config = 1
[MARK] ===== libusb_get_active_config_descriptor =====
  libusb_get_active_config_desc -> 0 (LIBUSB_SUCCESS / LIBUSB_TRANSFER_COMPLETED), bConfigurationValue = 1, bNumInterfaces = 1
[MARK] ===== teardown =====

```

Full log of the run, every command with its output:

- https://raw.githubusercontent.com/drpars/vfioctl/9a50d0c012dfbd634bb3b305e2a1676420011d64/docs/libusb-1926-run-2026-08-24-en.log
- https://raw.githubusercontent.com/drpars/vfioctl/9a50d0c012dfbd634bb3b305e2a1676420011d64/docs/libusb-1926-run-2026-08-24-tr.log (first run; Turkish scaffolding, same results)

The harness the logs call, so the run can be repeated:

- https://github.com/drpars/vfioctl/tree/9a50d0c012dfbd634bb3b305e2a1676420011d64/docs/libusb-1926-harness

libusb/os/linux_usbfs.c
drpars: Keep it. On this radio the request answers 0 and `dev_has_config0()` is false, so `usbfs_get_active_config()` stores -1 and `libusb_get_configuration()` reports 0; the arm writes the kernel's 1 over it, in all eight radio runs of the measurement I just posted.

Youw (approved):
This looks like a good improvement.

PR #1932, opened 2026-08-13
suppress some clang static analyzer false positives with asserts

seanm:
The clang static analyzer warning of possible NULL deref, but in fact the API contract for these functions requires passing non-NULL.

Make that obvious to both the analyzer and human readers with asserts.


Youw (approved):
This is even better than a documented contract.

seanm:
I'm a big fan of functions asserting their input parameter requirements.

If others like this too, I might do a pass and do even more of it...

PR #1935, opened 2026-08-15
core: reduce variable scope in core.c

seanm:


libusb/core.c
```diff
@@ -1864,7 +1831,7 @@ int API_EXPORTED libusb_claim_interface(libusb_device_handle *dev_handle,
 	if (dev_handle->claimed_interfaces & (1U << interface_number))
 		goto out;
 
-	r = usbi_backend.claim_interface(dev_handle, (uint8_t)interface_number);
+	int r = usbi_backend.claim_interface(dev_handle, (uint8_t)interface_number);
```
Youw: two lines above there is `goto out` and 6 lines below there is `return r` whihc might be uninitialized

Youw:
Long story short - if we're to avoid using C89 variables declaration (or at least narrow variable declaration to only where it is used) - it is hard to do so if we're to use "goto end" idiom in general.

libusb/core.c
seanm: Whoops, I should have looked at warnings before sharing this. :) Fixed.

Youw (approved):

PR #1936, opened 2026-08-15
emscripten: fix transfer helpers and version warnings

Youw:
## Summary

- Fix the WebUSB backend call to the existing `usbi_is_xferin()` helper after the macro-to-inline-function conversion in #1729.
- Fix the internal compatibility aliases to call the existing endpoint and transfer direction helpers.
- Map current and legacy Emscripten version macros to project-local `USBI_EMSCRIPTEN_*` aliases and use one minimum-version assertion.

## Context and compatibility

The WebUSB backend builds `emscripten_webusb.cpp` as C++. Emscripten 6.0.6 deprecates the mixed-case version macros. Emscripten 3.1.48 provides only those mixed-case forms, so the project-local aliases preserve the existing minimum SDK requirement without defining implementation-reserved identifiers. The aliases remain in this implementation translation unit.

This change does not modify the public API or ABI. The primary C library build is unaffected.

The hosted link steps still emit the pre-existing `-Wpthreads-mem-growth` performance notice. WebUSB requires `-pthread`, and this build intentionally permits memory growth. Changing either setting would change runtime behavior, so that notice is outside this PR.

## Validation

- [Hosted Linux PR run 31888182222](https://github.com/libusb/libusb/actions/runs/31888182222) passed. Its Emscripten step passed, and its log contains none of the reported mixed-case macro warnings or the invalid `usbi_is_xfer_in` call.
- Emscripten 3.1.48: configure, compile, and link passed through the legacy version-macro path.
- Current and legacy version-alias paths: C++20 syntax checks with `-Werror` passed; the minimum-version assertion accepts 3.1.48 and rejects 3.1.47.
- Primary C build: `.private/ci-build.sh --werror --build-dir build-local-primary-2 -- --enable-debug-log` passed.
- Internal compatibility aliases: GCC syntax check with `-Werror` passed.
- `git diff --check origin/master` passed.

Observed in [master Actions run 31880910116](https://github.com/libusb/libusb/actions/runs/31880910116).

Related to #1729

### DIFF
--- a/libusb/os/emscripten_webusb.cpp
+++ b/libusb/os/emscripten_webusb.cpp
@@ -25,8 +25,19 @@
 
 #include <emscripten/version.h>
 
-static_assert((__EMSCRIPTEN_major__ * 100 * 100 + __EMSCRIPTEN_minor__ * 100 +
-			   __EMSCRIPTEN_tiny__) >= 30148,
+#if defined(__EMSCRIPTEN_MAJOR__)
+#define USBI_EMSCRIPTEN_MAJOR __EMSCRIPTEN_MAJOR__
+#define USBI_EMSCRIPTEN_MINOR __EMSCRIPTEN_MINOR__
+#define USBI_EMSCRIPTEN_TINY __EMSCRIPTEN_TINY__
+#else
+/* Emscripten 3.1.48 provides only the legacy mixed-case version macros. */
+#define USBI_EMSCRIPTEN_MAJOR __EMSCRIPTEN_major__
+#define USBI_EMSCRIPTEN_MINOR __EMSCRIPTEN_minor__
+#define USBI_EMSCRIPTEN_TINY __EMSCRIPTEN_tiny__
+#endif
+
+static_assert((USBI_EMSCRIPTEN_MAJOR * 100 * 100 + USBI_EMSCRIPTEN_MINOR * 100 +
+			   USBI_EMSCRIPTEN_TINY) >= 30148,
 			  "Emscripten 3.1.48 or newer is required.");
 
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


Youw:
@RReverser your review on the part regarding version macro is welcome

libusb/os/emscripten_webusb.cpp
```diff
@@ -25,8 +25,19 @@
 
 #include <emscripten/version.h>
 
-static_assert((__EMSCRIPTEN_major__ * 100 * 100 + __EMSCRIPTEN_minor__ * 100 +
-			   __EMSCRIPTEN_tiny__) >= 30148,
+#if defined(__EMSCRIPTEN_MAJOR__)
+#define USBI_EMSCRIPTEN_MAJOR __EMSCRIPTEN_MAJOR__
```
RReverser, line 29: Personally I wouldn't introduce yet another custom macro name - just polyfill new ones via old ones like

```c
#ifndef __EMSCRIPTEN_MAJOR__
#define __EMSCRIPTEN_MAJOR__ ...
...
#endif
```

Makes the code both simpler now, and easier to rip out the shim in the future.

Youw: That's exactly how I initially done, but then I don't believe that is a very good idea to for the "user code" to define any macros that starts with `__` since technically that is an undefined behavior as per C and C++ standard.

Youw:
Team, this is a simple change and it fixes a broken master build.
I'm planning to merge this later today even if there will be no approve from other maintainers.

mcuee:
This github action build is okay -- so this PR seems to be okay.
https://github.com/libusb/libusb/actions/runs/31888182222/job/95020415324

@Youw 
But I have no idea why many github actions are now auto-skipped, including the latest build of this CR.
https://github.com/libusb/libusb/actions/runs/31925814077

Youw:
> I have no idea why many github actions are now auto-skipped

Unfortunately I do not know why that happens either

mcuee (approved):
CI passed now.

PR #1940, opened 2026-08-20
OSS-Fuzz: Revamp the fuzzers building process for OSS-Fuzz

arthurscchan:
At present OSS-Fuzz builds and runs one target only for the libusb. That target lives in the OSS-Fuzz repository (https://github.com/google/oss-fuzz/tree/master/projects/libusb), and the two targets added in this repository are not built or fuzz at all. This PR groups the three existing fuzz targets and builds them from a new tests/fuzz/Makefile.am, so that OSS-Fuzz can collect and fuzz all of them. After this infra revamp is merged in, more fuzz targets will be followed in later PRs.


tests/Makefile.am
```diff
@@ -1,3 +1,5 @@
+SUBDIRS = fuzz
```
Youw: I might not have a full picture here, but it seem that right now we have a separate CI job for fuzz tests, not sure it it relates to all these

does it mean the CI now would run both this and the other one?
and I believe the fuzz we have runs only on a single linux build, and looks like with this change each commit on each platform will be checked; that is probably increase the CI time a lot...

arthurscchan: The short answer: libusb currently has two separate sets of fuzz targets. One set lives in the OSS-Fuzz repository, which OSS-Fuzz builds and fuzzes continuously; the other set lives in this repository under tests/fuzz, but it has no build rule, so nothing builds it and nothing fuzzes it. This PR merges both sets into this repository so that OSS-Fuzz can pick up all of the targets and fuzz them continuously. The change does add a Makefile.am for the tests/fuzz directory, but building it requires the --enable-fuzzers flag, which defaults to off, and no workflow in this repository passes that flag, so the existing CI jobs build nothing new. The targets are also noinst_PROGRAMS rather than TESTS, so make check does not run them either. This Makefile.am exists purely so that OSS-Fuzz, and any other build that enables the flag, can pick up and build the fuzz targets.

**Remark: CIFuzz currently runs the single fuzzer that resides in the OSS-Fuzz repository (libusb_fuzzer). This PR renames that fuzzer to fuzz_control_setup, with identical logic. Once the OSS-Fuzz side is updated to use this Makefile.am, CIFuzz will run three fuzzers instead of one, but the run time barely changes: fuzz-seconds: 600 is a total budget that CIFuzz divides between the targets, so each one gets about 200 seconds rather than the full 600.**

Youw: in such case this is a fine housekeeping, thanks

Youw: > but building it requires the --enable-fuzzers flag

my confusing was that the subdirectory is added w/o an explicit condition, and I missed it while looking (not being very thorough, apparently) at the fuzz/Makefile

maybe lets move the condition on this level:
```suggestion
if BUILD_FUZZERS

SUBDIRS = fuzz

endif
```

Youw (commented):
Other than that - this looks solid. I'd update a CI to use this as well, but maybe later as a separate step, unless the author is willing to check on it right now too.

tests/Makefile.am
arthurscchan: I agree, fixed.

Youw (approved):
Overall looks good

Youw:
Codex review findings:

- **P2 — Fuzz seed corpora are absent from release archives.** `make dist` includes the fuzz target sources and Makefiles but omits every file under `tests/fuzz/corpus`. Add the corpus files to `EXTRA_DIST` so release-tarball builds retain the seed inputs.
- **P3 — `ENABLE_LOGGING` is redefined.** `fuzz_descriptor_parsers.c` defines `ENABLE_LOGGING` before including `config.h`, which then redefines it and emits a Clang warning. Define it after `config.h` with an `#undef`, or remove the local override.

Youw (approved):

PR #1942, opened 2026-08-21
darwin: fix interface leaks around re-claim, reset and close

### LINKED ISSUE #1923: darwin: `capture_count` is stranded when a re-claim fails during state restore
Several related defects in the darwin backend around claiming an interface again after the device state was rebuilt. All predate #1919. Items 1 to 4 affect macOS. In item 5, only the `capture_count` part does not.
### 1. A re-claim overwrites the previous interface plug-in
`darwin_claim_interface` writes the result of `QueryInterface` into `IOINTERFACE(cInterface)` without releasing what was there. `darwin_restore_state` leaves the pre-reset interfaces open, because `darwin_close_locked` only touches the device-level event source and `USBDeviceClose`. So every `libusb_reset_device` with interfaces claimed leaks one `IOUSBInterfaceInterface` per interface. Same for the re-claim in `darwin_set_interface_altsetting` after a pipe stall, and the set-configuration fallback inside `darwin_claim_interface`. Success path, no privileges needed, so this is the easiest one to hit.

### 2. Closing an auto-detached handle claims the interface back
`darwin_capture_release_interface` releases the interface, then reattaches the kernel driver, which re-enumerates. `darwin_reenumerate_device` reads `claimed_interfaces` while the bit for that interface is still set, because the core clears it only after the backend returns. `darwin_restore_state` claims it again, `darwin_close` has already passed that index, and the plug-in, event source and run loop reference all leak.
Seen on macOS 15.7.4 as root, claiming a CDC interface with `auto_detach_kernel_driver` and letting `libusb_close` release it:
```
darwin_attach_kernel_driver_locked: reenumerating device for kernel driver attach
darwin_restore_state: reclaiming interfaces
darwin_restore_state: re-claiming interface 1
darwin_restore_state: device state restored
```
The device does not come back as it started: `kernel_driver_active` for interface 2 reads 1, and read 0 before. Needs capture rights, which is why it is unreported.

### 3. Two failure returns in `darwin_restore_state` clean up nothing
The re-open failure and the `SetConfiguration` failure return with the claim bits cleared and every interface still open, so nothing releases them. #1919 added an unwind for the re-claim failure only.

### 4. `darwin_set_configuration_locked` ignores re-claim failures
It returns `LIBUSB_SUCCESS` either way, leaving a claim bit set with no plug-in behind it.

### 5. After a failed restore, the claim bits are left clear
The caller cannot release interfaces it still holds. `libusb_release_interface` returns `LIBUSB_ERROR_NOT_FOUND`, `darwin_close` skips them, and everything they hold is lost.
This never happens on working hardware, because re-enumeration either finishes or the device disappears and is caught earlier. I reproduced it with fault injection, forcing `darwin_claim_interface` to fail on its second call. Same result on x86_64/macOS 15.7.4 and arm64/macOS 26.4.1:
```
STEP 1 claim(2) -> 0 (LIBUSB_SUCCESS)
STEP 2 reset -> -5 (LIBUSB_ERROR_NOT_FOUND)
STEP 3 release after failed restore -> -5 (LIBUSB_ERROR_NOT_FOUND)     <-- the bug
STEP 4 close
```
For handles using `auto_detach_kernel_driver`, `darwin_capture_release_interface` also never runs, so `dpriv->capture_count` never reaches zero and the kernel driver is never reattached. The count lives on the shared `darwin_cached_device`, so it outlives `libusb_close`.

The `capture_count` strand affects iOS, tvOS and watchOS only. Not being able to release an interface you still hold, shown above, affects macOS too.

### Fix
In preparation, stacked on #1919 because it uses the helpers that PR adds. Five commits: 
1. release the interface before claiming it again
2. clear the claim bit before the reattach re-enumerates
3. release the interfaces on every `darwin_restore_state` failure return
4. return re-claim failures from `darwin_set_configuration_locked`
5. hand the claim bits back

Reference: #1798

sonatique:
Five fixes in the darwin backend. All are about claiming an interface again after the device state was rebuilt.

- `darwin_claim_interface` releases the interface it is about to replace. Before, it overwrote the plug-in pointer and leaked it. This runs on every `libusb_reset_device` with an interface claimed.
- `darwin_capture_release_interface` clears the claim bit before the kernel-driver reattach re-enumerates, so the restore no longer claims back the interface the core just released.
- `darwin_restore_state` releases its interfaces on all three failure returns, not just the re-claim one.
- `darwin_set_configuration_locked` releases an interface it failed to claim again, and returns the failure instead of always reporting success.
- `darwin_restore_state` hands the claim bits back on failure. On all platforms this lets `libusb_release_interface` succeed on an interface the caller still holds. On iOS, tvOS and watchOS it also lets close drain `capture_count`. That part is not needed on macOS, where `capture_count` is always zero during a restore.

### Testing

x86_64 / macOS 15.7.4 / Xcode 16.2 and arm64 / macOS 26.4.1 / Xcode 26.4.1, with a SEGGER J-Link. Clean builds on both, no warnings from `darwin_usb.c`.

The failure paths never happen on working hardware, so I forced them with a scratch build that makes `darwin_claim_interface` fail on its second call. That code is not part of this PR. Both machines gave the same result.

Before, with #1919 alone:

```
STEP 1 claim(2) -> 0 (LIBUSB_SUCCESS)
STEP 2 reset -> -5 (LIBUSB_ERROR_NOT_FOUND)
STEP 3 release after failed restore -> -5 (LIBUSB_ERROR_NOT_FOUND)
STEP 4 close
```

After:

```
STEP 1 claim(2) -> 0 (LIBUSB_SUCCESS)
STEP 2 reset -> -5 (LIBUSB_ERROR_NOT_FOUND)
STEP 3 release after failed restore -> 0 (LIBUSB_SUCCESS)
STEP 4 close
```

That `NOT_FOUND` is the bug. The claim bit was left clear, so the core refuses the release, `darwin_close` skips the interface, and everything it holds is lost.

The close problem is visible as root with `auto_detach_kernel_driver`: claim a CDC interface, then let `libusb_close` release it. Before:

```
darwin_attach_kernel_driver_locked: reenumerating device for kernel driver attach
darwin_restore_state: reclaiming interfaces
darwin_restore_state: re-claiming interface 1      <-- claimed back behind the core
darwin_restore_state: device state restored
```

The device does not come back as it started: `kernel_driver_active` for interface 2 reads 1, and read 0 before. After the fix that line is gone and the device comes back clean.

Normal use: 120 resets with an interface claimed and 40 claim/release cycles across both machines, all clean, no `Release final refCount` warnings.

ThreadSanitizer on arm64, library and tests instrumented: `stress`, `stress_mt` (8 threads), the reset loop, claim/release and `hotplugtest` all report zero races. I ran a deliberate race first to prove TSan was really detecting. TSan cannot see the `in_reenumerate` handshake, which the kernel orders rather than C11 atomics.

One cosmetic point: `darwin_release_interface` now logs `USBInterfaceClose: no connection to an IOService` once per reset, at warning level. It is closing the pre-reset interface, whose IOKit connection is already gone. That is the fix working, but it is new in debug output and easy to demote if reviewers find it noisy.

Stacked on #1919 and carries its commit until that merges, so review the top five only.

Fixes: #1923


Youw (commented):
Result: no material findings in the five PR #1942 commits.

Scope: `1a29fdda..6e5f6e4e`. The two stacked commits from #1919 were context only.

Codex checked interface and Core Foundation lifetimes, claim-mask and `capture_count` transitions, restore and configuration failure cleanup, lock order, and Apple SDK guards. Claude Fable 5 performed a separate strict read-only review of the same range and reached the same result.

Verification:

- `git diff --check 1a29fdda..6e5f6e4e` passed.
- All 12 current GitHub Actions runs and AppVeyor pass, including macOS.
- No inline fix suggestion is needed.

Remaining risk: The IOKit failure paths and a re-claim with an in-flight transfer have no automated test. The PR provides manual fault-injection and hardware results. The review environment had no Apple hardware and did not run the iOS, tvOS, or watchOS paths.

Generated by Codex (GPT-5.6 Sol) with an independent review by Claude Code (`claude-fable-5`).

mcuee:
Using the first test code from PR #1919 (generated by Claude Sonnet 5 Medium free chatbot).

```
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % cat issue1798_point4_runloop_pr1919.c
/*
 * issue1798_point4_runloop_pr1919.c
 *
 * Same reproducer as issue1798_point4_runloop.c, retargeted to build against
 * libusb PR #1919 ("darwin: hold retained run loop references instead of
 * rereading the global", branch sonatique:darwin-runloop-lifetime, commit
 * 203529f1fd5f1a3ad021e584d20bc4113835a7ac) instead of system libusb, so a
 * clean run is actual evidence the fix holds and not just "your installed
 * libusb happened not to hit it".
 *
 * Build libusb itself first (from the libusb source tree, on the pinned
 * commit above):
 *
 *   cd libusb
 *   git fetch origin pull/1919/head:pr1919
 *   git checkout 203529f1fd5f1a3ad021e584d20bc4113835a7ac   # pin exact commit
 *   ./bootstrap.sh
 *   CC=clang CFLAGS="-g -O1 -fsanitize=thread" \
 *   LDFLAGS="-fsanitize=thread" \
 *
 * Then build this test against that tree's headers/static lib directly
 * (NOT pkg-config, which would pick up your installed libusb instead):
 *
 *       -o issue1798_point4_runloop_pr1919 issue1798_point4_runloop_pr1919.c \
 *       -framework CoreFoundation -framework IOKit -lpthread \
 *       -o issue1798_point4_runloop_pr1919
 *
 * Recommended: run once against unpatched master (same commands, checkout
 * master instead of the PR commit) so you have a "still fails here" baseline
 * before confirming "clean on the PR branch".
 *
 * Reproducer for libusb/libusb#1798, item 4:
 *   "libusb_darwin_acfl accessed without libusb_darwin_at_mutex"
 *
 * darwin_open / darwin_close / darwin_claim_interface / darwin_release_interface
 * touch the global CFRunLoopRef libusb_darwin_acfl (CFRetain/CFRunLoopAddSource
 * on open+claim, CFRunLoopRemoveSource/CFRelease on close+release) WITHOUT
 * holding libusb_darwin_at_mutex. That mutex is only held on the event-thread
 * side when acfl is set (darwin_event_thread_main) and NULLed (event thread
 * shutdown, reached via the last libusb_exit()).
 *
 * This test intentionally races libusb_exit() against a pool of worker
 * threads that are mid-flight in open/claim/release/close on the SAME
 * context, to try to catch:
 *   - a NULL CFRunLoopRef dereference (acfl read as non-NULL, then NULLed,
 *     then used), or
 *   - a use-after-free / CF crash (acfl CFRelease'd by the exit path while
 *     a worker is still doing CFRetain/CFRunLoopAddSource/RemoveSource on it).
 *
 * IMPORTANT: this test deliberately does something the libusb docs do not
 * clearly forbid but also do not clearly allow -- calling libusb_exit()
 * while other threads still have in-flight calls on the same context. That
 * ambiguity is exactly what item 4 (and the item 7 discussion in #1798) is
 * about. Treat any crash/TSan report here as evidence for the audit, not
 * as "misuse" -- if this is misuse, it should be documented as such.
 *
 * (See the build instructions above this comment for the PR-pinned build --
 * do not use pkg-config here, it will silently link your installed libusb.)
 *
 * Usage:
 *   ./issue1798_point4_runloop_pr1919 <vid_hex> <pid_hex> [iface=0] [workers=8] [pre_exit_ms=50]
 *
 * Because libusb_exit() is called while workers may still be inside libusb
 * calls, this single run only covers ONE shutdown race window. Run it in a
 * loop (see the companion issue1798_point4_stress_pr1919.sh) to cover many
 * windows and increase the odds of hitting the narrow race.
 */

#include <stdio.h>
#include <stdlib.h>
#include <stdatomic.h>
#include <string.h>
#include <unistd.h>
#include <pthread.h>

static libusb_context      *g_ctx        = NULL;
static libusb_device       *g_dev        = NULL;
static int                  g_iface      = 0;
static atomic_int           g_stop       = 0;
static atomic_long          g_open_ok    = 0;
static atomic_long          g_open_fail  = 0;
static atomic_long          g_claim_ok   = 0;
static atomic_long          g_claim_fail = 0;

/* Each worker hammers open -> claim -> (tiny hold) -> release -> close,
 * back to back, with no delay, so that at any given instant several
 * threads are likely to be inside darwin_open/darwin_close/
 * darwin_claim_interface/darwin_release_interface -- exactly the four
 * functions item 4 flags as touching libusb_darwin_acfl unlocked. */
static void *worker_main(void *arg)
{
    (void)arg;
        int rc = libusb_open(g_dev, &handle);
        if (rc != LIBUSB_SUCCESS || handle == NULL) {
            atomic_fetch_add(&g_open_fail, 1);
            /* Context may already be torn down; back off briefly instead
             * of busy-spinning once libusb_exit() has clearly won. */
            usleep(200);
            continue;
        }
        atomic_fetch_add(&g_open_ok, 1);

        rc = libusb_claim_interface(handle, g_iface);
        if (rc == LIBUSB_SUCCESS) {
            atomic_fetch_add(&g_claim_ok, 1);
            /* Give the race window a little more room without stalling
             * the whole loop; keep it tiny so we cycle fast. */
            usleep(50);
            libusb_release_interface(handle, g_iface);
        } else {
            atomic_fetch_add(&g_claim_fail, 1);
        }

        libusb_close(handle);
    }
    return NULL;
}

{
    if (argc < 3) {
        fprintf(stderr,
            "usage: %s <vid_hex> <pid_hex> [iface=0] [workers=8] [pre_exit_ms=50]\n",
            argv[0]);
        return 2;
    }

    unsigned vid = (unsigned)strtoul(argv[1], NULL, 16);
    unsigned pid = (unsigned)strtoul(argv[2], NULL, 16);
    g_iface        = (argc > 3) ? atoi(argv[3]) : 0;
    int workers    = (argc > 4) ? atoi(argv[4]) : 8;
    int pre_exit_ms = (argc > 5) ? atoi(argv[5]) : 50;

    int rc = libusb_init(&g_ctx);
    if (rc != LIBUSB_SUCCESS) {
        return 1;
    }

    libusb_device **list = NULL;
    if (n < 0) {
                libusb_error_name((int)n));
        libusb_exit(g_ctx);
        return 1;
    }

            continue;
        if (desc.idVendor == vid && desc.idProduct == pid) {
            /* Take our own reference so the device survives
             * libusb_free_device_list()'s unref below. */
            g_dev = libusb_ref_device(list[i]);
            break;
        }
    }

    if (g_dev == NULL) {
        fprintf(stderr, "device %04x:%04x not found\n", vid, pid);
        libusb_exit(g_ctx);
        return 1;
    }

    fprintf(stderr,
        "point4: %d workers hammering open/claim(iface=%d)/release/close, "
        "libusb_exit() fires after ~%d ms\n",
        workers, g_iface, pre_exit_ms);

    pthread_t *tids = calloc((size_t)workers, sizeof(pthread_t));
    for (int i = 0; i < workers; i++)
        pthread_create(&tids[i], NULL, worker_main, NULL);

    /* Let workers ramp up so several are guaranteed to be mid-call
     * (inside darwin_open / darwin_claim_interface / darwin_close /
     * darwin_release_interface) when we pull the rug out from under
     * them below. */
    usleep((useconds_t)pre_exit_ms * 1000);

    /* This is the race: libusb_exit() runs darwin_exit(), which stops the
     * event thread and (per the audit) NULLs/CFReleases libusb_darwin_acfl
     * under libusb_darwin_at_mutex -- while worker threads above may be
     * inside darwin_open/darwin_close/darwin_claim_interface/
     * darwin_release_interface touching the same CFRunLoopRef WITHOUT that
     * mutex. */
    fprintf(stderr, "point4: calling libusb_exit() now (race point)\n");
    libusb_exit(g_ctx);
    fprintf(stderr, "point4: libusb_exit() returned (no crash inside exit itself)\n");

    /* Tell workers to stop and give them a moment to notice; any calls
     * they make on the now-dead context are exactly what we're trying to
     * catch, so we intentionally do NOT synchronize this stop with the
     * exit above. */
    atomic_store(&g_stop, 1);
    usleep(200 * 1000);

    for (int i = 0; i < workers; i++)
        pthread_join(tids[i], NULL);

    fprintf(stderr,
        "point4: done. open ok=%ld fail=%ld, claim ok=%ld fail=%ld "
        "(fails after the exit point are expected; a crash or TSan report "
        "anywhere above is the actual finding)\n",
        (long)g_open_ok, (long)g_open_fail, (long)g_claim_ok, (long)g_claim_fail);

    libusb_unref_device(g_dev);
    free(tids);
    return 0;
}

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1942_asan_debug/include issue1798_point4_runloop_pr1919.c -o issue1798_point4_runloop_pr1942_asan -L /Users/mcuee/mybin/bin_pr1942_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1942_asan
usage: ./issue1798_point4_runloop_pr1942_asan <vid_hex> <pid_hex> [iface=0] [workers=8] [pre_exit_ms=50]
mcuee in ~/build/libusb/libusb_test/issue1798_point4_pr1919 % ./issue1798_point4_runloop_pr1942_asan 04b4 8613
point4: 8 workers hammering open/claim(iface=0)/release/close, libusb_exit() fires after ~50 ms
point4: calling libusb_exit() now (race point)
point4: libusb_exit() returned (no crash inside exit itself)
=================================================================
==57905==ERROR: AddressSanitizer: heap-use-after-free on address 0x617000000080 at pc 0x00010029d074 bp 0x00016ff56210 sp 0x00016ff56208
READ of size 4 at 0x617000000080 thread T6
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
    #0 0x00010029d070 in log_v core.c:2954
    #1 0x000100294ff4 in usbi_log core.c:3052
    #2 0x0001002cabf0 in darwin_claim_interface darwin_usb.c:2366
    #3 0x0001002bbdcc in darwin_capture_claim_interface darwin_usb.c:3675
    #4 0x000100299e8c in libusb_claim_interface core.c:1867
    #5 0x0001001ed4e8 in worker_main issue1798_point4_runloop_pr1919.c:112
    #6 0x000100bbdd64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #7 0x000181ef7c54 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c54)
    #8 0x000181ef2c18 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1c18)

0x617000000080 is located 0 bytes inside of 752-byte region [0x617000000080,0x617000000370)
freed by thread T0 here:
    #0 0x000100bc1258 in free+0x7c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x41258)
    #1 0x00010029ce9c in libusb_exit core.c:2827
    #2 0x0001001ed0d4 in main issue1798_point4_runloop_pr1919.c:199
    #3 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

    #0 0x000100bc1450 in calloc+0x80 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x41450)
    #1 0x00010029bea4 in libusb_init_context core.c:2633
    #2 0x00010029bd50 in libusb_init core.c:2589
    #3 0x0001001ecc4c in main issue1798_point4_runloop_pr1919.c:143
    #4 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

Thread T6 created by T0 here:
    #0 0x000100bb9198 in pthread_create+0x60 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x39198)
    #1 0x0001001ed068 in main issue1798_point4_runloop_pr1919.c:184
    #2 0x000181b2c4e0 in start+0x1b4c (dyld:arm64e+0x204e0)

SUMMARY: AddressSanitizer: heap-use-after-free core.c:2954 in log_v
Shadow bytes around the buggy address:
  0x616ffffffe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616ffffffe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616fffffff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x616fffffff80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x617000000000: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x617000000080:[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000100: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000180: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000200: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000280: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x617000000300: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==57905==ABORTING
zsh: abort      ./issue1798_point4_runloop_pr1942_asan 04b4 8613
```
```

**Comment by mcuee** [comment]
New test code for this PR from Claude Sonnet 5 Medium free chatbot.

```
mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % cat darwin_reclaim_stress_pr1942.c
/* -*- Mode: C; indent-tabs-mode:t ; c-basic-offset:8 -*- */
/*
 * darwin_reclaim_stress - regression/stress test for libusb PR #1942
 * "darwin: fix interface leaks around re-claim, reset and close"
 * (stacked on PR #1919, "darwin: hold a run loop reference instead of
 * reading the global").
 *
 * This is a standalone, hardware-driving test tool, not part of the
 * portable tests/ suite: every scenario here only exists on the Darwin
 * backend and needs a real device attached. It is meant to be built and
 * run by hand against a `darwin_usb.c` built with and without #1942 on
 * top of #1919, and ideally under both `leaks`/`MallocStackLogging` and
 * ThreadSanitizer.
 *
 * What each scenario targets:
 *
 *  - reclaim_cycle: repeated claim/release of the same interface with no
 *    intervening reset. Exercises the new "release before claim" guard
 *    darwin_claim_interface() got in #1942 (previously a re-claim
 *    overwrote cInterface->interface without closing the old plug-in,
 *    leaking it and its run loop reference). Watch this under `leaks`
 *    with #1942 reverted vs applied.
 *
 *  - reset_while_claimed: this is the exact STEP 1-4 sequence from the
 *    #1942 PR description. Claim an interface, reset the device, then
 *    release. Before #1942, when a re-claim during darwin_restore_state
 *    fails (or more generally whenever the restore path returns early),
 *    the claim bit was left clear so the core's libusb_release_interface
 *    refuses with LIBUSB_ERROR_NOT_FOUND and darwin_close then skips the
 *    interface outright, leaking everything it held. After #1942, release
 *    must always report LIBUSB_SUCCESS here.
 *
 *  - detach_reset_close (root only, needs a kernel-driver-claimable
 *    interface such as a CDC ACM interface, opt in with -k): enables
 *    auto_detach_kernel_driver, claims the interface, resets, and checks
 *    libusb_kernel_driver_active() before close. This is the
 *    "close problem visible as root" scenario from the PR: before #1942,
 *    darwin_restore_state's re-claim reattached the kernel driver behind
 *    the core, so kernel_driver_active read 1 for an interface that read
 *    0 before the reset; also exercises darwin_capture_release_interface
 *    clearing dev_handle->claimed_interfaces before the kernel-driver
 *    reattach re-enumerates, which #1942 also fixes.
 *
 * None of these scenarios can fully confirm the IOKit plug-in/CFRunLoop
 * reference-count leaks are gone from return codes alone -- that is what
 * `leaks --atExit -- ./darwin_reclaim_stress ...` and a diff of `lsof -p`
 * pipe/port counts across iterations are for. What this program *can*
 * catch on every run is the functional regression: wrong return codes,
 * and libusb_kernel_driver_active() disagreeing with the state before
 * the reset.
 *
 * Build (Xcode command line tools, against an installed/prefixed libusb):
 *
 *   clang -Wall -Wextra -O0 -g \
 *       -I/path/to/libusb-build/include/libusb-1.0 \
 *       darwin_reclaim_stress.c \
 *       -L/path/to/libusb-build/lib -lusb-1.0 \
 *       -o darwin_reclaim_stress
 *
 * For a TSan run, build libusb itself with
 *   ./configure --enable-debug-log CFLAGS="-fsanitize=thread -g -O1" \
 *               LDFLAGS="-fsanitize=thread"
 * then link this program the same way and add -fsanitize=thread here too.
 *
 * Usage:
 *   ./darwin_reclaim_stress -d VID:PID [-i IFACE] [-n COUNT] [-k] [-v]
 *
 * SPDX-License-Identifier: LGPL-2.1-or-later
 */

/* Needed for getopt()/optarg on strict C standards; harmless on macOS,
 * where <unistd.h> already exposes them via _DARWIN_C_SOURCE. */
#ifndef _POSIX_C_SOURCE
#define _POSIX_C_SOURCE 200809L
#endif

#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <stdbool.h>
#include <inttypes.h>
#include <unistd.h> /* getopt */

#include <libusb.h>

static bool verbose = false;

#define LOGV(...) do { if (verbose) { printf(__VA_ARGS__); fflush(stdout); } } while (0)

struct test_stats {
	const char *name;
	int iterations;
	int failures;
};

static void stats_init(struct test_stats *s, const char *name)
{
	s->name = name;
	s->iterations = 0;
	s->failures = 0;
}

static void stats_report(const struct test_stats *s)
{
	printf("[%-20s] %d/%d iterations OK%s\n",
	       s->name, s->iterations - s->failures, s->iterations,
	       s->failures ? "  <-- FAILURES" : "");
}

/* Open dev_handle's device by VID:PID. Returns NULL on failure. */
static libusb_device_handle *open_target(libusb_context *ctx,
					  uint16_t vid, uint16_t pid)
{
	libusb_device_handle *handle;

	handle = libusb_open_device_with_vid_pid(ctx, vid, pid);
	if (!handle) {
		fprintf(stderr, "could not open device %04x:%04x "
			"(not found, or needs root/authorization)\n", vid, pid);
	}
	return handle;
}

/*
 * Scenario 1: reclaim_cycle
 *
 * Repeated claim/release of the same interface, no reset in between.
 * Every iteration must succeed. This is the cheapest way to exercise
 * darwin_claim_interface()'s new
 *     darwin_release_interface (dev_handle, iface);
 * guard before it re-does QueryInterface/USBInterfaceOpen -- run it
 * under `leaks` to see the difference #1942 makes to the IOKit plug-in
 * and CFRunLoopSourceRef reference counts.
 */
static void test_reclaim_cycle(libusb_device_handle *handle, int iface,
				int count, struct test_stats *stats)
{
	stats_init(stats, "reclaim_cycle");

	for (int i = 0; i < count; i++) {
		int r;

		stats->iterations++;

		r = libusb_claim_interface(handle, iface);
		if (r != LIBUSB_SUCCESS) {
			fprintf(stderr, "  [%d] claim failed: %s\n", i, libusb_error_name(r));
			stats->failures++;
			continue;
		}

		r = libusb_release_interface(handle, iface);
		if (r != LIBUSB_SUCCESS) {
			fprintf(stderr, "  [%d] release failed: %s\n", i, libusb_error_name(r));
			stats->failures++;
			continue;
		}

		LOGV("  [%d] claim -> release OK\n", i);
	}
}

/*
 * Scenario 2: reset_while_claimed
 *
 * This is the STEP 1-4 sequence from the #1942 PR description:
 *   STEP 1 claim(iface)
 *   STEP 2 libusb_reset_device()
 *   STEP 3 release after (possibly failed) restore
 *   STEP 4 close is done by the caller after the loop
 *
 * libusb_reset_device() itself may legitimately return
 * LIBUSB_ERROR_NOT_FOUND on some devices/OS combinations if the restore
 * can't re-claim before the caller gets control back -- that is not what
 * this test is checking. What must always hold, on every iteration and
 * regardless of the reset's own return code, is that the *release*
 * afterwards succeeds and a subsequent re-claim works. Before #1942 the
 * release after a failed restore returned LIBUSB_ERROR_NOT_FOUND because
 * the claim bit had been left clear.
 */
static void test_reset_while_claimed(libusb_device_handle *handle, int iface,
				      int count, struct test_stats *stats)
{
	stats_init(stats, "reset_while_claimed");

	for (int i = 0; i < count; i++) {
		int r;

		stats->iterations++;

		r = libusb_claim_interface(handle, iface);
		if (r != LIBUSB_SUCCESS) {
			fprintf(stderr, "  [%d] STEP 1 claim failed: %s\n",
				i, libusb_error_name(r));
			stats->failures++;
			continue;
		}

		r = libusb_reset_device(handle);
		LOGV("  [%d] STEP 2 reset -> %d (%s)\n", i, r, libusb_error_name(r));

		/* STEP 3: release must succeed no matter what reset returned. */
		r = libusb_release_interface(handle, iface);
		if (r != LIBUSB_SUCCESS) {
			fprintf(stderr,
				"  [%d] STEP 3 release after reset failed: %s "
				"(this is the bug #1942 fixes if you see NOT_FOUND here)\n",
				i, libusb_error_name(r));
			stats->failures++;
			continue;
		}
		LOGV("  [%d] STEP 3 release -> 0 (LIBUSB_SUCCESS)\n", i);

		/* Confirm the device handle is actually usable again: a
		 * fresh claim on the same interface must work. */
		r = libusb_claim_interface(handle, iface);
		if (r != LIBUSB_SUCCESS) {
			fprintf(stderr,
				"  [%d] re-claim after release failed: %s\n",
				i, libusb_error_name(r));
			stats->failures++;
			continue;
		}
		r = libusb_release_interface(handle, iface);
		if (r != LIBUSB_SUCCESS) {
			fprintf(stderr,
				"  [%d] final release failed: %s\n",
				i, libusb_error_name(r));
			stats->failures++;
			continue;
		}
	}
}

/*
 * Scenario 3: detach_reset_close (opt-in, typically needs root)
 *
 * Reproduces the "close problem visible as root" case: claim an
 * interface with auto_detach_kernel_driver enabled, reset, and check
 * that libusb_kernel_driver_active() reports the same thing it did
 * before the reset. Before #1942, darwin_restore_state's re-claim could
 * reattach the kernel driver behind the core (visible as
 * kernel_driver_active flipping 0 -> 1 across the reset), and
 * darwin_capture_release_interface's reattach could race the claim bit
 * against a re-enumeration.
 *
 * This only exercises anything interesting on an interface the kernel
 * actually wants to drive (e.g. a CDC ACM interface) and typically needs
 * root to detach/reattach kernel drivers on macOS, hence it is gated
 * behind -k and simply skipped otherwise.
 */
static void test_detach_reset_close(libusb_context *ctx, uint16_t vid, uint16_t pid,
				     int iface, struct test_stats *stats)
{
	libusb_device_handle *handle;
	int r, before, after;

	stats_init(stats, "detach_reset_close");
	stats->iterations = 1;

	handle = open_target(ctx, vid, pid);
	if (!handle) {
		stats->failures = 1;
		return;
	}

	r = libusb_set_auto_detach_kernel_driver(handle, 1);
	if (r != LIBUSB_SUCCESS) {
		fprintf(stderr, "  set_auto_detach_kernel_driver failed: %s "
			"(not supported on this platform build?)\n",
			libusb_error_name(r));
		libusb_close(handle);
		stats->failures = 1;
		return;
	}

	before = libusb_kernel_driver_active(handle, iface);
	LOGV("  kernel_driver_active before claim: %d\n", before);

	r = libusb_claim_interface(handle, iface);
	if (r != LIBUSB_SUCCESS) {
		fprintf(stderr, "  claim failed: %s\n", libusb_error_name(r));
		libusb_close(handle);
		stats->failures = 1;
		return;
	}

	r = libusb_reset_device(handle);
	LOGV("  reset -> %d (%s)\n", r, libusb_error_name(r));

	r = libusb_release_interface(handle, iface);
	if (r != LIBUSB_SUCCESS) {
		fprintf(stderr, "  release after reset failed: %s\n",
			libusb_error_name(r));
		stats->failures = 1;
		libusb_close(handle);
		return;
	}

	after = libusb_kernel_driver_active(handle, iface);
	LOGV("  kernel_driver_active after release: %d\n", after);

	if (before >= 0 && after >= 0 && before != after) {
		fprintf(stderr,
			"  kernel_driver_active changed across reset+release: "
			"%d -> %d (device did not come back clean)\n",
			before, after);
		stats->failures = 1;
	}

	/* libusb_close drains capture_count; this must not hang or assert. */
	libusb_close(handle);
}

static void usage(const char *prog)
{
	fprintf(stderr,
		"usage: %s -d VID:PID [-i IFACE] [-n COUNT] [-k] [-v]\n"
		"  -d VID:PID  target device, hex, e.g. -d 1a86:7523\n"
		"  -i IFACE    interface number to claim (default 0)\n"
		"  -n COUNT    iterations per scenario (default 40 for reclaim_cycle,\n"
		"              120 for reset_while_claimed, matching the PR's manual run)\n"
		"  -k          also run detach_reset_close (typically needs root and a\n"
		"              kernel-driver-claimable interface, e.g. CDC ACM)\n"
		"  -v          verbose per-iteration logging\n",
		prog);
}

int main(int argc, char *argv[])
{
	libusb_context *ctx = NULL;
	uint16_t vid = 0, pid = 0;
	int iface = 0;
	int reclaim_count = 40;
	int reset_count = 120;
	bool run_detach = false;
	bool have_device = false;
	int opt, r, exit_code = EXIT_SUCCESS;
	struct test_stats s1, s2, s3;

	while ((opt = getopt(argc, argv, "d:i:n:kv")) != -1) {
		switch (opt) {
		case 'd':
			if (sscanf(optarg, "%hx:%hx", &vid, &pid) != 2) {
				fprintf(stderr, "bad -d value, expected VID:PID in hex\n");
				return EXIT_FAILURE;
			}
			have_device = true;
			break;
		case 'i':
			iface = atoi(optarg);
			break;
		case 'n':
			reclaim_count = atoi(optarg);
			reset_count = reclaim_count;
			break;
		case 'k':
			run_detach = true;
			break;
		case 'v':
			verbose = true;
			break;
		default:
			usage(argv[0]);
			return EXIT_FAILURE;
		}
	}

	if (!have_device) {
		usage(argv[0]);
		return EXIT_FAILURE;
	}

	r = libusb_init_context(&ctx, NULL, 0);
	if (r != LIBUSB_SUCCESS) {
		fprintf(stderr, "libusb_init_context failed: %s\n", libusb_error_name(r));
		return EXIT_FAILURE;
	}

	const struct libusb_version *ver = libusb_get_version();
	printf("libusb %d.%d.%d.%d%s\n", ver->major, ver->minor, ver->micro,
	       ver->nano, ver->rc ? ver->rc : "");

	handle = open_target(ctx, vid, pid);
	if (!handle) {
		libusb_exit(ctx);
		return EXIT_FAILURE;
	}

	/* Detach kernel driver up front for the two return-code scenarios
	 * too, so a driver-claimed interface (e.g. AppleUSBCDCACMData)
	 * doesn't make every claim fail before we even get to exercise the
	 * reclaim/reset paths. */
	(void)libusb_set_auto_detach_kernel_driver(handle, 1);

	printf("=== scenario 1: claim/release cycling (no reset) ===\n");
	test_reclaim_cycle(handle, iface, reclaim_count, &s1);
	stats_report(&s1);

	printf("=== scenario 2: claim, reset, release (PR's STEP 1-4) ===\n");
	test_reset_while_claimed(handle, iface, reset_count, &s2);
	stats_report(&s2);

	libusb_close(handle);
	handle = NULL;

	if (run_detach) {
		printf("=== scenario 3: auto-detach + reset + close, root ===\n");
		test_detach_reset_close(ctx, vid, pid, iface, &s3);
		stats_report(&s3);
	} else {
		stats_init(&s3, "detach_reset_close");
		printf("=== scenario 3: skipped (pass -k to run) ===\n");
	}

	if (s1.failures || s2.failures || s3.failures)
		exit_code = EXIT_FAILURE;

	libusb_exit(ctx);

	printf("\n%s\n", exit_code == EXIT_SUCCESS ?
	       "PASS: all scenarios completed without a functional regression" :
	       "FAIL: see per-iteration errors above");

	return exit_code;
}
```

**Comment by mcuee** [comment]
Test results with this PR #1942, PR #1919 only and libusb git -- all are okay.

```
mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1942_asan_debug/include/libusb-1.0 darwin_reclaim_stress_pr1942.c -o darwin_reclaim_stress_pr1942_asan -L /Users/mcuee/mybin/bin_pr1942_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_pr1919_asan_debug/include/libusb-1.0 darwin_reclaim_stress_pr1942.c -o darwin_reclaim_stress_pr1919_asan -L /Users/mcuee/mybin/bin_pr1919_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % clang -Wall -Wextra -g -fsanitize=address -I /Users/mcuee/mybin/bin_git_asan_debug/include/libusb-1.0 darwin_reclaim_stress_pr1942.c -o darwin_reclaim_stress_git_asan -L /Users/mcuee/mybin/bin_git_asan_debug/lib -lusb-1.0 -pthread

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % ./darwin_reclaim_stress_git_asan -d 04b4:8613
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: skipped (pass -k to run) ===

PASS: all scenarios completed without a functional regression

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % ./darwin_reclaim_stress_pr1919_asan -d 04b4:8613
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: skipped (pass -k to run) ===

PASS: all scenarios completed without a functional regression

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % ./darwin_reclaim_stress_pr1942_asan -d 04b4:8613
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: skipped (pass -k to run) ===

PASS: all scenarios completed without a functional regression
```

With sudo and `-k` optoon.

```
mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_git_asan -d 04b4:8613 -k
Password:
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_git_asan -d 04b4:8613 -k
Password:
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression
mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_pr1919_asan -d 04b4:8613 -k
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression
mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_pr1942_asan -d 04b4:8613 -k
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression
```

**Comment by mcuee** [comment]
Test results with a USB Composite Device (custom + HID). Still no issues with this PR #1942, PR #1919 only or latest git.

```
mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_pr1942_asan -d 16c0:05dc -k
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_pr1919_asan -d 16c0:05dc -k
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression

mcuee in ~/build/libusb/libusb_test/issue1923_pr1942 % sudo ./darwin_reclaim_stress_git_asan -d 16c0:05dc -k
libusb 1.0.30.12037
=== scenario 1: claim/release cycling (no reset) ===
[reclaim_cycle       ] 40/40 iterations OK
=== scenario 2: claim, reset, release (PR's STEP 1-4) ===
[reset_while_claimed ] 120/120 iterations OK
=== scenario 3: auto-detach + reset + close, root ===
[detach_reset_close  ] 1/1 iterations OK

PASS: all scenarios completed without a functional regression
```

**Comment by mcuee** [comment]
Merge master branch to trigger PR run again to avoid the CI related issues.

**Comment by mcuee** [comment]
Looks like I will not be able to carry out real HW based tests with this PR (may need fault injection to the codes).
> The failure paths never happen on working hardware

As per the analysis mentioned in PR #1919, the first test code from PR #1919 may not be valid.

At least the new test code does not cause regressions.

Therefore I will still approve this PR based on the testing results.

**Review by mcuee — APPROVED** [review]

**Comment by mcuee** [comment]
@seanm 

Please review the changes as well.

**Comment by mcuee** [comment]
Not so sure why the CIFuzz/Fuzzing github action failed.
https://github.com/libusb/libusb/actions/runs/32642463037/job/97201557338?pr=1942

```
Run google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
  with:
    oss-fuzz-project-name: libusb
    fuzz-seconds: 600
    dry-run: false
    language: c
    sanitizer: address
    mode: code-change
    report-unreproducible-crashes: false
    minimize-crashes: false
    parallel-fuzzing: false
    output-sarif: false
    report-timeouts: true
    report-ooms: true
/usr/bin/docker run --name e24e1420780355f4259a4b1f3caa0e5917d_1a2720 --label 068e24 --workdir /github/workspace --rm -e "INPUT_OSS-FUZZ-PROJECT-NAME" -e "INPUT_FUZZ-SECONDS" -e "INPUT_DRY-RUN" -e "INPUT_LANGUAGE" -e "INPUT_SANITIZER" -e "INPUT_MODE" -e "INPUT_GITHUB-TOKEN" -e "INPUT_REPORT-UNREPRODUCIBLE-CRASHES" -e "INPUT_MINIMIZE-CRASHES" -e "INPUT_PARALLEL-FUZZING" -e "INPUT_OUTPUT-SARIF" -e "INPUT_REPORT-TIMEOUTS" -e "INPUT_REPORT-OOMS" -e "OSS_FUZZ_PROJECT_NAME" -e "LANGUAGE" -e "FUZZ_SECONDS" -e "DRY_RUN" -e "SANITIZER" -e "MODE" -e "GITHUB_TOKEN" -e "LOW_DISK_SPACE" -e "REPORT_UNREPRODUCIBLE_CRASHES" -e "MINIMIZE_CRASHES" -e "CIFUZZ_DEBUG" -e "CFL_PLATFORM" -e "PARALLEL_FUZZING" -e "OUTPUT_SARIF" -e "REPORT_TIMEOUTS" -e "REPORT_OOMS" -e "HOME" -e "GITHUB_JOB" -e "GITHUB_REF" -e "GITHUB_SHA" -e "GITHUB_REPOSITORY" -e "GITHUB_REPOSITORY_OWNER" -e "GITHUB_REPOSITORY_OWNER_ID" -e "GITHUB_RUN_ID" -e "GITHUB_RUN_NUMBER" -e "GITHUB_RETENTION_DAYS" -e "GITHUB_RUN_ATTEMPT" -e "GITHUB_ACTOR_ID" -e "G
2026-08-23 13:30:40,217 - root - DEBUG - Is github: True.
2026-08-23 13:30:40,218 - root - DEBUG - base_commit: None
2026-08-23 13:30:40,218 - root - DEBUG - pr_ref: refs/pull/1942/merge
2026-08-23 13:30:40,218 - root - DEBUG - No PROJECT_SRC_PATH.
2026-08-23 13:30:40,218 - root - INFO - Deleting builder docker images to save disk space.
2026-08-23 13:30:40,236 - root - DEBUG - Stderr of command "docker rmi -f gcr.io/oss-fuzz/libusb gcr.io/oss-fuzz-base/base-builder gcr.io/oss-fuzz-base/base-builder-go gcr.io/oss-fuzz-base/base-builder-javascript gcr.io/oss-fuzz-base/base-builder-jvm gcr.io/oss-fuzz-base/base-builder-python gcr.io/oss-fuzz-base/base-builder-rust gcr.io/oss-fuzz-base/base-builder-ruby gcr.io/oss-fuzz-base/base-builder-swift" is: Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-go:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-javascript:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-jvm:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-python:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-rust:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-ruby:latest
Error response from daemon: No such image: gcr.io/oss-fuzz-base/base-builder-swift:latest
.
2026-08-23 13:30:41,956 - root - INFO - ClusterFuzzDeployment: <clusterfuzz_deployment.OSSFuzz object at 0x7f59c38a9010>.
2026-08-23 13:30:41,957 - root - INFO - run fuzzers MODE is: code-change. Runner: <run_fuzzers.CiFuzzTargetRunner object at 0x7f59c377c510>.
2026-08-23 13:30:41,957 - root - INFO - Using address sanitizer.
2026-08-23 13:30:41,961 - root - INFO - Fuzz targets: ['/github/workspace/build-out/libusb_fuzzer']
2026-08-23 13:30:41,961 - root - INFO - Running fuzzer: libusb_fuzzer.
2026-08-23 13:30:41,961 - root - INFO - Downloading corpus from OSS-Fuzz: https://storage.googleapis.com/libusb-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/libusb_fuzzer/public.zip
2026-08-23 13:30:41,997 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): storage.googleapis.com:443
2026-08-23 13:30:42,159 - urllib3.connectionpool - DEBUG - [https://storage.googleapis.com:443](https://storage.googleapis.com/) "GET /libusb-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/libusb_fuzzer/public.zip HTTP/1.1" 200 2562618
2026-08-23 13:30:42,511 - root - INFO - Starting fuzzing
SIGBUS: bus error
PC=0x559e4e4408db m=7 sigcode=2 addr=0x559e4f194ec0
goroutine 0 gp=0xc0003061c0 m=7 mp=0xc000302008 [idle]:
```

PR #1947, opened 2026-08-23
ci: Do not trigger workflows on PR label events

### LINKED ISSUE #1945: CI builds were skipped for no reasons
Recently some CI builds for PRs are skipped for no reasons. Not so sure why.

Image

Youw:
## Summary

Remove the `labeled` trigger from the nine CI workflows.

Each label added to a PR started all nine workflows that subscribe to the `labeled` event (added in #1913 for the `force_testcore` / `force_long_tests` opt-in). For all other labels, every job was skipped by the label guard. Each label change still created nine "skipped" workflow runs, and their skipped check runs replaced the green check results on the PR head commit. This is what made CI look skipped in #1945. No real CI run was lost.

Evidence:

- PR #1942: label `macOS` added 2026-08-23 02:29:41Z → nine skipped runs at 02:29:44Z.
- PR #1940: label `test` added 2026-08-21 01:14:54Z → nine skipped runs at 01:14:57Z.
- All 18 skipped runs in the last 200 runs map 1:1 to these two label events.
- Only the nine workflows with the `labeled` trigger show skipped runs; windows, android and CIFuzz never do.
- The concurrent-job limit theory does not match the data: quota pressure leaves runs queued, never "skipped", and these runs completed within about 1 second of creation.

## Changes

- Drop `types: [opened, synchronize, reopened, labeled]`. A bare `pull_request:` trigger has the same three default types, minus `labeled`.
- Restore the simple concurrency group, identical to windows.yml and android.yml.
- Remove the per-job label guards.

The `force_testcore` / `force_long_tests` labels still work. The workflows read the label state from the event payload, so a new label takes effect on the PR's next push or close/reopen. GitHub search shows no PR has carried either label so far, so no workflow use depends on the immediate label-triggered run.

Fixes: #1945

### DIFF
--- a/.github/workflows/haiku.yml
+++ b/.github/workflows/haiku.yml
@@ -9,9 +9,10 @@ on:
     tags:
       - '**'
   pull_request:
-    types: [opened, synchronize, reopened, labeled]
 
-# Keep the long-running stress test on master unless a PR explicitly opts in.
+# Keep the long-running stress test on master unless a PR opts in with a
+# force_testcore or force_long_tests label. Labels are read from the event
+# payload, so a newly added label takes effect on the next push or reopen.
 env:
   TESTCORE_CONFIGURE_FLAG: >-
     ${{ (github.ref == 'refs/heads/master' ||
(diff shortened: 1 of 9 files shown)


Youw:
Basically at a time I've tried to force starting a build once a label is added, and that caused https://github.com/libusb/libusb/issues/1945. So the fix is to fall back to behavior when adding a lable doesn't re-trigged a build, but next build would respect the label as intended.

mcuee:
* PR #1948 with label `Documentation` also has the 9 tests skipped.

I think the analysis should be correct.

mcuee (approved):

mcuee:
No more skipped tests from this PR.

PR #1657, opened 2025-06-09
windows: Remove winsock include from libusb.h

tormodvolden:
From looking up commit 0ded9c1 , commit eea7ebe , commit 2442719 it seems winsock was introduced only for WinCE and for WDK, which we don't support any longer.

References #1509

PR #1955, opened 2026-08-28
windows: restore struct timeval in the public header

### LINKED ISSUE #1937: removing winsock.h break libusb_handle_events_timeout_completed/timeval
timeval is defined in winsock.h
the code with libusb 1.0.29 compile ok, but with libusb 1.0.30 failed.

the libusb.h should include timeval's definition for all platform.

on windows, it should be winsock.h, but commit e4baa9fa77a04f0cf168a178ed7a4c9679138027 removed it.

Youw:
`libusb.h` again gives applications a complete `struct timeval` on MSVC, and it
no longer forces Winsock 1 on them.

#1657 removed `#include <winsock.h>` from `libusb.h` for 1.0.30. On MSVC
`libusb.h` skips `<sys/time.h>` and `<time.h>` has no `struct timeval`, so the
only remaining source was the `winsock.h` that `windows.h` pulls in. An
application that defines `WIN32_LEAN_AND_MEAN` gets neither header, so it cannot
declare the `struct timeval` that `libusb_handle_events_timeout_completed()` and
the related functions take (#1937).

That same `winsock.h` also collides with a `winsock2.h` that the application
includes after `libusb.h`. #1509 reported this and was closed as fixed by #1657,
but it was not: `windows.h` still pulls in Winsock 1.

This change includes `<winsock2.h>` before `windows.h`, which is what #1509 first
proposed. The `_WINSOCKAPI_` guard skips the include when winsock is already in
scope. libusb's own Windows backends include `windows.h` first, so they are
unchanged, and every application include order that works today keeps working.
No public declaration changes.

Two effects on applications, both unavoidable if #1509 is to stay fixed:

- Code that includes `libusb.h` now gets Winsock 2 in place of Winsock 1. A call
  to a legacy name such as `inet_addr()` gets the C4996 deprecation warning,
  which is an error under `/WX`.
- Code that declared its own `struct timeval` to work around #1937 must remove
  that declaration, or `winsock2.h` reports C2011.

CI now compiles `.private/win32-header-check.c` twice, with and without
`WIN32_LEAN_AND_MEAN`, on MSVC and on MinGW. Both compiles fail on master.

Verified locally on Windows 11 with the Windows SDK 10.0.26100:

- MSVC (VS 2026, v145): full `msvc/libusb.sln` build, Debug x64, clean.
- MSVC, clang-cl, and mingw-w64 (x86_64 and i686): the new check fails on master
  and passes here, at `/W4 /WX` and `-Wall -Wextra -Werror`, as C and as C++20.
  MSVC also checked for x86, x64, and ARM64.
- Application include orders unchanged: `libusb.h` alone; `windows.h` first;
  `windows.h` first with `WIN32_LEAN_AND_MEAN`; `winsock2.h` first; `winsock.h`
  first; `windows.h` after `libusb.h`; `_WINSOCKAPI_` defined first.

Known limit: under a non-desktop `WINAPI_FAMILY`, an application that includes
`windows.h` before `libusb.h` still has no `struct timeval`, because the SDK
`winsock.h` sets `_WINSOCKAPI_` outside its desktop-only body. That case also
fails on master, and libusb does not support that target.

Fixes: #1937
Closes: #1509

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -90,6 +90,18 @@ jobs:
       - name: Setup MSBuild
         uses: microsoft/setup-msbuild@v3
 
+      # Syntax-only, so this needs no build. WIN32_LEAN_AND_MEAN makes windows.h
+      # skip winsock.h, which is the case that regressed in 1.0.30.
+      - name: Check public header include order
+        shell: cmd
+        run: |
+          for /f "usebackq tokens=*" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do set "VSDIR=%%i"
+          call "%VSDIR%\VC\Auxiliary\Build\vcvarsall.bat" x64
+          if errorlevel 1 exit /b 1
+          cl /nologo /W4 /WX /Zs /I libusb .private\win32-header-check.c
+          if errorlevel 1 exit /b 1
+          cl /nologo /W4 /WX /Zs /I libusb /D WIN32_LEAN_AND_MEAN .private\win32-header-check.c
+
       - name: Build ${{ matrix.configuration_family.first }}
         id: first_build
         continue-on-error: true
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -53,6 +53,15 @@ jobs:
           msystem: ${{ matrix.msystem }}
           update: true
           install: ${{ matrix.packages }}
+      # Syntax-only check of the public header's include contract; needs the
+      # source tree only. -Werror matters here: mingw-w64 reports a wrong
+      # winsock order with #warning, not an error.
+      - name: Check public header include order
+        run: >-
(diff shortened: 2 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -151,3 +151,34 @@ jobs:
           if ($env:FIRST_OUTCOME -ne 'success' -or $env:SECOND_OUTCOME -ne 'success') {
             throw "Build results: $env:FIRST_CONFIGURATION=$env:FIRST_OUTCOME, $env:SECOND_CONFIGURATION=$env:SECOND_OUTCOME"
           }
+
+  header-check:
```
Youw: instead of having another job in a matrix, maybe only add another step into existing builds?

Youw: Done, moved into the existing build jobs. It now runs as a step after "Setup MSBuild", so it needs no extra runner slot, and the windows.yml diff drops from 31 lines to 11. The MinGW check in msys2.yml was already a step in the existing job.

The step is still syntax-only (`cl /Zs`), so it costs about a second on top of the two MSBuild passes.

PR #1957, opened 2026-08-31
CMake support

pbosetti:
I have added support for building via CMake as static or shared library. It enables the library to be consumed via FetchContent by other CMake projects.
I have not added CI actions, because there are many ways to do that and I'd rather see if you accept this PR and discuss how to possibly deal with CI (e.g. separate actions for CMake building, vs. also repeat CMake building in the existing per-platform actions).


mcuee:
Please use libusb-cmake.

Youw:
https://github.com/libusb/libusb-cmake

